PR #5, opened 2019-04-02
DROOLS-3833: Ensure CDI annotations are only generated in Maven plugin

evacchi:


mswiderski:
we have now PR Jenkins jobs for submarine-runtimes so maybe it's work to use them ...

PR #7, opened 2019-04-03
DROOLS-3834: Process generation in jBPM Mojo should not cause DRL compilation

evacchi:
Previously: the KieBuilderImplWhatever infrastracture is slow to boot (lots of classes, lots of dynamic loading, assemblers, kie.confs etc etc etc) and also hard do customize. In this case we want to just run *only* the BPMN process builder, but our infrastructure requires DRL files to be processed as well. This is useless work (there's a different MOJO for DRLs) and it's also the root cause of some weird behaviors (DRL evaluation might error out even though the file is correct only because the KieBuilder is misconfigured).

In short: instead of using `kieBuilder`, I am bypassing the building API completely and just using XmlProcessReader because all of the KieBuild* infrastructure is already provided by Maven. So, bye-bye assemblers, KieServices and all their friends 👋👋👋

I see no downsides in this for now, and we can enhance as we go. 

Build has become much shorter, and it's easier to follow.


evacchi:
does it build the examples as well?

mswiderski:
nope, only runtimes are currently covered

PR #13, opened 2019-04-08
updates generated swagger tags and required definitions

tsurdilo:


tsurdilo:
* generates proper tags that match proper definition onb
* adds notnull on model id
this helps generate proper swagger.json from model

PR #15, opened 2019-04-09
annotation updates to process model generation

tsurdilo:
fixes parameter for post method in generated model so valid swagger json/yaml are generated for it


mswiderski:
@tsurdilo could you please make changes on top of master as this is now in conflict?

PR #32, opened 2019-04-18
JBPM-8433 - Support for process event listeners

mswiderski:


mswiderski:
@evacchi would you mind to have a look?

Added support for process event listeners and used new BpmnProcess as the base for new tests

jbpm/jbpm-bpmn2/src/main/java/org/kie/submarine/process/bpmn2/BpmnProcessInstance.java
```diff
@@ -29,11 +29,19 @@ public BpmnProcessInstance(AbstractProcess<BpmnVariables> process, BpmnVariables
 
     @Override
     protected Map<String, Object> bind(BpmnVariables variables) {
+        
+        if (variables == null) {
```
evacchi: why the null check? can we always expect a (possibly empty) BpmnVariables instance instead?

jbpm/jbpm-bpmn2/src/main/java/org/kie/submarine/process/bpmn2/BpmnProcessInstance.java
```diff
@@ -29,11 +29,19 @@ public BpmnProcessInstance(AbstractProcess<BpmnVariables> process, BpmnVariables
...
+            return null;
+        }
         return variables.asMap();
     }
 
     @Override
     protected void unbind(BpmnVariables variables, Map<String, Object> vmap) {
+        
+        if (variables == null) {
```
evacchi: same as above

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/LightProcessRuntime.java
```diff
@@ -451,7 +451,7 @@ public void abortProcessInstance(long processInstanceId) {
     }
 
     public WorkItemManager getWorkItemManager() {
-        return runtimeContext.getWorkItemManager();
+        return knowledgeRuntime.getWorkItemManager();
```
evacchi: the reason why ProcessRuntimeContext was used here is to avoid the dependency from the KnowledgeRuntime -- so the proper way to make this work would be to let LightProcessRuntimeContext return the right `WorkItemManager` instance. It's ok we don't do this right now, but I have opened a JIRA so we do it properly https://issues.jboss.org/browse/JBPM-8437

jbpm/jbpm-bpmn2/src/main/java/org/kie/submarine/process/bpmn2/BpmnProcessInstance.java
mswiderski: Due to data is optional and can be null, both for process ans work item. So check needs to happen somewhere

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/LightProcessRuntime.java
mswiderski: Ok, didn’t know exactly what was the purpose for runtimecontext thiugh it makes sense so let’s fix it in the other jira/pr

mswiderski:
@evacchi are you ok to merge this?

evacchi:
Go ahead

PR #33, opened 2019-04-19
DROOLS-3882: Align Drools Submarine API to jBPM Submarine API, provide bridge between the two 

evacchi:
Quite a big PR, but the meat of it is that I have dropped the separate processMojo and consolidated both into one GenerateModelMojo, which basically delegates all to shiny new `submarine-codegen` module, where I moved most of the codegen that was inside kie-maven-plugin. 

The entry point is ApplicationGenerator.

Also, it is now generating `my.group.id.Application` instead of `my.group.id.Module` and both RuleUnits and ProcessUnits are all together in the same Application. i.e. as discussed *Application Is The New Module™*.

The idea is to move all/most of our codegen there so that our runtime modules can stay agnostic from codegen.

For review, I suggest to import in an IDE, it's probably better navigation.

All examples build successfully, for tests, let's wait for CI.

I don't expect this to be merged soon because of Easter and all that, but let's start taking a look at it :-)

PS: please squash if you merge


submarine-codegen/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+    <version>8.0.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>submarine-codegen</artifactId>
+  <dependencies>
+    <dependency>
+      <groupId>org.jbpm</groupId>
+      <artifactId>jbpm-flow</artifactId>
+      <version>8.0.0-SNAPSHOT</version>
```
mswiderski: Use managed versions from bom

kie-maven-plugin/src/main/java/org/kie/maven/plugin/GenerateModelMojo.java
```diff
@@ -71,74 +69,152 @@
...
+//                cfg.withProcessConfig(
+//                        new ProcessConfigGenerator()
+//                                .withWorkItemConfig(
+//                                        processCodegen.moduleGenerator()
+//                                                .workItemConfigClass()));
+//                writeAll(processGeneratedFiles);
+//                getLog().info("Process Model successfully generated");
+//            }
+//        }
+//    }
```
mswiderski: Let’s remove the commented code if not needed any more

kie-maven-plugin/src/main/java/org/kie/maven/plugin/GenerateModelMojo.java
```diff
@@ -71,74 +69,152 @@
...
+    @Parameter(defaultValue = "${project.build.directory}/generated-sources/submarine")
+    private File generatedSources;
 
     @Parameter(property = "generateModel", defaultValue = "no")
     private String generateModel;
 
+    @Parameter(property = "generateProcessModel", defaultValue = "yes")
+    private String generateProcessModel;
+
     @Parameter(property = "dependencyInjection", defaultValue = "true")
```
mswiderski: Shall we align the naming for the property value true -> yes so we have the same as for generate model parameters

kie-maven-plugin/src/main/java/org/kie/maven/plugin/GenerateModelMojo.java
```diff
@@ -71,74 +69,152 @@
...
-
-    @Parameter(defaultValue = "${project.source.directory}")
-    private File projectSourceDirectory;
+    @Parameter(defaultValue = "${project.build.directory}/generated-sources/submarine")
+    private File generatedSources;
 
     @Parameter(property = "generateModel", defaultValue = "no")
     private String generateModel;
 
+    @Parameter(property = "generateProcessModel", defaultValue = "yes")
```
mswiderski: shouldn’t we have generate model enabled by default for drools and jbpm? Having this to be explicitly is not neccessary I believe...

kie-maven-plugin/pom.xml
```diff
@@ -188,5 +188,11 @@
       <artifactId>logback-classic</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.kie</groupId>
+      <artifactId>submarine-codegen</artifactId>
+      <version>8.0.0-SNAPSHOT</version>
+      <scope>compile</scope>
```
mswiderski: Same here, use managed version from bom

drools/drools-core-static/src/main/java/org/drools/statics/common/StaticServiceRegistry.java
```diff
@@ -39,38 +50,76 @@
...
+
+    static class StaticKieAssemblers implements KieAssemblers {
+
+        private Map<ResourceType, KieAssemblerService> assemblers = new HashMap<>();
+
+        public StaticKieAssemblers() {
+            // insert here reflective instantiation to assembler services
+            // this is an ad-interim solution
+            // e.g. assemblers.put(ResourceType.DMN, (KieAssemblerService)
+            //         instance("org.kie.dmn.core.assembler.DMNAssemblerService"));
```
mswiderski: Do we need to add anything now?

submarine-codegen/pom.xml
evacchi: ok

submarine-codegen/src/main/java/org/kie/submarine/codegen/process/ModuleGenerator.java
```diff
@@ -41,18 +40,24 @@
...
     private final List<ProcessGenerator> processes;
     private final List<ProcessInstanceGenerator> processInstances;
+    private final List<MethodDeclaration> factoryMethods;
     private String targetTypeName;
     private boolean hasCdi;
     private String workItemConfigClass = DefaultWorkItemHandlerConfig.class.getCanonicalName();
 
-    public ModuleGenerator(String groupId) {
-        this.packageName = groupId;
+    public ModuleGenerator(String packageName) {
```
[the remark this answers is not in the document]
evacchi: I think that at this point ModuleGenerator is used for little more than carrying data around, the real codegen happens in ApplicationGenerator. I have opened a JIRA to do some cleanup later https://issues.jboss.org/browse/DROOLS-3929

kie-maven-plugin/src/main/java/org/kie/maven/plugin/GenerateModelMojo.java
evacchi: I have opened this JIRA a few days ago so that we revisit all these conventions https://issues.jboss.org/browse/DROOLS-3924 if you prefer I can already change that to YES instead. Point is, I think a boolean should be used instead of a String (even for the other parameter values), and that that property should be namespaced (e.g. kie.dependencyInjection, kie.generateModel etc.)

kie-maven-plugin/src/main/java/org/kie/maven/plugin/GenerateModelMojo.java
evacchi: I'll probably revisit this while addressing this Maven-related JIRA https://issues.jboss.org/browse/DROOLS-3924 to adopt plugin conventions better. Right now I have just consolidated the 2 Mojos into one, for better codegen reuse (e.g. Quarkus extension), without considering defaults. 

Anyway I think I agree that codegen should be default, but right now Drools codegen relies on the "legacy" KnowledgeBuilder thingie, which behaves in a weird way when both BPMN and DRL resources are found, so for now I left it as it is.

kie-maven-plugin/pom.xml
evacchi: ok

drools/drools-core-static/src/main/java/org/drools/statics/common/StaticServiceRegistry.java
evacchi: I really hope we'll drop this altogether, but @tarilabs might need it while he's working on the DMN part that currently relies on assemblers (it should be easily to refactor it, but not a priority)

evacchi:
@mswiderski I have addressed your comments and/or deferred solution to a later JIRA. Can you take a look that my manual merge is working as expected? archetypes and examples are compiling. I'll wait for jenkins to finish for tests.

kie-maven-plugin/src/main/java/org/kie/maven/plugin/GenerateModelMojo.java
mswiderski: ok, fair enough, let's make sure we have this covered as soon as possible, I guess would make sense to have it before any release ...

kie-maven-plugin/src/main/java/org/kie/maven/plugin/GenerateModelMojo.java
mswiderski: +1 to both ... we can leave it as is for now and fix all at the same time

submarine-codegen/src/main/java/org/kie/submarine/codegen/process/ProcessCodegen.java
```diff
@@ -0,0 +1,300 @@
...
+        return moduleGenerator;
+    }
+
+    public ProcessCodegen withWorkItemHandlerConfig(String workItemHandlerConfigClass) {
+        this.workItemHandlerConfigClass = workItemHandlerConfigClass;
+        return this;
+    }
+
+    public ProcessCodegen withProcessEventListenerConfig(String customProcessListenerConfigExists) {
+        this.processEventListenerConfigClass = DefaultProcessEventListenerConfig.class.getCanonicalName();
```
mswiderski: shouldn't this use the customProcessListenerConfigExists (after it is bit renamed)?

submarine-codegen/src/main/java/org/kie/submarine/codegen/process/ProcessCodegen.java
```diff
@@ -0,0 +1,282 @@
...
+
+        for (ProcessInstanceGenerator pi : pis) {
+            storeFile(pi.generatedFilePath(), pi.generate().getBytes());
+        }
+
+        if (workItemHandlerConfigClass != null) {
+            moduleGenerator.setWorkItemHandlerClass(workItemHandlerConfigClass);
+        }
+
+        // I don't love that labels is a kind of "separate" return value, but let's keep it this way for now
```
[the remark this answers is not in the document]
mswiderski: maybe remove the comment or change the comment that this is needed to have s2i work properly

kie-maven-plugin/src/main/java/org/kie/maven/plugin/GenerateModelMojo.java
evacchi: plan to start working on this for this week. I can also do a short-term change just for the sake of property naming, then complete it with other improvements.

submarine-codegen/src/main/java/org/kie/submarine/codegen/process/ProcessCodegen.java
evacchi: whoopsss

evacchi:
@mswiderski addressed latest comments, please checkout again because I think some files were left out from commit while rebasing

mswiderski:
@evacchi just tried on my onboarding sample projects and all works as expected and the code looks even better :) Any further changes planned from your end?

evacchi:
cool! not for this PR, I'll do further PR with improvements as detailed above :-)

PR #37, opened 2019-04-27
add type of generated file so it can be easily examined

mswiderski:


submarine-codegen/src/main/java/org/kie/submarine/codegen/GeneratedFile.java
```diff
@@ -16,10 +16,22 @@
...
+    
+    public enum Type {
+        APPLICATION,
+        PROCESS,
+        PROCESS_INSTANCE,
+        REST,
+        RULE,
+        MODEL;
+    }
+
```
evacchi: I think it would be "more oop" if we used subclasses instead of an enum; i.e. ApplicationGeneratedFile extends GeneratedFile. More boilerplate to write, but it makes it easier to add new file types. so instead of doing
```storeFile(type, path, data)```
we would do
```storeFile(new ApplicationGeneratedFile(path, data))``` 
etc

evacchi:
just left one comment. we can address this separately on another PR but just wanted to point it out

submarine-codegen/src/main/java/org/kie/submarine/codegen/GeneratedFile.java
mswiderski: using subclass is nice on "this end" but on the other side where you need to check the type you would have to use instance of...

evacchi: heh, I'd think of exposing some method to invoke on the GenreatedFile subclass, which every subclass would override, but it's ok for now

PR #38, opened 2019-04-29
removed kie-dmn-bom leftovers

mswiderski:
depends on https://github.com/kiegroup/submarine-bom/pull/14


evacchi:
asking @mariofusco and @tarilabs to review, since the build has broken, but I am not sure what the current layout for DMN is on submarine

PR #39, opened 2019-04-30
quick fix: ProcessCodegen.ofFiles() should accept any Collection and not just List

evacchi:
one-liner. If builds pass, we merge.


evacchi:
no need to fix this, added fix in PR #40

PR #40, opened 2019-05-02
JBPM-8458: Code generation in subprocess node may use local data type info

evacchi:
when calling into a SubProcess we need to have info about the variables that we are going to bind/unbind. Instead of requiring full pre-analysis of all processes, we can use local type information that we have in our BPMN2 files:

e.g.:

```xml
    <bpmn2:callActivity calledElement="demo.orderItems"> 
      <bpmn2:ioSpecification>
        <bpmn2:dataInput drools:dtype="com.myspace.demo.Order" name="order"/>
```

### DIFF
--- a/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/LambdaSubProcessNodeVisitor.java
+++ b/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/LambdaSubProcessNodeVisitor.java
@@ -51,20 +51,26 @@ public class LambdaSubProcessNodeVisitor extends AbstractVisitor {
         Expression retValue = JavaParser.parse(resourceAsStream).findFirst(Expression.class).get();
 
         SubProcessNode subProcessNode = (SubProcessNode) node;
+        String subProcessId = subProcessNode.getProcessId();
         String nodeVar = "subProcessNode" + node.getId();
+
         addFactoryMethodWithArgsWithAssignment(body, SubProcessNodeFactory.class, nodeVar, "subProcessNode", new LongLiteralExpr(subProcessNode.getId()));
         addFactoryMethodWithArgs(body, nodeVar, "name", new StringLiteralExpr(getOrDefault(subProcessNode.getName(), "Call Activity")));
-        addFactoryMethodWithArgs(body, nodeVar, "processId", new StringLiteralExpr(subProcessNode.getProcessId()));
+        addFactoryMethodWithArgs(body, nodeVar, "processId", new StringLiteralExpr(subProcessId));
         addFactoryMethodWithArgs(body, nodeVar, "processName", new StringLiteralExpr(getOrDefault(subProcessNode.getProcessName(), "")));
         addFactoryMethodWithArgs(body, nodeVar, "waitForCompletion", new BooleanLiteralExpr(subProcessNode.isWaitForCompletion()));
         addFactoryMethodWithArgs(body, nodeVar, "independent", new BooleanLiteralExpr(subProcessNode.isIndependent()));
 
-        ModelMetaData subProcessModel = processToModel.get(subProcessNode.getProcessId());
+        Map<String, String> inputTypes = (Map<String, String>) subProcessNode.getMetaData("BPMN.InputTypes");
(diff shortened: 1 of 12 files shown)


jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/LambdaSubProcessNodeVisitor.java
```diff
@@ -51,20 +51,26 @@ public void visitNode(Node node, BlockStmt body, VariableScope variableScope, Pr
...
         addFactoryMethodWithArgs(body, nodeVar, "name", new StringLiteralExpr(getOrDefault(subProcessNode.getName(), "Call Activity")));
-        addFactoryMethodWithArgs(body, nodeVar, "processId", new StringLiteralExpr(subProcessNode.getProcessId()));
+        addFactoryMethodWithArgs(body, nodeVar, "processId", new StringLiteralExpr(subProcessId));
         addFactoryMethodWithArgs(body, nodeVar, "processName", new StringLiteralExpr(getOrDefault(subProcessNode.getProcessName(), "")));
         addFactoryMethodWithArgs(body, nodeVar, "waitForCompletion", new BooleanLiteralExpr(subProcessNode.isWaitForCompletion()));
         addFactoryMethodWithArgs(body, nodeVar, "independent", new BooleanLiteralExpr(subProcessNode.isIndependent()));
 
-        ModelMetaData subProcessModel = processToModel.get(subProcessNode.getProcessId());
+        Map<String, String> inputTypes = (Map<String, String>) subProcessNode.getMetaData("BPMN.InputTypes");
+        Map<String, String> outputTypes = (Map<String, String>) subProcessNode.getMetaData("BPMN.OutputTypes");
```
mswiderski: what if this is not given? I mean the BPMN file was created with for example eclipse modeller? Can we maybe fallback to search for variable by the mapping to see what type it is? Either here or at parsing time?

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ProcessToExecModelGenerator.java
```diff
@@ -76,6 +65,7 @@
 import org.kie.api.definition.process.Node;
 import org.kie.api.definition.process.NodeContainer;
 import org.kie.api.definition.process.WorkflowProcess;
+import org.w3c.dom.NamedNodeMap;
```
mswiderski: looks like not used import

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/VariableDeclarations.java
```diff
@@ -0,0 +1,51 @@
...
+
+package org.jbpm.compiler.canonical;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.jbpm.process.core.context.variable.Variable;
+import org.jbpm.process.core.context.variable.VariableScope;
+
+public interface VariableDeclarations {
```
mswiderski: any reason why this is interface and then we create anonymous class?

evacchi: right, I originally though to have 2 different impls, but I can make it a class now.

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/LambdaSubProcessNodeVisitor.java
evacchi: so in that case we should resolve using dataInput -> itemSubjectRef -> itemDefinition -> structureRef, right? or did you mean something else?

mswiderski: yes pretty much like that. Variable already has a data type so can also be used... if available

evacchi: ok I'm just keeping it like this because I don't have time to work further on this before summit (I am opening a new separate PR since this has been invalidated by the history import with the same changes)

PR #41, opened 2019-05-03
JBPM-8458: Code generation in subprocess node may use local data type info

evacchi:
supersedes #40


evacchi:
if this is ok we could merge it @mswiderski

mswiderski:
@evacchi catching up so let me give this a try today to see how it looks like...

mswiderski:
looks and works fine, merging

PR #43, opened 2019-05-03
Fix Drools codegen to avoid refleciton in class bind

evacchi:
should fix runtime NPE on native-image


evacchi:
will create Jira, please @mariofusco  take a look if it works with native image

PR #45, opened 2019-05-07
[DROOLS-3968] Move off JUnit 4 and onto JUnit Vintage

triceo:
Only merge after https://github.com/kiegroup/submarine-bom/pull/16.

### DIFF
--- a/drools/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/ComplexRulesTest.java
+++ b/drools/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/ComplexRulesTest.java
@@ -517,7 +517,7 @@ public class ComplexRulesTest extends BaseModelTest {
     }
 
     @Test
-    public void testCompareDateWithString() {
+    public void testCompareDateWithString() { // FIXME fails on non-US locales
         String str =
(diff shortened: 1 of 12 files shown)


drools/drools-model/drools-model-compiler/src/test/java/org/drools/modelcompiler/ComplexRulesTest.java
```diff
@@ -517,7 +517,7 @@ public void testCompareDate() {
     }
 
     @Test
-    public void testCompareDateWithString() {
+    public void testCompareDateWithString() { // FIXME fails on non-US locales
```
triceo: This is not related to my PR. I just found out that this particular test fails when I use cs_CZ locale on Linux.

triceo:
Failure appears unrelated to my changes. More importantly, it wasn't there last week, when I was first preparing this PR.

baldimir:
"[ERROR] 'dependencies.dependency.version' for junit:junit:jar is missing." Please provide similar PR also for submarine-cloud. Otherwise the PR pipeline will be broken.

triceo:
@baldimir I forgot about the cloud repo. Fixed now.

PR #47, opened 2019-05-15
DROOLS-4032: Parameterize Kogito source directory (default: src/main/resources)

evacchi:
This is currently required for the Quarkus extension to work properly.

### DIFF
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/ApplicationGenerator.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/ApplicationGenerator.java
@@ -41,6 +41,8 @@ import com.github.javaparser.ast.body.VariableDeclarator;
 
 public class ApplicationGenerator {
 
+    public static final String DEFAULT_GROUP_ID = "org.kie.kogito";
+    public static final String DEFAULT_PACKAGE_NAME = "org.kie.kogito.app";
     private static final String RESOURCE = "/class-templates/ApplicationTemplate.java";
     private final static String LABEL_PREFIX = "org.kie/";
     
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
@@ -41,6 +41,7 @@ import org.jbpm.compiler.xml.XmlProcessReader;
 import org.kie.api.definition.process.Process;
 import org.kie.api.definition.process.WorkflowProcess;
 import org.kie.api.io.Resource;
+import org.kie.kogito.codegen.ApplicationGenerator;
 import org.kie.kogito.codegen.ConfigGenerator;
 import org.kie.kogito.codegen.GeneratedFile;
 import org.kie.kogito.codegen.Generator;
@@ -117,7 +118,8 @@ public class ProcessCodegen implements Generator {
         for (Process process : processes) {
             this.processes.put(process.getId(), (WorkflowProcess) process);
         }
-
+        // set default package name
+        setPackageName(ApplicationGenerator.DEFAULT_PACKAGE_NAME);
     }
 
     public static String defaultWorkItemHandlerConfigClass(String packageName) {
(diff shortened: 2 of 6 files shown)


evacchi:
@mariofusco @mswiderski caveat: if a Java file is found in `src/main/resources` then Maven may try to build it. 

My proposal: what, if for submarine, we make this a configurable path default to `src/main/kie` -- this will need @mariofusco to fix the KieBuilder to make it configurable right now `resources` is hardcoded --; for "legacy" use cases, we can ask users to override the configuration to `src/main/resources`.

WDYT?

gitgabrio:
@evacchi 
out of curiosity: why not using build-helper:add-source to achieve that ? In what context it is used?

evacchi:
@gitgabrio this is the already existing Maven plugin, where we already add compiler sources (generated sources); the patch add to the source roots the directory that we traditionally use for DRL, BPMN sources (i.e. src/main/resources -- but we're thinking of making it src/main/kie).

The build helper plugin is used when you want to declare a new subdirectory, independently from other plugins you might have installed. e.g. if you have the kotlin plugin, that is supposed to add src/main/kotlin as a compiler source root [1]. 

However this change in particular is required because Quarkus currently will only monitor Maven source roots (and before the PR I submitted to them, only *one* single source root). This adds the root of the "drl/bpmn/etc" sources so that the quarkus extension hot code reload works as expected. It is a transitory measure, we should iterate on this a bit more because it's still not the best of all worlds.


[1] I may be lying a bit here for the purpose of the explanation -- kotlin plugin does this differently

gitgabrio:
@evacchi 
I'm not completely following you, surely my fault; beside the kotlin stuff (as you wrote, it works differently) this is the source code of original plugin

https://github.com/mojohaus/build-helper-maven-plugin/blob/master/src/main/java/org/codehaus/mojo/buildhelper/AddSourceMojo.java

and *it seems* to me (maybe I'm reading only a part of it, and this may be source of my misunderstanding) it does exactly the same thing as your modification

_this.project.addCompileSourceRoot( source.getAbsolutePath() );_

So, if Quarkus monitors Maven sources of the project using Maven, this is the scope of the AddSource goal. I'm keep talking of this because I had to do that on the two plugins I've recently done but there was a specific needs not covered by the builder-helper one.

evacchi:
I know what that plugin does, it's basically exposing the Maven API via a plug-in. Because that API is available for all plug-ins (we are already using to register our generated sources), and because our users will add our KIE plugin anyway (to do the first round of compilation before entering dev:mode) I thought it would be more convenient (for now) to just add that source root instead of requiring them to add *another* plugin. 

I may have not explained the Kotlin thing in the best way, but that *is* what the [kotlin plugin does in some cases](https://github.com/JetBrains/kotlin/blob/57d8e9457cf1ec33235d7cbedc2df42d5486e1aa/libraries/tools/kotlin-maven-plugin/src/main/java/org/jetbrains/kotlin/maven/kapt/KaptJVMCompilerMojo.java#L183) (although not in all cases, that's the reason for the footnote)
and the [scala plugin does in others](https://github.com/davidB/scala-maven-plugin/blob/fcade4c9507710ba5fe8cba662fb1f330d1a6e24/src/main/java/scala_maven/AddSourceMojo.java)

so I figured that's the best way to add our DRL, BPMN source directory to Maven, without requiring users to install *another* plugin, besides our own.

I am not entirely convinced that's the best thing to do, but remember this is just to push a release out of the window, then we will definitely revisit this one-liner

gitgabrio:
@evacchi 
Thanks for explanation - now I understood the missing part!

evacchi:
I have repurposed this issue. src/main/resources is still the default directory; the change is more extensive. I have copy/pasted `KieBuilderImpl` and butchered it into a new `KogitoBuilder` inside `kogito-codegen`. The new thingie does not rely on hardcoded paths anymore. It only compiles the DRLs it finds in the given directory, and produces Java source code, nothing else. It should also work for single files and not only root directories.

cc @mariofusco

baldimir:
@evacchi please rebase this as it has conflicts now.

evacchi:
the most relevant class of this PR got merged in commit c5f34c1a269c521adf831cb123eca4987a5c5bab so I can close this and revisit it as part of [DROOLS-4256: Make Quarkus extension hot reload to also work with new rules](https://issues.jboss.org/browse/DROOLS-4256)

PR #51, opened 2019-05-22
JBPM-8327 - Business rule task for local decision service, JBPM-8328 - Process Runtime integration with Rule Runtime, added tests in codegen module to test generated code

mswiderski:
support for business rule task to allow to execute rules (not decisions yet). The generated code for rules will be based on the generated ProjectRuntime

```
        org.jbpm.ruleflow.core.factory.RuleSetNodeFactory ruleSetNode3 = factory.ruleSetNode(3);
        ruleSetNode3.name("run rules");
        ruleSetNode3.ruleFlowGroup("employee", () -> {
            return app.ruleRuntimeBuilder().newKieSession("defaultStatelessKieSession");
        });
        ruleSetNode3.inMapping("employee", "employee");
        ruleSetNode3.outMapping("employee", "employee");
        ruleSetNode3.done();
```

Currently it uses stateless ksession to make it the most efficient, further on this can be extended to add support for statefull but for now I don't see much of a value for it.

In addition, this PR introduces a change in KieBuilderImpl that allows to skip the src/main/resources condition for resources to be added to kbase. Mainly to allow to use it for selected files use cases like using them from src/test/resources for unit testing.


api/kie-api/src/main/java/org/kie/submarine/process/Process.java
```diff
@@ -15,11 +15,17 @@
...
 
     ProcessInstance<T> createInstance(T workingMemory);
 
     ProcessInstances<T> instances();
 
     <S> void send(Signal<S> sig);
+    
+    T createModel();
+
+    ProcessInstance<? extends Model> createInstance(Model m);
```
evacchi: what's the difference with createInstance(T) ? I saw that you put the constraint on the type argument, but only in the abstract class `AbstractProcess< T extends Model >`, why not put it here?

api/kie-api/src/main/java/org/kie/submarine/Application.java
```diff
@@ -0,0 +1,37 @@
...
+ *
+ */
+public interface Application {
+    
+    Config config();
+
+    Process<? extends Model> processById(String processId);
+    
+    Collection<String> processIds();
+}
```
evacchi: I think it might be better to regroup these under `interface Processes {}`
as in 

```
public interface Application {
   Processes processes();
}
```
or something like that, because I fear an explosion in methods as it happened with KieSession. Not necessary right now, but we might want to revisit this soon, before we hit a stable release.

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
```diff
@@ -100,15 +101,16 @@ public void internalTrigger(final NodeInstance from, String type) {
                 throw new IllegalArgumentException("A RuleSetNode only accepts default incoming connections!");
             }
             RuleSetNode ruleSetNode = getRuleSetNode();
-            KieRuntime kruntime = getProcessInstance().getKnowledgeRuntime();
+            
+            KieRuntime kruntime = Optional.ofNullable(getRuleSetNode().getKieRuntime()).orElse(() -> getProcessInstance().getKnowledgeRuntime()).get();
```
evacchi: I think this will probably become something like `application.someRuleUnit().createInstance(RuleModel).fire()`

cf: the way subprocesses are now code-generated

mswiderski: yes, I agree here but I guess that is not there yet. So for the time being this enables us to use rules from within processes. Once we have the rule unit in place for not explicitly defined units then we can refactor

api/kie-api/src/main/java/org/kie/submarine/Application.java
mswiderski: very good point, will refactor this

mswiderski: added processes as new type

PR #52, opened 2019-05-23
DROOLS-3978 Add first bunch of unit tests for submarine-codegen

baldimir:
- Adds first few unit tests
- Fixes writing image metadata file in ApplicationGenerator - it failed to create the file when any of its path directories was not created 

Please squash if merging.


baldimir:
Thanks for review. License headers added.

PR #59, opened 2019-05-29
JBPM-8437: LightProcessRuntime should not pull WorkItemManager from kruntime

evacchi:
kruntime is there for legacy purposes. This PR removes useless API from ProcessRuntimeContext (getWorkItemManager()) because it's already in the ProcessRuntimeServiceProvider, which is already consumed by LightProcessRuntime, and returns the workItemManager instance that its in the service provider.

kruntime delegates to LightProcessRuntime

Rule of thumb: new code should not rely on kruntime, because that's a legacy interface that should go away. The dummy kruntime (if necessary) should only delegate to LightProcessRuntime


mswiderski:
is this ready to be merged?

evacchi:
since it's green, I'd say yes

PR #68, opened 2019-06-12
DROOLS-4160: Enable DRL codegen by default, rename config properties

evacchi:
MAKE CODEGEN GREAT AGAIN 💣

- enable model generation by default.

contextually rename properties as follows:

- `kogito.codegen.rules` : boolean -- enable codegen for rules. Defaults to null, meaning that codegen is enabled iff DRLs are found
- `kogito.codegen.processes` : boolean -- enable codegen for processes. Defaults to null, meaning that codegen is enabled iff BPMNs are founds 
- `kogito.sources.keep` : boolean -- keep DRL files in the generated artifact. Defaults to false.
- `kogito.di.enabled` : boolean  -- generate dependency injection annotation. Defaults to true.

### DIFF
--- a/archetypes/kogito-springboot-archetype/src/main/resources/archetype-resources/pom.xml
+++ b/archetypes/kogito-springboot-archetype/src/main/resources/archetype-resources/pom.xml
@@ -20,7 +20,7 @@
     <kogito.version>${version.org.kie.kogito}</kogito.version>
     <swagger.version>${version.swagger}</swagger.version>
     <cxf.version>3.2.6</cxf.version>
-    <dependencyInjection>false</dependencyInjection>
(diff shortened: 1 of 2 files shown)


mswiderski:
yay!!! shall we update the examples then too? remove the extra property?

evacchi:
yes, but first I want to make sure this does not break any tests in CI

evacchi:
I fear the reason why this break is that it's pulling kogito-codegen from master instead of using the local repo version. On my machine it's working because I manually installed kogito-codegen and api 🤔

evacchi:
jenkins please retest this

evacchi:
@baldimir you can take a look if you want

PR #70, opened 2019-06-21
DROOLS-4197 Fix Sonarcloud problems in kogito-maven-plugin

baldimir:
This fixes majority of issues reported by SonarCloud in kogito-maven-plugin (originally there were 102 issues, now there are 5).


baldimir:
Ok, so we have the option of the plugin cleanup. Should I delete everything irrelevant? If the decision about this would be long, I prefer we merge this. The code would still remain in the repository for some time in that case along with its problems.

evacchi:
code is fine; we can merge it and remove code later. Sorry for the useless work though!

PR #71, opened 2019-06-24
DROOLS-4242: NPE in Kogito Quickstart: broken codegen

evacchi:
Verified that this work by adding the following to my local copy of https://github.com/quarkusio/quarkus-quickstarts/tree/development/using-kogito

```
     <dependency>
      <artifactId>kogito-codegen</artifactId>
      <version>8.0.0-SNAPSHOT</version>
    </dependency>
    
```

new release? @mswiderski @mariofusco

### DIFF
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/AbstractApplicationSection.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/AbstractApplicationSection.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.codegen;
+
+import com.github.javaparser.ast.Modifier;
+import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
+import com.github.javaparser.ast.body.FieldDeclaration;
+import com.github.javaparser.ast.body.MethodDeclaration;
+import com.github.javaparser.ast.body.VariableDeclarator;
+import com.github.javaparser.ast.expr.NameExpr;
+import com.github.javaparser.ast.expr.ObjectCreationExpr;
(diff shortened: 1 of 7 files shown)


evacchi:
This is shared by maven and quarkus extension

Il Lun 24 Giu 2019, 18:10 Maciej Swiderski <notifications@github.com> ha
scritto:

mswiderski:
That’s fine by why it needs to be declared explicitly? It’s already included in kogito quarkus extension, isn’t it?

evacchi:
sure, I just added it to override the version number

kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/RuleUnitContainerGenerator.java
```diff
@@ -114,8 +115,10 @@ public ClassOrInterfaceDeclaration classDeclaration() {
         FieldDeclaration kieRuntimeFieldDeclaration = new FieldDeclaration();
 
         if (hasCdi) {
```
mswiderski: Isn’t it that this if else is doing exactly same thing now? Maybe just remove the if else and stick with just one

mswiderski:
Another thing is I am not sure we should always return new instance for both RuleUnits and then KieRuntimeBuilder- wouldn’t it make more sense to have just single instance for it?

evacchi:
@mswiderski good catches. It should be better now

kogito-codegen/src/main/java/org/kie/kogito/codegen/AbstractApplicationSection.java
```diff
@@ -0,0 +1,65 @@
...
+                .setName(innerClassName)
+                .addImplementedType(classType);
+    }
+
+    @Override
+    public FieldDeclaration fieldDeclaration() {
+        return new FieldDeclaration()
+                .addVariable(
+                        new VariableDeclarator()
+                                .setType(classType)
```
mswiderski: you should use innerClassName instead otherwise methods won't be found as they do not exist on the interface

kogito-codegen/src/main/java/org/kie/kogito/codegen/AbstractApplicationSection.java
```diff
@@ -0,0 +1,65 @@
...
+                                .setName(methodName)
+                                .setInitializer(new ObjectCreationExpr().setType(innerClassName)));
+    }
+
+    public MethodDeclaration factoryMethod() {
+        return new MethodDeclaration()
+                .setModifiers(Modifier.Keyword.PUBLIC)
+                .setType(classType)
```
mswiderski: you should use innerClassName instead otherwise methods won't be found as they do not exist on the interface

kogito-codegen/src/main/java/org/kie/kogito/codegen/AbstractApplicationSection.java
evacchi: should be fixed

PR #72, opened 2019-06-26
Update Kogito README

cristianonicolai:


jbpm/README.md
```diff
@@ -9,50 +9,64 @@
 [![License](https://img.shields.io/github/license/kiegroup/jbpm.svg)](https://github.com/kiegroup/jbpm/blob/master/LICENSE-ASL-2.0.txt)
 [![Twitter Follow](https://img.shields.io/twitter/follow/jbossjbpm.svg?label=Follow&style=social)](https://twitter.com/jbossjbpm?lang=en)
 
+Kogito jBPM
```
mswiderski: maybe we should have a overall readme at the root level? I'd prefer to avoid like to Jbpm website as it might get people confused... meaning kogito will have its own website

cristianonicolai: we would still keep this one then? or just one unified readme at the top level.

mswiderski: I'd say we would keep the top level only

cristianonicolai: @mswiderski moved most changes to the top level and removed the jbpm specific one now.

README.md
```diff
@@ -1,13 +1,70 @@
...
+git clone https://github.com/kiegroup/kogito-runtimes.git
+```
+
+2. Build with Maven:
+```
+cd kogito-runtimes
+mvn clean install -DskipTests
+```
+
+Contributing to jBPM
```
mswiderski: maybe change this to `Contributing to Kogito`

PR #75, opened 2019-06-28
KOGITO-111: kogito-codegen-tests: alternative testing strategy for codegen

evacchi:
Currently we are doing runtime compilation and dynamic class loading, but I think this is a better way, because we are reusing a toolchain:

- we setup a new kogito-codegen-tests (could have drools, jbpm submodules)
- we write data classes, rules, processes under src/main
- we write tests under src/test
- this way we can refer directly to generated code, as such:

```java
    Application app = new Application();

    @Test
    public void fireRuleTest() {
        KieSession kieSession =
                app.ruleUnits()
                        .ruleRuntimeBuilder()
                        .newKieSession();

        kieSession.insert(new Person());
        kieSession.fireAllRules();

        assertEquals(1, kieSession.getObjects(o -> o instanceof Result).size());
    }
```

or 

```java
        Application app = new Application();
        assertThat(app).isNotNull();

        BusinessRuleTaskProcess p = app.processes().createBusinessRuleTaskProcess();


        BusinessRuleTaskModel businessRuleTaskModel = new BusinessRuleTaskModel();
        businessRuleTaskModel.setPerson(new Person("john", 25));

        BusinessRuleTaskProcessInstance processInstance = p.createInstance(businessRuleTaskModel);
        processInstance.start();
```

benefits: 
- we reuse the maven toolchain, without running in-memory compiler and class loader wizardry, which are always error prone; so, less code to fix
- it doubles as a verification check to the codegeneration step because that code **must** compile, otherwise we have broken the codegen

caveats:
- not sure if the maven plugin that is being picked up is the one in the source tree, or the one from nexus, so we need to verify that
- you always need to run `mvn compile` before `mvn test` (and/or set-up your IDE to work accordingly -- there might be better ways to automate this)

thoughts? I believe overall this is a superior approach @mswiderski @baldimir because it's cleaner (no classloading wizardry) and at some point I think these are also APIs that we might want to allow end-users to code against (once they are stabler)


evacchi:
@mswiderski [this flag in the Maven plugin](https://github.com/evacchi/kogito-runtimes/blob/498195371e32c9d45af8df2a09ee3c83ef74617d/kogito-maven-plugin/src/main/java/org/kie/kogito/maven/plugin/GenerateModelMojo.java#L79-L86) should allow for partial recompilation; e.g.

```
mvn compile -Dkogito.codegen.partial  
```

but I realize now we should expose a file filter for the sub-directory we want to recompile

mswiderski:
nice to see progress on this :)

and how about IDE integration? I know you had in mind the IntelliJ feature that invokes maven goals, is this still the case...

evacchi:
that unfortunately would have to be done by hand (for now). Unless we come up with our own runner (we could even reuse the one for Quarkus but there's the problem of the extension being out of our source tree) 

Screenshot from 2019-08-01 11-12-19

mswiderski:
ok, would have to see how we could do the same in eclipse and then later in vscode :)

evacchi:
I am sure there's a way to do that

evacchi:
the problem with this is that the `kogito-maven-plugin` in `kogito-codegen-tests` is being pulled from nexus instead of the reactor. Do you guys know how to fix this? 🤔

evacchi:
main issue with this, is that the plugin in `kogito-codegen-tests` is being pulled from Nexus, so it breaks (or it is out of sync). 

Only solution -- to me -- seems to remove `kogito-codegen-tests` from the `<modules>` then `mvn install` and finally `cd kogito-codegen-tests && mvn install`. I don't totally hate this, but it means doing `mvn install` won't run all tests by default.

mswiderski:
> doing mvn install won't run all tests by default.

I guess that is a blocker, we need to have a way to run tests by default otherwise we will run into many issues people just not knowing that tests didn't run and send PRs

so what is the workflow now? do we always have to run maven goals to make use of this work? Wondering how it will impact regular working scheme ... modifying classes from both engine and code gen that are all in single IDE workspace but in different maven modules.

evacchi:
> I guess that is a blocker, we need to have a way to run tests by default otherwise we will run into many issues people just not knowing that tests didn't run and send PRs

last version of this works as expected. I did found a weird error when `<extension>true</extensions>`. In fact, the fix amounted to avoid doing that, and hook to a different lifecycle phase (see https://github.com/kiegroup/kogito-runtimes/pull/75/commits/ea4366a086e4c511f31b1f110b7c95f4bb759bcb) 

the way this works, it requires to do `mvn compile` before running the tests, that's all. We can also provide a flag to compile only part of the packages (partial support is already in, the "only-this-package" option is missing, but it should be pretty easy to implement)

mswiderski:
since this is a maven build wouldn't it mean that Jbpm or drools modules have to be also compiled/installed if there are any changes in them?

evacchi:
well yes, just as much as doing `mvn test` on any other submodule. You can also do `mvn test -pl drools,...,kogito-codegen-tests`. I am not necessarily advocating for putting all tests in this submodule, rather I am trying to see if there's a better workflow than invoking codegen and compile on-the-fly into memory.

mswiderski:
same goal here ... though I try to avoid too much hassle for our regular workflow when developing... right now with the compilation on the fly we have completely working solution for "in IDE". No need to invoke any of the maven goals on any module which makes the work much more efficient. If I would have to run maven build for every change in modules to try out the test that would be quite significant drawback.

evacchi:
I think this is pretty cool: I have added a `@KogitoTest` annotation, defined an annotation processor (which is usually picked up automatically by IDEs -- it does for IntelliJ for instance -- AND by Maven because it's run by `javac`) and start the codegen process from there. This allows to start tests directly from the IDE as expected. Works like a charm for me 🎉

Screenshot from 2019-08-07 14-08-17

mswiderski:
@evacchi great news, will give this a try tomorrow morning in eclipse and let you know how this worked out, thanks

evacchi:
@mswiderski last run failed with NPE because the last changed you guys pushed for config discovery break stand-alone usage. Can't do `new Application()` anymore because the configs are not injected in that case, that is the reason why we were doing the discovery "by hand" :-) in last commit I have added (as an experiment) a JUnit extension that uses Weld to do CDI in test cases, but I think we should consider not requiring a CDI impl in all cases

mswiderski:
@evacchi cdi is not required and the problem why it fails is that you set it explicitly to be used see here https://github.com/kiegroup/kogito-runtimes/pull/75/files#diff-04e288d94dbceff0f94aabee2aff4089R47

Just set it to null and it should work just fine

evacchi:
zomg. Ok I'll take a look

evacchi:
outdated for now

PR #79, opened 2019-07-02
[DROOLS-4241] first rule unit implementation

mariofusco:


api/kogito-api/src/main/java/org/kie/kogito/rules/DataStream.java
```diff
@@ -1,11 +1,11 @@
 /*
- * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2005 JBoss Inc
```
evacchi: I think we need to fix this back to RHT

drools/drools-core/src/main/java/org/drools/core/ruleunit/InternalDataStore.java
```diff
@@ -1,11 +1,11 @@
 /*
- * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2005 JBoss Inc
```
evacchi: same here

PR #83, opened 2019-07-03
[DROOLS-3960] Executable Model - kie-maven-plugin fails to generate model

lucamolteni:
with a mix of Java and DRL facts (#2327) - Tests in DROOLS-4011 (droolsjbpm-integration)


mariofusco:
We don't want to use the Eclipse compiler (especially if only for an edge case like this). I'm even not sure that the same problem holds on kogito.

PR #86, opened 2019-07-04
KOGITO-42 - Use Spring specific dependencies for injection in generat…

mswiderski:
…ed code


kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/RuleUnitContainerGenerator.java
```diff
@@ -134,8 +136,9 @@ public static ClassOrInterfaceType ruleUnitType(String canonicalName) {
                 .setTypeArguments(new ClassOrInterfaceType(null, canonicalName));
     }
 
-    public RuleUnitContainerGenerator withCdi(boolean hasCdi) {
+    public RuleUnitContainerGenerator withCdi(boolean hasCdi, DependencyInjectionAnnotator annotator) {
```
evacchi: I suppose we could just remove the boolean, if this is !=null then, we add the annotations, otherwise we omit them. WDYT? (same comment everywhere else)

mswiderski: yes, we can do that

mswiderski: updates done

PR #88, opened 2019-07-05
KOGITO-43: IncrementalRuleCodegen should generate all related rules when one DRL file is changed

evacchi:
This PR brings in most logic from ModelWriter [1] **into** IncrementalRuleCodegen, so that it used the ModelGenerator, without depending on ModelWriter to do the file writing. It also consolidates all the non-model-related codegen (ProjectModel, ProjectRuntime, ...) that is currently scattered across many different components (e.g. KieProject etc.) in one single point, separate from the model compiler.

[1] until this cannot be forward-ported https://github.com/kiegroup/drools/pull/2428


evacchi:
jenkins please retest this

evacchi:
jenkins please retest this

PR #98, opened 2019-07-17
[KOGITO-56] first implementation of automatic rest endpoint generatio…

mariofusco:
…n for queries


mariofusco:
ok to test

api/kogito-api/src/main/java/org/kie/kogito/rules/RuleUnitInstance.java
```diff
@@ -15,7 +15,11 @@
 
 package org.kie.kogito.rules;
 
+import java.util.List;
+import java.util.Map;
+
 public interface RuleUnitInstance<T extends RuleUnitMemory> {
     RuleUnit<T> unit();
     int fire();
+    List<Map<String, Object>> executeQuery( String query, Object... arguments);
```
evacchi: I wonder if we could do a less "generic" API and tie it more with the `T` that we have there, like... if the query is declared somehow in the unit, so that we actually return something meaningful instead of a List<Map<...>>.

drools/drools-core/src/main/java/org/drools/core/base/SequentialKnowledgeHelper.java
```diff
@@ -111,8 +111,8 @@ public Activation getMatch() {
         return this.activation;
     }
 
-    //    public QueryResults getQueryResults(final String query) {
-    //        return this.workingMemory.getQueryResults( query );
+    //    public QueryResults executeQuery(final String query) {
```
evacchi: I guess we may delete lines that have been commented out

drools/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/PackageModel.java
```diff
@@ -134,6 +134,7 @@
 
     private final String pkgUUID;
     private Set<Class<?>> ruleUnits = new HashSet<>();
+    private Map<Class<?>, List<QueryDescr>> queriesByRuleUnit = new HashMap<>();
```
evacchi: for simpler code, consider (if it makes sense) to create e RuleUnitDescr (any name is fine) that contains only the queries for that rule unit, instead of using a Map.  `private Set<Class<?>> ruleUnits = new HashSet<>(); ` may be fields of that class too.

drools/drools-model/drools-model-compiler/src/main/java/org/drools/modelcompiler/builder/PackageModel.java
```diff
@@ -349,6 +350,15 @@ public void addRuleUnit(Class<?> ruleUnitType) {
         return ruleUnits;
     }
 
+    public void addQueryInRuleUnit(Class<?> ruleUnitType, QueryDescr query) {
```
evacchi: and these could be methods for that class

kogito-codegen/src/main/java/org/kie/kogito/codegen/FileGenerator.java
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+
+package org.kie.kogito.codegen;
+
+import java.nio.charset.StandardCharsets;
+
+import static org.kie.kogito.codegen.ApplicationGenerator.log;
+
+public interface FileGenerator {
```
evacchi: what's the interface for? do we ever use this except for implement clauses?

kogito-codegen/src/main/java/org/kie/kogito/codegen/FileGenerator.java
```diff
@@ -0,0 +1,33 @@
...
+    String generatedFilePath();
+
+    String generate();
+
+    default GeneratedFile generateFile(GeneratedFile.Type fileType) {
```
evacchi: If I've understood the purpose of this interface fileType() can be just another method, and this one can be made zero args.

kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/RuleCodegenProject.java
```diff
@@ -35,10 +35,9 @@
...
 
-import com.github.javaparser.ast.CompilationUnit;
-import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
-import com.github.javaparser.ast.body.FieldDeclaration;
-import com.github.javaparser.ast.expr.StringLiteralExpr;
+import static com.github.javaparser.StaticJavaParser.parse;
+import static org.kie.kogito.codegen.ApplicationGenerator.log;
+import static org.kie.kogito.codegen.rules.RuleUnitsRegisterClass.RULE_UNIT_REGISTER_SOURCE;
 
 public class RuleCodegenProject extends CanonicalModelCodeGenerationKieProject implements KieBuilder.ProjectType {
```
evacchi: as discussed, we can drop this class, if you want

PR #100, opened 2019-07-24
[KOGITO-70] kogito-quarkus-archetype pom dependencyManagement

tkobayas:
Applying the same dependencyManagement from kogito-examples/pom.xml. It may not be fully required for the test-process but aligning with kogito-examples sounds good to me.


tkobayas:
Removed redundant <version>

PR #106, opened 2019-07-31
KOGITO-97: Internal message-based API for DataProcessor

evacchi:
note: this is really similar to jBPM's usage of signals, so we might want to unify here.

the DataStore+InternalDataStore interface pattern tends to be leaky, and it exposes details about the implementation in the public interface. I am proposing an alternative design where only a simple method is exposed:

```
interface DataProcessor<T> {
    void process(Message<T> m)
}
```

which also makes it nice to implement as a lambda (e.g. when you add a subscriber just for "peeking" at the data that's passing by -- very useful for testing; see https://github.com/kiegroup/kogito-examples/pull/59 where I came up with a factory for DataProcessor exactly to avoid implementing N methods)

a public hierarchy of messages is provided:

- Insert<T>
- Update<T>
- Delete<T>

end users are not supposed to interact with these messages directly, unless they want to write their own DataProcessor; they instead use public methods on DataSources. If they do want to implement a DataProcessor, they can easily choose which message they want to listen to, and discard the others e.g.:

```
DataProcessor<T> p = m -> { 
  if (m instanceof Insert) { /* do something */ }
  // ignore everything else
}
```

at the same time, this allows us to define internal messages, travelling on the same "logical" bus; e.g. we can provide a public implementation for Update<T> that any Processor can use, but also extend it to an internal EntryPointUpdate<T> that is only recognized by EntryPointProcessor instances (and processed by any other DataProcessor as a regular Update otherwise)

upsides: 
  - our interface is cleaner
  - it's probably a pattern we will reuse in other places (signal for yield/suspend on units)

downsides: 
  - we are now using instanceof instead of method dispatch (minor problem probably)
  - testing harder ? not sure...


mswiderski:
@evacchi isn't this pretty much the same as command pattern? Or maybe more of what we have both in drools and jbpm as command executor/runner? 

I am not a big fun of commands as they move a lot of logic from the service itself into the commands and then we start exposing very generic api like `execute(Command<?>)` or `process(Message<?>)` and but that we delegate a lot of responsibilities to end users to know the internals of each command etc.

We have in v7 already [KieCommands](https://github.com/kiegroup/droolsjbpm-knowledge/blob/39a0db88718a7b227c9b40f52a3f2c6f4136538c/kie-api/src/main/java/org/kie/api/command/KieCommands.java) class that looks to me as the same as these Insert, Update etc. Or am I just being paranoid? ;)

evacchi:
> @evacchi isn't this pretty much the same as command pattern? Or maybe more of what we have both in drools and jbpm as command executor/runner?

It's more similar to jBPM's signals than the command pattern, as the logic is not in the Message (as it would be for Command), but entirely in the receiver.

mswiderski:
so the message it self defines the destination - meaning what is going to consume it and data itself. But has no logic in it whatsoever. Is that right?

in jBPM signals we have it split into type which is kind of topic/destination and event which is data.

evacchi:
> so the message it self defines the destination - meaning what is going to consume it and data itself. But has no logic in it whatsoever. Is that right?

in this case, the destination is given by the sender, who specifically invokes `target.process(message)`, so there is no topic or channel. It's not much different from actually doing `target.method()`, but messages can be discarded and subtyped while methods cannot.

the point would be to avoid having things like
```
update(x)
```

and then 

```
update(x, and other parameters that are public but you should not really use it, it's internal and you should not even see it in the public interface)
```
on the one hand messages are less discoverable than public methods, on the other hand, that's kind of the point.

I am not saying this MUST be merged at all costs. We can still use the old Interface+InternalInterface pattern, but I don't love it :P

mswiderski:
was reviewing it few more times and have some concerns/comments/questions
- Message which seems to be the DataEvent can actually carry over the logic as well - see Insert that defines sender which is a processor that can be quite complex and might not be "wire friendly" with will make it really difficult to be transferred over network.. maybe that's not needed but at least message as name would suggest it should be :)
- what's the reason to associate processor with message itself like it's done in the Insert example, shouldn't these two be separated? Let's say I want to have multiple processors...
- DataEvent ... come on ... I wanted to use that name ... but since it's an empty interface I can still do it... but seriously why it is an empty interface? just to allow all and then distinguish based on instance of?

maybe I am missing many things just by looking at the PR itself without knowing how it worked before. 

P.S.
I am not big fun of Interface + InternalInterface approach so that's why asking so many questions :)

evacchi:
> was reviewing it few more times and have some concerns/comments/questions
> 
>     * Message which seems to be the DataEvent 

right I initially called it Message, then renamed with something more specific

> - can actually carry over the logic as well - see Insert that defines sender which is a processor that can be quite complex and might not be "wire friendly" with will make it really difficult to be transferred over network.. maybe that's not needed but at least message as name would suggest it should be :)

ok, reason for that is that I wanted to refer to the callee, specifically because I needed to [set it here](
https://github.com/kiegroup/kogito-runtimes/pull/106/commits/e06ffd6cbae6bfde58056c26a515547eaea8194a#diff-99e93aa6fdea1a7498e2feab2702296fR71)

now this is very specific to this piece of code, and maybe we can improve it further and remove that reference to the DataProcessor to the FactHandle. 

I did not intend this piece of code to be wire-friendly, but I see your point. Read on

>     * what's the reason to associate processor with message itself like it's done in the Insert example, shouldn't these two be separated? Let's say I want to have multiple processors...

in this case -- I was unsure about this -- the sender that it's being used is actually delegating to a collection of DataProcessors. Maybe we may define a DataProcessorManager/SignalManager. The interface is essentially the same though. I am borrowing the reasoning from actor processing here; you usually send a message to some target; then *it* decides whether it should forward to its children, which is similar here (although, in actors you would a mailbox, while here processing is synchronous). In this case, think of it as a chain of listeners. 

Consider that if you really wanted DataProcessor to be wire-friendly, in that case the underlying implementation would be a stub to something that sends the message over the wire; so it would be more of an address, than a proper object. Ok, now I am even [reinventing ActorRefs from first principles](https://doc.akka.io/japi/akka/current/akka/actor/ActorRef.html)

>     * DataEvent ... come on ... I wanted to use that name ... but since it's an empty interface I can still do it... but seriously why it is an empty interface? just to allow all and then distinguish based on instance of? Yes. There not much they have in common here, but maybe the DataHandle could be shared.

I am thinking of this kind of like of a "sealed" hierarchy. I do not expect people to actually define really "new" operations, rather "extensions" of the original; 

e.g. EntryPointUpdate: the message can be both seen as a simple Update, or as an extension, enriched with more info, but only if you know about that specific type; in that case the code path is different, but only if you know that type exists: see [EntryPointDataProcessor](https://github.com/kiegroup/kogito-runtimes/pull/106/commits/e06ffd6cbae6bfde58056c26a515547eaea8194a#diff-99e93aa6fdea1a7498e2feab2702296fR53). In fact, I think I should make [EntryPointUpdate](https://github.com/kiegroup/kogito-runtimes/pull/106/commits/e06ffd6cbae6bfde58056c26a515547eaea8194a#diff-2762cccc7f8f9fdd17d15f2a401e9044)  package-private, even. 

This basically realizes a sort-of "friend" class, which is allowed to send a special message that only the 2 parties are aware of, with specialized meaning. But at the same time, all other observers -- DataProcessors which are not EntryPointDataProcessors -- will be able to see the message and treat it as any other message.

> maybe I am missing many things just by looking at the PR itself without knowing how it worked before.
> 
> P.S.
> I am not big fun of Interface + InternalInterface approach so that's why asking so many questions :)

I am still not 100% convinced this is the best solution, but I still kind of prefer it to that  :)

mswiderski:
what I meant with wire-friendly was to make the Message/DataEvent as such not the data processor which indeed has little or no value to be wire-friendly. Moreover, I would actually remove the data processor from the Message/DataEvent and "look it up" based on the event type once received via the entry point like you said DataProcessorManager/SignalManager.

That way we have it clear separation and we could make the Message/DataEvent wire-friendly if needed as it can carry data and not logic/code so to say.

I like the extension part as additional processors can just consume if it does have extra info on it.
Will look later at the other patterns.

mariofusco:
I discussed this with @evacchi (@mswiderski when will you learn Italian so I could copy/paste our chats here?) and we decided to leave this here for a bit more time and further investigate if we could find a more satisfying implementation.

evacchi:
let's close it for now

PR #110, opened 2019-08-07
KOGITO-101: Add Kube labels for rule endpoints

evacchi:
very simple change. Add rule unit labels to the collection, should be written in the output file, like processes.

### DIFF
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/RuleCodegen.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/RuleCodegen.java
@@ -38,6 +38,10 @@ import org.kie.kogito.codegen.GeneratedFile;
 import org.kie.kogito.codegen.di.DependencyInjectionAnnotator;
 import org.kie.kogito.codegen.rules.config.RuleConfigGenerator;
 
+/**
+ * @deprecated use {@link IncrementalRuleCodegen}
+ */
+@Deprecated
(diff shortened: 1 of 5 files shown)


kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/RuleCodegen.java
```diff
@@ -38,6 +38,10 @@
 import org.kie.kogito.codegen.di.DependencyInjectionAnnotator;
 import org.kie.kogito.codegen.rules.config.RuleConfigGenerator;
 
+/**
+ * @deprecated use {@link IncrementalRuleCodegen}
+ */
+@Deprecated
 public class RuleCodegen extends AbstractGenerator {
```
mswiderski: why not removing this class directly? Is there any use of it....

mswiderski: why not removing this class directly? Is there any use of it....

evacchi: good point. I left it around just for the Quarkus extension, but we might delete and rename. @mariofusco any objections?

evacchi: opening JIRA to remove class

PR #113, opened 2019-08-08
[KOGITO-104] automatically generate DTO for unit to be used as input …

mariofusco:
…in rest endpoint


mswiderski:
I am not sure we can reuse much of the model class generator which is currently based on a process definition and its variables. Plus it uses some additional fields (like id) and method (like to/fromMap) which are most likely of no use for rule units based endpoints.

PR #116, opened 2019-08-13
KOGITO-140, KOGITO-139, KOGITO-141, KOGITO-142

mswiderski:
@cristianonicolai here are the fixes and improvements to allow to better integration with data index service

### DIFF
--- a/addons/persistence/infinispan-persistence-addon/src/main/java/org/kie/kogito/infinispan/CacheProcessInstances.java
+++ b/addons/persistence/infinispan-persistence-addon/src/main/java/org/kie/kogito/infinispan/CacheProcessInstances.java
@@ -38,7 +38,7 @@ public class CacheProcessInstances implements MutableProcessInstances {
     
     public CacheProcessInstances(Process<?> process, RemoteCacheManager cacheManager, String proto, MessageMarshaller<?>...marshallers) {
         this.process = process;    
-        this.cache = cacheManager.administration().getOrCreateCache(process.id(), (String)null);
+        this.cache = cacheManager.administration().getOrCreateCache(process.id() + "_store", (String)null);
         
         this.marshaller = new ProcessInstanceMarshaller(new ProtoStreamObjectMarshallingStrategy(proto, marshallers));
     }
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/process/persistence/proto/ReflectionProtoGenerator.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/process/persistence/proto/ReflectionProtoGenerator.java
@@ -24,7 +24,7 @@ public class ReflectionProtoGenerator implements ProtoGenerator<Class<?>> {
             Proto proto = new Proto(packageName, headers);
 
             for (Class<?> clazz : dataModel) {
-                messageFromClass(proto, clazz, null);
+                messageFromClass(proto, clazz, null, null, null);
             }
             return proto;
         } catch (Exception e) {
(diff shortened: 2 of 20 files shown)


cristianonicolai:
jenkins please retest this

addons/persistence/infinispan-persistence-addon/src/main/java/org/kie/kogito/infinispan/CacheProcessInstances.java
```diff
@@ -38,7 +38,7 @@
     
     public CacheProcessInstances(Process<?> process, RemoteCacheManager cacheManager, String proto, MessageMarshaller<?>...marshallers) {
         this.process = process;    
-        this.cache = cacheManager.administration().getOrCreateCache(process.id(), (String)null);
+        this.cache = cacheManager.administration().getOrCreateCache(process.id() + "_store", (String)null);
```
cristianonicolai: should I also add a _doman suffix for the data index caches?

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/persistence/proto/ReflectionProtoGenerator.java
```diff
@@ -116,11 +116,15 @@ protected void generateModelClassProto(Class<?> modelClazz, String targetDirecto
...
 
             String processId = generatedData.reference();            
-            Proto modelProto = generate(modelClazz.getPackage().getName() + "." + processId, modelClazz, "import \"kogito-index.proto\";", 
+            Proto modelProto = generate("@Indexed",
+                                        "@Field(store = Store.YES, analyze = Analyze.YES)",
+                                        modelClazz.getPackage().getName() + "." + processId, modelClazz, "import \"kogito-index.proto\";",                                        
                                         "import \"kogito-types.proto\";", 
-                                        "option kogito_model = \"" + generatedData.name() +"\";");
+                                        "option kogito_model = \"" + generatedData.name() +"\";", 
+                                        "option kogito_id = \"" + processId +"\";");
```
cristianonicolai: nice, I wonder if we actually need both options. Perhaps only adding _kogito_id_ to the message itself is enough to tell that this is the model message. And also allows that we have multiple process into the same proto file. wdyt?

addons/persistence/infinispan-persistence-addon/src/main/java/org/kie/kogito/infinispan/CacheProcessInstances.java
mswiderski: Would be good as it would then show the ownership directly

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/persistence/proto/ReflectionProtoGenerator.java
mswiderski: I would keep it both for now as it gives more clarity without inspecting individual messages. And I don’t think we want to have multiple domains in single Proto file, do we?

cristianonicolai: definitely not multiple domains, but if we have multiple processes that share the same domain, that would be easier to maintain, as we could use just a single proto file. Now we need one proto per public process.

addons/persistence/infinispan-persistence-addon/src/main/java/org/kie/kogito/infinispan/CacheProcessInstances.java
cristianonicolai: will do

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/persistence/proto/ReflectionProtoGenerator.java
mswiderski: that should be considered as domain itself, public process is like a domain ... just to ensure it has business meaning. Kind of like we have the case definition.

PR #117, opened 2019-08-19
KOGITO-143 - Support error state as process instance state with optio…

mswiderski:
…n to retrigger or skipped

depends on https://github.com/kiegroup/kogito-bom/pull/46

### DIFF
--- a/addons/persistence/infinispan-persistence-addon/src/test/java/org/kie/kogito/infinispan/MockCacheProcessInstancesTest.java
+++ b/addons/persistence/infinispan-persistence-addon/src/test/java/org/kie/kogito/infinispan/MockCacheProcessInstancesTest.java
@@ -15,12 +15,11 @@
 
 package org.kie.kogito.infinispan;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.kie.api.runtime.process.ProcessInstance.STATE_ACTIVE;
 import static org.kie.api.runtime.process.ProcessInstance.STATE_COMPLETED;
+import static org.kie.api.runtime.process.ProcessInstance.STATE_ERROR;
 
 import static org.mockito.Mockito.*;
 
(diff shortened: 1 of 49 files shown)


addons/process-management-addon/src/test/java/org/kie/kogito/process/management/ProcessInstanceManagementResourceTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+    }
+    
+    @Test
+    public void testRetriggerErrorInfo() {
+        
+        ProcessInstanceManagementResource resource = new ProcessInstanceManagementResource();
+        resource.processes = this.processes;
+        
+        doAnswer(new Answer<Void>() {
```
cristianonicolai: just a coding style preference but you can shorten these statements using lambda style.

api/kogito-api/src/main/java/org/kie/kogito/process/MutableProcessInstances.java
```diff
@@ -22,6 +22,6 @@
     void remove(String id);
     
     default boolean isActive(ProcessInstance<T> instance) {
-        return instance.status() == ProcessInstance.STATE_ACTIVE;        
+        return instance.status() == ProcessInstance.STATE_ACTIVE || instance.status() == ProcessInstance.STATE_ERROR;
```
cristianonicolai: so this means that error state is only when the process instance is active, meaning no end state for a process as in error, it gets moved to aborted or completed, makes sense?

kogito-codegen/src/test/java/org/kie/kogito/codegen/data/HelloService.java
```diff
@@ -20,11 +20,11 @@
 	
     public String hello(String name) {
         System.out.println("Service invoked with " + name + " on service " + this.toString());
-        return "Hello " + name + "!";
+        return "Hello " + name.toString() + "!";
```
cristianonicolai: shouldn't need these .toString right?

addons/persistence/infinispan-persistence-addon/src/test/java/org/kie/kogito/infinispan/MockCacheProcessInstancesTest.java
```diff
@@ -15,27 +15,34 @@
 
 package org.kie.kogito.infinispan;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.assertj.core.api.Assertions.assertThat;
```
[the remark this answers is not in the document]
mswiderski: it's more powerful and just easier to follow ...or it's a matter of getting used to it.

addons/process-management-addon/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+    </dependency>
+
+    <dependency>
+      <groupId>org.jboss.spec.javax.ws.rs</groupId>
+      <artifactId>jboss-jaxrs-api_2.1_spec</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>javax.inject</groupId>
+      <artifactId>javax.inject</artifactId>
```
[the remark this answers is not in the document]
mswiderski: if it is provided then it does not work with spring

addons/process-management-addon/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+    </dependency>
+
+    <!-- test dependencies -->
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-engine</artifactId>
```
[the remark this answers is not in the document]
mswiderski: I believe so as I copied it from another module :)

api/kogito-api/src/main/java/org/kie/kogito/process/MutableProcessInstances.java
mswiderski: correct, error state means the instance failed at execution and did not reached its final state. So it should be saved and later on fixed and retried.

kogito-codegen/src/test/java/org/kie/kogito/codegen/data/HelloService.java
mswiderski: actually it is needed to cause the failure :) if I give name as null it will fail ... otherwise it won't ;)

PR #118, opened 2019-08-19
KOGITO-35 - Kogito Data Index - Initial commit

cristianonicolai:
Depends on https://github.com/kiegroup/kogito-bom/pull/47


data-index/data-index-service/src/main/java/org/kie/kogito/index/event/KogitoCloudEvent.java
```diff
@@ -0,0 +1,243 @@
...
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public String getSpecVersion() {
+        return specVersion;
+    }
+
+    public void setSpecVersion(String specVersion) {
```
mswiderski: do we need setters if we have the builder?

data-index/data-index-service/src/main/java/org/kie/kogito/index/graphql/JsonPropertyDataFetcher.java
```diff
@@ -0,0 +1,53 @@
...
+            try {
+                JsonValue value = jsonObject.getValue(jsonPointer);
+                switch (value.getValueType()) {
+                    case OBJECT:
+                        return value.asJsonObject();
+                    case ARRAY:
+                        return value.asJsonArray();
+                    case NUMBER:
+                        return ((JsonNumber) value).numberValue();
+//              TODO: Confirm if booleans as returned as String or Number
```
mswiderski: so was it confirmed? ;)

data-index/data-index-service/src/main/java/org/kie/kogito/index/messaging/MessageEmitter.java
```diff
@@ -0,0 +1,28 @@
...
+
+import io.quarkus.vertx.web.Route;
+import io.smallrye.reactive.messaging.annotations.Emitter;
+import io.smallrye.reactive.messaging.annotations.Stream;
+import io.vertx.core.http.HttpMethod;
+import io.vertx.ext.web.RoutingContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class MessageEmitter {
```
mswiderski: is this test kind of emitter? I guess data index is just consumer and not producer of any events...

data-index/data-index-service/src/main/java/org/kie/kogito/index/service/IndexingService.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+        manager.getProcessInstancesCache().put(event.getProcessInstanceId(), pi);
+    }
+
+    public void indexProcessInstanceModel(KogitoCloudEvent event) {
+        String processId = event.getRootProcessId() == null ? event.getProcessId() : event.getRootProcessId();
+        String type = getModelFromProcessId(processId);
+        if (type == null) {
+//          Unknown process type, ignore
+            LOGGER.warn("Ignoring Kogito cloud event for Process Id: {}", event.getProcessId());
```
mswiderski: maybe change it to debug instead of warning as it is not really a warning when we deal with private processes

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/IndexingRouter.java
```diff
@@ -0,0 +1,41 @@
...
+import io.quarkus.vertx.web.Route;
+import io.vertx.core.http.HttpMethod;
+import io.vertx.ext.web.RoutingContext;
+import org.kie.kogito.index.cache.CacheService;
+import org.kie.kogito.index.event.KogitoCloudEvent;
+import org.kie.kogito.index.service.IndexingService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class IndexingRouter {
```
mswiderski: another test related router?

data-index/data-index-service/src/main/resources/application.properties
```diff
@@ -0,0 +1,22 @@
+# Quarkus
+quarkus.log.level=INFO
+quarkus.log.console.enable=true
+quarkus.log.console.level=DEBUG
+quarkus.log.category."org.kie.kogito".level=DEBUG
+#quarkus.log.category."io.vertx".level=DEBUG
+quarkus.log.category."graphql".level=DEBUG
```
mswiderski: maybe change to info by default

data-index/data-index-service/src/main/resources/application.properties
```diff
@@ -0,0 +1,22 @@
...
+# Infinispan
+quarkus.infinispan-client.server-list=localhost:11222
+
+# Kafka
+mp.messaging.incoming.processinstance.connector=smallrye-kafka
+mp.messaging.incoming.processinstance.value.deserializer=org.kie.kogito.index.messaging.KogitoCloudEventDeserializer
+
+mp.messaging.outgoing.emitter.connector=smallrye-kafka
+mp.messaging.outgoing.emitter.topic=kogito-processinstances-events
+mp.messaging.outgoing.emitter.value.serializer=org.apache.kafka.common.serialization.StringSerializer
```
mswiderski: these are test related only, right?

data-index/data-index-service/src/main/resources/schema.graphqls
```diff
@@ -0,0 +1,55 @@
...
+  state: ProcessInstanceState!
+  endpoint: String!
+}
+
+enum ProcessInstanceState {
+  PENDING,
+  ACTIVE,
+  COMPLETED,
+  ABORTED,
+  SUSPENDED
```
mswiderski: I have an open PR to add ERROR state as well.

data-index/data-index-storage/data-index-storage-api/pom.xml
```diff
@@ -0,0 +1,25 @@
...
+    <artifactId>data-index-storage</artifactId>
+    <version>8.0.0-SNAPSHOT</version>
+  </parent>
+  <artifactId>data-index-storage-api</artifactId>
+  <name>Kogito :: Data Index Storage API</name>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.glassfish</groupId>
+      <artifactId>javax.json</artifactId>
```
mswiderski: why do we need to have explicit depenency to `org.glassfish`?

data-index/data-index-storage/data-index-storage-infinispan/src/main/java/org/kie/kogito/index/infinispan/protostream/ProtoDomainModelProducer.java
```diff
@@ -0,0 +1,90 @@
...
+                case INT:
+                    return Integer.class.getName();
+                case LONG:
+                    return Long.class.getName();
+                case BOOLEAN:
+                    return Boolean.class.getName();
+                case MESSAGE:
+                    return fd.getMessageType().getFullName();
+                case STRING:
+                    //TODO Missing to map more types
```
mswiderski: shouldn't string still return `String.class.getName()`

pom.xml
```diff
@@ -57,8 +57,29 @@
...
+      </dependency>
+      <dependency>
+        <groupId>org.kie.kogito</groupId>
+        <artifactId>data-index-storage-api</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.kie.kogito</groupId>
+        <artifactId>data-index-storage-infinispan</artifactId>
+        <version>${project.version}</version>
```
mswiderski: these should be moved to kogito-bom

data-index/data-index-storage/data-index-storage-infinispan/src/main/java/org/kie/kogito/index/infinispan/protostream/ProtobufService.java
```diff
@@ -0,0 +1,190 @@
...
+
+    @Inject
+    CacheService manager;
+
+    @Inject
+    FileDescriptorSource kogitoDescriptors;
+
+    @Inject
+    @ConfigProperty(name = "kogito.protobuf.folder")
+    Optional<String> protoFiles;
```
mswiderski: so there is no watcher service attached to the folder, right?

data-index/data-index-service/src/main/resources/application.properties
```diff
@@ -0,0 +1,22 @@
...
+# Vert.x
+quarkus.vertx-http.port=8180
+quarkus.vertx-http.test-port=8181
+
+# Infinispan
+quarkus.infinispan-client.server-list=localhost:11222
+
+# Kafka
+mp.messaging.incoming.processinstance.connector=smallrye-kafka
+mp.messaging.incoming.processinstance.value.deserializer=org.kie.kogito.index.messaging.KogitoCloudEventDeserializer
```
mswiderski: to make it work out of the box I had to replace above two lines with

```
mp.messaging.incoming.kogito-processinstances-events.connector=smallrye-kafka
mp.messaging.incoming.kogito-processinstances-events.topic=kogito-processinstances-events
mp.messaging.incoming.kogito-processinstances-events.value.deserializer=org.kie.kogito.index.messaging.KogitoCloudEventDeserializer
```

data-index/data-index-service/src/main/java/org/kie/kogito/index/messaging/MessageEmitter.java
cristianonicolai: correct, just for testing

data-index/data-index-service/src/main/java/org/kie/kogito/index/event/KogitoCloudEvent.java
cristianonicolai: I see the builder just as a fluent way to build the object, some of the setters still have logic in it, so not necessarily a replacement.

data-index/data-index-service/src/main/java/org/kie/kogito/index/graphql/JsonPropertyDataFetcher.java
cristianonicolai: not yet hehe

cristianonicolai: fixed.

data-index/data-index-service/src/main/java/org/kie/kogito/index/service/IndexingService.java
cristianonicolai: done

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/IndexingRouter.java
cristianonicolai: yep, need to integrate messaging for testing this

data-index/data-index-service/src/main/resources/application.properties
cristianonicolai: +1

data-index/data-index-service/src/main/resources/application.properties
cristianonicolai: yeah for the test emitter

data-index/data-index-service/src/main/resources/schema.graphqls
cristianonicolai: added the new status here as well.

data-index/data-index-storage/data-index-storage-api/pom.xml
cristianonicolai: moved to javax.json

data-index/data-index-storage/data-index-storage-infinispan/src/main/java/org/kie/kogito/index/infinispan/protostream/ProtoDomainModelProducer.java
cristianonicolai: it will as it falls under the default anyway.

data-index/data-index-storage/data-index-storage-infinispan/src/main/java/org/kie/kogito/index/infinispan/protostream/ProtobufService.java
cristianonicolai: yep, will handle that in a separated PR

pom.xml
cristianonicolai: done

data-index/data-index-service/src/main/resources/application.properties
cristianonicolai: hm right, I renamed I didnt update it here, just weird that it needs to set the topic name. anyway will update to match yours.

cristianonicolai:
@mswiderski regarding rootProcessInstanceId, would that an issue with the capital _I_? I got the mapping with all lowercase, see: https://github.com/cristianonicolai/kogito-runtimes/blob/index/data-index/data-index-service/src/main/java/org/kie/kogito/index/event/KogitoCloudEvent.java#L46

cristianonicolai:
the issue with using kogito.Date is that it needs to be encoded as a type and not as simple String in the JSON level. I would have to reformat the variables json to handle that before pushing into the domain cache. If we could use simple string type ( in the generated proto ) for now would be best in my opinion.

cristianonicolai:
@mswiderski to give you an example, this is how the JSON would need to look like: https://gist.github.com/cristianonicolai/c3449626ca2da52fb4176b9ce8327443
For every custom type we add, the value in json needs to encapsulate a "data" field, which maps to the attribute in the proto type;

data-index/data-index-storage/data-index-storage-infinispan/src/main/java/org/kie/kogito/index/infinispan/protostream/ProtoDomainModelProducer.java
mswiderski: then what's the point of `case STRING` then? if it goes into `default`

data-index/data-index-storage/data-index-storage-infinispan/src/main/java/org/kie/kogito/index/infinispan/protostream/ProtobufService.java
mswiderski: perfect

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/IndexingRouter.java
mswiderski: do we want to have it in the code base or we just need to keep it for now and will be removed later on?

data-index/data-index-service/src/main/java/org/kie/kogito/index/event/KogitoCloudEvent.java
mswiderski: true, though the issue with setters is that the instance becomes mutable so might not be what we want to achieve with builder

data-index/data-index-service/src/main/resources/application.properties
mswiderski: the topic is kind of optional but allows easier abstraction and increases clarity in my opinion. The name that is referenced in code can't be easily changed but the topic could be

data-index/data-index-service/src/main/java/org/kie/kogito/index/vertx/IndexingRouter.java
cristianonicolai: my plan is to remove later on, once I find a better way to cover the end-to-end testing. That would require having the messaging setup now.

mswiderski: fair enough

data-index/data-index-service/src/main/java/org/kie/kogito/index/event/KogitoCloudEvent.java
cristianonicolai: true, I would use the constructor instead for that, but the builder would have to hold the value for all fields. Perhaps moving setter to package level instead of public would be a good compromise?

mswiderski: +1 for package level

cristianonicolai: jsob marshalling did not like the package level, will keep it public for now.

mswiderski:
@cristianonicolai works almost perfect, last remaining issue is with rootProcessinstanceid ... it's still null for me even though I see it in the event

```
  "id": "b50a3dd4-a955-49c4-a055-f30a4d768f87",
  "kogitoParentProcessinstanceId": "f8868a2e-1bbb-47eb-93cf-fa46ff9dbfee",
  "kogitoProcessId": "hotelBooking",
  "kogitoProcessinstanceId": "c2fa5c5e-3002-44c7-aef7-bce82297e3fe",
  "kogitoProcessinstanceState": "2",
  "kogitoRootProcessId": "travels",
  "kogitoRootProcessinstanceId": "f8868a2e-1bbb-47eb-93cf-fa46ff9dbfee",
  "specversion": "0.3",
  "time": "2019-08-20T21:26:02.110668+02:00[Europe/Warsaw]",
  "type": "ProcessInstanceEvent"
```

and noticed that there is no parentProcessinstanceId defined in the process instance at all, is this on purpose? I believe we need to have both parent and root instance ids

cristianonicolai:
hm ok, I missed the kogitoParentProcessinstanceId.

PR #122, opened 2019-08-21
Rename Exteption

ruromero:
Rename ProcessInstanceNotFoundException and ProcessInstanceExecutionException classes

Signed-off-by: Ruben Romero Montes <rromerom@redhat.com>

### DIFF
--- a/drools/drools-core/src/main/java/org/drools/core/addon/ClassTypeResolver.java
+++ b/drools/drools-core/src/main/java/org/drools/core/addon/ClassTypeResolver.java
@@ -30,7 +30,7 @@ public class ClassTypeResolver
 
     private String defaultPackagName;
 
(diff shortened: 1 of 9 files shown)


drools/drools-core/src/main/java/org/drools/core/addon/ClassTypeResolver.java
```diff
@@ -30,7 +30,7 @@
 
     private String defaultPackagName;
 
-    private Set<String> imports = Collections.emptySet();
+    private Set<String> imports;
```
mswiderski: why this is needed here?

ruromero: It is not related to this renaming but it seems it's just not needed. It will always be set in the constructor.

PR #123, opened 2019-08-21
KOGITO-166 - SpringBoot archetype does not work with custom groupId

mswiderski:
@krisv could you please have a look?

### DIFF
--- a/archetypes/kogito-springboot-archetype/src/main/resources/archetype-resources/src/main/java/KogitoApplication.java
+++ b/archetypes/kogito-springboot-archetype/src/main/resources/archetype-resources/src/main/java/KogitoApplication.java
@@ -3,7 +3,7 @@ package ${package};
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 
-@SpringBootApplication(scanBasePackages={"${package}.**"})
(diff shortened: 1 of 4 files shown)


archetypes/kogito-springboot-archetype/src/main/resources/archetype-resources/src/main/java/KogitoApplication.java
```diff
@@ -3,7 +3,7 @@
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 
-@SpringBootApplication(scanBasePackages={"${package}.**"})
+@SpringBootApplication(scanBasePackages={"${package}.**", "${groupId}.**"})
```
krisv: It seems this now generates:
@SpringBootApplication(scanBasePackages={"com.company.**", "com.company.**"})
It seems other updates in master probably made sure package was updated to use com.company as well.
Anyway, this works, not sure that package and groupId are ever expected to be different or not.

mswiderski: Yes, they can be different as archetype allows you to specify package separately

PR #124, opened 2019-08-21
KOGITO-102: jBPM Rule Unit integration 

evacchi:
Initial implementation for runtime logic of https://github.com/kiegroup/kogito-examples/pull/72. this is still **WIP**

- refactored BusinessRuleTask implementation so that it includes a RuleType, which may be RuleFlowGroup, RuleUnit, Decision, with respective parameters.

- added code-generation:
  ```java
        org.jbpm.ruleflow.core.factory.RuleSetNodeFactory ruleSetNode2 = factory.ruleSetNode(2);
        ruleSetNode2.name("Evaluate person");
        // no longer a lambda, the RuleUnit is given
        ruleSetNode2.ruleUnit("PersonValidationService", 
              app.ruleUnits().createPersonValidationServiceRuleUnit());
        ruleSetNode2.inMapping("persons", "person");
        ruleSetNode2.outMapping("persons", "person");
        ...
        ruleSetNode2.done();
  ```
- added handling logic for runtime
  ```
            } else if (ruleType.isRuleUnit()) {
                RuleUnitMemory> workingMemory = /* reflective magic to instantiate: we need to improve this */;
                RuleUnit<RuleUnitMemory> ruleUnit = ruleSetNode.getRuleUnit();
                RuleUnitInstance<RuleUnitMemory> instance = ruleUnit.createInstance(workingMemory);
                instance.fire();
                triggerCompleted();
   ```

the following rules for i/o parameter mapping currently apply:

Mapping | Process Variable | Rule Unit field | action
--------|------------------|-----------------|----------------------------------------------- 
IN      | scalar           | scalar          | assignment
IN      | scalar           | data source     | add to (i.e. insert into) data source
IN      | collection       | data source     | add (assignment would probably make more sense)
OUT     | scalar           | scalar          | assignment
OUT     | scalar           | data source     | get 1 value off the data source
OUT     | collection       | data source     | get 1 value off the data source (would probably make more sense to collect into the collection)

### DIFF
--- a/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/RuleSetNodeVisitor.java
+++ b/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/RuleSetNodeVisitor.java
@@ -81,35 +81,31 @@ public class RuleSetNodeVisitor extends AbstractVisitor {
         RuleSetNode.RuleType ruleType = ruleSetNode.getRuleType();
 
         if (ruleSetNode.getLanguage().equals(RuleSetNode.DRL_LANG)) {
-            if (ruleType.isRuleFlowGroup()) {
-                MethodCallExpr ruleRuntimeBuilder = new MethodCallExpr(
-                        new MethodCallExpr(new NameExpr("app"), "ruleUnits"), "ruleRuntimeBuilder");
-                MethodCallExpr ruleRuntimeSupplier = new MethodCallExpr(ruleRuntimeBuilder, "newKieSession", NodeList.nodeList(new StringLiteralExpr("defaultStatelessKieSession"), new NameExpr("app.config().rule()")));
-                actionBody.addStatement(new ReturnStmt(ruleRuntimeSupplier));
-                addFactoryMethodWithArgs(body, "ruleSetNode" + node.getId(), "ruleFlowGroup", new StringLiteralExpr(ruleType.getName()), lambda);
-            } else if (ruleType.isRuleUnit()) {
+            MethodCallExpr ruleRuntimeBuilder = new MethodCallExpr(
+                    new MethodCallExpr(new NameExpr("app"), "ruleUnits"), "ruleRuntimeBuilder");
+            MethodCallExpr ruleRuntimeSupplier = new MethodCallExpr(ruleRuntimeBuilder, "newKieSession", NodeList.nodeList(new StringLiteralExpr("defaultStatelessKieSession"), new NameExpr("app.config().rule()")));
+            actionBody.addStatement(new ReturnStmt(ruleRuntimeSupplier));
+            addFactoryMethodWithArgs(body, "ruleSetNode" + node.getId(), "ruleFlowGroup", new StringLiteralExpr(ruleType.getName()), lambda);
+        } else if (ruleSetNode.getLanguage().equals(RuleSetNode.RULE_UNIT_LANG)) {
+            InputStream resourceAsStream = this.getClass().getResourceAsStream("/class-templates/RuleUnitFactoryTemplate.java");
+            Expression ruleUnitFactory = parse(resourceAsStream).findFirst(Expression.class).get();
 
-                InputStream resourceAsStream = this.getClass().getResourceAsStream("/class-templates/RuleUnitFactoryTemplate.java");
-                Expression ruleUnitFactory = parse(resourceAsStream).findFirst(Expression.class).get();
+            String unitName = ruleType.getName();
+            Class<?> unitClass = loadUnitClass(unitName);
 
-                String unitName = ruleType.getName();
-                Class<?> unitClass = loadUnitClass(unitName);
+            ruleUnitFactory.findAll(ClassOrInterfaceType.class)
+                    .stream()
+                    .filter(t -> t.getNameAsString().equals("$Type$"))
+                    .forEach(t -> t.setName(unitName));
 
-                ruleUnitFactory.findAll(ClassOrInterfaceType.class)
-                        .stream()
-                        .filter(t -> t.getNameAsString().equals("$Type$"))
-                        .forEach(t -> t.setName(unitName));
+            ruleUnitFactory.findFirst(MethodDeclaration.class, m -> m.getNameAsString().equals("bind"))
+                    .ifPresent(m -> m.setBody(bind(variableScope, ruleSetNode, unitClass)));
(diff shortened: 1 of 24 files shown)


mswiderski:
@evacchi so effectively we have now three types of "rules" supported in business rule task
- rule flow group
- dmn
- rule units
is this correct?

Overall looks good to me, just this reflection stuff to get the working memory. Can't we get it from the rule unit itself?

evacchi:
I think we should be able to pre-generate everything at compile time, will work on it

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/RuleSetNodeVisitor.java
```diff
@@ -74,8 +122,94 @@ public void visitNode(String factoryField, Node node, BlockStmt body, VariableSc
...
+                new ObjectCreationExpr().setType(unitName),
+                AssignExpr.Operator.ASSIGN);
+
+        BlockStmt actionBody = new BlockStmt();
+        actionBody.addStatement(assignExpr);
+
+        try {
+
+            // warning: load class here
+            Class<?> unitClass = contextClassLoader.loadClass(unitName);
```
mswiderski: if the rule unit comes from the same project this might not work in quarkus extensions as the class might not be there since it is generated, isn't it?

evacchi: in the current framework, the rule unit class must always be user-provided, because it defines the shape (data sources) of the working memory

evacchi: in current framework, users must always provide a class because it is used to define the shape of the working memory (data sources)

mswiderski: so that means the class will be already available on build classpath ... is this the case for quarkus extension as it mainly relies on jandex index.

evacchi: I think we should be safe enough as long as we set correctly the class loader, but that's certainly a good point. let's keep an eye on this. I'll open a JIRA for quarkus integration when this gets merged

evacchi:
rebased over https://github.com/kiegroup/kogito-runtimes/pull/128

evacchi:
updated first post with details on how parameter mapping is currently working

evacchi:
code is still pretty rough/duplicated, so I'll open a JIRA for refactoring; however: tests pass. I'd say we could merge it, WDYT? @mswiderski @mariofusco

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/node/RuleUnitFactory.java
```diff
@@ -0,0 +1,11 @@
+package org.jbpm.workflow.core.node;
```
mswiderski: missing license header

kogito-codegen/src/test/java/org/kie/kogito/codegen/tests/BusinessRuleUnit.java
```diff
@@ -0,0 +1,14 @@
+package org.kie.kogito.codegen.tests;
```
mswiderski: same here .. license header missing

kogito-codegen/src/test/resources/org/kie/kogito/codegen/tests/BusinessRuleUnit.bpmn2
```diff
@@ -0,0 +1,84 @@
...
+    <bpmn2:endEvent id="EndEvent_1">
+      <bpmn2:extensionElements>
+        <tns:metaData name="elementname">
+          <tns:metaValue><![CDATA[]]></tns:metaValue>
+        </tns:metaData>
+      </bpmn2:extensionElements>
+      <bpmn2:incoming>SequenceFlow_1</bpmn2:incoming>
+    </bpmn2:endEvent>
+    <bpmn2:sequenceFlow id="SequenceFlow_2" tns:priority="1" name="" sourceRef="StartEvent_1" targetRef="BusinessRuleTask_2"/>
+    <bpmn2:businessRuleTask id="BusinessRuleTask_2" tns:ruleFlowGroup="unit:org.kie.kogito.codegen.tests.BusinessRuleUnit" name="Business Rule Task" implementation="##unspecified">
```
mswiderski: just wondering if we should not use the implementation attribute of the rule task element instead of prefixes inside the ruleFlowGroup, wdyt?

kogito-codegen/src/test/resources/org/kie/kogito/codegen/tests/BusinessRuleUnit.drl
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+package org.kie.kogito.codegen.tests;
+unit BusinessRuleUnit;
```
mswiderski: so unit:FQCN is composed of package + unit from drl, right?

evacchi: yes. It's actually the FQCN of the class found in src/main/java

kogito-codegen/src/test/resources/org/kie/kogito/codegen/tests/BusinessRuleUnit.bpmn2
evacchi: oooh, me likey. Is it used for anything already, or is it free? if it's free then it would be a veeeerrrry good fit

mswiderski: I don't think it's being used right now, so go ahead for it... we could actually skip the ruleFlowGroup completely and use the implementation only ... show this will have impact on the editor ... right now it uses ruleFlowGroup and language ...

evacchi:
ok, let's wait for CI and see

evacchi:
jenkins please retest this

PR #128, opened 2019-08-22
KOGITO-174: drop "explicit" factory methods for processes and rules

evacchi:
define methods (see also [JIRA](https://issues.jboss.org/browse/KOGITO-174))

```
interface RuleUnits {
  ...
  RuleUnit<RuleUnitMemory> findById(String ruleUnitCanonicalName)
   <T extends RuleUnitMemory> RuleUnit<T> RuleUnits.create(Class<T> ruleUnitMemory)
}
```

(same can be done for processes)

and Application load stub, thus it is possible to write:

```
public class Blah {
    public static void main(String[] args) {
        Application application = Application.create();
        RuleUnit<LoggerService> loggerServiceRuleUnit =
                application.ruleUnits().create(LoggerService.class);
         ...
    }
}
```

if CDI or Spring are present, then the usual injection mechanism can be used.


api/kogito-api/src/main/java/org/kie/kogito/Application.java
```diff
@@ -28,6 +31,16 @@
  */
 public interface Application {
 
+    static Application create() {
+        try {
+            Class<Application> appClass = (Class<Application>) Class.forName("org.kie.kogito.$ApplicationStub$");
+            Method create = appClass.getMethod("create");
+            return (Application) create.invoke(null);
+        } catch (Exception e) {
```
mswiderski: why do we need this? can't we generate it with direct use of the API instead of reflection?

kogito-codegen/src/main/java/org/kie/kogito/codegen/ApplicationGenerator.java
```diff
@@ -213,6 +217,18 @@ public ApplicationGenerator withPersistence(boolean persistence) {
...
     }
 
+    public GeneratedFile generateApplicationStub() {
+        CompilationUnit compilationUnit =
+                parse(this.getClass().getResourceAsStream(RESOURCE_STUB));
+
+        compilationUnit.findFirst(ObjectCreationExpr.class).map(oce -> oce.setType(targetCanonicalName));
+
+        return new GeneratedFile(GeneratedFile.Type.CLASS,
+                                 "org/kie/kogito/$ApplicationStub$.java",
```
mswiderski: will this `$ApplicationStub$` be somehow resolved to actual class name

evacchi: I actually wanted an ugly name, because that's not supposed to invoked by user code, it's just the stub that's invoked by the static method

api/kogito-api/src/main/java/org/kie/kogito/Application.java
evacchi: We *could* avoid it by writing a stub to link against during compilation, and then remove it from the generated artifact. I suppose there should be a way to do it.

api/kogito-api/pom.xml
```diff
@@ -20,6 +20,13 @@
   </properties>
 
   <dependencies>
+    <dependency>
+      <groupId>org.kie.kogito</groupId>
+      <artifactId>kogito-stubs</artifactId>
+      <version>8.0.0-SNAPSHOT</version>
+      <scope>provided</scope>
+    </dependency>
```
evacchi: if we like this approach this will go in the bom instead

api/kogito-stubs/src/main/java/org/kie/kogito/application/stubs/ApplicationStub.java
```diff
@@ -0,0 +1,9 @@
+package org.kie.kogito.application.stubs;
+
+public class ApplicationStub {
+    public Object create() {
+        throw new NoClassDefFoundError(
+                "The Java linker has picked up the default ApplicationStub. " +
+                        "This should not happen: the default ApplicationStub is excluded from the Jar.");
+    }
```
evacchi: the method has to return Object to avoid a circular dependency

kogito-codegen/src/main/resources/class-templates/ApplicationStubTemplate.java
```diff
@@ -1,9 +1,9 @@
-package org.kie.kogito;
+package org.kie.kogito.application.stubs;
 
 import org.kie.kogito.Application;
 
-class $ApplicationStub$ {
-    public static final Application create() {
+public class ApplicationStub {
+    public Object create() {
```
evacchi: the method of the template must still return Object to avoid a linking error (NoSuchMethodError). However, this is transparent to the user, because they will only use `Application.create()`.

mswiderski:
@evacchi overall I like the idea of unifying the Processes and RUleUnits to be able to load by id. Though just wondering how much value the `Application.create()` has as in most of the cases users will use either Arc or Spring dependency injection. Would it make sense to avoid this `ApplicationStub` and just add support for `RuleUnits.findById`?

evacchi:
@mswiderski ah, sure, that can be done in a separate PR if it makes sense

evacchi:
if we all agree, we can merge this. @mariofusco ?

kogito-codegen/src/main/java/org/kie/kogito/codegen/ApplicationGenerator.java
```diff
@@ -61,6 +63,7 @@
     private static final Logger logger = LoggerFactory.getLogger(ApplicationGenerator.class);
 
     private static final String RESOURCE = "/class-templates/ApplicationTemplate.java";
+    private static final String RESOURCE_STUB = "/class-templates/ApplicationStubTemplate.java";
```
mswiderski: I guess this is not needed anymore

PR #129, opened 2019-08-23
KOGITO-168 - Data Index developer docs

cristianonicolai:


data-index/data-index-service/README.md
```diff
@@ -0,0 +1,139 @@
+# Instructions for developers
+
+Data Index service is a Quarkus based application that aims to consume [CloudEvents](https://cloudevents.io) based messages from Kogito runtimes, process and index the information for later consumption via GraphQL queries.
+These events contains information about units of work executed for a process. Visit [process-runtime-events](https://github.com/kiegroup/kogito-runtimes/wiki/Configuration#process-runtime-events) for more details about the payload of the messages.
+This data is then parsed and pushed into different Infinispan caches.
+These caches are structured as follows:
+ - Domain cache: This cache is a generic cache, one per process id, where the process instance variables are pushed as the root content. This cache also includes
```
mswiderski: > This cache is a generic cache

is this really a generic cache, I see it more as specific to given domain/process rather than generic

data-index/data-index-service/README.md
```diff
@@ -0,0 +1,139 @@
+# Instructions for developers
+
+Data Index service is a Quarkus based application that aims to consume [CloudEvents](https://cloudevents.io) based messages from Kogito runtimes, process and index the information for later consumption via GraphQL queries.
+These events contains information about units of work executed for a process. Visit [process-runtime-events](https://github.com/kiegroup/kogito-runtimes/wiki/Configuration#process-runtime-events) for more details about the payload of the messages.
```
mswiderski: > about units of work executed for a process

it can contain multiple process instances, though each process instance will have its own event

data-index/data-index-service/README.md
```diff
@@ -0,0 +1,139 @@
...
+Once the service is up and running, it will start consuming messages from the  Kafka topic named: `kogito-processinstances-events`.
+For more details on how to enable a Kogito runtime to produce events, please visit: https://github.com/kiegroup/kogito-runtimes/wiki/Configuration#publishing-events
+
+When running on dev mode, the [GraphiQL](https://github.com/graphql/graphiql) UI is available, on http://localhost:8180/, which allows
+exploring and querying the available data model.
+
+In there you can explore the current types available using the Docs section on the top right and execute queries on the model.
+Some examples:
+
+### Querying the technical cache
```
mswiderski: maybe add some examples with filtering

data-index/data-index-service/README.md
```diff
@@ -0,0 +1,139 @@
...
+      id
+      name
+      type
+    }
+  }
+}
+```  
+
+### Querying the domain cache
+Assuming a Travels model is deployed
```
mswiderski: same here, can we add some filtering like show travels to US only ...?

data-index/data-index-service/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+Messaging integration is also provided by Quarkus using `quarkus-smallrye-reactive-messaging-kafka`. For more details and configution details, visit
+https://quarkus.io/guides/kafka-guide and https://smallrye.io/smallrye-reactive-messaging/.
+
+## Prerequisites
+
+You will need:
+ - Infinispan server:
+   Download and run https://downloads.jboss.org/infinispan/10.0.0.Beta4/infinispan-server-10.0.0.Beta4.zip.
+ Service will be available on port 11222. This should match the setting `quarkus.infinispan-client.server-list=localhost:11222` on `application.properties`.
```
mswiderski: I believe this should be overridden on startup with `-Dquarkus.infinispan-client.server-list=anotherhost:11222`

is that correct?

data-index/data-index-service/README.md
```diff
@@ -0,0 +1,139 @@
...
+https://quarkus.io/guides/kafka-guide and https://smallrye.io/smallrye-reactive-messaging/.
+
+## Prerequisites
+
+You will need:
+ - Infinispan server:
+   Download and run https://downloads.jboss.org/infinispan/10.0.0.Beta4/infinispan-server-10.0.0.Beta4.zip.
+ Service will be available on port 11222. This should match the setting `quarkus.infinispan-client.server-list=localhost:11222` on `application.properties`.  
+
+ - Kafka messaging server:
```
mswiderski: maybe add info on how to point to external (non localhost) Kafka? I guess there is so mp... property to be defined

data-index/data-index-service/README.md
cristianonicolai: only if needed right.

data-index/data-index-service/README.md
cristianonicolai: +1

data-index/data-index-service/README.md
cristianonicolai: +1

cristianonicolai:
will add content to wiki, closing PR.

PR #136, opened 2019-08-28
KOGITO-216 Rename into kogito-dmn

tarilabs:
Requires

https://github.com/kiegroup/kogito-bom/pull/49


drools/kogito-dmn/src/main/java/org/kie/kogito/dmn/rest/DMNDecisionInfo.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package org.kie.dmn.kogito.rest.quarkus;
+package org.kie.kogito.dmn.rest;
```
mswiderski: are these classes really rest specific? maybe the rest package is not really needed here...

tarilabs: they are indeed REST specific, they are only used when a DMN evaluation is happening over REST.

mswiderski:
one more thing I noticed that you use JSONB annotations, have you tried it out with spring boot runtime to see if that still works?

maybe skip the annotations completely and rely on field names instead... this will certainly help to avoid issues in consuming JSON on client side, for example `model.namespace`  instead of what is now `model.model-namespace`

tarilabs:
unfortunately I cannot rely on field names, as those are annotations are needed to ensure consistency between XML and JSON serialization, as well as with the previous payload _result_ from kie-server (I mean the wrapper object which list result context, result decision messages, result decision result wrapper object)

But I take your point, I will make a task to check it eventually with spring boot

mswiderski:
just so you know that having `-` in json names is of a huge pain for (JS) developers as you can't access them directly. So you have to use constructs like `model['model-namespace']` kind of thing. 
Personally I think that goal of having xml and json payload unified was the worse decision we made in Kie Server. it has practically no value as the payload is different anyway. I'd say you should reconsider this in context of Kogito and make it more JSON natural

tarilabs:
thanks for the pointers @mswiderski : I will definitely make use of them as I'll pitch this to @etirelli ;)

tarilabs:
p.s.: for the reference, my own take of the "worse decision" was that we had ultra-generics payload, but that is already being "fixed" with this Kogito isn't it ;)

mswiderski:
exactly, that's why this is the best opportunity to make it better for people consuming it

tarilabs:
https://github.com/kiegroup/kogito-runtimes/pull/136#issuecomment-525642848 captured as KOGITO-223

PR #143, opened 2019-09-01
KOGITO-238 - Separate consumer for model and process instance

cristianonicolai:
Depends on https://github.com/kiegroup/kogito-bom/pull/53

### DIFF
--- a/data-index/data-index-service/src/main/java/org/kie/kogito/index/messaging/ReactiveMessagingEventConsumer.java
+++ b/data-index/data-index-service/src/main/java/org/kie/kogito/index/messaging/ReactiveMessagingEventConsumer.java
@@ -29,19 +29,29 @@ import org.slf4j.LoggerFactory;
 public class ReactiveMessagingEventConsumer {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ReactiveMessagingEventConsumer.class);
-    private static final String TOPIC = "kogito-processinstances-events";
-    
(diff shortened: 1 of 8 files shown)


cristianonicolai:
@ricardozanini we dont need to get rid of if then :)

data-index/data-index-service/src/main/java/org/kie/kogito/index/messaging/ReactiveMessagingEventConsumer.java
```diff
@@ -29,19 +29,29 @@
 public class ReactiveMessagingEventConsumer {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ReactiveMessagingEventConsumer.class);
-    private static final String TOPIC = "kogito-processinstances-events";
-    
+    private static final String KOGITO_PROCESSINSTANCES_EVENTS = "kogito-processinstances-events";
+    private static final String KOGITO_PROCESSDOMAIN_EVENTS = "kogito-processdomain-events";
```
mswiderski: is the expectation now to have two topics? I thought we will just have two consumers attached to the same topic as these will process the same messages

cristianonicolai: it is using the same topic, as defined in the properties file. Somehow if I reuse the same name in the annotation one of the methods ends up not receiving any messages.

mswiderski: Ah that’s perfect. Missed that

PR #145, opened 2019-09-04
KOGITO-17 - Enhancement to Work Items and their handlers

mswiderski:


mswiderski:
@cristianonicolai if you could have a look as this is related to the work on user tasks on the data index side.

@evacchi it's more for your information in case you have some spare cycles to review as well.

PR #146, opened 2019-09-04
KOGITO-245: Rule Unit implementation for non-codegenerated use cases

evacchi:
This is the exact equivalent of [BpmnProcess](https://github.com/kiegroup/kogito-runtimes/blob/79cec20bd06195840671d1528f07d18e8d5ec695/jbpm/jbpm-bpmn2/src/main/java/org/kie/kogito/process/bpmn2/BpmnProcess.java#L35-L34) but for rule units. See test for example usage.


evacchi:
I'll merge this since it's just an addition and breaks nothing

PR #148, opened 2019-09-06
[DROOLS-4292] executable-model fails with more than 5 arguments query

lucamolteni:
Forward port 

@mariofusco @baldimir


lucamolteni:
jenkins test again

PR #149, opened 2019-09-06
KOGITO-250 - Generate REST endpoint for private processes but without…

mswiderski:
… option to create new instances, KOGITO-251 - Expose service url as part of the kogito events


mswiderski:
Sonar was complaining about not high enough test coverage but the examples or should add to that to cover it

cristianonicolai:
@mswiderski I think the real issue is the reported "code smells".

mswiderski:
Ok, will take a look once back from PTO

mswiderski:
@cristianonicolai [code smells eliminated ](https://github.com/kiegroup/kogito-runtimes/pull/149/checks?check_run_id=223031472)

PR #153, opened 2019-09-10
KOGITO-260 DMN include kogito-dmn dep inside of quarkus archetype

tarilabs:


tarilabs:
> Any reason why this is not added to springboot archetype?

First I will throughoutfully check it works, then will add example and archetype update ;) that's all

PR #164, opened 2019-09-19
KOGITO-113 Fix major issues reported by SonarCloud

baldimir:
Sorry for the large PR. If needed, I can split it to more smaller PRs.


mariofusco:
@baldimir I gave a look at this PR and it looks "generally" ok. To be honest it is a bit hard to review a so big PR, so in future it would be better if we could do this sort of changes in smaller batches. Please resolve the conflicts and I'll give a 2nd look and merge. Thanks.

baldimir:
@mariofusco agree that this should have been split into more PRs. Sorry about that. I force-pushed. Let's wait for the PR check and review please.

drools/drools-core/src/main/java/org/drools/core/definitions/impl/KnowledgePackageImpl.java
```diff
@@ -142,15 +141,15 @@ public KnowledgePackageImpl() {
      */
     public KnowledgePackageImpl(final String name) {
         this.name = name;
-        this.accumulateFunctions = Collections.emptyMap();
```
mariofusco: This change is questionable. The reason why we initialized those variables with empty collections was to save both memory and time for collections that in many cases are not used. I'd revert this change if you don't see a strong reason to keep it.

baldimir: OK, I will revert this change.

baldimir: Done.

PR #165, opened 2019-09-20
KOGITO-313 - Validate if process id can be used for generated class n…

mswiderski:
…ames to improve error handling

### DIFF
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessGenerator.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessGenerator.java
@@ -21,6 +21,8 @@ import java.util.List;
 import java.util.Map.Entry;
 import java.util.NoSuchElementException;
 
+import javax.lang.model.SourceVersion;
+
 import com.github.javaparser.ast.CompilationUnit;
 import com.github.javaparser.ast.Modifier;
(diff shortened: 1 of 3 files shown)


kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessGenerator.java
```diff
@@ -112,6 +114,10 @@ public ProcessGenerator(
         this.generatedFilePath = targetCanonicalName.replace('.', '/') + ".java";
         this.completePath = "src/main/java/" + generatedFilePath;
         this.context = context;
+        
+        if (!SourceVersion.isName(targetTypeName)) {
+            throw new IllegalArgumentException("Process id '" + typeName + "' is not valid");
```
evacchi: since we are improving error message, we should probably add some better description of what's wrong with it. I'd say something like: 

```suggestion
            throw new IllegalArgumentException("Process id '" + typeName + "' is not valid. "
                "A valid identifier may only contain alphanumeric characters or the underscore symbol and it cannot start with a number.");   
```

mswiderski: I was thinking about it but it can simply be too much details to describe it in the exception message, for instance it can't start with underscore so there are bit more rules... and that's why I kept it to minimal to point people where the problem is.

PR #166, opened 2019-09-20
KOGITO-178 - Use Vert.x bus for synchronizing model indexing

cristianonicolai:


data-index/data-index-service/src/main/java/org/kie/kogito/index/messaging/ReactiveMessagingEventConsumer.java
```diff
@@ -16,65 +16,80 @@
...
-    public void onProcessInstanceEvent(KogitoProcessCloudEvent event) {
-        try {
-            LOGGER.debug("Process instance consumer received KogitoCloudEvent: \n{}", event);
-            indexingService.indexProcessInstance(event);
-        } catch (Exception ex) {
-            LOGGER.error("Error processing KogitoCloudEvent: {}", ex.getMessage(), ex);
-        }
+    public CompletionStage<Void> onProcessInstanceEvent(KogitoProcessCloudEvent event) {
+        LOGGER.debug("Process instance consumer received KogitoCloudEvent: \n{}", event);
+        return CompletableFuture.runAsync(() -> indexingService.indexProcessInstance(event.getData()));
```
mswiderski: Why do we need to do it asynchronously? And when is the message ack done?

data-index/data-index-service/src/main/java/org/kie/kogito/index/messaging/ReactiveMessagingEventConsumer.java
```diff
@@ -16,65 +16,80 @@
...
-    public void onUserTaskInstanceEvent(KogitoUserTaskCloudEvent event) {
-        try {
-            LOGGER.debug("Task instance received KogitoUserTaskCloudEvent \n{}", event);
-            indexingService.indexUserTaskInstance(event);
-        } catch (Exception ex) {
-            LOGGER.error("Error processing KogitoUserTaskCloudEvent: {}", ex.getMessage(), ex);
-        }
+    public CompletionStage<Void> onUserTaskInstanceEvent(KogitoUserTaskCloudEvent event) {
+        LOGGER.debug("Task instance received KogitoUserTaskCloudEvent \n{}", event);
+        return CompletableFuture.runAsync(() -> indexingService.indexUserTaskInstance(event.getData()));
```
mswiderski: Same here about async need and message ack

data-index/data-index-service/src/main/java/org/kie/kogito/index/messaging/ReactiveMessagingEventConsumer.java
```diff
@@ -16,65 +16,80 @@
...
+    public void onDomainEvent(Message<String> message) {
         try {
-            LOGGER.debug("Task domain received KogitoUserTaskCloudEvent \n{}", event);
-            indexingService.indexUserTaskInstanceDomain(event);
+            indexingService.indexModel(message.body());
+            message.reply(null);
         } catch (Exception ex) {
-            LOGGER.error("Error processing KogitoUserTaskCloudEvent: {}", ex.getMessage(), ex);
+            LOGGER.error("Error processing domain event: {}", ex.getMessage(), ex);
+            message.fail(0, ex.getMessage());
```
mswiderski: Will this rollback message back to event bus or?

data-index/data-index-service/src/main/java/org/kie/kogito/index/messaging/ReactiveMessagingEventConsumer.java
```diff
@@ -16,65 +16,80 @@
...
-    public void onUserTaskInstanceDomainEvent(KogitoUserTaskCloudEvent event) {
+    public CompletionStage<Void> onUserTaskInstanceDomainEvent(KogitoUserTaskCloudEvent event) {
+        LOGGER.debug("Task domain received KogitoUserTaskCloudEvent \n{}", event);
+        JsonObject json = new UserTaskInstanceMetaMapper().apply(event);
+        return sendMessage(json);
+    }
+
+    private CompletableFuture<Void> sendMessage(JsonObject json) {
+        CompletableFuture<Void> cf = new CompletableFuture<>();
+        eventBus.send(KOGITO_DOMAIN_EVENTS, json.toString(), async -> cf.complete(null));
```
mswiderski: Can you explain how this is going to resolve the race condition? It looks like we just slightly decreased probability under small load but with higher number of messages we still run into the issue. Or am I missing something?

data-index/data-index-service/src/main/java/org/kie/kogito/index/service/IndexingService.java
```diff
@@ -52,121 +47,72 @@
...
-    private void updateTaskInstances(JsonObjectBuilder builder, KogitoUserTaskCloudEvent event, JsonObject value) {
-        JsonObject jsonValue = new UserTaskInstanceMetaMapper().apply(event);
-        JsonArrayBuilder utBuilder = Json.createArrayBuilder().add(jsonValue);
-        if (value != null) {
-            JsonArray array = value.getJsonArray(USER_TASK_INSTANCES_DOMAIN_ATTRIBUTE);
-            if (array != null) {
-                array.stream().filter(json -> !event.getUserTaskInstanceId().equals(json.asJsonObject().getString("id"))).forEach(utBuilder::add);
-            }
+    private void copyJsonArray(String attribute, JsonObject model, JsonObjectBuilder builder, JsonArray indexTIArray) {
+        JsonArrayBuilder tiBuilder = Json.createArrayBuilder(indexTIArray);
```
mswiderski: Names of the variables and params are bit misleading as they refer to tasks but the method is more of generic usage

data-index/data-index-service/src/main/java/org/kie/kogito/index/messaging/ReactiveMessagingEventConsumer.java
cristianonicolai: This was done mainly to facilitate tests. According to SmallRye, ack is done when CompletionStage finishes, so although the API change it should behave exactly the same, next message will only be delivered when processing here is done.

data-index/data-index-service/src/main/java/org/kie/kogito/index/messaging/ReactiveMessagingEventConsumer.java
cristianonicolai: will mark this message as failed for the vert.x bus so a response can be sent back to https://github.com/kiegroup/kogito-runtimes/pull/166/files#diff-b5acac58b35ad44d1e0a45c6406c5418R81 but similarly to the previous behaviour we ack message from Kafka anyway, simply ignoring it.

data-index/data-index-service/src/main/java/org/kie/kogito/index/messaging/ReactiveMessagingEventConsumer.java
cristianonicolai: so now I consolidate the domain processing/update into a single consumer https://github.com/kiegroup/kogito-runtimes/pull/166/files#diff-b5acac58b35ad44d1e0a45c6406c5418R85. So even we receive multiple Kafka messages for domain update, they are pushed into a single queue using vert.x bus. That ensures that only one consumer is processing these updates at any time.

data-index/data-index-service/src/main/java/org/kie/kogito/index/service/IndexingService.java
cristianonicolai: fixed.

data-index/data-index-service/src/main/java/org/kie/kogito/index/messaging/ReactiveMessagingEventConsumer.java
mswiderski: won't this mean we can miss some messages? meaning that we didn't process it from the event bus but ack in Kafka and by that it will not be delivered any more?

data-index/data-index-service/src/main/java/org/kie/kogito/index/messaging/ReactiveMessagingEventConsumer.java
mswiderski: how do we guarantee that... I mean event bus will hold all messages and process them one by one meaning in synchronous way.. it sounds kind of weird as this would not scale at all. Is this somehow configured or?

data-index/data-index-service/src/main/java/org/kie/kogito/index/messaging/ReactiveMessagingEventConsumer.java
mswiderski: same here it looks like a big bottleneck in general if we can't have parallel processing from the same topic

data-index/data-index-service/src/main/java/org/kie/kogito/index/messaging/ReactiveMessagingEventConsumer.java
cristianonicolai: ok so changed the approach to have a single consumer per process id. Consumer is created via api so we can ensure there is only one running at the time.

data-index/data-index-service/src/main/java/org/kie/kogito/index/messaging/ReactiveMessagingEventConsumer.java
cristianonicolai: not really, we are not missing messages, we only ack Kafka messages after they are processed by the vert.x bus. The only point now is that we ack the Kafka message anyway, independently if the result of the processing in the vert.x bus fails or not, which matches with the previous behaviour, meaning even if we try to index it and an exception is thrown.

data-index/data-index-service/src/main/java/org/kie/kogito/index/messaging/ReactiveMessagingEventConsumer.java
cristianonicolai: true, changed the strategy now for a consumer per process id.

data-index/data-index-service/src/main/java/org/kie/kogito/index/messaging/ReactiveMessagingEventConsumer.java
```diff
@@ -38,43 +55,67 @@
...
-    public void onUserTaskInstanceDomainEvent(KogitoUserTaskCloudEvent event) {
+    public CompletionStage<Void> onUserTaskInstanceDomainEvent(KogitoUserTaskCloudEvent event) {
+        LOGGER.debug("Task domain received KogitoUserTaskCloudEvent \n{}", event);
+        JsonObject json = new UserTaskInstanceMetaMapper().apply(event);
+        return sendMessage(json);
+    }
+
+    private CompletableFuture<Void> sendMessage(JsonObject json) {
+        CompletableFuture<Void> cf = new CompletableFuture<>();
+        String processId = json.getString("processId");
```
mswiderski: just last thing to confirm - is this taking into consideration rootProcessId?

cristianonicolai: yep, comes from:
- https://github.com/kiegroup/kogito-runtimes/pull/166/files#diff-69e4b437e8193a8b5845a67bf3e56d14R43
- https://github.com/kiegroup/kogito-runtimes/pull/166/files#diff-7d761e40f8f592a02375793138c669d1R43

PR #176, opened 2019-09-30
[KOGITO-309] Adding support for external hotrod-client.properties

ricardozanini:
See:
https://issues.jboss.org/browse/KOGITO-309

In this PR we added the possibility to users to set their custom `hotrod-client.properties` via `kogito.cache.hotrod.property.path` property. 

This way, the Operator would mount a property file inside a ConfigMap and set this property to where the file is.

We also removed the older placeholder since those variables will be vanished, being replaced by the custom hothod-client file. To keep the compatibility with older versions, the Operator will continue to accept the authentication options, setting them to the ConfigMap instead of using env vars.

The image will change too, since it won't need to inject auth vars to env.

### DIFF
--- a/data-index/data-index-service/src/main/resources/META-INF/hotrod-client.properties
+++ b/data-index/data-index-service/src/main/resources/META-INF/hotrod-client.properties
@@ -1,9 +1,0 @@
-# Docker 4 Mac workaround
-infinispan.client.hotrod.client_intelligence=BASIC
-# https://docs.jboss.org/infinispan/9.4/apidocs/org/infinispan/client/hotrod/configuration/package-summary.html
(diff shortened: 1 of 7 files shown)


ricardozanini:
Blocked by: https://github.com/quarkusio/quarkus/issues/4285

We can't read nor manipulate the `RemoteCacheManager` since the extension is initializing it before we get the chance to do it.

mswiderski:
@ricardozanini I believe this can be closed? The Infinispan related changed was merged into quarkus

ricardozanini:
@mswiderski you bet. I'm going to close it and will work on a new one with this Quarkus changes once they release it.

PR #181, opened 2019-10-02
KOGITO-366 DMN fix OCP asset resolution errors

tarilabs:
The scope of this PR is to fix OCP resolution errors.


tarilabs:
thanks @mswiderski

PR #183, opened 2019-10-03
KOGITO-370: Codegen: remove outdated Drools codegen classes

evacchi:
delete outdated classes that are no longer used


mswiderski:
@evacchi could you please rebase/resolve conflicts so we can merge this in?

PR #187, opened 2019-10-07
KOGITO-380: Better error reporting on process codegen

evacchi:


jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/MalformedNodeException.java
```diff
@@ -0,0 +1,11 @@
+package org.jbpm.bpmn2.xml;
```
mswiderski: missing license header

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessCodegenException.java
```diff
@@ -0,0 +1,10 @@
+package org.kie.kogito.codegen.process;
```
mswiderski: same here, missing license header

evacchi:
sure will do

PR #197, opened 2019-10-18
Kogito Maven dev mode

cristianonicolai:
Assembly:
- https://github.com/kiegroup/kogito-runtimes/pull/197
- https://github.com/kiegroup/kogito-bom/pull/63

### DIFF
--- a/kogito-maven-plugin/src/main/java/org/kie/kogito/maven/plugin/DevMojo.java
+++ b/kogito-maven-plugin/src/main/java/org/kie/kogito/maven/plugin/DevMojo.java
@@ -1,0 +1,147 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.maven.plugin;
+
+import java.io.File;
+import java.net.InetSocketAddress;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Optional;
+
+import org.apache.maven.artifact.DefaultArtifact;
+import org.apache.maven.artifact.handler.DefaultArtifactHandler;
+import org.apache.maven.execution.MavenSession;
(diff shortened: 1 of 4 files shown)


cristianonicolai:
@nmirasch @mswiderski just experimenting to get your feedback, still needs some tweaking, but this would spin up all required services ( data index, infinispan and kafka ) by running maven kogito:dev. Let me know what you think.

evacchi:
so I don't know how would it be feasible but if we moved this under `kogito-codegen` (of course, lack of a better name: it may become something like `kogito-dev` or something) we could possibly get it to work both as apart of the usual Maven plug-in but especially in **quarkus dev mode**. 

Otherwise we would have this "split world" situation where you have kogito:dev and quarkus:dev. in the case of quarkus:dev we could just enable this with some flag... or make it a part of a separate extension 

WDYT

cristianonicolai:
@evacchi agree, I was thinking on how to integrate both, we could make it even so our dev mode call quarkus:dev. For now, I just wanted to get feedback on what people think about this idea, if it is worth exploring it.

baldimir:
Based on the SonarCloud report, this has 0% test coverage on new code. Could you please add some unit tests that cover this PR?

kogito-maven-plugin/src/main/java/org/kie/kogito/maven/plugin/DevMojo.java
```diff
@@ -0,0 +1,147 @@
...
+                .adminOperationsHandler(new EmbeddedServerAdminOperationHandler())
+                .build();
+        hotRodServer.start(cfg, cacheManager);
+    }
+
+    private void startDataIndex() throws MojoExecutionException {
+        PluginDescriptor plugin = (PluginDescriptor) getPluginContext().get("pluginDescriptor");
+        Optional<ComponentDependency> dep = plugin.getDependencies().stream().filter(d -> d.getGroupId().equals("org.kie.kogito") &&
+                d.getArtifactId().equals("data-index-service")).findFirst();
+        if (dep.isPresent()) {
```
mswiderski: shouldn't we have the same dependency check for Infinispan and Kafka?

evacchi:
> For instance we should not spin up Kafka, Infinispan etc when we have only rules or decisions in the project.

@mswiderski I think there are a few ways to do that. 
1) Simple way: expose flags those through System properties
2) more complex way: use the quarkus extension / build step mechanism to look for extensions in quarkus

either way: if the common code base is instantiated via bool/flags/options then we can do 1) for maven plugin 2) for quarkus extension

cristianonicolai:
@mswiderski correct, my idea is to look in the project dependencies to know which services to start.

kogito-maven-plugin/src/main/java/org/kie/kogito/maven/plugin/DevMojo.java
cristianonicolai: not really, as this is only to know the data-service version that the plugin depends on, so it can be started.

PR #198, opened 2019-10-18
KOGITO-388: ApplicationGenerator breaks when packageName contains hyphens

evacchi:
now it throws if the package name is invalid. On a slightly-related note, we found out about this bug because a user was using dashes inside their groupId. I was wondering whether we really need to use a packageName == groupId instead of a default package name like quarkus. Doing so could simplify a few assumptions we make sometimes (e.g., where `Application` class is)

### DIFF
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/ApplicationGenerator.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/ApplicationGenerator.java
@@ -17,11 +17,14 @@ package org.kie.kogito.codegen;
 
 import java.io.File;
 import java.nio.charset.StandardCharsets;
+import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.stream.Collectors;
+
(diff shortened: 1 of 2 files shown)


mswiderski:
isn't this kind of standard practice to use group id as package name? I like the approach to be it more tailored when it comes to package names as it feels more integrated. Plus in case we would support composition at some point we won't run into same class kind of conflicts.

evacchi:
> isn't this kind of standard practice to use group id as package name? I like the approach to be it more tailored when it comes to package names as it feels more integrated. Plus in case we would support composition at some point we won't run into same class kind of conflicts

yes although it makes a little less easy to know "automatically where an Application class is at runtime: the group id is a property of the Maven artifact... (although maybe there is some info in Modules ?). Besides, I thought there would have been only 1 `Application` per uber-jar/executable? Other cases when this would make sense if the `Application` class were not hidden but user-provided/customizable scaffolding, but we are not doing that yet

mswiderski:
@evacchi either way I am fine, not strongly opinionated about it. So if you prefer to have it as fixed package name then fine

evacchi:
anyway, subject unrelated, we can merge this

PR #201, opened 2019-10-21
KOGITO-455 - Allow retrieving all subprocesses from a process instance

cristianonicolai:


data-index/data-index-service/src/main/java/org/kie/kogito/index/graphql/GraphQLSchemaManager.java
```diff
@@ -113,12 +123,18 @@ private GraphQLSchema createSchema() {
         return schemaGenerator.makeExecutableSchema(typeDefinitionRegistry, runtimeWiring);
     }
 
-    private Collection<ObjectNode> getProcessInstancesValues(DataFetchingEnvironment env) {
+    private Set<String> getChildProcessInstancesValues(DataFetchingEnvironment env) {
+        ProcessInstance source = env.getSource();
+        Collection<ProcessInstance> pil = queryService.queryProcessInstances(ProcessInstanceFilter.builder().parentProcessInstanceId(singletonList(source.getId())).build());
```
mswiderski: so is this performing a query for given process instance to load its subprocesses? If so what about a case when first is the parent process and then subprocesses send in the event lists?

cristianonicolai: the attribute will be resolved using a query whenever graphql asks for it, so it will be kept up-to-date with the result of the query.

mswiderski: ok, clear now, thanks

PR #206, opened 2019-10-23
KOGITO-430: Complete Rule Unit input/output mapping

evacchi:
I have added better test coverage. Now the cases are handled as follows:

| Mapping | Process Variable | Rule Unit field   | Action
|-----------|-----------------------|--------------------|-----------|
| IN      | scalar           | scalar            | Assignment
| IN      | scalar           | data source 	    | Add to (i.e. insert into) data source
| IN      | collection       | data source 	    | Add all contents from data source
| OUT     | scalar           | scalar 	        | Assignment
| OUT     | scalar           | data source 	    | get 1 value off the data source
| OUT     | collection       | data source 	    | Add all contents to the data source

there might be still too much magic, so we might want to reconsider at some point


jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/LambdaSubProcessNodeVisitor.java
```diff
@@ -38,6 +38,7 @@
 import org.kie.api.definition.process.WorkflowProcess;
 
 import static com.github.javaparser.StaticJavaParser.parse;
+import static com.github.javaparser.StaticJavaParser.parseStatement;
```
mswiderski: is this needed?

kogito-codegen/src/test/resources/ruletask/ExampleP.bpmn
```diff
@@ -0,0 +1 @@
...ardDeviation="0"/></bpsim:ProcessingTime></bpsim:TimeParameters><bpsim:ResourceParameters><bpsim:Availability><bpsim:FloatingParameter value="0"/></bpsim:Availability><bpsim:Quantity><bpsim:FloatingParameter value="0"/></bpsim:Quantity></bpsim:ResourceParameters><bpsim:CostParameters><bpsim:UnitCost><bpsim:FloatingParameter value="0"/></bpsim:UnitCost></bpsim:CostParameters></bpsim:ElementParameters><bpsim:ElementParameters elementRef="_5293F325-A06A-469B-8C31-1E79EE36A3E2"><bpsim:TimeParameters><bpsim:ProcessingTime><bpsim:NormalDistribution mean="0" standardDeviation="0"/></bpsim:ProcessingTime></bpsim:TimeParameters><bpsim:ResourceParameters><bpsim:Availability><bpsim:FloatingParameter value="0"/></bpsim:Availability><bpsim:Quantity><bpsim:FloatingParameter value="0"/></bpsim:Quantity></bpsim:ResourceParameters><bpsim:CostParameters><bpsim:UnitCost><bpsim:FloatingParameter value="0"/></bpsim:UnitCost></bpsim:CostParameters></bpsim:ElementParameters></bpsim:Scenario></bpsim:BPSimData></bpmn2:extensionElements><bpmn2:source>_08903786-89F6-4098-917D-E1E5AD98CA9C</bpmn2:source><bpmn2:target>_08903786-89F6-4098-917D-E1E5AD98CA9C</bpmn2:target></bpmn2:relationship></bpmn2:definitions>
```
mswiderski: can we format it?

kogito-codegen/src/test/java/ruletask/Example.java
```diff
@@ -0,0 +1,32 @@
+package ruletask;
```
mswiderski: missing license header?

PR #208, opened 2019-10-24
KOGITO-481 - Fix for filtering using null values

cristianonicolai:


data-index/data-index-service/src/test/java/org/kie/kogito/index/GraphQLUtils.java
```diff
@@ -114,13 +114,13 @@ private static String toGraphQLString(String root, Class targetClass, Class filt
     }
 
     private static String getObjectString(Object object) {
-        return object instanceof String ? "\\\"" + object + "\\\"" : object.toString();
+        return object == null ? "null" : object instanceof String ? "\\\"" + object + "\\\"" : object.toString();
```
mswiderski: so does this allow you to search for things that have null values or?

data-index/data-index-service/src/test/java/org/kie/kogito/index/GraphQLUtils.java
```diff
@@ -114,13 +114,13 @@ private static String toGraphQLString(String root, Class targetClass, Class filt
...
     }
 
     private static Function<Object, String> getObjectString() {
         return object -> {
             if (object instanceof Collection) {
-                return ((Collection) object).stream().map(o -> getObjectString(o)).collect(joining(", ")).toString();
+                return "[ " + ((Collection) object).stream().map(o -> getObjectString(o)).collect(joining(", ")).toString() + " ] ";
```
mswiderski: so this is like JSON array? just done manually?

cristianonicolai: yep, just for test.

data-index/data-index-service/src/test/java/org/kie/kogito/index/GraphQLUtils.java
cristianonicolai: exactly, in the list of values (ids), you need to input null in there, otherwise, we cant know if the value is just null or you actually want to filter by null. But this class here is just a helper for tests.

PR #211, opened 2019-10-31
[KOGITO-496] Writing compressed labels for protobuf files

ricardozanini:
See:
https://issues.jboss.org/browse/KOGITO-496

Now we are compressing the protobuf file content as a workaround for the OpenShift labels limit size until we finish the new solution to those files during runtime.

Related:
https://github.com/kiegroup/kogito-cloud-operator/pull/99


ricardozanini:
Fixed conflicts.

PR #213, opened 2019-11-05
KOGITO-539 - Do not create protobuf file if no persistence is used

mswiderski:


kogito-codegen/src/main/java/org/kie/kogito/codegen/process/persistence/PersistenceGenerator.java
```diff
@@ -99,33 +99,33 @@ public ApplicationSection section() {
...
-                    !KOGITO_APPLICATION_PROTO.equalsIgnoreCase(name) && name.toLowerCase().endsWith(PersistenceLabeler.PROTO_FILE_EXT))))
-                this.persistenceLabeler.processProto(protoFile);
-        }
-        
-        if (!protoFilePath.toFile().exists()) {
-            try {
-                // generate proto file based on known data model
-                Proto proto = protoGenerator.generate(packageName, dataModelClasses, "import \"kogito-types.proto\";");
-                protoFilePath = Paths.get(targetDirectory.toString(), "classes", "/persistence", KOGITO_APPLICATION_PROTO);
+        if (persistence) {
```
ricardozanini: @mswiderski the `persistence=true` is coming from looking for `persistence add-on` in the classpath by the maven plugin?

mswiderski: That’s correct

PR #214, opened 2019-11-05
KOGITO-549 Explanatory toString() for Drools executable model

triceo:
This is a port of https://github.com/kiegroup/drools/pull/2632 to Kogito. Identical code.


triceo:
No longer necessary.

PR #221, opened 2019-11-19
KOGITO-621 - Remove GraphiQL resources and do a browser redirect to /…

cristianonicolai:
…graphiql


cristianonicolai:
@mswiderski I addded these in the past because of an issue on vertx which missed to include it. It has now been fixed with the update on Quarkus so we dont need to duplicate it.

PR #226, opened 2019-11-21
[KOGITO-625] - Kogito quickstart Person example fails with list of ob…

spolti:
…jects

Signed-off-by: Filippe Spolti <fspolti@redhat.com>


spolti:
thanks :)

PR #227, opened 2019-11-22
KOGITO-639: kogito-runtimes pom should not depend on quarkus

evacchi:
otherwise it will bring an older version of kogito-runtimes
as a transitive dependency

### DIFF
--- a/addons/pom.xml
+++ b/addons/pom.xml
@@ -19,4 +19,18 @@
     <module>jobs</module>
   </modules>
 
+
+  <dependencyManagement>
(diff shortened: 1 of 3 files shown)


evacchi:
fix here is only partial. Proper fix is either NOT depending on quarkus-bom and enumerate the dependencies, or NOT depending on our internal versions. Please also notice that we are implicitly dependending on the Spring version declared in the quarkus bom. This has also to be fixed.

/cc @cristianonicolai

PR #231, opened 2019-11-25
KOGITO-517 - Index last update and consolidate Kogito metadata attributes

cristianonicolai:


data-index/data-index-service/src/test/resources/visaApplications.proto
```diff
@@ -32,8 +32,6 @@ message VisaApplications {
 	optional string id = 1; 
 	/* @Field(store = Store.YES) */ 
 	optional VisaApplication visaApplication = 2; 
-	/* @Field(store = Store.YES) */ 
-	repeated org.kie.kogito.index.model.ProcessInstanceMeta processInstances = 3; 
-	/* @Field(store = Store.YES) */ 
-	repeated org.kie.kogito.index.model.UserTaskInstanceMeta userTasks = 4; 
+    /* @Field(store = Store.YES) */
+    optional org.kie.kogito.index.model.KogitoMeta kogito = 3;
```
mswiderski: would this be the best name to be used? kogito sounds bit to generic and does not provide a information about the content of it... maybe using metadata or so would be more readable

cristianonicolai: yep, that might be better, will update to metadata instead.

cristianonicolai:
@mswiderski updated, please take a nother look.

PR #234, opened 2019-11-26
KOGITO-641: Build fails processing dtable, when no drools-decisiontables

evacchi:
An helpful error message is now printed when XLS are present, but the DTABLE dependency is missing

```
2019-11-26 10:56:33,120 [main] ERROR A Decision Table resource was found, but a necessary dependency is missing. 
Verify you have added decision table support to your project dependencies: 

<dependency>
      <artifactId>drools-decisiontables</artifactId>
</dependency>

```

### DIFF
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/IncrementalRuleCodegen.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/IncrementalRuleCodegen.java
@@ -173,15 +173,7 @@ public class IncrementalRuleCodegen extends AbstractGenerator {
         ReleaseIdImpl dummyReleaseId = new ReleaseIdImpl("dummy:dummy:0.0.0");
         if (!decisionTableSupported &&
                 resources.stream().anyMatch(r -> r.getResourceType() == ResourceType.DTABLE)) {
-            ApplicationGenerator.logger.error(
-                    "A Decision Table resource was found, but a necessary dependency is missing. \n" +
-                    "Verify you have added decision table support to your project dependencies: \n" +
-                    "\n" +
-                    "<dependency>\n" +
-                    "      <groupId>org.kie.kogito</groupId>\n" +
-                    "      <artifactId>drools-decisiontables</artifactId>\n" +
-                    "</dependency>");
-            throw new RuleCodegenError();
+            throw new MissingDecisionTableDependencyError();
         }
 
         moduleGenerator = new RuleUnitContainerGenerator();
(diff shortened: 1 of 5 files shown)


kogito-codegen/src/test/java/org/kie/kogito/codegen/rules/IncrementalRuleCodegenTest.java
```diff
@@ -21,14 +21,26 @@
...
 import org.kie.api.io.ResourceType;
 import org.kie.kogito.codegen.GeneratedFile;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 
 public class IncrementalRuleCodegenTest {
 
+    @BeforeEach
+    public void blah() {
```
mswiderski: maybe not the best method name?

kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/RuleCodegenError.java
```diff
@@ -0,0 +1,8 @@
+package org.kie.kogito.codegen.rules;
```
mswiderski: missing license header

kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/IncrementalRuleCodegen.java
```diff
@@ -182,7 +198,7 @@ public ApplicationSection section() {
 
         if (modelBuilder.hasErrors()) {
             ApplicationGenerator.logger.error( modelBuilder.getErrors().toString() );
-            return Collections.emptyList();
+            throw new RuleCodegenError();
```
mswiderski: shoudn't we pass the errors to the exception?

kogito-codegen/src/test/java/org/kie/kogito/codegen/rules/IncrementalRuleCodegenTest.java
evacchi: whoopsie :P

kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/IncrementalRuleCodegen.java
evacchi: Yes, maybe that's a better way to do it? Currently the error just says "please scroll up to read more"

mswiderski: isn't that like duplicated error? one that has the actual info and the the other that just says... try to find it somewhere up the scroll?

PR #236, opened 2019-11-27
KOGITO-535: Adding quarkus security conf and disabling graphiQL when …

nmirasch:
…auth is enabled


mswiderski:
just one question - does this mean that as soon as we enable auth we won't be able to use graphiql interface? isn't this too big limitation.... just asking

nmirasch:
@mswiderski the answer is yes, it won't be accessible if auth is enabled, for now. We have got the quarkus security limitation of maintain web auth (that call to keycloak web login)  and service auth in the same app. We have postpossed this UI auth work, prioritizing  the managemente console. Giving some time to quarkus oidc that is still evolving and we'll back to expose this UI with the auth control.

nmirasch:
rebased

PR #237, opened 2019-11-27
KOGITO-645 - Multi parameter Service Task support

mswiderski:
@krisv @DuncanDoyle here is an implementation that allows to use service task (generic service task) to invoke method with multiple parameters. Since the BPMN2 defines messages where there could be only one in message and one out message, this PR introduces an alternative to not define messages at all on operations but control this based on data inputs and outputs

So with that, you can have 0 or many data inputs and 0 or 1 data output for a service task.


mswiderski:
@krisv I used stunner in vscode to design a sample process and didn't find any issues with that.

PR #244, opened 2019-12-03
KOGITO-681 - Upgrade Quarkus to version 1.0.1

cristianonicolai:
Depends on:
- https://github.com/kiegroup/kogito-bom/pull/72

### DIFF
--- a/addons/jobs/jobs-service/pom.xml
+++ b/addons/jobs/jobs-service/pom.xml
@@ -1,13 +1,29 @@
 <?xml version="1.0"?>
-<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
-         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
+         xmlns="http://maven.apache.org/POM/4.0.0">
   <modelVersion>4.0.0</modelVersion>
   <parent>
     <groupId>org.kie.kogito</groupId>
     <artifactId>jobs</artifactId>
     <version>8.0.0-SNAPSHOT</version>
   </parent>
+
   <artifactId>jobs-service</artifactId>
+  <name>Jobs Service</name>
+  <description>Jobs Service (Timers and Async Jobs)</description>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>io.quarkus</groupId>
+        <artifactId>quarkus-bom</artifactId>
+        <version>${version.io.quarkus}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
   <dependencies>
     <dependency>
       <groupId>org.kie.kogito</groupId>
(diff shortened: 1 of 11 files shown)


mswiderski:
@cristianonicolai would you mind to resolve conficts?

mswiderski:
@cristianonicolai I think you also need to update Jenkins file in this repo to use the new property to enable Infinispan tests https://github.com/kiegroup/kogito-runtimes/blob/master/Jenkinsfile#L38

btw, is this considered temp solution or?

addons/jobs/jobs-service/pom.xml
```diff
@@ -130,6 +146,46 @@
     </plugins>
   </build>
   <profiles>
+    <profile>
+      <id>infinispan</id>
+      <activation>
+        <property>
+          <name>infinispan</name>
```
tiagodolphine: The profile name could be something more generic like "integration" to represent the execution of integration tests, in case we have more external services other than infinispan, to be mocked/embedded like kafka, etc, WDYT?

cristianonicolai:
@mswiderski @nmirasch @tiagodolphine please take another look, I've changed the approach to not rely on Infinispan jars but using testcontainers to start the service. This should avoid any version conflict.

addons/jobs/jobs-service/src/test/java/org/kie/kogito/jobs/service/resource/InfinispanServerTestResource.java
```diff
@@ -18,68 +18,34 @@
...
+import org.testcontainers.containers.FixedHostPortGenericContainer;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
 
 public class InfinispanServerTestResource implements QuarkusTestResourceLifecycleManager {
 
-    private HotRodServer hotRodServer;
-    private EmbeddedCacheManager cacheManager;
-    private static final Integer PORT = 11232;
+    private static final String INFINISPAN_VERSION = System.getProperty("infinispan.version", "latest");
```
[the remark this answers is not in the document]
cristianonicolai: @mswiderski makes sense...

addons/jobs/jobs-service/src/test/java/org/kie/kogito/jobs/service/resource/InfinispanServerTestResource.java
```diff
@@ -18,68 +18,37 @@
...
-                .proxyPort(PORT)
-                .adminOperationsHandler(new EmbeddedServerAdminOperationHandler())
-                .build();
-        hotRodServer.start(cfg, cacheManager);
-
-        Stream.of(InfinispanConfiguration.Caches.SCHEDULED_JOBS)
-                .forEach(name -> cacheManager.administration().getOrCreateCache(name, configuration));
-
+        if (INFINISPAN_VERSION == null) {
+            throw new RuntimeException("Please define a valid Infinispan image version in system property infinispan.version");
```
tiagodolphine: It is possible to run the tests using maven since the property is set on the sunfire plugin but in case you want to run the test outside maven, for instance directly from the IDE to debug, we get the exception. Just a comment, I don't know if it makes sense to have a default parameter or someway else to the version value.

cristianonicolai: @tiagodolphine that was my concern too with not having the default value, but this setting actually works, at least on IntelliJ, when running the tests individually.

tiagodolphine: ok great then, that was my concern but if it works, no problem!

PR #249, opened 2019-12-09
KOGITO-648 - Generate reactive REST endpoint instead of the default one

DuncanDoyle:
Introduces Reactive REST resources when running on Quarkus.

### DIFF
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/context/QuarkusKogitoBuildContext.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/context/QuarkusKogitoBuildContext.java
@@ -1,0 +1,20 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.codegen.context;
+
+public class QuarkusKogitoBuildContext implements KogitoBuildContext {
+    
+}
--- a/kogito-codegen/src/main/resources/class-templates/ReactiveRestResourceTemplate.java
+++ b/kogito-codegen/src/main/resources/class-templates/ReactiveRestResourceTemplate.java
@@ -1,0 +1,152 @@
+package com.myspace.demo;
+
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+
+import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.DELETE;
+import javax.ws.rs.GET;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.DefaultValue;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+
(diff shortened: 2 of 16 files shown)


DuncanDoyle:
@mswiderski I'm not sure I'm happy with how I implemented the passing of the GeneratorContext to the ProcessCodegen. Basically I now pass the context as a second argument in the `ofPath` method. I couldn't really come up with anything cleaner at this point.

kogito-codegen/src/main/resources/class-templates/RestResourceTemplate.java
```diff
@@ -27,13 +30,28 @@
...
 public class $Type$Resource {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger($Clazz$.class);
+
     Process<$Type$> process;
     
     Application application;
 
+    @PostConstruct
+    public void onStartup() {
```
mswiderski: I guess we can remove these two methods from here?

kogito-codegen/src/main/resources/class-templates/ReactiveRestResourceTemplate.java
```diff
@@ -0,0 +1,164 @@
...
+    Application application;
+
+    @PostConstruct
+    public void onStartup() {
+        LOGGER.info("Starting Reactive REST Resource '/$name$'.");
+    }
+
+    @PreDestroy
+    public void onShutdown() {
+        LOGGER.info("Stopping Reactive REST Resource '/$name$'.");
```
mswiderski: same here, these methods do not make much sense...

kogito-codegen/src/main/resources/class-templates/ReactiveRestResourceTemplate.java
```diff
@@ -0,0 +1,164 @@
...
+    
+    protected $Type$ getModel(ProcessInstance<$Type$> pi) {
+        if (pi.status() == ProcessInstance.STATE_ERROR && pi.error().isPresent()) {
+            throw new ProcessInstanceExecutionException(pi.id(), pi.error().get().failedNodeId(), pi.error().get().errorMessage());
+        }
+        
+        return pi.variables();
+    }
+    
+    protected Policy[] policies(String user, List<String> groups) {
```
mswiderski: can't this be shared between both types of resources?

kogito-codegen/src/main/java/org/kie/kogito/codegen/context/QuarkusKogitoBuildContext.java
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.codegen.context;
+
+public class QuarkusKogitoBuildContext implements KogitoBuildContext {
```
mswiderski: so in the end you didn't want to move the dependency injector here or just simply due to lack of time?

DuncanDoyle: No, I actually took it out of the PR because I wanted this to be a clean implementation for Reactive REST, and not pollute it with a discussion on where those DependencyInjectors should go. We can add that at a later point. The DependencyInjector is just one aspect that could go into these classes.

kogito-codegen/src/main/resources/class-templates/ReactiveRestResourceTemplate.java
DuncanDoyle: I like logging here and there for info and debugging purposes (which Kogito is a bit lacking atm IMO), so that's why I added them (it was also convenient to see WHICH implementation was being used without having to go into the source code).

kogito-codegen/src/main/resources/class-templates/ReactiveRestResourceTemplate.java
DuncanDoyle: Tbh, I haven't tried that, but could give that a go.

kogito-codegen/src/main/resources/class-templates/RestResourceTemplate.java
DuncanDoyle: If we don't want lifecycle logging, yes we can.

mswiderski:
@DuncanDoyle could you please rebase with master as this should make the build green so we could merge it

mswiderski:
ok, rebased with master and squashed commits, let's see how it goes now...

PR #254, opened 2019-12-10
[KOGITO-649] Query based REST endpoint generation to work on single i…

mariofusco:
…nput and output


mariofusco:
jenkins test

mariofusco:
ok to test

PR #255, opened 2019-12-11
[KOGITO-336] Include DMN inside of Spring Boot archetype

tkobayas:
I added "http*" for archetype application.properties so that generated DMN REST classes can be scanned. A package name of a DMN REST class is generated from its namespace. So this configuration works when DMN namespace starts from "http" or "https". According to DMN spec, namespace is anyURI. So it's not constrained to http/https but it's a reasonable default value for archetype?

> namespace: anyURI [1] | This attribute identifies the namespace associated with this Definitions and follows the convention established by XML Schema.

I manually tested by dropping TrafficViolation.dmn (https://github.com/kiegroup/kogito-runtimes/blob/master/drools/kogito-dmn/src/test/resources/org/kie/kogito/dmn/TrafficViolation.dmn) into the generated project and confirmed that the generated REST endpoint works.

```
$ curl -d '{ "Driver": { "Points": 2}, "Violation": { "Type": "speed", "Actual Speed": 120, "Speed Limit": 100 }}' -H "Content-Type: application/json" -X POST http://localhost:8080/Traffic%20Violation

{"Violation":{"Type":"speed","Speed Limit":100,"Actual Speed":120},"Driver":{"Points":2},"Fine":{"Points":3,"Amount":500},"Should the driver be suspended?":"No"}
```

### DIFF
--- a/archetypes/kogito-springboot-archetype/src/main/resources/archetype-resources/src/main/resources/application.properties
+++ b/archetypes/kogito-springboot-archetype/src/main/resources/archetype-resources/src/main/resources/application.properties
@@ -2,4 +2,4 @@ server.address=0.0.0.0
 
 spring.mvc.servlet.path=/docs
 
-resteasy.jaxrs.scan-packages=org.kie.**,\${groupId}
+resteasy.jaxrs.scan-packages=org.kie.**,\${groupId},http*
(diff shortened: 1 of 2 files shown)


tkobayas:
I think the Jenkins failure is not related to this PR.

tarilabs:
Jenkins test again

PR #258, opened 2019-12-12
KOGITO-770 - Align use of micro profile config to latest specificatio…

mswiderski:
…n requirements

### DIFF
--- a/kogito-codegen/src/main/resources/class-templates/MessageConsumerTemplate.java
+++ b/kogito-codegen/src/main/resources/class-templates/MessageConsumerTemplate.java
@@ -1,6 +1,7 @@
 package com.myspace.demo;
 
 import java.util.TimeZone;
+import java.util.Optional;
 
 import javax.annotation.PostConstruct;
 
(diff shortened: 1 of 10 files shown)


kogito-codegen/src/main/resources/class-templates/MessageConsumerTemplate.java
```diff
@@ -35,7 +36,7 @@ public void consume(String payload) {
 	    final String trigger = "$Trigger$";
         try {
             
-            if (useCloudEvents) {
+            if (useCloudEvents.orElse(true)) {
```
spolti: Does it needs to set it to true here as it is already being set to true on line 27?

kogito-codegen/src/main/resources/class-templates/MessageProducerTemplate.java
```diff
@@ -29,7 +30,7 @@ public void produce(ProcessInstance pi, $Type$ eventData) {
 	private String marshall(ProcessInstance pi, $Type$ eventData) {
 	    try {
 	        
-	        if (useCloudEvents) {
+	        if (useCloudEvents.orElse(true)) {
```
spolti: same here.

kogito-codegen/src/main/resources/class-templates/MessageConsumerTemplate.java
mswiderski: yes, as the initial value will be overridden by dependency injection

PR #260, opened 2019-12-12
KOGITO-773: Re-align Kogito to Drools model compiler (source-level)

evacchi:


mswiderski:
@evacchi there are bunch of classes that don't have license headers ... so would be worth to add them. Plus have we decided to skip reviews/approvals for PRs?

evacchi:
@mswiderski sorry @mariofusco said to go ahead

PR #264, opened 2019-12-18
KOGITO-807 - Prediction service as add on for user tasks

mswiderski:
depends on https://github.com/kiegroup/kogito-bom/pull/75

examples can be found in kogito quick starts 
https://github.com/mswiderski/kogito-quickstarts/tree/master/kogito-usertasks-with-predictions-quarkus
https://github.com/mswiderski/kogito-quickstarts/tree/master/kogito-usertasks-with-predictions-springboot

best to use spring boot as it does not depend on master of quarkus so does not require additional work.

@ruivieira @DuncanDoyle you might be interested in this....

### DIFF
--- a/addons/pom.xml
+++ b/addons/pom.xml
@@ -17,6 +17,7 @@
     <module>events</module>
     <module>process-management-addon</module>
     <module>jobs</module>
+    <module>predictions</module>
   </modules>
 
   <dependencyManagement>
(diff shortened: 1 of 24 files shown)


spolti:
Will this need a new image?

mswiderski:
@spolti not, at least not now. Depending if there will be need to have external service for predictions then we might need new image

mswiderski:
ok, this now is ready from my point of view ... with some extra test coverage.

ruivieira:
@mswiderski It seems to work as expected. lgtm!

PR #269, opened 2019-12-23
KOGITO-590: Rules should "see" process variables

evacchi:
It is now possible to declare RuleUnits implicitly. The DataSources (DataStores actually) are inferred from the process variables. e.g. if process variables are

```
String foo;
Person bar;
```

the rule unit will declare:

```
DataStore<String> foo;
DataStore<Person> bar;
```

mutation occurs in-place on the reference that is shared with the process variable.
Future work: 
- explicitly (possibly automatically) copy back and forth (useful especially when the rule service is a remote service)
- improve codegen: right now a DataStore is always used


jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/RuleSetNodeVisitor.java
```diff
@@ -102,17 +111,58 @@ public void visitNode(String factoryField, Node node, BlockStmt body, VariableSc
...
+
+        if (ruleType.isRuleUnit()) {
+            if (ruleSetNode.getInMappings().isEmpty()) {
+                GeneratedRuleUnitDescription generatedRuleUnitDescription = new GeneratedRuleUnitDescription(ruleType.getName(), contextClassLoader);
+                for (Variable v : variableScope.getVariables()) {
+                    generatedRuleUnitDescription.putDatasourceVar(v.getName(), DataStore.class.getCanonicalName(), v.getType().getStringType());
+                }
+                RuleUnitComponentFactoryImpl impl = (RuleUnitComponentFactoryImpl) RuleUnitComponentFactory.get();
+                impl.registerRuleUnitDescription(generatedRuleUnitDescription);
+            }
```
evacchi: I don't love this cast, but at the same time, I did not want to expose this in the public API. Maybe we can introduce some other interface

api/kogito-internal/src/main/java/org/kie/internal/ruleunit/RuleUnitComponentFactory.java
```diff
@@ -30,6 +30,8 @@ static RuleUnitComponentFactory get() {
 
     RuleUnitDescription createRuleUnitDescription( KiePackage pkg, Class<?> ruleUnitClass );
 
+    RuleUnitDescription createRuleUnitDescription( KiePackage pkg, String ruleUnitSimpleName );
```
evacchi: I guess we should backport this, although it will be a no-op

drools/drools-core/src/main/java/org/drools/core/ruleunit/RuleUnitDescriptionLoader.java
```diff
@@ -62,11 +62,17 @@ private RuleUnitDescription findDescription(final String ruleUnit) {
...
         } catch (final ClassNotFoundException e) {
-            nonExistingUnits.add(ruleUnit);
-            return null;
+            RuleUnitDescription ruleUnitDescription = ruleUnitComponentFactory.createRuleUnitDescription(pkg, ruleUnit);
+            if (ruleUnitDescription == null) {
+                nonExistingUnits.add(ruleUnit);
+                return null;
+            } else {
+                return ruleUnitDescription;
+            }
```
evacchi: I am not sure this change should be backported, There is no way `createRuleUnitDescription(pkg, String) != null` in the 7 series

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/RuleSetNodeVisitor.java
```diff
@@ -127,44 +177,47 @@ public void visitNode(String factoryField, Node node, BlockStmt body, VariableSc
...
-        for (Entry<String, String> entry : ruleSetNode.getOutMappings().entrySet()) {
-            addFactoryMethodWithArgs(body, "ruleSetNode" + node.getId(), "outMapping", new StringLiteralExpr(entry.getKey()), new StringLiteralExpr(entry.getValue()));
-        }
+                factory.findFirst(MethodDeclaration.class, m -> m.getNameAsString().equals("bind"))
+                        .ifPresent(m -> m.setBody(bind(variableScope, ruleSetNode, unitName)));
+                factory.findFirst(MethodDeclaration.class, m -> m.getNameAsString().equals("unit"))
+                        .ifPresent(m -> m.setBody(unit(unitName)));
+//                factory.findFirst(MethodDeclaration.class, m -> m.getNameAsString().equals("unbind"))
+//                        .ifPresent(m -> m.setBody());
+            });
```
evacchi: this is duplicated code, really. We should schedule a refactoring later

kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/IncrementalRuleCodegen.java
```diff
@@ -286,25 +288,31 @@ public ApplicationSection section() {
...
-        SessionsPool sessionsPoolAnn = ruleUnit.getAnnotation( SessionsPool.class );
-        if (sessionsPoolAnn != null && sessionsPoolAnn.value() > 0) {
-            unitKieBaseModel.setSessionsPool( SessionsPoolOption.get( sessionsPoolAnn.value() ) );
-        }
-        EventProcessing eventAnn = ruleUnit.getAnnotation( EventProcessing.class );
-        if (eventAnn != null && eventAnn.value() == EventProcessing.Type.STREAM) {
-            unitKieBaseModel.setEventProcessingMode( EventProcessingOption.STREAM );
+        unitKieBaseModel.addPackage(ruleUnitDescription.getPackageName());
+
+        // fixme config should go in the description as well
```
evacchi: these fixmes are now captured in [KOGITO-827](https://issues.redhat.com/browse/KOGITO-827)

kogito-codegen/src/test/java/org/kie/kogito/codegen/AbstractCodegenTest.java
```diff
@@ -99,6 +99,16 @@ protected Application generateCode(List<String> processResources, List<String> r
...
+        if (logger.isDebugEnabled()) {
+            Path temp = Files.createTempDirectory("KOGITO_TESTS");
+            logger.debug("Dumping generated files in " + temp);
+            for (GeneratedFile entry : generatedFiles) {
+                Path fpath = temp.resolve(entry.relativePath());
+                fpath.getParent().toFile().mkdirs();
+                Files.write(fpath, entry.contents());
+            }
+        }
+
```
evacchi: I have added this utility: when DEBUG mode is on, the files are also dumped to a temp directory

api/kogito-internal/src/main/java/org/kie/internal/ruleunit/RuleUnitComponentFactory.java
evacchi: see [DROOLS-4895](https://issues.redhat.com/browse/DROOLS-4895)

drools/drools-core/src/main/java/org/drools/core/ruleunit/RuleUnitDescriptionLoader.java
evacchi: see [DROOLS-4895](https://issues.redhat.com/browse/DROOLS-4895)

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/RuleSetNodeVisitor.java
evacchi: (the comment is there to show that the *entire* snippet is duplicated

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/RuleSetNodeVisitor.java
```diff
@@ -127,44 +177,47 @@ public void visitNode(String factoryField, Node node, BlockStmt body, VariableSc
...
-        }
-        for (Entry<String, String> entry : ruleSetNode.getOutMappings().entrySet()) {
-            addFactoryMethodWithArgs(body, "ruleSetNode" + node.getId(), "outMapping", new StringLiteralExpr(entry.getKey()), new StringLiteralExpr(entry.getValue()));
-        }
+                factory.findFirst(MethodDeclaration.class, m -> m.getNameAsString().equals("bind"))
+                        .ifPresent(m -> m.setBody(bind(variableScope, ruleSetNode, unitName)));
+                factory.findFirst(MethodDeclaration.class, m -> m.getNameAsString().equals("unit"))
+                        .ifPresent(m -> m.setBody(unit(unitName)));
+//                factory.findFirst(MethodDeclaration.class, m -> m.getNameAsString().equals("unbind"))
+//                        .ifPresent(m -> m.setBody());
```
mswiderski: maybe remove the commented code

kogito-codegen/src/test/java/org/kie/kogito/codegen/tests/BusinessRuleUnitTest.java
```diff
@@ -188,6 +188,38 @@ public void ioMapping() throws Exception {
 
     }
 
+
+    @Test
+    public void ioMappingGenerated() throws Exception {
```
mswiderski: would suggest to provide bit more meaningful method name

kogito-codegen/src/test/resources/ruletask/ExampleGenerated.bpmn
```diff
@@ -0,0 +1,156 @@
...
+        </drools:metaData>
+        <drools:metaData name="isAutoConnection.target">
+          <drools:metaValue>true</drools:metaValue>
+        </drools:metaData>
+      </bpmn2:extensionElements>
+    </bpmn2:sequenceFlow>
+    <bpmn2:endEvent id="_B89C98A9-F577-4B07-BB72-52283EE38A77">
+      <bpmn2:incoming>_D8A8D35C-3BBE-461E-8C67-682A3CEE6ACD</bpmn2:incoming>
+    </bpmn2:endEvent>
+    <bpmn2:businessRuleTask id="_5293F325-A06A-469B-8C31-1E79EE36A3E2" drools:ruleFlowGroup="unit:ruletask.Generated" name="Rule2" implementation="http://www.jboss.org/drools/rule">
```
mswiderski: so this rule flow group value has to map to package and unit name defined in DRL or?

kogito-codegen/src/test/resources/ruletask/Generated.drl
```diff
@@ -0,0 +1,11 @@
+package ruletask;
+unit Generated;
+
+import org.kie.kogito.codegen.data.Person;
+
+rule singlePerson when
+    p: /singlePerson[ age >= 18 ]
```
mswiderski: can you use standard (no oopath) syntax? I tried it with just Person(age >= 18) and it didn't work...

kogito-codegen/src/test/resources/ruletask/Generated.drl
```diff
@@ -0,0 +1,11 @@
...
+then
+    p.setAdult(true);
```
mswiderski: can you also add or remove data which would result in setting or removing process variables?

kogito-codegen/src/test/resources/ruletask/Generated.drl
danielezonca: @mswiderski 
DS are named entry point so you can rewrite that OOPath expression as
`p: Person( age >= 18 ) from singlePerson`

mswiderski: thanks @danielezonca

kogito-codegen/src/test/resources/ruletask/Generated.drl
evacchi:
```
singlePerson.delete(p)
```

kogito-codegen/src/test/resources/ruletask/ExampleGenerated.bpmn
evacchi: yes it is not actually a ruleFlowGroup that's just special syntax that's interpreted as a rule unit identifier. I chose this so that it can be retrofitted for the designer without changes. However, an alternative syntax is to just use `implemenation="ruletask.Generated" and language="http://.../rule-unit"` (don't remember the URI right now) but that requires a change in the designer UI

kogito-codegen/src/test/java/org/kie/kogito/codegen/tests/BusinessRuleUnitTest.java
evacchi: ok

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/RuleSetNodeVisitor.java
evacchi: ok

kogito-codegen/src/test/resources/ruletask/Generated.drl
evacchi: (I think you can also use `remove(p)`)

mswiderski: `singlePerson.remove(p)` does not seem to remove the value from process.

what I had in mind is that we should allow to set one process variable based on others that are evaluated as rules. So we have the person and string here so based on person we should be able to set the string - like if there is adult person then we should set "you can drink" as string.

similar we should be able to remove given value like if person is not adult we should set the string to null.

so it's not about adding new variables but more of changing values of defined variables.

evacchi: oh in that case, you can refer to any variable (as long as the process declares it), but it's turned into a data source so you always have to use this dot notation (variables are always wrapped in a data source of some kind). We may refine this further with syntactic sugar, though

e.g. if you have a variable `string` of type `String`

you may (currently) write `string.add("you can drink")`

we could iteratively refine this by allowing "singleton" values and add syntactic sugar so that you can effectively write

`string = "you can drink"`

or something like that

evacchi: the reason why the process is not updated when you remove/add is probably because the `unbind` part is not implemented yet (the lines that were commented away), so that's why some changes are not reflected

mswiderski: ok, made some implementation to provide support for it, sent PR to your branch https://github.com/evacchi/kogito-runtimes/pull/1 please have a look and see if it is of any use

PR #275, opened 2020-01-07
bring back removed persistence generator for kogito maven plugin

mswiderski:
@evacchi these two where removed by https://github.com/kiegroup/kogito-runtimes/pull/260 which breaks persistence completely for kogito maven plugin.

### DIFF
--- a/kogito-maven-plugin/src/main/java/org/kie/kogito/maven/plugin/ProcessClassesMojo.java
+++ b/kogito-maven-plugin/src/main/java/org/kie/kogito/maven/plugin/ProcessClassesMojo.java
@@ -1,0 +1,158 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.maven.plugin;
+
+
+import java.io.File;
+import java.io.IOException;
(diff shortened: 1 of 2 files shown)


kogito-maven-plugin/src/main/java/org/kie/kogito/maven/plugin/ProcessClassesMojo.java
```diff
@@ -0,0 +1,160 @@
...
+import org.reflections.util.ConfigurationBuilder;
+
+@Mojo(name = "process-model-classes",
+      requiresDependencyResolution = ResolutionScope.RUNTIME,
+      requiresProject = true,
+      defaultPhase = LifecyclePhase.PROCESS_CLASSES)
+public class ProcessClassesMojo extends AbstractKieMojo {
+    
+    @SuppressWarnings("deprecation")
+    private static final JavaCompiler JAVA_COMPILER = JavaCompilerFactory.getInstance().loadCompiler(JavaDialectConfiguration.CompilerType.NATIVE, "1.8");
```
MarianMacik: SonarCloud has noticed:
From [drools](https://github.com/kiegroup/drools/blob/734e83196b1ec8c40b2cc8c24872e130f5e09c82/drools-compiler/src/main/java/org/drools/compiler/commons/jci/compilers/JavaCompilerFactory.java#L42-L47):
```
@deprecated will be remove after the next release, please create an instance yourself
public static JavaCompilerFactory getInstance() {
```

Though it has been there for ~9 years or even more, so not sure if it is relevant. But it would be better to not use deprecated methods in new code.

mswiderski: @MarianMacik updated, hope this time will be good

MarianMacik: In the end Mario decided to make it a singleton using enum and leave that method, so you can revert back. Sorry for the confusion :) But at least we dismissed the deprecation which shouldn't have been there.

mswiderski: reverted

MarianMacik:
Looks good now, we can merge this.

PR #279, opened 2020-01-13
[KOGITO-432] allow to declare RuleUnit data sources in drl

mariofusco:


mariofusco:
Won't merge: we decided to generate rule units from declared types instead of globals.

PR #280, opened 2020-01-15
KOGITO-887 - User tasks from embedded subprocess should be exposed as REST endpoints

mswiderski:
…cuirsevely by default


jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/CompositeContextNodeVisitor.java
```diff
@@ -40,10 +41,15 @@ public CompositeContextNodeVisitor(Map<Class<?>, AbstractVisitor> nodesVisitors)
 
     @Override
     public void visitNode(String factoryField, Node node, BlockStmt body, VariableScope variableScope, ProcessMetaData metadata) {
-        CompositeContextNode compositeContextNode = (CompositeContextNode) node;
+        CompositeContextNode compositeContextNode = (CompositeContextNode) node;
```
Christopher-Chianelli: Formatting

PR #287, opened 2020-01-17
KOGITO-828: RuleSetNodeVisitor: refactor duplicated code 

evacchi:
Improved code everywhere in rule unit handling. All tests should still pass.

I actually spotted a bug in the way "singleton" process variables are handled. Because currently we are using a "collection" data source ("DataStore"), the variable may be set to a "previous" value (ordering is not guaranteed). I have fixed this in the test case by first removing the contents of the variable and then setting it (see Generated.drl)

This will be fixed when we introduce "singleton" data sources (KOGITO-892)


kogito-codegen/src/test/resources/logback-test.xml
```diff
@@ -17,7 +17,7 @@
 
 <!--  <logger name="org.kie.kogito.codegen" level="debug"/>-->
 
-  <root level="info">
+  <root level="debug">
```
mswiderski: maybe let's get back to info

kogito-codegen/src/test/resources/ruletask/Generated.drl
```diff
@@ -13,8 +13,10 @@ end
 
 rule singlePerson_add_note when
     p: /singlePerson[ age == 50 ]
+    s: /singleString[ this == "hello" ]
 then
     p.setAdult(true);
+    singleString.remove(s);
```
mswiderski: what's the reason for this change?

evacchi: description above
> Because currently we are using a "collection" data source ("DataStore"), the variable may be set to a "previous" value (ordering is not guaranteed). I have fixed this in the test case by first removing the contents of the variable and then setting it (see Generated.drl)

kogito-codegen/src/test/resources/logback-test.xml
evacchi: thanks

kogito-codegen/src/test/resources/ruletask/Generated.drl
evacchi: basically, previous version was ordering dependent and the refactoring showed that it was broken (it was sometimes green but most times red)
this will be fixed in KOGITO-892

mswiderski: ok, thanks for explanation

PR #293, opened 2020-01-23
KOGITO-885 [Jobs Service] Add health check for Jobs Service

tiagodolphine:
Depends on https://github.com/kiegroup/kogito-bom/pull/81


addons/persistence/infinispan-quarkus-health/pom.xml
```diff
@@ -0,0 +1,74 @@
...
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.testcontainers</groupId>
+      <artifactId>testcontainers</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
```
ricardozanini: The scope is also inherited?

tiagodolphine: no, it should be `test` scope, great catch.

tiagodolphine: done

addons/persistence/pom.xml
```diff
@@ -11,5 +11,6 @@
   <description>Kogito Persistence Implementations</description>
   <modules>
     <module>infinispan-persistence-addon</module>
+    <module>infinispan-quarkus-health</module>
```
MarianMacik: Just an organizational question. Since this is really not an addon, rather a permanent part of jobs-service, wouldn't it make more sense to put it in the job-service module? Or do we want to include it into the Kogito generated app as well in the future? Then I guess it won't make sense for Spring Boot runtime.

addons/persistence/infinispan-quarkus-health/src/test/java/org/kie/kogito/infinispan/InfinispanServerTestResource.java
```diff
@@ -0,0 +1,58 @@
...
+
+import io.quarkus.test.common.QuarkusTestResourceLifecycleManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.FixedHostPortGenericContainer;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
+import org.testcontainers.containers.wait.strategy.Wait;
+
+public class InfinispanServerTestResource implements QuarkusTestResourceLifecycleManager {
```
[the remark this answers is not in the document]
cristianonicolai: +1 to unify, we need to create a shared module for these test resources.

addons/persistence/pom.xml
cristianonicolai: it will be included into data index as well.

addons/persistence/infinispan-quarkus-health/src/test/java/org/kie/kogito/infinispan/InfinispanServerTestResource.java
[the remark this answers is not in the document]
cristianonicolai: I'm happy for that to be handled on another JIRA/PR.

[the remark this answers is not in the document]
MarianMacik: Sure, I can probably plan something for next week/sprint.

addons/persistence/pom.xml
tiagodolphine: the reason as Cristiano mentioned for this module, is to be shared with whatever service that uses infinispan. I chose this place to insert but we can move it to any other module.

addons/persistence/infinispan-quarkus-health/src/test/java/org/kie/kogito/infinispan/InfinispanServerTestResource.java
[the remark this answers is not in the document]
tiagodolphine: +1

addons/persistence/infinispan-quarkus-health/src/main/java/org/kie/kogito/infinispan/health/InfinispanHealthCheck.java
```diff
@@ -0,0 +1,103 @@
...
+                                               return false;
+                                           }
+                                       })
+                                       .allMatch(Boolean.FALSE::equals))
+                    .map(allDown -> buildDownResponse(channelFactory, !allDown))
+                    .orElse(buildDownResponse(channelFactory, false));
+        }).orElse(null);
+    }
+
+    private HealthCheckResponse buildDownResponse(ChannelFactory channelFactory, boolean state) {
```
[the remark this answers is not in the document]
tiagodolphine: Done.

addons/jobs/jobs-service/src/main/resources/application.properties
```diff
@@ -64,8 +64,10 @@ kogito.service.url=http://localhost:8080
 kogito.jobs-service.events-support=false
 
 #enabled with the profile: 'events-support' (-Dquarkus.profile=events-support)
+%events-support.quarkus.kafka.health.enabled=true
+%events-support.quarkus.kafka.bootstrap-servers=localhost:9092
```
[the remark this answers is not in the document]
tiagodolphine: yes, it must be overridden on openshift and it could be overridden in case, in your local testing you use a Kafka cluster on a different host other than the localhost, np. 
This property is set with localhost just to make things easier, to see the property name and override, this does not affect any behavior just explicitly show the default is `localhost:9092`.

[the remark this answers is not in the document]
MarianMacik: ok, thanks

addons/persistence/pom.xml
MarianMacik: OK, makes sense, then I would just make the name consistent with other addons, e.g. `infinispan-quarkus-health-addon`. Wdyt? So users can in future add it into the kogito app knowing it is an addon. I guess then we will provide code generation to expose the healthcheck endpoint in case addon is on classpath.

tiagodolphine: done

PR #294, opened 2020-01-28
KOGITO-892: Create Singleton DataSource

evacchi:


evacchi:
the "bugs" reported by SonarCloud are not actually bugs; they are deliberate
the field is initially `null`, when the first assingment is made the value is wrapped in an option which may become empty. When the field is null we do not propagate "inserts" to subscribers, because there are no values inside.

drools/kogito-ruleunits/src/main/java/org/kie/kogito/rules/units/FieldDataStore.java
```diff
@@ -0,0 +1,106 @@
...
+import org.drools.core.spi.Activation;
+import org.drools.core.util.bitmask.BitMask;
+import org.kie.api.runtime.rule.FactHandle;
+import org.kie.kogito.rules.DataHandle;
+import org.kie.kogito.rules.DataProcessor;
+import org.kie.kogito.rules.SingletonDataStore;
+import org.kie.kogito.rules.units.impl.DataHandleImpl;
+
+public class FieldDataStore<T> implements SingletonDataStore<T>, InternalDataStore<T> {
+    private Optional<T> value = null;
```
mariofusco: Replace this Optional with a plain value plus an enum to track in which of the possible 3 states UNINITALIZED, EMPTY, PRESENT.

drools/kogito-ruleunits/src/main/java/org/kie/kogito/rules/units/FieldDataStore.java
```diff
@@ -0,0 +1,106 @@
...
+    private DataHandle handle = null;
+
+    private final List<EntryPointDataProcessor> entryPointSubscribers = new ArrayList<>();
+    private final List<DataProcessor<T>> subscribers = new ArrayList<>();
+
+    public DataHandle set(T t) {
+        value = Optional.of(t);
+        handle = new DataHandleImpl( t );
+        entryPointSubscribers.forEach(s -> internalInsert(handle, s));
```
mariofusco: Invoke insert only the first time and hold the handle. The subsequent times invoke update on that handle.

evacchi: apparently DataHandle has no `update()` method 🤔

evacchi: I have added a setObject to the DataHandle

PR #297, opened 2020-01-29
KOGITO-888: Use properties to configure rules

evacchi:
add support for config properties in `application.conf`

```
kogito.rules."my.rule.Unit".event-processing-type = [CLOUD | STREAM] # default = CLOUD
kogito.rules."my.rule.Unit".clock-type" = [REALTIME | PSEUDO] # default = REALTIME
kogito.rules."my.rule.Unit".sessions-pool = [>1] # default = null
```

### DIFF
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/IncrementalRuleCodegen.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/IncrementalRuleCodegen.java
@@ -21,12 +21,12 @@ import java.io.UncheckedIOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.OptionalInt;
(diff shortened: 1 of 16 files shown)


kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/IncrementalRuleCodegen.java
```diff
@@ -295,31 +308,26 @@ public ApplicationSection section() {
...
-        if (ruleUnit != null) {
-            SessionsPool sessionsPoolAnn = ruleUnit.getAnnotation(SessionsPool.class);
-            if (sessionsPoolAnn != null && sessionsPoolAnn.value() > 0) {
-                unitKieBaseModel.setSessionsPool(SessionsPoolOption.get(sessionsPoolAnn.value()));
-            }
-            EventProcessing eventAnn = ruleUnit.getAnnotation(EventProcessing.class);
-            if (eventAnn != null && eventAnn.value() == EventProcessing.Type.STREAM) {
-                unitKieBaseModel.setEventProcessingMode(EventProcessingOption.STREAM);
-            }
+        RuleUnitConfig config = ruleUnitDescription.getConfig();
```
mariofusco: You should get the config from the map and override this is it is present.

PR #299, opened 2020-02-03
KOGITO-889: Uniform names for process/rules data definitions (WIP)

evacchi:


api/kogito-api/src/main/java/org/kie/kogito/process/ProcessData.java
```diff
@@ -13,15 +13,17 @@
...
 import java.util.Map;
 
+import org.kie.kogito.DataContext;
+
 /**
  * Represents data model type of objects that are usually descriptor of data holders.
  *
  */
-public interface Model {
+public interface ProcessData extends DataContext {
```
mswiderski: This is backward incompadible change that will break most of already created applications, mostly around tests that rely on Model as generic type

evacchi: I know, this is why I would like to do this before 1.0

evacchi: in fact, I would even go as far as just using DataContext and share it between processes and rule units

mswiderski: so why not leaving it as model? it is generic enough name in my opinion and the Context is way overloaded term especially in our code base

evacchi: that is true, but also `model` is quite overloaded :/

mswiderski: don't think it is extensively used within our code base ... overall Model is overloaded but provides bit more info than context

evacchi:
will redo for 0.10+

PR #301, opened 2020-02-03
KOGITO-1015: Fix usages of javax.annotation.{PostConstruct, ...}

evacchi:
master is currently broken (not sure why), this change also removes usages of such annotations in some places where it's not necessary (e.g. PostConstruct can be usually substituted by a generated constructor)

### DIFF
--- a/kogito-codegen/src/main/resources/class-templates/MessageConsumerTemplate.java
+++ b/kogito-codegen/src/main/resources/class-templates/MessageConsumerTemplate.java
@@ -25,9 +25,13 @@ public class $Type$MessageConsumer {
     Optional<Boolean> useCloudEvents = Optional.of(true);
     
     private ObjectMapper json = new ObjectMapper();
-        
+
     {
         json.setDateFormat(new StdDateFormat().withColonInTimeZone(true).withTimeZone(TimeZone.getDefault()));
+    }
+
+    public void configure() {
+
     }
     
 	public void consume(String payload) {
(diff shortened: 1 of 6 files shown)


kogito-codegen/src/main/resources/class-templates/MessageConsumerTemplate.java
```diff
@@ -28,7 +26,7 @@
     
     private ObjectMapper json = new ObjectMapper();
         
-    public void configure() {
+    {
```
evacchi: I have substituted the explicitly-annotated method with an anonymous constructor. This works even if the class gets renamed.

kogito-codegen/src/main/resources/class-templates/MessageConsumerTemplate.java
```diff
@@ -27,7 +25,9 @@
     Optional<Boolean> useCloudEvents = Optional.of(true);
     
     private ObjectMapper json = new ObjectMapper();
-        
+
+    { configure(); }
+
```
evacchi: instead of relying on annotations, I have added an anonymous constructor (i.e., an empty constructor) that invokes `configure();` the anonymous constructor is always invoked when you instantiate the class, but it does not require the name to be the same as the class, so codegen does not break

evacchi:
I have left the configure() method as it is, moved the ObjectMapper init to a constructor (no need for postcostruct for that), left the annotations where necessary, added back the dependency on javax.annotation to kogito-codegen tests

PR #302, opened 2020-02-04
KOGITO-458 Add health check for Data Index service

tiagodolphine:
Adding the health check for kafka and infinispan in the same way as the jobs-service.

### DIFF
--- a/data-index/data-index-storage/data-index-storage-infinispan/src/main/java/org/kie/kogito/index/infinispan/InfinispanHealthCheckProducer.java
+++ b/data-index/data-index-storage/data-index-storage-infinispan/src/main/java/org/kie/kogito/index/infinispan/InfinispanHealthCheckProducer.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.index.infinispan;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.inject.Instance;
+import javax.enterprise.inject.Produces;
+
+import org.eclipse.microprofile.health.Readiness;
+import org.infinispan.client.hotrod.RemoteCacheManager;
+import org.kie.kogito.infinispan.health.InfinispanHealthCheck;
+
+@ApplicationScoped
+public class InfinispanHealthCheckProducer {
+
+    @Produces
+    @Readiness
+    public InfinispanHealthCheck infinispanHealthCheck(Instance<RemoteCacheManager> cacheManagerInstance) {
+        return new InfinispanHealthCheck(cacheManagerInstance);
+    }
+}
--- a/data-index/data-index-service/pom.xml
+++ b/data-index/data-index-service/pom.xml
@@ -52,6 +52,12 @@
       <artifactId>quarkus-oidc</artifactId>
     </dependency>
 
+    <!-- Health Check -->
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-smallrye-health</artifactId>
+    </dependency>
+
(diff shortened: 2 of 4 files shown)


tiagodolphine:
@MarianMacik do you mind to take a look at this, it is basically the same as the jobs-service health-check.

ricardozanini:
@tiagodolphine Sonar is complaining about coverage of new code. Might be worth adding some?

data-index/data-index-storage/data-index-storage-infinispan/src/main/java/org/kie/kogito/index/infinispan/InfinispanHealthCheckProducer.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.eclipse.microprofile.health.Readiness;
+import org.infinispan.client.hotrod.RemoteCacheManager;
+import org.kie.kogito.infinispan.health.InfinispanHealthCheck;
+
+@ApplicationScoped
+public class InfinispanHealthCheckProducer {
+
+    @Produces
+    @Readiness
```
spolti: is there a liveness check as well?

tiagodolphine: @spolti, in this case, this is a `Readiness` check, because infinispan may be down and up, in the same way as kafka, so the state may change over time depending on external services, it will give the information if the application is `ready`. The `Liveness` could make sense in case our service (in this case data-index) achieved an unrecoverable state (out of memory for instance) and must be restarted/killed, giving the information if the application is `alive`.
For now, we are adding the Readiness health check, and if we face any case to check the Liveness we can add in the future.

tiagodolphine:
@ricardozanini well the only new code we are adding on the PR is a `Producer` that basically instantiates a new object for the `InfinispanHealthCheck`, there is no logic to be tested. IMHO there is no need to test it, wdyt?

data-index/data-index-storage/data-index-storage-infinispan/src/main/java/org/kie/kogito/index/infinispan/InfinispanHealthCheckProducer.java
spolti: It is important for OpenShift IMHO, with a proper liveness probe OCP can use it to guarantee that the pod will be restarted in case the liveness fails.

tiagodolphine: For OpenShift/k8s point for view the liveness health check to be used is `/health/live`, and the implementation is up to each service, and for now it will return HTTP 200 ok for liveness, the default implementation on quarkus health, in the future if we face any case that may break the application to a "dead" state we can add this Liveness implementation, but the API remains the same.

spolti: Ok, sounds good then, thanks for the explanation.

tiagodolphine: Sure, it was worth our discussion in this way we are all aligned, and maybe we can add this on runtime as well, thank you for bringing this up.

@mswiderski @cristianonicolai do we have the health check on the runtime?

mswiderski: No, as we expect to define it by users to properly handle that. We could think on what might be valid to check from engines point of view as most of the cases are compile checks when it comes to pure engine but there might be more depending whar components are used, eg infinispan, jobs service etc

tiagodolphine: Yeah, maybe we can have some default health check that checks the used dependencies like you said, infinispan, jobs service etc, in case the user wants to define more, he can extend or customize, anyway just brainstorming, not the scope of this PR :)

MarianMacik: +1 for connection with other components. We can detect which components the kogito app will be connected to when compiling assets/endpoints and based on that generate health checks, i.e. if we use persistence, have a persistence health check as we now have for jobs and data index.

ricardozanini:
> @ricardozanini well the only new code we are adding on the PR is a `Producer` that basically instantiates a new object for the `InfinispanHealthCheck`, there is no logic to be tested. IMHO there is no need to test it, wdyt?

Well, tell that to SonarCloud :p

PR #303, opened 2020-02-04
KOGITO-19 - Tagging of process variables

mswiderski:
@cristianonicolai @tiagodolphine @evacchi 
Supported/interpreted tags
- internal - makes variable to not be exposed in the model and consider as process instance internal variable
- required - requires to be given to start process instance
- readonly - can be set only once
- input - marks given variable as an input of the process and by that not exposed in the returned data model
- output - marks given variable as output and thus not expecting it on starting and will be included in the returned data model

When variable is meant for both input and output no tag shall be assigned.

This also provides dedicated input and output model that is exposed on REST api and by that open api spec

Screenshot 2020-02-04 at 13 49 00

in the screenshot above
- name - is both input and output
- traveller - is just input
- review is just output
- id - is an extra field that represents unique id of the process instance

### DIFF
--- a/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractCompositeNodeVisitor.java
+++ b/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractCompositeNodeVisitor.java
@@ -24,6 +24,7 @@ import org.jbpm.process.core.datatype.impl.type.ObjectDataType;
 import org.kie.api.definition.process.Node;
 
 import com.github.javaparser.ast.NodeList;
+import com.github.javaparser.ast.expr.NullLiteralExpr;
 import com.github.javaparser.ast.expr.ObjectCreationExpr;
 import com.github.javaparser.ast.expr.StringLiteralExpr;
 import com.github.javaparser.ast.stmt.BlockStmt;
@@ -68,9 +69,10 @@ public class AbstractCompositeNodeVisitor extends AbstractVisitor {
                 if (!visitedVariables.add(variable.getName())) {
                     continue;
                 }
+                String tags = (String) variable.getMetaData(Variable.VARIABLE_TAGS);
                 ClassOrInterfaceType variableType = new ClassOrInterfaceType(null, ObjectDataType.class.getSimpleName());
                 ObjectCreationExpr variableValue = new ObjectCreationExpr(null, variableType, new NodeList<>(new StringLiteralExpr(variable.getType().getStringType())));
-                addFactoryMethodWithArgs(contextNode, body, "variable", new StringLiteralExpr(variable.getName()), variableValue);
+                addFactoryMethodWithArgs(contextNode, body, "variable", new StringLiteralExpr(variable.getName()), variableValue, new StringLiteralExpr(Variable.VARIABLE_TAGS), (tags != null ? new StringLiteralExpr(tags) : new NullLiteralExpr()));
             }
         }
     }
--- a/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ProcessToExecModelGenerator.java
+++ b/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ProcessToExecModelGenerator.java
@@ -65,6 +65,7 @@ import com.github.javaparser.ast.expr.BooleanLiteralExpr;
 import com.github.javaparser.ast.expr.LongLiteralExpr;
 import com.github.javaparser.ast.expr.MethodCallExpr;
 import com.github.javaparser.ast.expr.NameExpr;
+import com.github.javaparser.ast.expr.NullLiteralExpr;
 import com.github.javaparser.ast.expr.ObjectCreationExpr;
 import com.github.javaparser.ast.expr.StringLiteralExpr;
 import com.github.javaparser.ast.expr.VariableDeclarationExpr;
@@ -155,7 +156,26 @@ public class ProcessToExecModelGenerator extends AbstractVisitor {
         String name = extractModelClassName(process.getId());
 
         return new ModelMetaData(process.getId(), packageName, name, process.getVisibility(),
-                                 VariableDeclarations.of((VariableScope) ((org.jbpm.process.core.Process) process).getDefaultContext(VariableScope.VARIABLE_SCOPE)));
+                                 VariableDeclarations.of((VariableScope) ((org.jbpm.process.core.Process) process).getDefaultContext(VariableScope.VARIABLE_SCOPE)),
+                                 false);
+    }
+    
+    public ModelMetaData generateInputModel(WorkflowProcess process) {
+        String packageName = process.getPackageName();
+        String name = extractModelClassName(process.getId()) + "Input";
+
+        return new ModelMetaData(process.getId(), packageName, name, process.getVisibility(),
+                                 VariableDeclarations.ofInput((VariableScope) ((org.jbpm.process.core.Process) process).getDefaultContext(VariableScope.VARIABLE_SCOPE)),
+                                 true, "/class-templates/ModelNoIDTemplate.java");
+    }
+    
+    public ModelMetaData generateOutputModel(WorkflowProcess process) {
+        String packageName = process.getPackageName();
+        String name = extractModelClassName(process.getId()) + "Output";
+
+        return new ModelMetaData(process.getId(), packageName, name, process.getVisibility(),
+                                 VariableDeclarations.ofOutput((VariableScope) ((org.jbpm.process.core.Process) process).getDefaultContext(VariableScope.VARIABLE_SCOPE)),
+                                 true);
     }
 
     public static String extractModelClassName(String processId) {
@@ -234,9 +254,10 @@ public class ProcessToExecModelGenerator extends AbstractVisitor {
                 if (!visitedVariables.add(variable.getName())) {
                     continue;
                 }
+                String tags = (String) variable.getMetaData(Variable.VARIABLE_TAGS);
                 ClassOrInterfaceType variableType = new ClassOrInterfaceType(null, ObjectDataType.class.getSimpleName());
                 ObjectCreationExpr variableValue = new ObjectCreationExpr(null, variableType, new NodeList<>(new StringLiteralExpr(variable.getType().getStringType())));
-                addFactoryMethodWithArgs(FACTORY_FIELD_NAME, body, "variable", new StringLiteralExpr(variable.getName()), variableValue);
+                addFactoryMethodWithArgs(FACTORY_FIELD_NAME, body, "variable", new StringLiteralExpr(variable.getName()), variableValue, new StringLiteralExpr(Variable.VARIABLE_TAGS), tags != null ? new StringLiteralExpr(tags) : new NullLiteralExpr());
             }
         }
     }
--- a/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/VariableDeclarations.java
+++ b/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/VariableDeclarations.java
@@ -17,6 +17,7 @@ package org.jbpm.compiler.canonical;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.function.Predicate;
 
 import org.jbpm.process.core.context.variable.Variable;
 import org.jbpm.process.core.context.variable.VariableScope;
@@ -26,6 +27,32 @@ public class VariableDeclarations {
     public static VariableDeclarations of(VariableScope vscope) {
         HashMap<String, String> vs = new HashMap<>();
         for (Variable variable : vscope.getVariables()) {
+            if (variable.hasTag(Variable.INTERNAL_TAG)) {
+                continue;
+            }
+            
+            vs.put(variable.getName(), variable.getType().getStringType());
+        }
+        return of(vs);
+    }
+    
+    public static VariableDeclarations ofInput(VariableScope vscope) {
+        
+        return of(vscope, variable -> variable.hasTag(Variable.INTERNAL_TAG) || variable.hasTag(Variable.OUTPUT_TAG));
+    }
+    
+    public static VariableDeclarations ofOutput(VariableScope vscope) {
+        
+        return of(vscope, variable -> variable.hasTag(Variable.INTERNAL_TAG) || variable.hasTag(Variable.INPUT_TAG));
+    }
+    
+    public static VariableDeclarations of(VariableScope vscope, Predicate<Variable> filterOut) {
+        HashMap<String, String> vs = new HashMap<>();
+        for (Variable variable : vscope.getVariables()) {
+            if (filterOut.test(variable)) {
+                continue;
+            }
+            
             vs.put(variable.getName(), variable.getType().getStringType());
         }
         return of(vs);
--- a/jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/Variable.java
+++ b/jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/Variable.java
@@ -17,7 +17,10 @@
 package org.jbpm.process.core.context.variable;
 
 import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import org.jbpm.process.core.TypeObject;
@@ -32,11 +35,21 @@ import org.jbpm.process.core.ValueObject;
 public class Variable implements TypeObject, ValueObject, Serializable {
 
     private static final long serialVersionUID = 510l;
+        
+    public static final String VARIABLE_TAGS = "customTags";
+    
+    public static final String READONLY_TAG = "readonly";
+    public static final String REQUIRED_TAG = "required";
+    public static final String INTERNAL_TAG = "internal";
+    public static final String INPUT_TAG = "input";
+    public static final String OUTPUT_TAG = "output";
 
     private String name;
     private DataType type;
     private Object value;
     private Map<String, Object> metaData = new HashMap<String, Object>();
+    
+    private List<String> tags = new ArrayList<>();
 
     public Variable() {
         this.type = UndefinedDataType.getInstance();
@@ -80,6 +93,10 @@ public class Variable implements TypeObject, ValueObject, Serializable {
 
     public void setMetaData(String name, Object value) {
         this.metaData.put(name, value);
+        
+        if (VARIABLE_TAGS.equals(name) && value != null) {
+            tags = Arrays.asList(value.toString().split(","));
+        }
     }
     
     public Object getMetaData(String name) {
@@ -92,5 +109,17 @@ public class Variable implements TypeObject, ValueObject, Serializable {
     
     public String toString() {
         return this.name;
+    }
+    
+    public List<String> getTags() {
+        if (tags.isEmpty() && this.metaData.containsKey(VARIABLE_TAGS)) {
+            tags = Arrays.asList(metaData.get(VARIABLE_TAGS).toString().split(","));
+            
+        }
+        return tags;
+    }
+    
+    public boolean hasTag(String tagName) {
+        return getTags().contains(tagName);
     }
 }
--- a/kogito-codegen/src/main/resources/class-templates/ReactiveRestResourceTemplate.java
+++ b/kogito-codegen/src/main/resources/class-templates/ReactiveRestResourceTemplate.java
@@ -41,14 +41,14 @@ public class $Type$ReactiveResource {
     @POST()
     @Produces(MediaType.APPLICATION_JSON)
     @Consumes(MediaType.APPLICATION_JSON)    
-    public CompletionStage<$Type$> createResource_$name$($Type$ resource) {
+    public CompletionStage<$Type$Output> createResource_$name$($Type$Input resource) {
         if (resource == null) {
-            resource = new $Type$();
+            resource = new $Type$Input();
         }
-        final $Type$ value = resource;
+        final $Type$Input value = resource;
         return CompletableFuture.supplyAsync(() -> {
             return org.kie.kogito.services.uow.UnitOfWorkExecutor.executeInUnitOfWork(application.unitOfWorkManager(), () -> {
-                ProcessInstance<$Type$> pi = process.createInstance(value);
+                ProcessInstance<$Type$> pi = process.createInstance(mapInput(value, new $Type$()));
                 pi.start();
                 return getModel(pi);
             });
@@ -57,10 +57,10 @@ public class $Type$ReactiveResource {
 
     @GET()
     @Produces(MediaType.APPLICATION_JSON)
-    public CompletionStage<List<$Type$>> getResources_$name$() {
+    public CompletionStage<List<$Type$Output>> getResources_$name$() {
         return CompletableFuture.supplyAsync(() -> {
             return process.instances().values().stream()
-                    .map(ProcessInstance::variables)
+                    .map(pi -> mapOutput(new $Type$Output(), pi.variables()))
                  .collect(Collectors.toList());
         });   
     }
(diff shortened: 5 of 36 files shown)


kogito-codegen/src/test/resources/usertask/approval.bpmn2
```diff
@@ -5,7 +5,7 @@
   <bpmn2:itemDefinition id="ItemDefinition_1" isCollection="false" structureRef="java.lang.Integer"/>
   <bpmn2:itemDefinition id="ItemDefinition_2" isCollection="false" structureRef="java.lang.Boolean"/>
   <bpmn2:process id="approvals" tns:packageName="org.acme.travels" name="approvals" isExecutable="true" processType="Public">
-    <bpmn2:property id="approver" itemSubjectRef="ItemDefinition_9" name="approver"/>
+    <bpmn2:property id="approver" itemSubjectRef="ItemDefinition_9" name="approver" />
```
cristianonicolai: perhaps can avoid changing this file completely

mswiderski:
@cristianonicolai comments addressed

cristianonicolai:
@mswiderski thanks, one final question, how about "business relevant" tag, is that still planned?

mswiderski:
@cristianonicolai yes, it is but I want to split it out into two jiras/PRs

ricardozanini:
@mswiderski how are you adding the tags to the BPMN file? Manually?

mswiderski:
@ricardozanini yes, manually, no support for metadata is available in any of the editors

cristianonicolai:
@mswiderski @ricardozanini https://issues.redhat.com/browse/KOGITO-1020

ricardozanini:
@cristianonicolai we need an EPIC for this story, do we already have one?

cristianonicolai:
@ricardozanini now we do https://issues.redhat.com/browse/KOGITO-1033 :1st_place_medal:

kogito-codegen/src/test/java/org/kie/kogito/codegen/tests/UserTaskTest.java
```diff
@@ -675,4 +677,129 @@ public void testBasicUserTaskProcessClaimReleaseClaimAndCompletePhases() throws
...
+        
+        Application app = generateCodeProcessesOnly("usertask/approval-with-readonly-variable-tags.bpmn2");        
+        assertThat(app).isNotNull();
+                
+        Class<?> resourceClazz = Class.forName("org.acme.travels.ApprovalsModel", true, testClassLoader());
+        assertNotNull(resourceClazz);
+        
+        Field approverField = resourceClazz.getDeclaredField("approver");
+        assertThat(approverField).isNotNull();
+        assertThat(approverField.getType().getCanonicalName()).isEqualTo(String.class.getCanonicalName());
```
evacchi: [I think there is a shorthand assertj method for this](https://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html#extracted-properties-assertion)

```java
// "name" needs to be either a property or a field of the TolkienCharacter class
assertThat(fellowshipOfTheRing).extracting("name")
                               .contains("Boromir", "Gandalf", "Frodo", "Legolas")
                               .doesNotContain("Sauron", "Elrond");
```

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ProcessToExecModelGenerator.java
```diff
@@ -155,7 +156,26 @@ public ModelMetaData generateModel(WorkflowProcess process) {
...
+                                 VariableDeclarations.of((VariableScope) ((org.jbpm.process.core.Process) process).getDefaultContext(VariableScope.VARIABLE_SCOPE), variable -> variable.hasTag(Variable.INTERNAL_TAG)),
+                                 false);
+    }
+    
+    public ModelMetaData generateInputModel(WorkflowProcess process) {
+        String packageName = process.getPackageName();
+        String name = extractModelClassName(process.getId()) + "Input";
+
+        return new ModelMetaData(process.getId(), packageName, name, process.getVisibility(),
+                                 VariableDeclarations.of((VariableScope) ((org.jbpm.process.core.Process) process).getDefaultContext(VariableScope.VARIABLE_SCOPE), variable -> variable.hasTag(Variable.INTERNAL_TAG) || variable.hasTag(Variable.OUTPUT_TAG)),
```
evacchi: I would move this Scope + Predicate method to `VariableDeclarations` static method; e.g. `ofOutput`

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ProcessToExecModelGenerator.java
```diff
@@ -155,7 +156,26 @@ public ModelMetaData generateModel(WorkflowProcess process) {
...
+                                 true, "/class-templates/ModelNoIDTemplate.java");
+    }
+    
+    public ModelMetaData generateOutputModel(WorkflowProcess process) {
+        String packageName = process.getPackageName();
+        String name = extractModelClassName(process.getId()) + "Output";
+
+        return new ModelMetaData(process.getId(), packageName, name, process.getVisibility(),
+                                 VariableDeclarations.of((VariableScope) ((org.jbpm.process.core.Process) process).getDefaultContext(VariableScope.VARIABLE_SCOPE), variable -> variable.hasTag(Variable.INTERNAL_TAG) || variable.hasTag(Variable.INPUT_TAG)),
+                                 true);
```
evacchi: I would move this Scope + Predicate method to `VariableDeclarations` static method; e.g. `ofInput`

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ProcessToExecModelGenerator.java
```diff
@@ -155,7 +156,26 @@ public ModelMetaData generateModel(WorkflowProcess process) {
         String name = extractModelClassName(process.getId());
 
         return new ModelMetaData(process.getId(), packageName, name, process.getVisibility(),
-                                 VariableDeclarations.of((VariableScope) ((org.jbpm.process.core.Process) process).getDefaultContext(VariableScope.VARIABLE_SCOPE)));
+                                 VariableDeclarations.of((VariableScope) ((org.jbpm.process.core.Process) process).getDefaultContext(VariableScope.VARIABLE_SCOPE), variable -> variable.hasTag(Variable.INTERNAL_TAG)),
+                                 false);
+    }
+    
+    public ModelMetaData generateInputModel(WorkflowProcess process) {
```
evacchi: maybe we can move this method entirely to InputModelClassGenerator ?

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ProcessToExecModelGenerator.java
```diff
@@ -155,7 +156,26 @@ public ModelMetaData generateModel(WorkflowProcess process) {
...
+        String packageName = process.getPackageName();
+        String name = extractModelClassName(process.getId()) + "Input";
+
+        return new ModelMetaData(process.getId(), packageName, name, process.getVisibility(),
+                                 VariableDeclarations.of((VariableScope) ((org.jbpm.process.core.Process) process).getDefaultContext(VariableScope.VARIABLE_SCOPE), variable -> variable.hasTag(Variable.INTERNAL_TAG) || variable.hasTag(Variable.OUTPUT_TAG)),
+                                 true, "/class-templates/ModelNoIDTemplate.java");
+    }
+    
+    public ModelMetaData generateOutputModel(WorkflowProcess process) {
```
evacchi: maybe we can move this method entirely to OutputModelClassGenerator ?

evacchi:
I would like (in a subsequent PR) to provide a annotations that mimick these tags, so that we can also tag rule units similarly

also, now that we have I/O model is the ProcessInstance be `ProcessInstance<I,O>` ? Could it be that we still have a `Model` that is the composition of both? so that we still only have `ProcessInstance<V>` ? then `V` could even have those annotations to represent the tags (I am thinking for a future programmatic interface here -- even if V were still generated)

e.g.:

```java
class MyProcVar implements (something) {
   @Input int var1;
   @Hidden String hidden;
   @Output int var2;
}
class MyProcInput implements InputModel {
   @Input int var1;
   @Hidden hidden
}
class MyProcOutput implements OutputModel {
   @Hidden hidden
   @Output int var2;
}
```

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ProcessToExecModelGenerator.java
mswiderski: moved

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ProcessToExecModelGenerator.java
mswiderski: moved

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ProcessToExecModelGenerator.java
mswiderski: done

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ProcessToExecModelGenerator.java
mswiderski: done

mswiderski:
@evacchi in general there is not really possible to use annotations as the input/output would only apply on process data model and not actual classes of that model. What I mean is if process has variable named `person` of type `org.test.Person` it can be tagged as input or output on process level but you can't annotate its fields like `name` or `address` as input or output as this is already too low in the hierarchy and will be ignored.

kogito-codegen/src/main/resources/class-templates/ReactiveRestResourceTemplate.java
```diff
@@ -41,14 +41,14 @@
     @POST()
     @Produces(MediaType.APPLICATION_JSON)
     @Consumes(MediaType.APPLICATION_JSON)    
-    public CompletionStage<$Type$> createResource_$name$($Type$ resource) {
+    public CompletionStage<$Type$Output> createResource_$name$($Type$Input resource) {
         if (resource == null) {
             resource = new $Type$();
         }
         final $Type$ value = resource;
```
MarianMacik: I guess this should be of type `$Type$Input` as well? Also the if above it.

kogito-codegen/src/test/java/org/kie/kogito/codegen/tests/UserTaskTest.java
```diff
@@ -675,4 +677,129 @@ public void testBasicUserTaskProcessClaimReleaseClaimAndCompletePhases() throws
         processInstance.abort();
         assertThat(processInstance.status()).isEqualTo(ProcessInstance.STATE_ABORTED);
     }
+    
+    @SuppressWarnings({"rawtypes", "unchecked"})
+    @Test
+    public void testApprovalWithVariableTags() throws Exception {
```
MarianMacik: Maybe rename to `*WithReadOnlyVariableTags`

kogito-codegen/src/test/java/org/kie/kogito/codegen/tests/UserTaskTest.java
```diff
@@ -675,4 +677,129 @@ public void testBasicUserTaskProcessClaimReleaseClaimAndCompletePhases() throws
...
+    
+    @Test
+    public void testApprovalWithIOVariableTags() throws Exception {
+        
+        Application app = generateCodeProcessesOnly("usertask/approval-with-io-variable-tags.bpmn2");        
+        assertThat(app).isNotNull();
+                
+        Class<?> modelClazz = Class.forName("org.acme.travels.ApprovalsModel", true, testClassLoader());
+        assertNotNull(modelClazz);
+        // internal variables are not exposed on the model
```
MarianMacik: But here they are exposed on the model. See `assertNotNull`.

kogito-codegen/src/test/java/org/kie/kogito/codegen/tests/UserTaskTest.java
```diff
@@ -675,4 +677,129 @@ public void testBasicUserTaskProcessClaimReleaseClaimAndCompletePhases() throws
...
+        assertNotNull(modelClazz.getDeclaredField("decision"));
+        assertNotNull(modelClazz.getDeclaredField("approver"));
+        
+        Class<?> inputModelClazz = Class.forName("org.acme.travels.ApprovalsModelInput", true, testClassLoader());
+        assertNotNull(inputModelClazz);
+        // internal variables are not exposed on the model
```
MarianMacik: Probably bad comment here?

kogito-codegen/src/test/java/org/kie/kogito/codegen/tests/UserTaskTest.java
```diff
@@ -675,4 +677,129 @@ public void testBasicUserTaskProcessClaimReleaseClaimAndCompletePhases() throws
...
+        assertNotNull(inputModelClazz.getDeclaredField("approver"));
+        assertThrows(NoSuchFieldException.class, () -> inputModelClazz.getDeclaredField("decision"));
+        assertThrows(NoSuchFieldException.class, () -> inputModelClazz.getDeclaredField("id"));
+        
+        Class<?> outputModelClazz = Class.forName("org.acme.travels.ApprovalsModelOutput", true, testClassLoader());
+        assertNotNull(outputModelClazz);
+        // internal variables are not exposed on the model
```
MarianMacik: Same here.

evacchi:
@mswiderski I understand that, but I was thinking that in rules, the "variables" are explicitly described by the end user in 1 data class; input/output classes are still generated internally (just like for processes, we generate for queries -> output)

at the same time, could it be that at some point you'll want to allow to declare data explicitly in a POJO? In this case, I would expect -- for convenience -- to require 1 POJO instead of In/Out (?)

And again, what about ProcessInstance<T> is it going to become ProcessInstance<In, Out> ?

mswiderski:
@evacchi how does below map to decisions (dmn)?

> I understand that, but I was thinking that in rules, the "variables" are explicitly described by the end user in 1 data class; input/output classes are still generated internally (just like for processes, we generate for queries -> output)

I always thought about DMN to be more like processes where it is defined and names so not that needed to be given explicitly by users. 

> at the same time, could it be that at some point you'll want to allow to declare data explicitly in a POJO? In this case, I would expect -- for convenience -- to require 1 POJO instead of In/Out (?)

kind to the other way around I'd say. Allow users to define their model in the process (similar as it is done in DMN) and free up users from any POJOs. This would then open it up even more for other developers/users than just java ones.

> And again, what about ProcessInstance is it going to become ProcessInstance<In, Out> ?
I thought about it but this proved to be way to complex and that's why I kept the process to be complete model based while narrowing down to input and output on the edge so to say - REST API.
Can be revisited later on when needed though for me the more important api is REST api as this must enable users to effectively consume service.

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/VariableDeclarations.java
```diff
@@ -26,6 +27,32 @@
...
+        return of(vscope, variable -> variable.hasTag(Variable.INTERNAL_TAG) || variable.hasTag(Variable.INPUT_TAG));
+    }
+    
+    public static VariableDeclarations of(VariableScope vscope, Predicate<Variable> filter) {
+        HashMap<String, String> vs = new HashMap<>();
+        for (Variable variable : vscope.getVariables()) {
+            if (filter.test(variable)) {
+                continue;
+            }
+
```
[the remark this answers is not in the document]
mswiderski: what would we gain by using another lambda?

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/VariableDeclarations.java
```diff
@@ -26,6 +27,32 @@
     public static VariableDeclarations of(VariableScope vscope) {
         HashMap<String, String> vs = new HashMap<>();
         for (Variable variable : vscope.getVariables()) {
+            if (variable.hasTag(Variable.INTERNAL_TAG)) {
+                continue;
+            }
+            
+            vs.put(variable.getName(), variable.getType().getStringType());
+        }
+        return of(vs);
```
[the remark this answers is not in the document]
mswiderski: what would be the gain?

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/VariableDeclarations.java
```diff
@@ -26,6 +27,32 @@
...
+    }
+    
+    public static VariableDeclarations ofInput(VariableScope vscope) {
+        
+        return of(vscope, variable -> variable.hasTag(Variable.INTERNAL_TAG) || variable.hasTag(Variable.OUTPUT_TAG));
```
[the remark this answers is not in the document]
mswiderski: parameter renamed

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractCompositeNodeVisitor.java
```diff
@@ -68,9 +69,10 @@ protected void visitVariableScope(String contextNode, VariableScope variableScop
                 if (!visitedVariables.add(variable.getName())) {
                     continue;
                 }
+                String tags = (String) variable.getMetaData("customTags");
                 ClassOrInterfaceType variableType = new ClassOrInterfaceType(null, ObjectDataType.class.getSimpleName());
                 ObjectCreationExpr variableValue = new ObjectCreationExpr(null, variableType, new NodeList<>(new StringLiteralExpr(variable.getType().getStringType())));
-                addFactoryMethodWithArgs(contextNode, body, "variable", new StringLiteralExpr(variable.getName()), variableValue);
+                addFactoryMethodWithArgs(contextNode, body, "variable", new StringLiteralExpr(variable.getName()), variableValue, new StringLiteralExpr("customTags"), (tags != null ? new StringLiteralExpr(tags) : new NullLiteralExpr()));
```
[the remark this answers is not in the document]
mswiderski: done

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ProcessToExecModelGenerator.java
```diff
@@ -234,9 +254,10 @@ private void visitVariableScope(VariableScope variableScope, BlockStmt body, Set
                 if (!visitedVariables.add(variable.getName())) {
                     continue;
                 }
+                String tags = (String) variable.getMetaData("customTags");
                 ClassOrInterfaceType variableType = new ClassOrInterfaceType(null, ObjectDataType.class.getSimpleName());
                 ObjectCreationExpr variableValue = new ObjectCreationExpr(null, variableType, new NodeList<>(new StringLiteralExpr(variable.getType().getStringType())));
-                addFactoryMethodWithArgs(FACTORY_FIELD_NAME, body, "variable", new StringLiteralExpr(variable.getName()), variableValue);
+                addFactoryMethodWithArgs(FACTORY_FIELD_NAME, body, "variable", new StringLiteralExpr(variable.getName()), variableValue, new StringLiteralExpr("customTags"), tags != null ? new StringLiteralExpr(tags) : new NullLiteralExpr());
```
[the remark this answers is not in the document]
mswiderski: done

kogito-codegen/src/main/resources/class-templates/ReactiveRestResourceTemplate.java
mswiderski: done

kogito-codegen/src/test/java/org/kie/kogito/codegen/tests/UserTaskTest.java
mswiderski: what do you mean? internal should not be exposed on model at all

kogito-codegen/src/test/java/org/kie/kogito/codegen/tests/UserTaskTest.java
mswiderski: removed

kogito-codegen/src/test/java/org/kie/kogito/codegen/tests/UserTaskTest.java
mswiderski: removed

kogito-codegen/src/test/java/org/kie/kogito/codegen/tests/UserTaskTest.java
mswiderski: renamed

mswiderski:
@evacchi @MarianMacik any other comments? or is this good enough to be merged....

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/VariableDeclarations.java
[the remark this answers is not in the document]
MarianMacik: Consistency :) And avoidance of `continue`. But I am not saying it's wrong now.

kogito-codegen/src/test/java/org/kie/kogito/codegen/tests/UserTaskTest.java
MarianMacik: I meant

```
assertNotNull(modelClazz.getDeclaredField("decision"));
assertNotNull(modelClazz.getDeclaredField("approver"));
```
So the comment was misleading.

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/Variable.java
```diff
@@ -93,4 +110,16 @@ public Object getMetaData(String name) {
     public String toString() {
         return this.name;
     }
+    
+    public List<String> getTags() {
+        if (tags.isEmpty() && this.metaData.containsKey(VARIABLE_TAGS)) {
+            tags = Arrays.asList(metaData.get(VARIABLE_TAGS).toString().split(","));
+            
+        }
+        return tags;
```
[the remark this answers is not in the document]
MarianMacik: Still holds, can you clarify?

kogito-codegen/src/test/java/org/kie/kogito/codegen/tests/UserTaskTest.java
mswiderski: comment is removed now

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/Variable.java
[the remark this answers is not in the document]
mswiderski: tags can be empty and in most of the cases will be empty. meta data can be given as one via `setMetaData` or it's done via the MetadataWrapper during bpmn2 parsing and it gets the metadata map and add data directly to that map. In such case it will need this extra processing.

> Moreover, if I pass null to setMetaData, then getTags will produce NPE.

isn't this expected?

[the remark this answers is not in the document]
MarianMacik: Because I only saw one option how to set metadata, the double-processing was not clear to me. Thanks.

> isn't this expected?

Well, you tell me if NPE is expected or not :) In general I think not, but if it's by design here, then OK. We could return empty list in case `metaData.get(VARIABLE_TAGS)` returns null.

evacchi:
sorry missed this. What are the contents of the T in ProcessInstance<T> now? :)

mswiderski:
same as before, the complete model

evacchi:
oh ok then

PR #304, opened 2020-02-04
KOGITO-536: Data-index-service Keycloak integration: GraphiQL (UI)

nmirasch:
Added a new service to be able to login on keycloak and interact with data-index-service when security is enabled.


nmirasch:
@cristianonicolai could you please take a look on this solution?
Regarding quarkus don't allow to maintain web / service authentication in the same service. I tried this interact with the graphiql service from outside quarkus app configured as open id web-app, forwarding the 'securitized' interaction with the data-index-service.

ricardozanini:
Let's also try to test this on OpenShift first before merging.

data-index/data-index-service-ui/src/main/java/org.kie.kogito.index.vertx/RouterSetup.java
```diff
@@ -0,0 +1,127 @@
...
+    private HttpRequest<Buffer> getRemoteRequest(RoutingContext ctx) {
+        String bearerToken = "";
+        if (ctx.user() != null) {
+            bearerToken = "Bearer " + ((QuarkusHttpUser) ctx.user()).getSecurityIdentity().getCredential(AccessTokenCredential.class).getToken();
+        } else {
+            bearerToken = ctx.request().headers().get("Bearer-token");
+        }
+        if (bearerToken != null) {
+
+            HttpRequest<Buffer> remoteRequest;
```
spolti: wouldn't be better use a switch case here?

data-index/data-index-service-ui/src/main/java/org.kie.kogito.index.vertx/RouterSetup.java
```diff
@@ -0,0 +1,127 @@
...
+                remoteRequest = client.deleteAbs(remoteURL);
+            } else if (method == HttpMethod.GET) {
+                remoteRequest = client.getAbs(remoteURL);
+            } else {
+                remoteRequest = client.putAbs(remoteURL);
+            }
+
+            //headers
+            remoteRequest.putHeader(HttpHeaders.AUTHORIZATION.toString(), bearerToken)
+                    .putHeader(HttpHeaders.ACCESS_CONTROL_ALLOW_ORIGIN.toString(), "*")
```
spolti: these 3 headers are used more than once, maybe create a method to add it instead have it duplicated?

nmirasch:
@ricardozanini @spolti @cristianonicolai  providing this extra service just to manage ui when security is enabled, seems not a good solution regarding this problem would be present on other runtimes that need both kind of authentication mechanisms. Regarding this current quarkus limitation, that it seems to be taken into account.. https://github.com/quarkusio/quarkus/issues/6997 I prefer to close this PR and try to provide a better solution. Thanks for your comments and reviews!

PR #307, opened 2020-02-07
[KOGITO-1023] consolidate kogito-bom and kogito-runtimes repos

mariofusco:


mariofusco:
@mswiderski @cristianonicolai I believe this PR is failing because the old bom project is still built before of this. Could you please give a look and let me know if I should do something about this?

mswiderski:
@mariofusco try removing this stage https://github.com/kiegroup/kogito-runtimes/blob/master/Jenkinsfile#L25-L34

boms/Jenkinsfile.deploy
```diff
@@ -0,0 +1,75 @@
...
+            script {
+                mailer.sendEmailFailure()
+            }
+        }
+        always {
+            junit '**/target/surefire-reports/**/*.xml'
+            cleanWs()
+        }
+    }
+}
```
mareknovotny: this file doesn't need to be merged as always we will have only Jenkinsfile in root of repository

boms/scripts/mvn-all.sh
```diff
@@ -0,0 +1,127 @@
+#!/bin/bash
```
mareknovotny: we should probably move scripts folder to the root of repository not to boms folder

mareknovotny:
I think  we should have boms on the first level of repository and merge the bom-parent with kogito-runtimes parent.

mariofusco:
jenkins test

mariofusco:
@mareknovotny I moved the scripts one folder up as you suggested. As for merging the kogito-bom-parent into kogito-runtimes-parent, I believe that we will need a kogito-bom-parent regardless as root module for kogito-bom and kogito-deps-bom and since I cannot remove it I guess it will be cleaner to keep bom related stuff in that pom. What do you think?

scripts/repository-list.txt
```diff
@@ -0,0 +1,4 @@
+kogito-bom
```
mareknovotny: this should be removed too now

mareknovotny:
I don't follow why the bom-parent is good to have in boms folder. 
Now root of kogito-runtimes has defined version properties so the kogito-bom-parent is useless imho. If we move individual boms into root it will be simpler.

boms/Jenkinsfile
```diff
@@ -0,0 +1,83 @@
+@Library('jenkins-pipeline-shared-libraries')_
```
mswiderski: this is not needed here, jenkinsfile is only needed in top level repo

boms/Jenkinsfile.deploy
```diff
@@ -0,0 +1,75 @@
+@Library('jenkins-pipeline-shared-libraries')_
```
mswiderski: same here, not needed file

boms/README.md
```diff
@@ -0,0 +1 @@
+# kogito-bom
```
mswiderski: since this is empty file maybe we don't need to at all, no longer it's own repo

mariofusco:
@mareknovotny better?

mareknovotny:
it is better @mariofusco almost ok, but one fact is that you merged kogito-bom and kogito-deps-bom which could make some trouble if user will import that and have a need to use some other 3rd party dependencies in their projects.

MarianMacik:
> so if i've understood correctly, now all submodules, except for bom (of course) are _importing_ bom -- instead of it being their parent. I am not an expert I don't think there are downsides (?) Anyway, we can always improve this later.

@evacchi 
One downside of it is that now you cannot override a dependency version (if needed) with just a simple Maven property. You have to declare a full `dependencyManagement` tag.

mareknovotny:
we can merge it as it is to not block this consolidation

cristianonicolai:
@mareknovotny should https://github.com/kiegroup/kogito-bom be archived?

Ginxo:
@cristianonicolai @mareknovotny @mariofusco the rest of the projects still inheriting from `kogito-bom-parent` and PRs like https://github.com/kiegroup/kogito-cloud/pull/89 or https://github.com/kiegroup/kogito-examples/pull/130/files don't seem to solve it. Am I missing something?

mareknovotny:
@Ginxo I think the follow up PRs are not merged, but they should be quickly finished to have all repos consistent

mareknovotny:
@cristianonicolai I archived kogito-bom now

PR #308, opened 2020-02-07
KOGITO-855 - [Jobs Service] improve persisted jobs loading

tiagodolphine:
Schedule and load jobs in chunks, expressed in a period of time that can be configured, for instance, next 10 minutes, 30 minutes,  60 minutes...
The scheduler just handles the jobs in the current chunk in this way it is not necessary to schedule jobs that are supposed to be executed in a long time in the future.

Besides, it was added the jobs load prioritization.

### DIFF
--- a/addons/jobs/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/BaseTimerJobScheduler.java
+++ b/addons/jobs/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/BaseTimerJobScheduler.java
@@ -18,9 +18,12 @@ package org.kie.kogito.jobs.service.scheduler;
 
 import java.time.Duration;
 import java.time.ZonedDateTime;
+import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Consumer;
 
 import javax.inject.Inject;
 
@@ -46,11 +49,18 @@ public abstract class BaseTimerJobScheduler implements ReactiveJobScheduler<Sche
 
     private static final Logger LOGGER = LoggerFactory.getLogger(BaseTimerJobScheduler.class);
 
-    @ConfigProperty(name = "kogito.jobs-service.backoffRetryMillis", defaultValue = "1000")
+    @ConfigProperty(name = "kogito.jobs-service.backoffRetryMillis")
     long backoffRetryMillis;
 
-    @ConfigProperty(name = "kogito.jobs-service.maxIntervalLimitToRetryMillis", defaultValue = "60000")
+    @ConfigProperty(name = "kogito.jobs-service.maxIntervalLimitToRetryMillis")
     long maxIntervalLimitToRetryMillis;
+
+    /**
+     * The current chunk size  in minutes the scheduler handles, it is used to keep a limit number of jobs scheduled
+     * in the in-memory scheduler.
+     */
+    @ConfigProperty(name = "kogito.jobs-service.schedulerChunkInMinutes")
+    long schedulerChunkInMinutes;
 
     @Inject
     JobExecutor jobExecutor;
@@ -58,7 +68,10 @@ public abstract class BaseTimerJobScheduler implements ReactiveJobScheduler<Sche
     @Inject
     ReactiveJobRepository jobRepository;
 
-    public BaseTimerJobScheduler() {
+    private final Map<String, ZonedDateTime> schedulerControl;
+
+    protected BaseTimerJobScheduler() {
+        this(null, null, 0, 0);
     }
 
     public BaseTimerJobScheduler(JobExecutor jobExecutor,
@@ -69,27 +82,45 @@ public abstract class BaseTimerJobScheduler implements ReactiveJobScheduler<Sche
         this.jobRepository = jobRepository;
         this.backoffRetryMillis = backoffRetryMillis;
         this.maxIntervalLimitToRetryMillis = maxIntervalLimitToRetryMillis;
+        this.schedulerControl = new ConcurrentHashMap<>();
     }
 
     @Override
     public Publisher<ScheduledJob> schedule(Job job) {
         LOGGER.debug("Scheduling {}", job);
         return ReactiveStreams
-                //1- check if the job is already scheduled
+                //check if the job is already scheduled and persisted
                 .fromCompletionStage(jobRepository.exists(job.getId()))
                 .flatMap(exists -> Boolean.TRUE.equals(exists)
                         ? handleExistingJob(job)
-                        : ReactiveStreams.of(Boolean.TRUE))
-                .filter(Boolean.TRUE::equals)
-                //2- calculate the delay (when the job should be executed)
-                .map(checked -> job.getExpirationTime())
+                        : ReactiveStreams.of(job))
+                .flatMap(j -> isOnCurrentSchedulerChunk(job)
+                        //in case the job is on the current bulk, proceed with scheduling process
+                        ? doJobScheduling(job)
+                        //in case the job is not on the current bulk, just save it to be scheduled later
+                        : ReactiveStreams.fromCompletionStage(jobRepository.save(ScheduledJob.builder()
+                                                                                         .job(job)
+                                                                                         .status(JobStatus.SCHEDULED)
+                                                                                         .build())))
+                .buildRs();
+    }
+
+    /**
+     * Performs the given job scheduling process on the scheduler, after all the validations already made.
+     * @param job to be scheduled
+     * @return
+     */
+    private PublisherBuilder<ScheduledJob> doJobScheduling(Job job) {
+        return ReactiveStreams.of(job)
+                //calculate the delay (when the job should be executed)
+                .map(current -> job.getExpirationTime())
                 .map(this::calculateDelay)
                 .peek(delay -> Optional
                         .of(delay.isNegative())
                         .filter(Boolean.FALSE::equals)
                         .orElseThrow(() -> new RuntimeException("Delay should be positive")))
-                //3- schedule the job
-                .map(delay -> doSchedule(delay, job))
+                //schedule the job on the scheduler
+                .map(delay -> schedule(delay, job))
                 .flatMap(p -> p)
                 .map(scheduleId -> ScheduledJob
                         .builder()
@@ -98,11 +129,18 @@ public abstract class BaseTimerJobScheduler implements ReactiveJobScheduler<Sche
                         .status(JobStatus.SCHEDULED)
                         .build())
                 .map(scheduledJob -> jobRepository.save(scheduledJob))
-                .flatMapCompletionStage(p -> p)
-                .buildRs();
+                .flatMapCompletionStage(p -> p);
     }
 
-    private PublisherBuilder<Boolean> handleExistingJob(Job job) {
+    /**
+     * Check if it should be scheduled (on the current chunk) or saved to be scheduled later.
+     * @return
+     */
+    private boolean isOnCurrentSchedulerChunk(Job job) {
+        return job.getExpirationTime().isBefore(DateUtil.now().plusMinutes(schedulerChunkInMinutes));
+    }
+
+    private PublisherBuilder<ScheduledJob> handleExistingJob(Job job) {
         //always returns true, canceling in case the job is already schedule
         return ReactiveStreams.fromCompletionStage(jobRepository.get(job.getId()))
                 //handle scheduled and retry cases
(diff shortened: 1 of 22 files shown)


addons/jobs/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/BaseTimerJobScheduler.java
```diff
@@ -52,13 +55,19 @@
     @ConfigProperty(name = "kogito.jobs-service.maxIntervalLimitToRetryMillis", defaultValue = "60000")
     long maxIntervalLimitToRetryMillis;
 
+    @ConfigProperty(name = "kogito.jobs-service.schedulerChunkInMinutes", defaultValue = "10")
```
spolti: Shouldn't we add these properties on the README?

addons/jobs/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/BaseTimerJobScheduler.java
```diff
@@ -79,30 +89,45 @@ public BaseTimerJobScheduler(JobExecutor jobExecutor,
...
-                .flatMap(p -> p)
-                .map(scheduleId -> ScheduledJob
-                        .builder()
-                        .job(job)
-                        .scheduledId(scheduleId)
-                        .status(JobStatus.SCHEDULED)
-                        .build())
-                .map(scheduledJob -> jobRepository.save(scheduledJob))
-                .flatMapCompletionStage(p -> p)
+                        : ReactiveStreams.of(job))
```
spolti: just an opinion, wouldn't be better break this step in smaller pieces for a better code maintainability?

addons/jobs/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/JobSchedulerManager.java
```diff
@@ -35,22 +40,49 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(JobSchedulerManager.class);
 
+    @ConfigProperty(name = "kogito.jobs-service.schedulerChunkInMinutes", defaultValue = "10")
+    long schedulerChunkInMinutes;
+
+    @ConfigProperty(name = "kogito.jobs-service.loadJobIntervalInMillis", defaultValue = "60000")
```
mswiderski: why to keep these two properties in different time units? one in minutes and the other in millis?

addons/jobs/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/JobSchedulerManager.java
```diff
@@ -35,22 +40,49 @@
...
-                //is is necessary to skip error on the publisher to continue processing, otherwise the subscribe
-                // terminated
+    @Inject
+    Vertx vertx;
+
+    void onStartup(@Observes StartupEvent startupEvent) {
+        //first execution
+        vertx.runOnContext(v -> loadScheduledJobs());
+        //periodic execution
+        vertx.setPeriodic(loadJobIntervalInMillis, id -> loadScheduledJobs());
```
mswiderski: so the load of jobs will run every minute by default and the jobs to be loaded are for 10 min from current time. so that means every minute we load next 10 min of jobs. Isn't that we will get lot of duplicates to filter out?

wouldn't it be better to run less frequently ... close to the chunk size/time and just to load next chunk and thus no need to filter out duplicates? This would make a real difference in case there are lot of jobs in given chunk.

I don't think we need to worry about scheduled jobs in between (like the chunk that was already loaded) as this is already checked upon scheduling.

addons/jobs/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/BaseTimerJobScheduler.java
mswiderski: +1 and to update wiki page about jobs service

tiagodolphine: +1, I'll update the wiki and insert some javadoc as well...

addons/jobs/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/JobSchedulerManager.java
tiagodolphine: np, let unify... I used milliseconds just because vertx scheduler is based on it, but this is an easy conversion, better to keep the same units.

addons/jobs/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/JobSchedulerManager.java
tiagodolphine: yeah, this is the point of getting the configuration property, it is easy to change, I think we can keep both intervals equals or very close to the chunk size, this is more efficient, I was testing with this more frequently load to see it the scheduling was running properly and avoiding duplication. 
So I'll change the default values to be equals, 10min is ok in your point of view?

mswiderski: yes, let's go with defaults to be the same value. in the end it is all configurable so can be adjusted as needed

addons/jobs/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/BaseTimerJobScheduler.java
tiagodolphine: done, missing the wiki.

addons/jobs/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/BaseTimerJobScheduler.java
tiagodolphine: done, I split into one more method to execute the scheduling process.

addons/jobs/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/JobSchedulerManager.java
tiagodolphine: done

addons/jobs/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/JobSchedulerManager.java
tiagodolphine: done

tiagodolphine:
@MarianMacik would you like to take a look at this PR?

tiagodolphine:
PR updated it was added KOGITO-532 to handle jobs prioritization on loading.

addons/jobs/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/BaseTimerJobScheduler.java
```diff
@@ -46,19 +49,29 @@
...
+    @ConfigProperty(name = "kogito.jobs-service.maxIntervalLimitToRetryMillis")
     long maxIntervalLimitToRetryMillis;
 
+    /**
+     * The current chunk size  in minutes the scheduler handles, it is used to keep a limit number of jobs scheduled
+     * in the in-memory scheduler.
+     */
+    @ConfigProperty(name = "kogito.jobs-service.schedulerChunkInMinutes")
+    long schedulerChunkInMinutes;
+
```
MarianMacik: Can you please just quickly explain these 3 config parameters?

addons/jobs/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/BaseTimerJobScheduler.java
```diff
@@ -98,11 +129,18 @@ public BaseTimerJobScheduler(JobExecutor jobExecutor,
...
-                .buildRs();
+                .flatMapCompletionStage(p -> p);
+    }
+
+    /**
+     * Check if it should be scheduled (on the current chunk) or saved to be scheduled later.
+     * @return
+     */
+    private boolean isOnCurrentSchedulerChunk(Job job) {
+        return job.getExpirationTime().isBefore(DateUtil.now().plusMinutes(schedulerChunkInMinutes));
```
MarianMacik: Does this mean that the chunk is constantly moving?

addons/jobs/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/JobSchedulerManager.java
```diff
@@ -35,22 +41,65 @@
...
+
+        //first execution
+        vertx.runOnContext(v -> loadScheduledJobs());
+        //periodic execution
+        vertx.setPeriodic(TimeUnit.MINUTES.toMillis(loadJobIntervalInMinutes), id -> loadScheduledJobs());
+    }
+
+    //Runs periodically loading the jobs from the repository in chunks
+    void loadScheduledJobs() {
+        loadJobsInCurrentChunk()
```
mswiderski: I guess this will run periodically so maybe changing log level to debug instead of info to not spam logs too much, wdyt?

tiagodolphine: I'll change to debug the log for each job and keep the completed job load with info, this is important to check it the loading jobs is running properly from time to time.

addons/jobs/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/BaseTimerJobScheduler.java
tiagodolphine: Yes, the chunk is based on time, so basically the chunk is the interval from **now** until the defined interval, for instance, 10 minutes. In this way, the scheduler keeps in memory only the next X jobs on the current chunk.

addons/jobs/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/BaseTimerJobScheduler.java
tiagodolphine: **schedulerChunkInMinutes** is the chunk size, represented in minutes, the chunk is an interval of time. This config will be on the [wiki,](https://github.com/kiegroup/kogito-runtimes/wiki/Jobs-Service#configuration-properties) the other two parameters are on the wiki already and represent the **Retry** configurations for backoff and the limit that is supposed to be retried.

addons/jobs/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/JobSchedulerManager.java
tiagodolphine: done

addons/jobs/jobs-service/src/main/java/org/kie/kogito/jobs/service/scheduler/BaseTimerJobScheduler.java
ricardozanini: @spolti  could you please open a jira to handle this property on operator side?

tiagodolphine:
Jenkins retest this

tiagodolphine:
Jenkins retest this please

ricardozanini:
@tiagodolphine @cristianonicolai  guys sorry being a PITA, but when you add a new feature to the services, could you please talk to us so we can open JIRAs on our side to support them? Or at least have a discussion with you about implementation details on operator side? Thanks!!

mswiderski:
@ricardozanini this was discussed when we met here in Krakow so I thought you were up to date with the details. Plus does this actually have impact on the operator? Is this about new properties or...

tiagodolphine:
@ricardozanini as @mswiderski pointed out, this feature was already discussed during the krakow f2f and on the planning, IMHO this is the place we should discuss what will be implemented on the feature and if it is necessary some new JIRAs be handled on the operator or any other component, maybe we missed this point, and besides I'm not sure these configurations should be mandatory to be on the operator since they are fine-tuning for the application runtime, or we can expose them with some pre-defined values or some limitations, for instance from 1 to 100, WDYT guys?

tiagodolphine:
Do you know how to retest the PR? it seems to be failed on a test with the integration test with infinispan, but locally it is running well. The "Jenkins retest this please" thing seems not to be working here.

mswiderski:
@tiagodolphine you need to retriever it manually in Jenkins console - there is a sort of refresh button next to login/logout button in right hand side top corner, just triggered it ...

ricardozanini:
> @ricardozanini this was discussed when we met here in Krakow so I thought you were up to date with the details. Plus does this actually have impact on the operator? Is this about new properties or...

I’m not talking about only about this feature, but any implantation in general. For this one we have now a JIRA.

@tiagodolphine any property that will reflect the behavior of the service in runtime should be reflected in the operator deployment. I’ll change that behavior though to have a config map mounted instead with default values, so we won’t need to take care about properties in the future (or not so much).

Plus would be nice if you start giving a chance to the operator. I’ll think on something to make your life easier or at least a tutorial of how to rebuild the images and test them with the operator. :)

tiagodolphine:
> not talking only about this feature, but any implantation in general. For this one we have now a JIRA.
> 
> @tiagodolphine any property that will reflect the behavior of the service in runtime should be reflected in the operator deployment. I’ll change that behavior though to have a config map mounted instead with

Sure +1, and the default values would be nice to have.

tiagodolphine:
I think we are ok to merge the PR.

PR #309, opened 2020-02-12
KOGITO-1145 - Business relevant process variables support

mswiderski:


api/kogito-internal/src/main/java/org/kie/internal/kogito/codegen/VariableInfo.java
```diff
@@ -0,0 +1,32 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.FIELD})
+public @interface VariableInfo {
+   /**
+    * Variable tags assigned to given property of the model
+    */
+   String tags() default "";
```
cristianonicolai: @mswiderski is the intention that we support "custom tags"? Im asking that because if not, I would suggest us to have a enum to define these values, avoiding potential mistakes and making it easy for developers to know the possible values.

api/kogito-services/src/main/java/org/kie/kogito/services/event/impl/ProcessInstanceEventBatch.java
```diff
@@ -37,10 +40,14 @@
 import org.kie.kogito.event.EventBatch;
 import org.kie.kogito.services.event.ProcessInstanceDataEvent;
 import org.kie.kogito.services.event.UserTaskInstanceDataEvent;
+import org.kie.kogito.services.event.VariableInstanceDataEvent;
 
 
 public class ProcessInstanceEventBatch implements EventBatch {
     
+    public static final String BUSINESS_RELEVANT = "business-relevant";
```
cristianonicolai: @mswiderski if we add an enum, I would suggest calling this simply _business_ then add a comment with the explanation, just to shorten it.

api/kogito-services/src/main/java/org/kie/kogito/services/event/impl/ProcessInstanceEventBatch.java
```diff
@@ -90,6 +98,12 @@ public void append(Object rawEvent) {
                 if (workItem instanceof HumanTaskWorkItem && workItemTransitionEvent.isTransitioned()) {
                     userTaskInstances.putIfAbsent(workItem.getId(), createUserTask(workItemTransitionEvent));                    
                 }
+            } else if (event instanceof ProcessVariableChangedEvent) {
+                ProcessVariableChangedEvent variableChangedEvent = (ProcessVariableChangedEvent) event;
+                
+                if (variableChangedEvent.hasTag(BUSINESS_RELEVANT) || variableChangedEvent.hasTag(TRACKED)) {
```
cristianonicolai: @mswiderski I couldn't get the default behaviour for variables. Would that be _tracked_? Should we make that configurable, so for instance if we want to enable audit for all variables changes we wouldnt need to mark all variables with this tag?

cristianonicolai:
@mswiderski overall looks good to me, also take a look on the sonar report, some issues might be worth fixing.

api/kogito-internal/src/main/java/org/kie/internal/kogito/codegen/VariableInfo.java
mswiderski: yes, we do support custom tags

api/kogito-services/src/main/java/org/kie/kogito/services/event/impl/ProcessInstanceEventBatch.java
mswiderski: we can shorten it anyway, just let me know if we have better name, as business itself sounds bit too generic

api/kogito-services/src/main/java/org/kie/kogito/services/event/impl/ProcessInstanceEventBatch.java
mswiderski: no, I'd prefer to be explicitly enabled ... maybe we can think about some global switch but I wouldn't go for it at first as this can be really chatty when it comes to events being emitted.

addons/events/kogito-events-reactive-messaging-addon/src/main/java/org/kie/kogito/events/rm/ReactiveMessagingEventPublisher.java
```diff
@@ -39,6 +39,7 @@
 public class ReactiveMessagingEventPublisher implements EventPublisher {
     private static final String PI_TOPIC_NAME = "kogito-processinstances-events";
     private static final String UI_TOPIC_NAME = "kogito-usertaskinstances-events";
+    private static final String VI_TOPIC_NAME = "kogito-variables-events";
```
cristianonicolai: @ricardozanini FYI, new topic for runtimes

mswiderski:
@cristianonicolai just want through sonar report and address some issues but some are not really relevant like equals method too complex... or too many arguments in constructor

any other comments?

api/kogito-internal/src/main/java/org/kie/internal/kogito/codegen/VariableInfo.java
tiagodolphine: In case of multiple tags for the same variable, the idea is to serialize them in an array? Otherwise, maybe it is worth making `String[] tags()` instead of a single String, WDYT?

mswiderski: since this is mainly data carrier and simply maps the customTags metaData to this annotation I'd stick to simple string ... I guess this would be also easier to deal with in the proto file that data index would interpret. @cristianonicolai what's your opinion?

mswiderski:
after conversation with @cristianonicolai made a change that variable events are only emitted for variables tagged with `tracked`

PR #312, opened 2020-02-14
Add support for Start Signal Catch Events.

Christopher-Chianelli:
This PR add support for Start Signal Catch Events. Before, an error was thrown since the code expected Messages to be the only Start Node kind with Triggers. I am guessing tests will revolve around creating BPMN diagrams and asserting the correct code is generated and is executed correctly?

(Why is this useful? It allows one to handle requests as they come in for a particular process without needing to creating a signal loop or use persistence with messages).

### DIFF
--- a/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/StartNodeVisitor.java
+++ b/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/StartNodeVisitor.java
@@ -16,9 +16,11 @@
 
 package org.jbpm.compiler.canonical;
 
+import java.util.HashMap;
 import java.util.Map;
 
 import org.jbpm.process.core.context.variable.VariableScope;
+import org.jbpm.ruleflow.core.factory.EventNodeFactory;
 import org.jbpm.ruleflow.core.factory.StartNodeFactory;
 import org.jbpm.workflow.core.node.StartNode;
 import org.kie.api.definition.process.Node;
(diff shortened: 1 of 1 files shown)


mswiderski:
@Christopher-Chianelli what is the rationale behind this change? I mean what are you trying to achieve?
I was thinking about using signal events to allow different endpoints to start the process instance via REST api. That means each signal start event should produce dedicated REST endpoint with the payload that maps to given variable. So that would require additional change in `kogito-codegen` project to support that.

Christopher-Chianelli:
@mswiderski The rationale behind this change is to provide a way to model incoming data **to a specific existing process.** This is different from Start Message, which models incoming data **that starts a new process**. With this change, a new endpoint that signals an existing process is generated, but no endpoint is generated that starts a new process from this node (although I think there a "Start From" endpoint somewhere that might be able to do it.

Before this change, to achieve something like this, I needed to create a BPMN diagram like this one below:
LoopBPMN
After this change, this BPMN can be converted to something like this:
BPMNWithStartSignal
Note messages aren't very good for this use case, since Passengers need to be approved/deny for a specific flight, which means we need to add some form of persistence so the passenger can be added to the flight (and at that point it might as well be it own process, despite being heavily related to the Create Flight process).

Christopher-Chianelli:
Note I tried using messages with `KogitoReferenceId`, but that doesn't really work (the message generates a signal to `Message-${messageId}`, but Messages cannot catch `Signals`, meaning I need to have a both a `Start Message` (which does nothing but create the Kafka listener) and a `Start Signal` (which handles the event).

mswiderski:
@Christopher-Chianelli as discussed on the meeting let's address this via event based subprocess  https://issues.redhat.com/browse/KOGITO-1217 so I'd opt at closing this one, are you ok with this?

Christopher-Chianelli:
@mswiderski I am okay with this; closing the PR.

PR #313, opened 2020-02-14
[KOGITO-1022] move quarkus extension back into kogito repo

mariofusco:


evacchi:
note: we should probably 

1) *remove* addons from the bom as they are stand-alone applications people are not supposed to depend on
2) create a different bom or parent-pom for addons so that they do not necessarily have to be aligned with the quarkus version of the extension 
3) create a different bom for examples (this is already a tracked issue)

PR #314, opened 2020-02-14
KOGITO-1194 - Improve data handling within process instance scope to …

mswiderski:
…allow to extract data for data input and outputs

@tarilabs here is the support for input and output expressions


jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/VariableScopeResolverFactory.java
```diff
@@ -26,18 +29,31 @@
...
+	    boolean found = variableScope.getVariable(name) != null;
+	    if (!found) {
+            return extraParameters.containsKey(name);
+        }
+        
+        return found;
 	}
 	
 	public VariableResolver getVariableResolver(String name) {
+	    if (extraParameters.containsKey(name)) {
```
MarianMacik: So these extra parameters are now used for resolvable mappings?

PR #317, opened 2020-02-17
KOGITO-1180 - Allow to use business key as an alternative to generate…

mswiderski:
…d process instance id

### DIFF
--- a/api/kogito-api/src/main/java/org/kie/api/runtime/process/WorkflowProcessInstance.java
+++ b/api/kogito-api/src/main/java/org/kie/api/runtime/process/WorkflowProcessInstance.java
@@ -79,5 +79,12 @@ public interface WorkflowProcessInstance
      * @return error message
      */
     String getErrorMessage();
+    
+    
+    /**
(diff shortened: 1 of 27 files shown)


api/kogito-api/src/main/java/org/kie/api/runtime/process/WorkflowProcessInstance.java
```diff
@@ -79,5 +79,12 @@
      * @return error message
      */
     String getErrorMessage();
+    
+    
+    /**
+     * Returns optional correlation key assigned to process instance
+     * @return correlation key if present otherwise null
+     */
+    String getCorrelationKey();
```
cristianonicolai: perhaps consolidate as _businessKey_?

mswiderski: this is done on purpose as to keep the correlation key on the "legacy" api as it is used already

mswiderski:
@cristianonicolai rebased and address sonar issues

PR #323, opened 2020-02-19
KOGITO-893: Use Singleton DataSource in processes+rules

evacchi:
We have introduced the SingletonStore: a data store for single-values. In this PR we use it for workflow-generated rule units; i.e. rule units that transparently "see" process variables

### DIFF
--- a/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractVisitor.java
+++ b/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractVisitor.java
@@ -154,11 +154,11 @@ public abstract class AbstractVisitor {
         }
     }
 
-    protected void addNodeMappings(Mappable workItemNode, BlockStmt body, String variableName) {
-        for (Entry<String, String> entry : workItemNode.getInMappings().entrySet()) {
+    protected void addNodeMappings(Mappable node, BlockStmt body, String variableName) {
+        for (Entry<String, String> entry : node.getInMappings().entrySet()) {
             addFactoryMethodWithArgs(body, variableName, "inMapping", new StringLiteralExpr(entry.getKey()), new StringLiteralExpr(entry.getValue()));
         }
-        for (Entry<String, String> entry : workItemNode.getOutMappings().entrySet()) {
+        for (Entry<String, String> entry : node.getOutMappings().entrySet()) {
             addFactoryMethodWithArgs(body, variableName, "outMapping", new StringLiteralExpr(entry.getKey()), new StringLiteralExpr(entry.getValue()));
         }
     }
--- a/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractVisitor.java
+++ b/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractVisitor.java
@@ -23,6 +23,7 @@ import java.util.Map.Entry;
 
 import org.drools.core.util.StringUtils;
 import org.jbpm.process.core.Work;
+import org.jbpm.process.core.context.variable.Mappable;
 import org.jbpm.process.core.context.variable.Variable;
 import org.jbpm.process.core.context.variable.VariableScope;
 import org.jbpm.workflow.core.node.WorkItemNode;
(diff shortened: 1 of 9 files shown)


jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/node/RuleSetNode.java
```diff
@@ -37,7 +38,8 @@
 /**
  * Default implementation of a RuleSet node.
  */
-public class RuleSetNode extends StateBasedNode implements ContextContainer {
+public class RuleSetNode extends StateBasedNode implements ContextContainer,
+                                                           Mappable {
```
evacchi: this did not implement Mappable, but it did provide all of the required methods

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/RuleUnitHandler.java
```diff
@@ -35,6 +35,19 @@
...
+ *
+ * | Mapping | Process Variable | Rule Unit field   | Action
+ * | IN      | scalar           | scalar            | Assignment
+ * | IN      | scalar           | data source 	    | Add to (i.e. insert into) data source
+ * | IN      | collection       | data source 	    | Add all contents from data source
+ * | OUT     | scalar           | scalar 	        | Assignment
+ * | OUT     | scalar           | data source 	    | get 1 value off the data source
+ * | OUT     | collection       | data source 	    | Add all contents to the data source
+ *
+ */
```
evacchi: moved the documentation of the mappings, as it was now handled here

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/RuleSetNodeVisitor.java
```diff
@@ -100,27 +88,10 @@ public void visitNode(String factoryField, Node node, BlockStmt body, VariableSc
...
-        if (ruleType.isRuleUnit()) {
-            if (ruleSetNode.getInMappings().isEmpty()) {
-                GeneratedRuleUnitDescription generatedRuleUnitDescription = new GeneratedRuleUnitDescription(ruleType.getName(), contextClassLoader);
-                for (Variable v : variableScope.getVariables()) {
-                    generatedRuleUnitDescription.putDatasourceVar(v.getName(), DataStore.class.getCanonicalName(), v.getType().getStringType());
-                }
-                RuleUnitComponentFactoryImpl impl = (RuleUnitComponentFactoryImpl) RuleUnitComponentFactory.get();
-                impl.registerRuleUnitDescription(generatedRuleUnitDescription);
-            }
-        }
```
evacchi: this code was duplicated in handleDecision+generateRuleUnitDescription

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractVisitor.java
```diff
@@ -153,8 +154,7 @@ protected void addWorkItemParameters(Work work, BlockStmt body, String variableN
         }
     }
 
-    protected void addWorkItemMappings(WorkItemNode workItemNode, BlockStmt body, String variableName) {
-
+    protected void addNodeMappings(Mappable workItemNode, BlockStmt body, String variableName) {
```
evacchi: renamed as it now takes any Mappable node

kogito-codegen/src/test/resources/ruletask/Generated.drl
```diff
@@ -16,14 +16,14 @@ rule singlePerson_add_note when
     s: /singleString[ this == "hello" ]
 then
     p.setAdult(true);
-    singleString.remove(s);
-	singleString.add("Now the life starts again");
+    singleString.clear();
+	singleString.set("Now the life starts again");
```
mswiderski: Wouldn’t it be enough to just set the new value instead of clear and then set? It’s sort of more natural to jest set value... does that have some side effects?

evacchi: ah you are right. I just rewrote all of the calls without thinking it twice. Indeed, set() is enough

PR #332, opened 2020-02-25
KOGITO-1264 - Allow to use Validation API to enhance generated REST api

mswiderski:
This PR brings support for validation api to enhance the generated REST api

In general it allows to use validation api annotations on model classes and then the generated api will enforce this validation. See some details in jira

So an example here could be to use following model

```
import javax.validation.constraints.Email;
import javax.validation.constraints.Min;
import javax.validation.constraints.NotNull;
import javax.validation.constraints.Size;

public class Person {

    @NotNull
    @Size(min = 2, max = 30)
    private String name;

    @NotNull
    @Min(value = 2, message = "Age must be higher than 5")
    private int age;

    @NotNull
    @Email
    private String email;

    private boolean adult;
```

and this will then generated following input data model for process that references this type as variable

```
public class PersonsModelInput implements org.kie.kogito.Model {


    @org.kie.internal.kogito.codegen.VariableInfo(tags = "")
    @javax.validation.Valid()
    private org.acme.travels.Person person;

    public org.acme.travels.Person getPerson() {
        return person;
    }

    public void setPerson(org.acme.travels.Person person) {
        this.person = person;
    }
}
```

And then the REST resource for creating new instances of the process

```
 @POST()
    public PersonsModelOutput createResource_persons(@Context HttpHeaders httpHeaders, 
@QueryParam("businessKey") String businessKey,
 @javax.validation.Valid() @javax.validation.constraints.NotNull() PersonsModelInput resource) {     
```

whenever user attempts to send invalid data such as following

```
curl -X POST -H 'Content-Type:application/json' -H 'Accept:application/json' -d '{"person" : {"name" : "john", "age" : 3, "email" : "user"}}' http://localhost:8080/persons
```

it will produce `400 Bad Request` response with following message (this can actually be configured)

```
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8080 (#0)
> POST /persons HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.54.0
> Content-Type:application/json
> Accept:application/json
> Content-Length: 59
>
* upload completely sent off: 59 out of 59 bytes
< HTTP/1.1 400 Bad Request
< Content-Length: 262
< validation-exception: true
< Content-Type: application/json
<
* Connection #0 to host localhost left intact
{"exception":null,"propertyViolations":[],"classViolations":[],"parameterViolations":[{"constraintType":"PARAMETER","path":"createResource_persons.resource.person.email","message":"must be a well-formed email address","value":"user"}],"returnValueViolations":[]}
```

this could be nice addition not only to processes but also to decision and rules 

In addition this also is exposed via the OpenAPI/swagger

image

@tarilabs @evacchi in case you would be interested

### DIFF
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/context/QuarkusKogitoBuildContext.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/context/QuarkusKogitoBuildContext.java
@@ -15,6 +15,18 @@
 
 package org.kie.kogito.codegen.context;
 
+import java.util.function.Predicate;
+
 public class QuarkusKogitoBuildContext implements KogitoBuildContext {
-    
+
+    private Predicate<String> classAvailabilityResolver;
+
+    public QuarkusKogitoBuildContext(Predicate<String> classAvailabilityResolver) {
+        this.classAvailabilityResolver = classAvailabilityResolver;
+    }
+
+    @Override
+    public boolean hasClassAvailable(String fqcn) {
+        return classAvailabilityResolver.test(fqcn);
+    }
 }
--- a/kogito-quarkus-extension/deployment/src/main/java/org/kie/kogito/quarkus/deployment/KogitoAssetsProcessor.java
+++ b/kogito-quarkus-extension/deployment/src/main/java/org/kie/kogito/quarkus/deployment/KogitoAssetsProcessor.java
@@ -323,7 +323,11 @@ public class KogitoAssetsProcessor {
                 .getClassByName(createDotName(persistenceFactoryClass)) != null;
 
         GeneratorContext context = GeneratorContext.ofResourcePath(projectPath.resolve("src/main/resources").toFile());
-        context.withBuildContext(new QuarkusKogitoBuildContext());
+        context.withBuildContext(new QuarkusKogitoBuildContext(className -> {
+                DotName classDotName = createDotName(className);
+                return !combinedIndexBuildItem.getIndex().getAnnotations(classDotName).isEmpty() || combinedIndexBuildItem.getIndex().getClassByName(classDotName) != null;
+                
+        }));
 
         ApplicationGenerator appGen = new ApplicationGenerator(appPackageName, new File(projectPath.toFile(), "target"))
                 .withDependencyInjection(new CDIDependencyInjectionAnnotator())
--- a/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ModelMetaData.java
+++ b/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ModelMetaData.java
@@ -27,6 +27,7 @@ import java.util.Optional;
 import java.util.stream.Collectors;
 
 import org.drools.core.util.StringUtils;
+import org.jbpm.process.core.context.variable.Variable;
 import org.kie.api.definition.process.WorkflowProcess;
 import org.kie.internal.kogito.codegen.Generated;
 import org.kie.internal.kogito.codegen.VariableInfo;
@@ -68,6 +69,8 @@ public class ModelMetaData {
     private String visibility;
     private boolean hidden;
     private String templateName;
+    
+    private boolean supportsValidation;
 
     public ModelMetaData(String processId, String packageName, String modelClassSimpleName, String visibility, VariableDeclarations variableScope, boolean hidden) {
         this(processId, packageName, modelClassSimpleName, visibility, variableScope, hidden, "/class-templates/ModelTemplate.java");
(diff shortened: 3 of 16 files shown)


mswiderski:
@cristianonicolai was thinking about this but then would have to make this too generic and just for the sake of reuse make the extra looping over artefacts to transform to files etc. So decided to make it not reusable but bit faster...

mswiderski:
@evacchi @mariofusco guys, is the move of quarkus extension done? As this time it looks like tests are failing due to missing constructor even though this PR updates extension in kogito-runtimes

```

Caused by: java.lang.RuntimeException: 
io.quarkus.builder.BuildException: Build failure: Build failed due to errors
	[error]: Build step io.quarkus.kogito.deployment.KogitoAssetsProcessor#generateModel threw an exception: java.lang.NoSuchMethodError: org.kie.kogito.codegen.context.QuarkusKogitoBuildContext: method <init>()V not found
	at io.quarkus.kogito.deployment.KogitoAssetsProcessor.createApplicationGenerator(KogitoAssetsProcessor.java:321)
	at io.quarkus.kogito.deployment.KogitoAssetsProcessor.generateModel(KogitoAssetsProcessor.java:172)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at io.quarkus.deployment.ExtensionLoader$2.execute(ExtensionLoader.java:929)
	at io.quarkus.builder.BuildContext.run(BuildContext.java:279)
	at org.jboss.threads.ContextClassLoaderSavingRunnable.run(ContextClassLoaderSavingRunnable.java:35)
	at org.jboss.threads.EnhancedQueueExecutor.safeRun(EnhancedQueueExecutor.java:2027)
	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.doRunTask(EnhancedQueueExecutor.java:1551)
	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1442)
	at java.lang.Thread.run(Thread.java:748)
	at org.jboss.threads.JBossThread.run(JBossThread.java:479)

```

kogito-quarkus-extension/deployment/src/main/java/org/kie/kogito/quarkus/deployment/KogitoAssetsProcessor.java
```diff
@@ -323,7 +323,11 @@ private ApplicationGenerator createApplicationGenerator(Path projectPath, Launch
                 .getClassByName(createDotName(persistenceFactoryClass)) != null;
 
         GeneratorContext context = GeneratorContext.ofResourcePath(projectPath.resolve("src/main/resources").toFile());
-        context.withBuildContext(new QuarkusKogitoBuildContext());
+        context.withBuildContext(new QuarkusKogitoBuildContext((className) -> {
+                DotName classNotName = createDotName(className);
```
MarianMacik: Probably the variable should be called `classDotName`?

kogito-codegen/src/main/java/org/kie/kogito/codegen/context/QuarkusKogitoBuildContext.java
```diff
@@ -15,6 +15,30 @@
...
+    public QuarkusKogitoBuildContext() {
+        this((className) -> {
+            try {
+                Thread.currentThread().getContextClassLoader().loadClass(className);
+                return true;
+            } catch (ClassNotFoundException e) {
+                return false;
+            }
+        });
+    }
```
MarianMacik: Is this constructor even used anywhere? I see that you actually use the other one, which means you override the resolver function everywhere in the end.

mswiderski: this was just to fix the kogito quarkus extension that was still (and most likely is the same case atm) the io.quarkus kogito extension ... so once that is fixed this constructor won't be used.

kogito-quarkus-extension/deployment/src/main/java/org/kie/kogito/quarkus/deployment/KogitoAssetsProcessor.java
mswiderski: fixed

mswiderski:
@MarianMacik 

> Can you explain more? Because right now both methods are looping over artifacts/files (so over the same thing - jars)

to make it generic I would have to make an argument of incoming jars to scan.. and that would have to be List<File> and in such case to have list of files from list of Artifacts I would have to iterate over the artefacts and extract files from them. So that would be this extra looping.

evacchi:
@mswiderski so if I have understood correctly, enabling this on rules and decisions amounts to just adding the annotation on the endpoint+JSON-pojo we generate, right? that is, this new method:

       public PersonsModelOutput createResource_persons(@Context HttpHeaders httpHeaders, 

that we weren't generating before. Is this correct?

kogito-codegen/src/main/java/org/kie/kogito/codegen/context/QuarkusKogitoBuildContext.java
MarianMacik: Thanks.

mswiderski:
@evacchi validation is currently done on inputs so adding this extra annotations on resource that accepts data to call rules/decisions and then depending if there is also model class generated additional ‚@Valid’ to cascade the validation

PR #333, opened 2020-02-25
fix hasClassOnClasspath (getDependencyArtifacts is deprecated)

r00ta:
`getDependencyArtifacts` is currently deprecated. In addition to that if a URL is malformed the current implementation will return `false`, but actually the class might be in the classpath -> with the new implementation even if a URL is malformed all the others are taken into consideration to provide a result.

### DIFF
--- a/kogito-maven-plugin/src/main/java/org/kie/kogito/maven/plugin/GenerateModelMojo.java
+++ b/kogito-maven-plugin/src/main/java/org/kie/kogito/maven/plugin/GenerateModelMojo.java
@@ -6,6 +6,7 @@ import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.io.IOException;
 import java.io.UncheckedIOException;
+import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.nio.file.FileSystems;
@@ -14,6 +15,7 @@ import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
 import java.nio.file.PathMatcher;
 import java.nio.file.Paths;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


kogito-maven-plugin/src/main/java/org/kie/kogito/maven/plugin/GenerateModelMojo.java
```diff
@@ -267,26 +269,25 @@ private void deleteDrlFiles() throws MojoExecutionException {
...
-                urls[i] = artifact.getFile().toURI().toURL();
-                i++;
+        for(Artifact artifact : artifacts){
+            try {
+                urls.add(artifact.getFile().toURI().toURL());
             }
-            try (URLClassLoader cl = new URLClassLoader(urls)) {
-                cl.loadClass(className);
+            catch(MalformedURLException e){
+                getLog().debug("Artifact has malformed URL: skipping the artifact.", e);
```
mswiderski: when would this happen?

kogito-maven-plugin/src/main/java/org/kie/kogito/maven/plugin/GenerateModelMojo.java
```diff
@@ -45,7 +47,7 @@
 import org.kie.kogito.maven.plugin.util.MojoUtil;
 
 @Mojo(name = "generateModel",
-        requiresDependencyResolution = ResolutionScope.NONE,
+        requiresDependencyResolution = ResolutionScope.COMPILE,
```
mswiderski: compile and runtime might be actually more suitable here, see same/similar changes being done here https://github.com/kiegroup/kogito-runtimes/pull/332

r00ta: Sorry I did not see your previous PR. Given that the change is already included in your PR we can abandon this one if it's ok for you

kogito-maven-plugin/src/main/java/org/kie/kogito/maven/plugin/GenerateModelMojo.java
MarianMacik: This would probably happen only if Maven Plugin API and impl had a bug when providing the paths to dependencies which I don't think is something we should be checking.

kogito-maven-plugin/src/main/java/org/kie/kogito/maven/plugin/GenerateModelMojo.java
MarianMacik: +1

kogito-maven-plugin/src/main/java/org/kie/kogito/maven/plugin/GenerateModelMojo.java
r00ta: In the previous implementation `Exception` was catched in general, so that's why if the first artifact (for example) has a malformed URL then `hasClassOnClasspath` would return `false` even if that class might be actually in the classpath. 
But yeah, fine for me anyhow. The important thing for me is that we move away from the deprecated method and we use `ResolutionScope.COMPILE`, because yesterday I was using `hasClassOnClasspath` for some stuff and it was not working as expected. 

Given that this change is included in the other PR I will abandon this one

PR #338, opened 2020-02-26
KOGITO-1266: Fix hot reload for rule units

evacchi:
Not the greatest of fix, but for now it will make do.
We don't use class literals for type-testing (e.g. DataSource.class), but we create a class holding all these types in the correct classloader (DataSourceTypes), then we use those to do such checks.

also, let's make sure these examples are not broken, especially the BPMN one:
https://github.com/evacchi/kogito-rules-example

### DIFF
--- a/drools/kogito-ruleunits/src/main/java/org/kie/kogito/rules/units/GeneratedRuleUnitDescription.java
+++ b/drools/kogito-ruleunits/src/main/java/org/kie/kogito/rules/units/GeneratedRuleUnitDescription.java
@@ -16,13 +16,9 @@
 
 package org.kie.kogito.rules.units;
 
-import java.lang.reflect.Method;
-import java.lang.reflect.ParameterizedType;
-import java.lang.reflect.Type;
 import java.util.function.Function;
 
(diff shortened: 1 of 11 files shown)


drools/kogito-ruleunits/src/main/java/org/kie/kogito/rules/units/GeneratedRuleUnitDescription.java
```diff
@@ -16,13 +16,9 @@
...
-import java.lang.reflect.Method;
-import java.lang.reflect.ParameterizedType;
-import java.lang.reflect.Type;
 import java.util.function.Function;
 
 import org.drools.core.addon.TypeResolver;
-import org.kie.kogito.rules.DataSource;
 import org.kie.kogito.rules.RuleUnitConfig;
 
 public class GeneratedRuleUnitDescription extends AbstractRuleUnitDescription {
```
evacchi: cleaning up some unused methods here

evacchi:
let's use mario's branch -- includes this one https://github.com/kiegroup/kogito-runtimes/pull/342#pullrequestreview-365596318

PR #346, opened 2020-02-28
KOGITO-1313: Override netty version to avoid BufferOverflow in infinispan client

evacchi:


evacchi:
@cristianonicolai as soon as you're ok with this, we can merge -- this should unbreak persistence for all users of the kogito quarkus extension (eg examples)

PR #348, opened 2020-03-03
[KOGITO-1348] allow kogito-codegen to compile resources embedded in a…

mariofusco:
… jar


mariofusco:
@evacchi unfortunately this is necessary for 0.8.0 and in particular for quarkus test integration that I'm preparing in this other branch https://github.com/mariofusco/quarkus-platform/tree/k1251

I don't want to have a simply copy/paste of our integration tests there but something more automatic in the style of what have been done by the camel team and also reusing their rpkgtests-maven-plugin. What happens there is that it downloads our integration test module from the maven repo and try to build the resources that this time are not contained in a folder as usual but in a jar. That's why I needed to allow our codegenerators to retrieve drools/jbpm/dmn artifacts also inside a jar and generate code out of them. 

To recap this PR should only allow our generators to also work on a jar and not only on a folder and shouldn't affect code generation in any other case.

PR #357, opened 2020-03-06
KOGITO-1163 Call to Jobs service fails on OpenShift 

tiagodolphine:
Fix URL without port on service configuration.

https://github.com/kiegroup/kogito-apps/pull/94

### DIFF
--- a/addons/jobs/jobs-api/pom.xml
+++ b/addons/jobs/jobs-api/pom.xml
@@ -11,6 +11,18 @@
   <name>Jobs API</name>
   <description>Jobs API used by both service and client</description>
 
+  <dependencies>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+    </dependency>
+  </dependencies>
+
   <build>
     <plugins>
       <plugin>
--- a/addons/jobs/jobs-api/pom.xml
+++ b/addons/jobs/jobs-api/pom.xml
@@ -20,6 +20,7 @@
     <dependency>
       <groupId>org.assertj</groupId>
       <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
     </dependency>
   </dependencies>
 
--- a/addons/jobs/jobs-management-common/src/main/java/org/kie/kogito/jobs/management/RestJobsService.java
+++ b/addons/jobs/jobs-management-common/src/main/java/org/kie/kogito/jobs/management/RestJobsService.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.jobs.management;
+
+import java.net.URI;
+import java.util.Objects;
+
+import org.kie.kogito.jobs.JobsService;
+import org.kie.kogito.jobs.ProcessInstanceJobDescription;
+import org.kie.kogito.jobs.api.URIBuilder;
+
+public abstract class RestJobsService implements JobsService {
+
+    public static final String JOBS_PATH = "/jobs";
+
+    private URI jobsServiceUri;
+    private String callbackEndpoint;
+
+    public RestJobsService(String jobServiceUrl, String callbackEndpoint) {
+        this.jobsServiceUri = Objects.nonNull(jobServiceUrl) ? buildJobsServiceURI(jobServiceUrl) : null;
+        this.callbackEndpoint = callbackEndpoint;
+    }
+
+    public String getCallbackEndpoint(ProcessInstanceJobDescription description) {
+        return URIBuilder.toURI(callbackEndpoint
+                                        + "/management/jobs/"
+                                        + description.processId()
+                                        + "/instances/"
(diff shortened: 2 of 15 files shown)


addons/jobs/jobs-management-common/src/main/java/org/kie/kogito/jobs/management/RestJobsService.java
```diff
@@ -0,0 +1,56 @@
...
+
+public abstract class RestJobsService implements JobsService {
+
+    public static final String JOBS_PATH = "/jobs";
+
+    private URI jobsServiceUri;
+    private String callbackEndpoint;
+
+    public RestJobsService(String jobServiceUrl, String callbackEndpoint) {
+        this.jobsServiceUri = Objects.nonNull(jobServiceUrl) ? buildJobsServiceURI(jobServiceUrl) : null;
```
sutaakar: Wouldn't it have more sense to thrown an exception in case jobServiceUrl is null?
From my point of view it is not valid value and we should fail fast.

tiagodolphine: That is a point but when initializing the runtime with quarkus, in some step it instantiates the bean on CDI using the default constructor, with no args and in this case throwing an exception will break the startup, I tested it =/ 
See https://quarkus.io/guides/cdi-reference on item 5.4

tiagodolphine:
@cristianonicolai when you have some time to take a look at this one.
Thanks !

addons/jobs/jobs-management-common/src/main/java/org/kie/kogito/jobs/management/RestJobsService.java
sutaakar: ok

addons/jobs/jobs-api/pom.xml
```diff
@@ -11,6 +11,18 @@
...
 
+  <dependencies>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
```
cristianonicolai: test scope?

addons/jobs/jobs-management-quarkus-addon/src/test/java/org/kie/kogito/jobs/management/quarkus/VertxJobsServiceTest.java
```diff
@@ -0,0 +1,105 @@
...
+        tested = new VertxJobsService(JOB_SERVICE_URL, CALLBACK_URL, vertx, instance);
+        tested.initialize();
+    }
+
+    @Test
+    void scheduleProcessJob() {
+        ProcessJobDescription processJobDescription = ProcessJobDescription.of(ExactExpirationTime.now(),
+                                                                               1,
+                                                                               "processId");
+        Assertions
```
cristianonicolai: maybe just do import static here as well

addons/jobs/jobs-management-springboot-addon/pom.xml
```diff
@@ -53,5 +56,28 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>commons-logging</groupId>
+      <artifactId>commons-logging</artifactId>
+      <version>1.2</version>
```
cristianonicolai: if you need to declare a depency version, that should be done via depencymanagement in the root pom.

addons/jobs/jobs-management-springboot-addon/src/test/java/org/kie/kogito/jobs/management/springboot/SpringRestJobsServiceTest.java
```diff
@@ -0,0 +1,89 @@
...
+        this.tested = new SpringRestJobsService(JOB_SERVICE_URL, CALLBACK_URL, restTemplate);
+        tested.initialize();
+    }
+
+    @Test
+    void scheduleProcessJob() {
+        ProcessJobDescription processJobDescription = ProcessJobDescription.of(ExactExpirationTime.now(),
+                                                                               1,
+                                                                               "processId");
+        Assertions
```
cristianonicolai: same here about using import static instead.

addons/jobs/pom.xml
```diff
@@ -13,5 +13,6 @@
     <module>jobs-api</module>
     <module>jobs-management-quarkus-addon</module>
     <module>jobs-management-springboot-addon</module>
+    <module>jobs-management-common</module>
```
cristianonicolai: should it be added before quarkus and springboot?

addons/jobs/jobs-api/pom.xml
sutaakar: AFAIK the test scope is inherited from dependency declaration.
Though it would be good to declare it here explicitly.

addons/jobs/jobs-management-quarkus-addon/src/test/java/org/kie/kogito/jobs/management/quarkus/VertxJobsServiceTest.java
tiagodolphine: done

addons/jobs/jobs-management-springboot-addon/src/test/java/org/kie/kogito/jobs/management/springboot/SpringRestJobsServiceTest.java
tiagodolphine: done

addons/jobs/jobs-management-springboot-addon/pom.xml
tiagodolphine: great point, I need to add this dependency because it is required running the spring tests, but I added the `commons-logging-jboss-logging` instead and it works fine now, there is already this dep on bom.

addons/jobs/pom.xml
tiagodolphine: I'll change it to be more semantically correct, but maven during the build has its own mechanism to sort the order the modules are built, http://maven.apache.org/guides/mini/guide-multiple-modules.html, on in the end it does not make difference the order we declare.

tiagodolphine:
@cristianonicolai thanks for the review I already applied your comments.

PR #360, opened 2020-03-10
KOGITO-1249: Maven plugin cleanup

evacchi:
Review maven plugin. e.g. following goals should be removed,

-    injectreactive
-    generateDMNModel

these are all superseded by generateDeclaredTypes and generateModel

we may also rename generateModel in something different as "Model" here meant "executable model" while now most kogito-related stuff gets generated here.

~~we may also remove the customized manifest, where all phases are overridden: we can just bind~~

-    ~~generateDeclaredTypes to generate-sources and~~
-    ~~generateModel to package or anything post compile~~

~~but we need to verify compilation is not needed anymore before generateModel (I have doubts here)~~

let's take a safe route for now and keep the double-compile pass


mswiderski:
@evacchi the description is rather worrying as it says that 'process-model-classes' is to be removed why the PR does not remove it. Moreover it should not be removed as it would then break persistence ...

evacchi:
true, I copy pasted from JIRA but when I realized it was used for persistence I did not actually remove it, removed other stuff. I will fix the description

evacchi:
Let's wait for https://github.com/kiegroup/kogito-examples/pull/142 to be merged then

PR #370, opened 2020-03-16
KOGITO-1460 - Move Maven build to use version 3.6.0

cristianonicolai:


cristianonicolai:
@mbiarnes this fixes the daily snapshot build failure.

mbiarnes:
Jenkins please retest this

archetypes/kogito-quarkus-archetype/pom.xml
```diff
@@ -42,17 +42,14 @@
       <plugins>
         <plugin>
           <artifactId>maven-archetype-plugin</artifactId>
-          <version>3.0.1</version>
```
ricardozanini: Those versions were added to the BOM?

evacchi: I think it's default when you bump maven version

evacchi: guys I think this might bite us if we don't add the mvnw and/or the enforcer plugin like Quarkus did. The reason why e.g. surefire version in particular is overridden is that otherwise Quarkus tests won't work as expected

mbiarnes:
@cristianonicolai can this me merged too?

archetypes/kogito-quarkus-archetype/pom.xml
cristianonicolai: good catch, added now

cristianonicolai: @evacchi having the version in the pluginmanagement would enforce that specific version to be used. Plus adding the required Maven version I think makes sense, also to make sure we run on the same setup as jenkins. Could you look into that?

evacchi: sure, just a heads up: plugin management won't work with bom import AFAIK https://issues.apache.org/jira/browse/MNG-5588 🙈

evacchi:
I have no idea why this keeps failing, can you try to rebase @cristianonicolai ?

PR #371, opened 2020-03-17
[KOGITO-1452] prometheus-monitoring-addon should use org.kie.kogito p…

tkobayas:
…ackage

Manually confirmed that metrics feature works with the new package class in jbpm-quarkus-example/jbpm-springboot-example.


tkobayas:
In kogito-examples, only onboarding-example is affected.

See https://github.com/kiegroup/kogito-examples/pull/148

Projects in kogito-travel-agency-tutorial also use PrometheusProcessEventListener/PrometheusMetricsDroolsListener but they use fixed kogito versions (e.g. 0.4.0) so not affected.

tkobayas:
@evacchi I just changed the package so SonarCloud gives some failures from the original codes. Shall I fix them on this PR, too?

PR #372, opened 2020-03-17
KOGITO-1476: Examples - setup develop+stable branch

evacchi:
update the Jenkinsfile so it fallbacks to `devel` in examples instead of `master`

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -25,6 +25,7 @@ pipeline {
         stage('Build kogito-runtimes') {
             steps {
                 script {
+		    println("$CHANGE_TARGET") // debugging var...
                     maven.runMavenWithSubmarineSettings('clean install -Prun-code-coverage', false)
                     /*
                        The analysis must happen before the other stages as these clone different projects into a root
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -25,7 +25,6 @@ pipeline {
         stage('Build kogito-runtimes') {
             steps {
                 script {
-		    println("$CHANGE_TARGET") // debugging var...
                     maven.runMavenWithSubmarineSettings('clean install -Prun-code-coverage', false)
                     /*
                        The analysis must happen before the other stages as these clone different projects into a root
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -48,14 +48,14 @@ pipeline {
             steps {
                 dir("kogito-examples") {
                     script {
-                        githubscm.checkoutIfExists('kogito-examples', "$CHANGE_AUTHOR", "$CHANGE_BRANCH", 'kiegroup', "$CHANGE_TARGET")
+                        githubscm.checkoutIfExists('kogito-examples', "$CHANGE_AUTHOR", "$CHANGE_BRANCH", 'kiegroup', "${CHANGE_TARGET ?: 'devel'}")
```
radtriste: I think the change here will never go to `devel` as $CHANGE_TARGET will always be set on a PR.
For example, if we are merging from author `radtriste`, branch `kogito-xxx` to repo `kiegroup`, branch `master`, then we have:
- CHANGE_AUTHOR = radtriste
- CHANGE_BRANCH = kogito-xxx
- CHANGE_TARGET = master
so it will always go to examples `master` branch.
A workaround would be to check CHANGE_TARGET env var and when it is `master`, then we change it to `devel`.

Or if we adopt stable/master, like said in our chat, with stable default, then there will be no problem I think.

evacchi: so does `CHANGE_TARGET` always fallback to `master`, or is it the _default_ branch? because, if we make `stable` the default checkout branch, then the problem will still be there

radtriste: I think that `CHANGE_TARGET` is the branch where your PR will be merged (kiegroup/master in my example)
so there is no default.

mareknovotny:
what if we change the default branch ?

Jenkinsfile
radtriste: so, as far as I understood, and taking my above example, the `githubscm.checkoutIfExists` method will first try to checkout branch `kogito-xxx` from `radtriste/kogito-examples` repository and if it does not exist, it will try to checkout branch `master` from `kiegroup/kogito-examples`.

source: https://github.com/kiegroup/jenkins-pipeline-shared-libraries/blob/684f38521727dc72710c1b0f2d1cdcf96ec2ca71/vars/githubscm.groovy#L14

radtriste:
it could. if we create `stable` version of `kogito-examples` to point to latest version, make it default to the `kogito-examples` repository and we keep `master` as our dev branch, then creating a PR to `kogito-runtimes` on branch `master` will anyway try to checkout `kogito-examples` on branch `master`

evacchi:
just doing a test here, I want to see if it does run against devel

evacchi:
so it looks like it _does_ pull from master @radtriste -- examples now defaults to `stable` branch which is currently same as tag `0.8.1`

radtriste:
Great ! Thx for checking all of that @evacchi :)

PR #374, opened 2020-03-18
[KOGITO-1485] add test case for hot reload

mariofusco:


kogito-quarkus-extension/integration-test/src/test/java/io/quarkus/it/kogito/drools/HotReloadTest.java
```diff
@@ -0,0 +1,67 @@
...
+    private static final String PACKAGE = "io.quarkus.it.kogito.drools";
+    private static final String RESOURCE_FILE = PACKAGE.replace( '.', '/' ) + "/adult.drl";
+
+    @RegisterExtension
+    final static QuarkusDevModeTest test = new QuarkusDevModeTest()
+            .setArchiveProducer(new Supplier<JavaArchive>() {
+                @Override
+                public JavaArchive get() {
+                    return ShrinkWrap.create(JavaArchive.class)
+                            .addAsResource( "adult.txt", RESOURCE_FILE );
```
evacchi: why a txt and not the drl directly ?

mariofusco: I could make it a drl file, but if I do so it will be also compiled together with other assets even when running unrelated tests, which is not harmful (I think), but imo useless.

evacchi: ok I was misunderstanding the code: this is renaming a txt into DRL, triggering compile, then the text replace action triggers hot reload. good

PR #376, opened 2020-03-18
KOGITO-1495: Fix process definition so it can be opened with the Kogi…

DuncanDoyle:
…to VSCode tooling.


jomarko:
Do we want to keep 'bpmn2' extension? What do you think about simplification just to `bpmn`?

PR #388, opened 2020-03-25
Add kogito-quarkus-deployment to kogito-bom

cristianonicolai:


cristianonicolai:
@evacchi kogito-quarkus and spring boot related dependencies are here too. I dont there is a problem in having it here and not in use.

evacchi:
right, but in order to use those you would still need all of the quarkus-bom dependencies (all quarkus extensions) so I am just not sure if this is useful

cristianonicolai:
@evacchi yeah, I'm not not fully convinced that this will help much, I guess just trying to be on the safe side based on the situation we had regarding overriding versions, etc. We can drop this if you feel it doest help.

mariofusco:
@cristianonicolai I agree with @evacchi's comment: if this is in kogito-quarkus-bom I don't see any reason to also put it in kogito-bom.

cristianonicolai:
np, thanks @evacchi @evacchi

PR #395, opened 2020-03-27
KOGITO-1575 DMN assembler-free ikb-free DMNRuntime

tarilabs:
Requires https://github.com/kiegroup/drools/pull/2831

PR #398, opened 2020-03-30
Refactoring - remove duplicated code in DMNKogito

r00ta:
Hi, 

While I was working on some stuff I found out that this class could have been refactored since two methods contains the same code. 

Hope this helps

### DIFF
--- a/drools/kogito-dmn/src/main/java/org/kie/kogito/dmn/DMNKogito.java
+++ b/drools/kogito-dmn/src/main/java/org/kie/kogito/dmn/DMNKogito.java
@@ -16,12 +16,25 @@
 
 package org.kie.kogito.dmn;
 
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.IOException;
+import java.io.Reader;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


drools/kogito-dmn/src/main/java/org/kie/kogito/dmn/DMNKogito.java
```diff
@@ -53,43 +56,26 @@ private DMNKogito() {
...
         try (Stream<Path> fileStream = Files.walk(Paths.get("."))) {
             List<java.nio.file.Path> files = fileStream
-                                                       .filter(path -> Files.isRegularFile(path) && path.toString().endsWith(".dmn"))
-                                                       .peek(x -> LOG.debug("Adding DMN model {} to runtime", x))
-                                                       .collect(Collectors.toList());
+                    .filter(path -> Files.isRegularFile(path) && path.toString().endsWith(".dmn"))
+                    .peek(x -> LOG.debug("Adding DMN model {} to runtime", x))
+                    .collect(Collectors.toList());
             for (java.nio.file.Path file : files) {
-                DMNModel m = compilerImpl.compile(new FileSystemResource(file.toFile()));
```
danielezonca: Please verify that output DMNModels will have `getResource()` properly populated after this refactoring. I remember I had to change from a normal reader to `FileSystemResource` exactly for this reason during Kogito test scenario implementation. If resource is null it could break test scenario execution

r00ta:
I will abandon this PR since this class is already refactored in another open pr https://github.com/kiegroup/kogito-runtimes/pull/395

PR #399, opened 2020-03-30
KOGITO-1599 - Update release script to support stable branch

cristianonicolai:


evacchi:
minor improvement, I suppose you can default to `master` if `$2` is undefined

cristianonicolai:
> minor improvement, I suppose you can default to `master` if `$2` is undefined

In that case we don't need to push the branch. Same would be when releasing 0.9.1, as 0.9.x is already there.

PR #403, opened 2020-03-31
KOGITO-1781: set modelName and modelNamespace in DMNResult

r00ta:
Hi, 

For accountability purposes we need the modelName and modelNamespace properly set in the DMNResult, this is a small PR to do that. 

Best

### DIFF
--- a/drools/kogito-dmn/src/main/java/org/kie/kogito/dmn/DMNKogito.java
+++ b/drools/kogito-dmn/src/main/java/org/kie/kogito/dmn/DMNKogito.java
@@ -113,7 +113,7 @@ public class DMNKogito {
                                                                            null,
                                                                            null,
                                                                            null);
-        return new DMNResult(evaluationResult.result);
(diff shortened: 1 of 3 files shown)


r00ta:
> you can even consider removing this additional constructor and just rely on the one accepting DMNResult for the same end-result. Up to you

Thx, done! I've also updated the dmn endpoint template accordingly

tarilabs:
no wait, I meant exaclty the opposite! :D

r00ta:
> no wait, I meant exaclty the opposite! :D

lol eheh. But I don't get it, the `rg.kie.dmn.api.core.DMNResult` does not contain the `modelName`, how can I set up the `modelName` in the constructor with just the `rg.kie.dmn.api.core.DMNResult` parameter?

r00ta:
@evacchi, Myself and @tarilabs checked and we are fine with this version: we can move forward with this

PR #404, opened 2020-03-31
KOGITO-1306 Error during code generation when using java reserved word on variable

tiagodolphine:
- Allow using reserved words as variable names
- Jackson annotation used to keep the variable name
- Sanitized name on variables

### DIFF
--- a/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ModelMetaData.java
+++ b/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ModelMetaData.java
@@ -16,9 +16,6 @@
 
 package org.jbpm.compiler.canonical;
 
-import static com.github.javaparser.StaticJavaParser.parse;
-import static com.github.javaparser.StaticJavaParser.parseClassOrInterfaceType;
-
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -26,12 +23,7 @@ import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
-import org.drools.core.util.StringUtils;
-import org.jbpm.process.core.context.variable.Variable;
-import org.kie.api.definition.process.WorkflowProcess;
-import org.kie.internal.kogito.codegen.Generated;
-import org.kie.internal.kogito.codegen.VariableInfo;
-
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.github.javaparser.ast.CompilationUnit;
 import com.github.javaparser.ast.Modifier;
 import com.github.javaparser.ast.NodeList;
@@ -58,6 +50,14 @@ import com.github.javaparser.ast.expr.VariableDeclarationExpr;
 import com.github.javaparser.ast.stmt.BlockStmt;
 import com.github.javaparser.ast.stmt.ReturnStmt;
 import com.github.javaparser.ast.type.ClassOrInterfaceType;
+import org.drools.core.util.StringUtils;
+import org.jbpm.process.core.context.variable.Variable;
+import org.kie.api.definition.process.WorkflowProcess;
+import org.kie.internal.kogito.codegen.Generated;
+import org.kie.internal.kogito.codegen.VariableInfo;
+
+import static com.github.javaparser.StaticJavaParser.parse;
+import static com.github.javaparser.StaticJavaParser.parseClassOrInterfaceType;
 
 public class ModelMetaData {
 
@@ -69,13 +69,13 @@ public class ModelMetaData {
     private String visibility;
     private boolean hidden;
     private String templateName;
-    
+
     private boolean supportsValidation;
 
     public ModelMetaData(String processId, String packageName, String modelClassSimpleName, String visibility, VariableDeclarations variableScope, boolean hidden) {
         this(processId, packageName, modelClassSimpleName, visibility, variableScope, hidden, "/class-templates/ModelTemplate.java");
     }
-    
+
     public ModelMetaData(String processId, String packageName, String modelClassSimpleName, String visibility, VariableDeclarations variableScope, boolean hidden, String templateName) {
         this.processId = processId;
         this.packageName = packageName;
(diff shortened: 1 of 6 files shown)


jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/Variable.java
```diff
@@ -64,6 +69,27 @@ public String getName() {
...
+     * @param name the input
+     * @return the output valid Java identifier
+     */
+    private static String sanitizeIdentifier(String name) {
+        return Optional.ofNullable(name)
+                        .filter(SourceVersion::isName)
+                        .orElseGet(() -> Optional.ofNullable(StringUtils.extractFirstIdentifier(name, 0))
+                                .filter(s -> !StringUtils.isEmpty(s))
+                                .filter(SourceVersion::isName)
+                                .orElseGet(() -> "attribute" + StringUtils.generateUUID()));
```
evacchi, line 92: I think we can simplify this by just prepending a "$" sign or a mix of "$" sign and some other stuff. This will keep it readable and make it safe. 

```suggestion
       // prepend v$ in front of the variable name to prevent clashing with reserved keywords
        return String.format("v$%s", name);
```

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ModelMetaData.java
```diff
@@ -112,26 +112,27 @@ public BlockStmt copyInto(String sourceVarName, String destVarName, ModelMetaDat
...
-            value = value.substring(2, value.length() -1);
+            value = value.substring(2, value.length() - 1);
         }
-        
+
         return callSetter(targetVar, destField, new NameExpr(value));
     }
 
     public MethodCallExpr callSetter(String targetVar, String destField, Expression value) {
+        String name = variableScope.getTypes().get(destField).getSanitizedName();
```
evacchi, line 133: because setters and getters are always prepended with `get` and `set` they can never be invalid. So maybe we can use the "regular" name?

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ModelMetaData.java
```diff
@@ -112,26 +112,27 @@ public BlockStmt copyInto(String sourceVarName, String destVarName, ModelMetaDat
         BlockStmt blockStmt = new BlockStmt();
 
         for (Map.Entry<String, String> e : mapping.entrySet()) {
-            String destField = e.getKey();
+            String destField = variableScope.getTypes().get(e.getKey()).getSanitizedName();
```
evacchi, line 115: same here: if we call a getter, then the getter may retain its original name

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ModelMetaData.java
```diff
@@ -175,36 +176,41 @@ private CompilationUnit compilationUnit() {
...
             fd.addAnnotation(new NormalAnnotationExpr(new Name(VariableInfo.class.getCanonicalName()), NodeList.nodeList(new MemberValuePair("tags", new StringLiteralExpr(tags.stream().collect(Collectors.joining(",")))))));
+            fd.addAnnotation(new NormalAnnotationExpr(new Name(JsonProperty.class.getCanonicalName()),
+                                                      NodeList.nodeList(new MemberValuePair("value",
+                                                                                            new StringLiteralExpr(varName)))));
 
             applyValidation(fd, tags);
-            
+
             fd.createGetter();
             fd.createSetter();
```
evacchi, line 196: If these accept params maybe we can override the name. If they don't it may be just enough to create the FD with the unsanitized name and then _after_ these lines add `fd.setName(sanitizedName)`

tiagodolphine: Great idea, I try to do that and see if it works. At the beginning, I tried but the `fd.createGetter();` was automatically generated based on the fd name.. so... 
Let's see.

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/Variable.java
tiagodolphine: done.

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ModelMetaData.java
tiagodolphine: @evacchi the way to make the getter/setter work was to manually change the method name after the `fd.createSetter()`, but even with this we may have problems in case the variable names contain invalid characters like `[]()!*?! ...` to the logic could be more complicated and it will not be 100% compatible with the original name and as @MarianMacik  pointed out it will not be consistent with Java Beans. So IMHO it is better to keep the default getter/setter name unless we found bugs or anything affecting other points on the runtime.

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ModelMetaData.java
tiagodolphine: https://github.com/kiegroup/kogito-runtimes/pull/404#discussion_r401821836

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ModelMetaData.java
tiagodolphine: https://github.com/kiegroup/kogito-runtimes/pull/404#discussion_r401821836

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ModelMetaData.java
evacchi: ok i am not too much opinionated on this, especially because at this time this is an impl detail. Maybe when this becomes public/scaffolding we can revisit if needed :)

evacchi:
@tiagodolphine would you like to cherry pick this to 0.9.x ?

tiagodolphine:
Good question, if we decide this is a bug fix that is good to be incorporated, to be honest, I don't know if using "reserved" word as variable name as that common use case, but it seems to be safe to cherry-pick. wdyt @evacchi @cristianonicolai ?

cristianonicolai:
+1 to cherry pick

PR #407, opened 2020-04-01
KOGITO-1655: Bump 0.9.x to 0.9.1-SNAPSHOT

evacchi:
should we do this for examples too? anything else?

### DIFF
--- a/addons/events/kogito-events-reactive-messaging-addon/pom.xml
+++ b/addons/events/kogito-events-reactive-messaging-addon/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.kie.kogito</groupId>
     <artifactId>events</artifactId>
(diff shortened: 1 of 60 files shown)


radtriste:
you should update all the pom in the repository ;)

radtriste:
and yes, should also be done on kogito-examples

cristianonicolai:
I still not sold on this, would really prefer to avoid this for now. But if this is the direction, please update all three repos, including runtimes, app and examples. Would be important to update the release script as well so this manual steps/prs aren't needed.

radtriste:
kogito-examples has been updated. kogito-apps is in progress https://github.com/kiegroup/kogito-apps/pull/140#pullrequestreview-385777587
I am focusing on deploy pipeline from kogito-runtimes
PR Jenkinsfiles should already handle correctly this as we are using `CHANGE_TARGET` variable

PR #408, opened 2020-04-01
[KOGITO-1651] prevent compilation of drl having a query with no binding

mariofusco:


evacchi:
@mariofusco please also cherry pick to 0.9.x branch when this is merged

PR #412, opened 2020-04-03
[KOGITO-1492] Allow direct injection of event listeners

kostola:
I also added some utility methods in `CodegenUtils` and refactored `DependencyInjectionAnnotator` to make it more flexible (e.g. it allows direct return of the annotated nodes).

Edit: moved the refactoring out of this PR since it was getting too big and out of scope.

Edit 2: [Related PR in kogito-examples](https://github.com/kiegroup/kogito-examples/pull/192)


kogito-codegen/src/main/java/org/kie/kogito/codegen/GeneratedFile.java
```diff
@@ -23,6 +23,7 @@
         APPLICATION,
         APPLICATION_SECTION,
         APPLICATION_CONFIG,
```
r00ta, line 25: Is `APPLICATION_CONFIG` still used?

drools/drools-core/src/main/java/org/drools/core/config/CachedRuleEventListenerConfig.java
```diff
@@ -24,22 +24,32 @@
...
+
+    public CachedRuleEventListenerConfig(List<AgendaEventListener> agendaListeners, List<RuleRuntimeEventListener> ruleRuntimeListeners) {
+        this.agendaListeners = agendaListeners;
+        this.ruleRuntimeListeners = ruleRuntimeListeners;
+    }
 
     public CachedRuleEventListenerConfig register(AgendaEventListener listener) {
-        agendaEventListeners.add(listener);
+        agendaListeners.add(listener);
         return this;
```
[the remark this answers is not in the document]
evacchi: this is unrelated to this PR. The API was supposed to allow chaining.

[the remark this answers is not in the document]
gitgabrio: @evacchi 
Yup, I saw it is not related to this PR.
I'm asking because it seems to lead to something like
`new CachedRuleEventListenerConfig(...).register(....).somethingelse(...)` that looks a lot like a mixed pattern (maybe better avoided, IMHO)

drools/drools-core/src/main/java/org/drools/core/config/CachedRuleEventListenerConfig.java
```diff
@@ -24,22 +24,32 @@
...
     }
-    
+
     public CachedRuleEventListenerConfig register(RuleRuntimeEventListener listener) {
         ruleRuntimeListeners.add(listener);
```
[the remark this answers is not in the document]
kostola: Same answer as above :smile:

kogito-codegen/src/main/java/org/kie/kogito/codegen/ApplicationGenerator.java
```diff
@@ -148,11 +151,11 @@ CompilationUnit compilationUnit() {
...
-                .addModifier(Modifier.Keyword.PROTECTED)
-                .addVariable(new VariableDeclarator()
-                                     .setType(Config.class.getCanonicalName())
-                                     .setName("config")
-                                     .setInitializer(configGenerator.newInstance()));
+                    .addModifier(Modifier.Keyword.PROTECTED)
+                    .addVariable(new VariableDeclarator()
+                            .setType(Config.class.getCanonicalName())
+                            .setName("config")
+                            .setInitializer(new NullLiteralExpr() /* TODO: fix this - configGenerator.newInstance() */));
```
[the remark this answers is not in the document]
kostola: Resolving since this code is outdated after moving the refactoring out of this PR.

kogito-codegen/src/main/java/org/kie/kogito/codegen/GeneratedFile.java
kostola: Resolving since this code is outdated after moving the refactoring out of this PR.

kogito-codegen/src/main/java/org/kie/kogito/codegen/CodegenUtils.java
```diff
@@ -0,0 +1,316 @@
...
+import com.github.javaparser.ast.expr.SimpleName;
+import com.github.javaparser.ast.expr.VariableDeclarationExpr;
+import com.github.javaparser.ast.stmt.BlockStmt;
+import com.github.javaparser.ast.stmt.IfStmt;
+import com.github.javaparser.ast.stmt.ReturnStmt;
+import com.github.javaparser.ast.type.ClassOrInterfaceType;
+import com.github.javaparser.ast.type.Type;
+import org.kie.kogito.codegen.di.DependencyInjectionAnnotator;
+
+public class CodegenUtils {
```
evacchi, line 41: I am not in love with this class, starting with the name CodegenUtils, it easily makes it the target for a "dump" containing anything. I see that you only moved it, but the number of methods has grown a lot.
In general, I usually create "topic" modules like `Fields`, `Types`, etc... 

but in this case I would argue that the constructors found in javaparser are in general more than enough. 
VariableDeclarator etc etc is definitely a bit boring to use, but then just add a shorthand for variable declarator :) 

besides, I think that this code is mostly unused, right?

kogito-codegen/src/main/java/org/kie/kogito/codegen/CodegenUtils.java
```diff
@@ -0,0 +1,316 @@
...
+        return new FieldDeclaration()
+                .setModifiers(modifiers)
+                .addVariable(variableDeclarator(type, name, expr));
+    }
+
+    public static FieldDeclaration field(Modifier.Keyword[] modifiers, Class<?> type, String name) {
+        return new FieldDeclaration()
+                .setModifiers(modifiers)
+                .addVariable(variableDeclarator(type, name));
+    }
```
evacchi, line 80: I mean is this more readable than:

```java
        new FieldDeclaration()
                .setModifiers(PUBLIC)
                .addVariable(new VariableDeclarator(type, name))
```

or even:
```java
        new FieldDeclaration(nodeList(PUBLIC, new VariableDeclarator(type, name))
```

kogito-codegen/src/main/java/org/kie/kogito/codegen/di/DependencyInjectionAnnotator.java
```diff
@@ -42,140 +43,184 @@
...
+    <T extends NodeWithAnnotations<?>> T withNamed(T node, String name);
+
     /**
      * Annotates given node with application level annotations e.g. ApplicationScoped, Component
+     *
      * @param node node to be annotated
      */
-    void withApplicationComponent(NodeWithAnnotations<?> node);
-    
+    <T extends NodeWithAnnotations<?>> T withApplicationComponent(T node);
```
evacchi, line 58: I am not against this change per se, but why is it needed?

kostola: This syntax allows to keep the old behavior but also chain calls together.
I used it and I think it makes the code more concise, like this: https://github.com/kiegroup/kogito-runtimes/pull/412/files#diff-50ac5c7365b5b0682209e3695be81b51R74

kogito-codegen/src/main/java/org/kie/kogito/codegen/CodegenUtils.java
kostola: In my idea, this class is useful to make the code-generation code more readable, since a lot of common patterns in JavaParser are constantly repeated (e.g. creating a variable, a generic type definition, etc.).

Here is a snippet in which some of those methods are used: https://github.com/kiegroup/kogito-runtimes/pull/412/files#diff-50ac5c7365b5b0682209e3695be81b51R105-R133

I would like to keep this class. I agree with you on splitting it into several `*Utils` classes to avoid having a big dump of everything in the end, maybe moving all of them in a dedicated `utils` subpackage.

I think I'm using all of these methods somehow, but not all the overloaded version, which could be cleaned up as well.

In addition, there was already a `CodegenUtils` class in the `process` subpackage that was instead used all around in rules and decisions as well. I moved those methods in this class.

Let me know what you think about my suggestions.

kogito-codegen/src/main/java/org/kie/kogito/codegen/CodegenUtils.java
kostola: Probably my knowledge of JavaParser APIs is not deep enough. I can clean this up if you want.

kogito-codegen/src/main/java/org/kie/kogito/codegen/di/DependencyInjectionAnnotator.java
evacchi: ok thanks for the clarification. I don't think the self-recursive type is required unless you need to explicitly to access metods of the subtype, which I don't think is the case. Let's keep it, but FYI it can be simplified for readability

kogito-codegen/src/main/java/org/kie/kogito/codegen/CodegenUtils.java
kostola: I simplified `CodegenUtils` keeping only the meaningful methods.

kogito-codegen/src/main/java/org/kie/kogito/codegen/CodegenUtils.java
kostola: I removed this utils method to use JavaParser APIs directly as you suggested.

kogito-codegen/src/main/java/org/kie/kogito/codegen/decision/config/DecisionConfigGenerator.java
```diff
@@ -0,0 +1,133 @@
...
+    private static final String VAR_DECISION_EVENT_LISTENER_CONFIG = "decisionEventListenerConfigs";
+    private static final String VAR_DEFAULT_DECISION_EVENT_LISTENER_CONFIG = "defaultDecisionEventListenerConfig";
+    private static final String VAR_DMN_RUNTIME_EVENT_LISTENERS = "dmnRuntimeEventListeners";
+
+    private DependencyInjectionAnnotator annotator;
+
+    private List<BodyDeclaration<?>> members = new ArrayList<>();
+
+    public ObjectCreationExpr newInstance() {
+        if (annotator != null) {
```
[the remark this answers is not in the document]
kostola: @danielezonca all the `*ConfigGenerator` classes (and a few others) don't implement the `useInjection` method but checks directly for `annotator != null`.
I agree that we should make it consistent.

Do you want me to change everywhere (or at least all the `*ConfigGenerator` classes)?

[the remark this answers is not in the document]
danielezonca: Ok so in this case it is fine to keep it. Can you please just create a new ticket to review/refactor this check?

evacchi:
the build is fine, coverage looks low because it's mostly about codegen.

PR #422, opened 2020-04-06
KOGITO-1564 kogito-runtimes: cloud events are empty in native mode

tiagodolphine:
kogito-runtimes: cloud events are empty in native mode (persistence and events)

### DIFF
--- a/kogito-quarkus-extension/deployment/src/main/java/org/kie/kogito/quarkus/deployment/KogitoAssetsProcessor.java
+++ b/kogito-quarkus-extension/deployment/src/main/java/org/kie/kogito/quarkus/deployment/KogitoAssetsProcessor.java
@@ -205,9 +205,13 @@ public class KogitoAssetsProcessor {
                     launchMode, resource, curateOutcomeBuildItem);
 
             reflectiveClass.produce(
+                    new ReflectiveClassBuildItem(true, true, "org.kie.kogito.event.AbstractDataEvent"));
+            reflectiveClass.produce(
                     new ReflectiveClassBuildItem(true, true, "org.kie.kogito.services.event.AbstractProcessDataEvent"));
             reflectiveClass.produce(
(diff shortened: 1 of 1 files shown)


cristianonicolai:
not sure if in time for 0.9.1 but we need some way to do integration tests to verify this sort of problems. Not sure if an integration-test module, or even relying on the examples ( travel-agency ) to make sure we can produce and consume these events.

radtriste:
I will try to test this today by EOD. if it works, it is not a big change to cherry-pick to 0.9.x that I will be able to test quickly then

radtriste:
you can implement some integration tests. keep in mind also that on Operator side we have the BDD tests which are testing this: https://github.com/kiegroup/kogito-cloud-operator/blob/00d7e526bb922d39d520370ad4cae330821ec191/test/features/install_dataindex.feature#L25
This feature test that message from app are sent to the data-index. We test this with both non native and native mode.

radtriste:
in the pipeline, we are missing the jar artifacts for kogito-apps.
@cristianonicolai @tiagodolphine do you think it would make sense also to archive artifacts from kogito-apps in kogito-runtimes PR Jenkinsfile ?
not a big fan but it would avoid to have to create a PR on kogito-apps just to be able to retrieve the artifacts ...
wdyt ?

tiagodolphine:
@radtriste see this PR I opened as well https://github.com/kiegroup/kogito-examples/pull/186 I needed some changes to be able to native build the travel agency

radtriste:
I am testing with the jbpm-quarkus-example for now ;)

tiagodolphine:
Ok, easier in this way... :)

radtriste:
let me know when cherry-picked to 0.9.x branch

cristianonicolai:
@radtriste I can add the artifacts in the pipeline, that is done for kogito-apps already, I think it makes sense.

PR #424, opened 2020-04-07
KOGITO-1635 Update Kogito to Quarkus 1.3.1.Final 

tiagodolphine:
Update Kogito to Quarkus 1.3.1.Final +  align versions on POM
Integration tests: http port change to avoid conflicts (instead of 8080)

I tested the Kogito Apps and Travel Agency, it seems all are working fine with Quarkus 1.3.1


cristianonicolai:
@radtriste, to be honest, I'm not sure we should at this point, as we might not have much time to fix any issues

radtriste:
oki. fair enough.
I updated the issue so we know.

PR #433, opened 2020-04-09
KOGITO-1799: kogito-runtimes: build using Java 8 compatibility

evacchi:
related apps https://github.com/kiegroup/kogito-apps/pull/169


pom.xml
```diff
@@ -956,7 +956,6 @@
           <configuration>
             <showDeprecation>true</showDeprecation>
             <showWarnings>true</showWarnings>
-            <release>${maven.compiler.release}</release>
```
radtriste, line 959: so this will make the compilation to be on Java 8 ?
where is set the fact that we want this to be compiled with Java 8 ?

evacchi: apparently i forgot a commit...

radtriste: why don't you use the same `maven.compiler.release` property ?

evacchi: only because it wouldn't build with JDK 8 otherwise. But i guess we can ignore that

PR #447, opened 2020-04-15
[KOGITO-1869] Assign unique id to DMN evaluations

kostola:
In this PR I enhanced `DmnDecisionModel` to inject a new `UUID` into the `DMNMetadata` (in a predefined field) right before the calls to `evaluateAll` and `evaluateDecisionService` of the underlying  `DMNRuntime`.

### DIFF
--- a/api/kogito-api/src/main/java/org/kie/kogito/ExecutionIdSupplier.java
+++ b/api/kogito-api/src/main/java/org/kie/kogito/ExecutionIdSupplier.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.kie.kogito;
 
 import java.util.function.Supplier;
--- a/api/kogito-api/src/main/java/org/kie/kogito/ExecutionIdSupplier.java
+++ b/api/kogito-api/src/main/java/org/kie/kogito/ExecutionIdSupplier.java
@@ -1,0 +1,11 @@
+package org.kie.kogito;
+
+import java.util.function.Supplier;
+
+/**
+ * Wraps logic to generate unique identifiers per execution (e.g. evaluation of DMN model)
+ *
+ * Each call of the {@link #get()} method is considered a separate request of a new ID.
+ */
+public interface ExecutionIdSupplier extends Supplier<String> {
+}
--- a/api/kogito-api/src/main/java/org/kie/kogito/decision/DecisionExecutionIdUtils.java
+++ b/api/kogito-api/src/main/java/org/kie/kogito/decision/DecisionExecutionIdUtils.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.kie.kogito.decision;
 
 import java.util.Optional;
--- a/api/kogito-api/src/main/java/org/kie/kogito/decision/DecisionExecutionIdUtils.java
+++ b/api/kogito-api/src/main/java/org/kie/kogito/decision/DecisionExecutionIdUtils.java
@@ -1,0 +1,25 @@
+package org.kie.kogito.decision;
+
+import java.util.Optional;
+
+import org.kie.dmn.api.core.DMNContext;
+import org.kie.kogito.ExecutionIdSupplier;
+
+public class DecisionExecutionIdUtils {
+
+    private static final String EXECUTION_ID_KEY = "__kogito_execution_id__";
+
+    public static String get(DMNContext context) {
+        return Optional.ofNullable(context)
+                .map(DMNContext::getMetadata)
+                .map(metadata -> metadata.get(EXECUTION_ID_KEY))
+                .map(Object::toString)
+                .orElse(null);
+    }
+
+    public static DMNContext inject(DMNContext context, ExecutionIdSupplier executionIdSupplier) {
+        context.getMetadata().set(EXECUTION_ID_KEY, executionIdSupplier.get());
+        return context;
+    }
+
+}
--- a/api/kogito-api/src/test/java/org/kie/kogito/decision/DecisionExecutionIdUtilsTest.java
+++ b/api/kogito-api/src/test/java/org/kie/kogito/decision/DecisionExecutionIdUtilsTest.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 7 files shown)


drools/kogito-dmn/src/main/java/org/kie/kogito/dmn/DmnDecisionModel.java
```diff
@@ -25,11 +27,17 @@ public DMNContext newContext(Map<String, Object> variables) {
...
     }
 
     @Override
     public DMNResult evaluateDecisionService(DMNContext context, String decisionServiceName) {
-        return dmnRuntime.evaluateDecisionService(dmnRuntime.getModel(namespace, name), context, decisionServiceName);
+        return dmnRuntime.evaluateDecisionService(dmnRuntime.getModel(namespace, name), injectEvaluationId(context), decisionServiceName);
+    }
+
+    private static DMNContext injectEvaluationId(DMNContext context) {
+        context.getMetadata().set(EVALUATION_ID_KEY, UUID.randomUUID().toString());
```
danielezonca, line 39: Can you please wrap the `UUID` in a `ExecutionIDProvider` interface or similar so that we can change ID logic in a single place? You can add this object to the constructor. This logic will become more complex when we will extend tracing to processes and rules because we want to have end to end tracing and this ID needs to be stable for the whole execution.

kostola:
> I suggest you can instead "inject" it while calling the `DmnDecisionModel.newContext(...)`

@tarilabs I think this is a good idea. I have a question about it: as far as I can see, the `newContext` method is called only inside the code-generated REST `*Resource` classes (one for DMN model).

What if we remove completely the `newContext` method from the `DecisionModels` interface and change `evaluateAll` and `evaluateDecisionService` to accept directly the `Map<String, Object>` map?

**Pros:**
- This would allow us to be sure that a new `DMNContext` is created before each evaluation (which is what we want I assume), then we can inject the evaluation ID and execute it.
- If the developer needs to access the `DMNContext` after the evaluation, it can still do it from the reference in `DMNResult`.

**Cons:**
- The developer would not be able to access the `DMNContext` before the evaluation begins (but based on the current code it would not be possible anyway since the only call is in the code-generated part).

What do you think?

tarilabs:
> What if we remove completely the newContext method from the DecisionModels interface and change evaluateAll and evaluateDecisionService to accept directly the Map<String, Object> map?

No unfortunately that doesn't fly because the "API contract" is about supplying a `DMNContext`.
I believe the false assumption there was about looking just to Kogito code ;)
Moreover that would also not fly when we will move to a more strongly typed API (where we will supply a `DMNContext` which is not a wrapper for a Map)

drools/kogito-dmn/src/main/java/org/kie/kogito/dmn/DmnDecisionModel.java
kostola: I think it's a good idea, but then I this `ExecutionIdProvider` in my opinion should also contain the logic to extract the evaluation ID and be injected as part of the `DecisionConfig` object (maybe we could call it `ExecutionIdManager`).

**Pros:**
- It completely encapsulates the execution ID logic, so that it can be completely swapped if needed. The only "contract" would be that the ID is a `String` and is injected/extracted in/from the `DMNContext`.
- Other components that need to deal with DMN execution IDs (e.g. [the tracing addon I'm implementing](https://issues.redhat.com/browse/FAI-85)) don't need to know how to extract it in order to read it, they will only rely on the provider object referenced in the `DecisionConfig`.

What do you think?

mswiderski:
+1 to what @tarilabs said and keep in mind that Kogito is not exclusively meant to be used with generated code, it can also be used without that on with extra layer on top of what is generated.

As a side note, why not setting the ID directly in DMNContext? I guess it does not have any side effects as it is set on metadata.

kostola:
@tarilabs thanks for the explanation. I clearly made too narrow assumptions and now I have the full picture. Let's stick to your suggestion then.


> As a side note, why not setting the ID directly in DMNContext? I guess it does not have any side effects as it is set on metadata.

@mswiderski I already had a discussion with @tarilabs about this and agreed to put the ID in the metadata because the metadata is clear area to store information that are not supposed to be involved in the evaluation, even if it's true that storing it directly in the context doesn't have any side effect as you said.

tarilabs:
we can always move the "UUID"-equivalent directly into the `DMNContext` later once the E2E use-case requirement is more clear ;)

mswiderski:
@kostola what I meant (and not clearly described, sorry for that) was to set the ID in the metadata but at the time when DMNContext is created so no need for extra setting inside the Kogito code.

drools/kogito-dmn/src/main/java/org/kie/kogito/dmn/DmnDecisionModel.java
kostola: Based on your suggestion, I wrapped the ID generation logic in `ExecutionIdSupplier`.

kogito-codegen/src/main/resources/class-templates/DMNApplicationClassDeclTemplate.java
```diff
@@ -8,7 +8,7 @@ public void init(org.kie.kogito.Application app) {
     }
 
     public org.kie.kogito.decision.DecisionModel getDecisionModel(java.lang.String namespace, java.lang.String name) {
-        return new org.kie.kogito.dmn.DmnDecisionModel(dmnRuntime, namespace, name);
+        return new org.kie.kogito.dmn.DmnDecisionModel(dmnRuntime, namespace, name, new org.kie.kogito.dmn.DmnExecutionIdSupplier());
```
tarilabs, line 11: Sorry, no. This shall be parametrised in some way, we cannot enforce all users to be impacted by this UUID, especially considering then also Trusty will need to plug their own supplied (for instance, process execution ID).

kostola: I created a [new ticket](https://issues.redhat.com/browse/KOGITO-1928) for the task of making the injection logic pluggable.

In fact the ID is not required in this phase, since the Trusty addon that requires it is not yet ready. I will revert this row back to the previous version and do the injection properly only later when really needed.

tarilabs: sounds ok, you can use JavaParser on the kogito codegen side to append the last argument when configured for the ID provider my2c

api/kogito-api/src/main/java/org/kie/kogito/ExecutionIdSupplier.java
```diff
@@ -0,0 +1,11 @@
+package org.kie.kogito;
```
tarilabs, line 1: missing license header.

tarilabs, line 1: missing license header.

api/kogito-api/src/main/java/org/kie/kogito/decision/DecisionExecutionIdUtils.java
```diff
@@ -0,0 +1,25 @@
+package org.kie.kogito.decision;
```
tarilabs, line 1: missing license header.

api/kogito-api/src/test/java/org/kie/kogito/decision/DecisionExecutionIdUtilsTest.java
```diff
@@ -0,0 +1,98 @@
+package org.kie.kogito.decision;
```
tarilabs, line 1: missing license header.

drools/kogito-dmn/src/main/java/org/kie/kogito/dmn/DmnExecutionIdSupplier.java
```diff
@@ -0,0 +1,14 @@
+package org.kie.kogito.dmn;
```
tarilabs, line 1: missing license header.

kogito-codegen/src/main/resources/class-templates/DMNApplicationClassDeclTemplate.java
kostola: Yes that could be a valid idea :+1:

api/kogito-api/src/main/java/org/kie/kogito/ExecutionIdSupplier.java
kostola: Added missing header

api/kogito-api/src/main/java/org/kie/kogito/decision/DecisionExecutionIdUtils.java
kostola: Added missing header

api/kogito-api/src/test/java/org/kie/kogito/decision/DecisionExecutionIdUtilsTest.java
kostola: Added missing header

drools/kogito-dmn/src/main/java/org/kie/kogito/dmn/DmnExecutionIdSupplier.java
kostola: Added missing header

PR #453, opened 2020-04-17
KOGITO-1892: Spring Boot Archetype: use version property (0.9.x)

evacchi:
fix archetype -- it is pointing an outdated version of spring boot; use property so it is always up-to-date


MarianMacik:
As I don't see it defined anywhere, I presume it will be provided from command line. So I would add it to the README which now currently has just groupId and artifactId.

evacchi:
it's provided from the parent pom /pom.xml

PR #463, opened 2020-04-20
KOGITO-1936 Support Typesafe endpoint generation DMN

lucamolteni:
Support Typesafe endpoint generation DMN

see https://issues.redhat.com/browse/DROOLS-5063
https://issues.redhat.com/browse/DROOLS-5061


lucamolteni:
jenkins test again

api/kogito-api/src/main/java/org/kie/kogito/decision/DecisionModel.java
```diff
@@ -15,15 +15,13 @@
...
-import java.util.Map;
-
 import org.kie.dmn.api.core.DMNContext;
 import org.kie.dmn.api.core.DMNModel;
 import org.kie.dmn.api.core.DMNResult;
 
 public interface DecisionModel {
 
-    DMNContext newContext(Map<String, Object> variables);
+    DMNContext newContext(Object variables);
```
evacchi, line 24: I assume we plan to make this generic in T ? maybe in a future PR? We may have one API for Map and one API for T

kogito-codegen/src/main/java/org/kie/kogito/codegen/decision/DecisionCodegen.java
```diff
@@ -53,9 +58,12 @@
 import static org.drools.core.util.IoUtils.readBytesFromInputStream;
 import static org.kie.api.io.ResourceType.determineResourceType;
 import static org.kie.kogito.codegen.ApplicationGenerator.log;
+import static org.kie.kogito.codegen.ApplicationGenerator.logger;
 
 public class DecisionCodegen extends AbstractGenerator {
 
+    public static String OPEN_API_CONFIGURATION_KEY = "quarkus.kogito.dmn.openapi";
```
evacchi, line 65: 1) currently all APIs are under bare `kogito.*` so I would use that for consistency
note: I _do_ I think we should move them under `quarkus.kogito.*`, when quarkus is in use -- but that's especially because of a hard limitation of the config extension API -- when we will do that change, we will fix this too.
2) I also think we should `kogito.decisions.*` is that fine @tarilabs ?
3) why is it called openapi? why not just call it "typed" or something like that?
4) would it make sense to make `true` the default for kogito?

kogito-codegen/src/main/resources/class-templates/DMNRestResourceTemplate.java
```diff
@@ -18,7 +20,7 @@
     @POST
     @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
-    public Object dmn(java.util.Map<String, Object> variables) {
+    public Object dmn(INPUT_TYPE variables) {
```
evacchi, line 23: I assume this is "template variable", so please use the same convention `$inputType$`)

api/kogito-api/src/main/java/org/kie/kogito/decision/DecisionModel.java
tarilabs, line 24: how about leaving this one as-it-was, and having the overload accepting `FEELPropertyAccessible` ?

api/kogito-api/src/main/java/org/kie/kogito/decision/PropertyValueResult.java
```diff
@@ -0,0 +1,5 @@
+package org.kie.kogito.decision;
```
tarilabs, line 1: dead class?

kogito-quarkus-extension/integration-test/src/main/resources/application.properties
```diff
@@ -14,4 +14,4 @@
 # limitations under the License.
 #
 
-quarkus.http.port=65535
\ No newline at end of file
+quarkus.kogito.dmn.openapi=true
```
tarilabs, line 17: Why is this overriding an existing property? Isn't that used by other Integration Tests?

pom.xml
```diff
@@ -134,7 +134,7 @@
     <version.org.keycloak.image>8.0.1</version.org.keycloak.image>
     <version.org.mockito>3.0.0</version.org.mockito>
     <version.org.mvel>2.4.7.Final</version.org.mvel>
-    <version.org.kie7>7.36.1.Final</version.org.kie7>
+    <version.org.kie7>7.37.0.Final</version.org.kie7>
```
tarilabs, line 137: @mariofusco @evacchi please especially validate this one for us, thank you.

kogito-quarkus-extension/integration-test/src/test/java/io/quarkus/it/kogito/dmn/DMNTest.java
```diff
@@ -0,0 +1,64 @@
+package io.quarkus.it.kogito.dmn;
```
tarilabs, line 1: missing license header

kogito-quarkus-extension/integration-test/src/main/resources/application.properties
lucamolteni: ? it's not overriding, it's defining the property.

kogito-codegen/src/main/java/org/kie/kogito/codegen/decision/DecisionCodegen.java
tarilabs: 4. nope, as there are known issues pending, so we want this flag as experimental

kogito-quarkus-extension/integration-test/src/main/resources/application.properties
tarilabs: Understood offline this file previously was ignored, as it was in a wrong location, so whatever was present in this file was not relevant. My original comment then has no basis. Resolving.

kogito-codegen/src/main/java/org/kie/kogito/codegen/decision/DecisionCodegen.java
lucamolteni: @evacchi feature is still in progress, therefore the false default but yes, it'd make total sense

pom.xml
lucamolteni: asked in chat yesterday @tarilabs

kogito-codegen/src/main/resources/class-templates/DMNRestResourceTemplate.java
```diff
@@ -10,6 +10,8 @@
 import org.kie.kogito.dmn.rest.DMNEvaluationErrorException;
 import org.kie.kogito.dmn.rest.DMNResult;
 
+import org.kie.dmn.api.core.FEELPropertyAccessible;
```
danielezonca, line 13: Do we need this import?
If it is optional I would prefer to add it if used (or use FQCN) instead of leave an unused import. Wdyt?

kogito-codegen/src/main/java/org/kie/kogito/codegen/decision/DecisionCodegen.java
```diff
@@ -168,6 +188,27 @@ public DecisionContainerGenerator moduleGenerator() {
...
+                    .generateSourceCodeOfAllTypes();
+
+            allTypesSourceCode.entrySet().stream()
+                    .map(kv -> {
+                        String key = kv.getKey().replace(".", "/") + ".java";
+                        return new GeneratedFile(GeneratedFile.Type.CLASS, key, kv.getValue());
+                    })
+                    .forEach(generatedFiles::add);
+        } catch(Exception e) {
+            logger.error("Unable to generate Strongly Typed Input for: {} {}", model.getNamespace(), model.getName());
```
danielezonca, line 208: When does this happen?
Btw I think we should just fail and not proceed: it is opt-in so I expect if user switch on this feature and it is not possible to make it work for a problem in the model it should fail. Wdyt?

lucamolteni: Agree

pom.xml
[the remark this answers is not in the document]
lucamolteni: To me it's the same, anyway I added in the comment

lucamolteni:
jenkins test again

lucamolteni:
jenkins test again

lucamolteni:
jenkins test again

lucamolteni:
jenkins test again

mareknovotny:
there is outage check https://status.quay.io/ before running it again

evacchi:
removed reference to bump as that was done in https://github.com/kiegroup/kogito-runtimes/commit/fb16446f6b631d5c0d8f604f5853698973b1a3c1

PR #467, opened 2020-04-21
[KOGITO-1248] Add casemgmt tests

ruromero:

Fix https://issues.redhat.com/browse/KOGITO-1248

Added case management processes from JBPM7.x and implemented visitors and missing methods for DynamicNodes, MilestoneNodes and StateNodes

### DIFF
--- a/api/kogito-api/src/main/java/org/kie/api/definition/process/Node.java
+++ b/api/kogito-api/src/main/java/org/kie/api/definition/process/Node.java
@@ -72,9 +72,9 @@ public interface Node {
     /**
      * The NodeContainer this Node lives in.
      *
-     * @return the NodeContainer
+     * @return the parent NodeContainer
      */
-    NodeContainer getNodeContainer();
+    NodeContainer getParentContainer();
 
     /**
      * Meta data associated with this Node.
--- a/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractCompositeNodeVisitor.java
+++ b/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractCompositeNodeVisitor.java
@@ -17,21 +17,27 @@ package org.jbpm.compiler.canonical;
 
 import com.github.javaparser.ast.stmt.BlockStmt;
 import org.jbpm.process.core.context.variable.VariableScope;
+import org.jbpm.workflow.core.node.CompositeContextNode;
 import org.kie.api.definition.process.Node;
 
 import java.util.Map;
 
-public abstract class AbstractCompositeNodeVisitor extends AbstractNodeVisitor {
+public abstract class AbstractCompositeNodeVisitor<T extends CompositeContextNode> extends AbstractNodeVisitor<T> {
 
-    protected Map<Class<?>, AbstractNodeVisitor> nodesVisitors;
+    protected Map<Class<?>, AbstractNodeVisitor<? extends Node>> nodesVisitors;
 
-    public AbstractCompositeNodeVisitor(Map<Class<?>, AbstractNodeVisitor> nodesVisitors) {
+    public AbstractCompositeNodeVisitor(Map<Class<?>, AbstractNodeVisitor<? extends Node>> nodesVisitors) {
         this.nodesVisitors = nodesVisitors;
     }
 
-    protected void visitNodes(String factoryField, Node[] nodes, BlockStmt body, VariableScope variableScope, ProcessMetaData metadata) {
-        for (Node node : nodes) {
-            AbstractNodeVisitor visitor = nodesVisitors.get(node.getClass());
+    @Override
+    public void visitNode(T node, BlockStmt body, VariableScope variableScope, ProcessMetaData metadata) {
+        super.visitNode(node, body, variableScope, metadata);
+    }
+
+    protected <U extends Node> void visitNodes(String factoryField, U[] nodes, BlockStmt body, VariableScope variableScope, ProcessMetaData metadata) {
+        for (U node : nodes) {
+            AbstractNodeVisitor<U> visitor = (AbstractNodeVisitor<U>) nodesVisitors.get(node.getClass());
             if (visitor == null) {
                 continue;
             }
--- a/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractCompositeNodeVisitor.java
+++ b/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractCompositeNodeVisitor.java
@@ -30,11 +30,6 @@ public abstract class AbstractCompositeNodeVisitor<T extends CompositeContextNod
         this.nodesVisitors = nodesVisitors;
     }
 
-    @Override
-    public void visitNode(T node, BlockStmt body, VariableScope variableScope, ProcessMetaData metadata) {
-        super.visitNode(node, body, variableScope, metadata);
-    }
-
     protected <U extends Node> void visitNodes(String factoryField, U[] nodes, BlockStmt body, VariableScope variableScope, ProcessMetaData metadata) {
         for (U node : nodes) {
             AbstractNodeVisitor<U> visitor = (AbstractNodeVisitor<U>) nodesVisitors.get(node.getClass());
--- a/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractNodeVisitor.java
+++ b/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractNodeVisitor.java
@@ -82,12 +82,10 @@ public abstract class AbstractNodeVisitor<T extends Node> extends AbstractVisito
             variableMethod.addArgument(arg);
         }
 
-        AssignExpr assignExpr = new AssignExpr(
+        return new AssignExpr(
                 new VariableDeclarationExpr(type, variableName),
                 variableMethod,
                 AssignExpr.Operator.ASSIGN);
-
-        return assignExpr;
     }
 
     public static Statement makeAssignment(Variable v) {
@@ -140,7 +138,7 @@ public abstract class AbstractNodeVisitor<T extends Node> extends AbstractVisito
 
     protected String extractVariableFromExpression(String variableExpression) {
         if (variableExpression.startsWith("#{")) {
-            return variableExpression.substring(2, variableExpression.indexOf("."));
+            return variableExpression.substring(2, variableExpression.indexOf('.'));
         }
         return variableExpression;
     }
--- a/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractNodeVisitor.java
+++ b/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractNodeVisitor.java
@@ -52,13 +52,6 @@ import static org.jbpm.ruleflow.core.factory.StateBasedNodeFactory.METHOD_TIMER;
 
 public abstract class AbstractNodeVisitor extends AbstractVisitor {
 
-    public static final String METADATA_TRIGGER_REF = "TriggerRef";
-    public static final String METADATA_MESSAGE_TYPE = "MessageType";
-    public static final String METADATA_TRIGGER_TYPE = "TriggerType";
-    public static final String METADATA_TRIGGER_MAPPING = "TriggerMapping";
-    public static final String METADATA_MAPPING_VARIABLE = "MappingVariable";
-    public static final String METADATA_EVENT_TYPE = "EventType";
-
     protected static final String EVENT_TYPE_SIGNAL = "signal";
     protected static final String EVENT_TYPE_MESSAGE = "message";
 
--- a/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractNodeVisitor.java
+++ b/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractNodeVisitor.java
@@ -50,25 +50,25 @@ import static org.jbpm.ruleflow.core.factory.NodeFactory.METHOD_DONE;
 import static org.jbpm.ruleflow.core.factory.NodeFactory.METHOD_NAME;
 import static org.jbpm.ruleflow.core.factory.StateBasedNodeFactory.METHOD_TIMER;
 
-public abstract class AbstractNodeVisitor extends AbstractVisitor {
+public abstract class AbstractNodeVisitor<T extends Node> extends AbstractVisitor {
 
     protected static final String EVENT_TYPE_SIGNAL = "signal";
     protected static final String EVENT_TYPE_MESSAGE = "message";
 
     protected abstract String getNodeKey();
 
-    public void visitNode(Node node, BlockStmt body, VariableScope variableScope, ProcessMetaData metadata) {
+    public void visitNode(T node, BlockStmt body, VariableScope variableScope, ProcessMetaData metadata) {
         visitNode(FACTORY_FIELD_NAME, node, body, variableScope, metadata);
     }
 
-    protected String getNodeId(Node node) {
+    protected String getNodeId(T node) {
         return getNodeKey() + node.getId();
     }
 
-    public void visitNode(String factoryField, Node node, BlockStmt body, VariableScope variableScope, ProcessMetaData metadata) {
+    public void visitNode(String factoryField, T node, BlockStmt body, VariableScope variableScope, ProcessMetaData metadata) {
     }
 
-    protected MethodCallExpr getNameMethod(Node node, String defaultName) {
+    protected MethodCallExpr getNameMethod(T node, String defaultName) {
         return getFactoryMethod(getNodeId(node), METHOD_NAME, new StringLiteralExpr(getOrDefault(node.getName(), defaultName)));
     }
 
@@ -170,18 +170,5 @@ public abstract class AbstractNodeVisitor extends AbstractVisitor {
         body.addStatement(getFactoryMethod(factoryField, "connection", new LongLiteralExpr(connection.getFrom().getId()),
                 new LongLiteralExpr(connection.getTo().getId()),
                 new StringLiteralExpr(getOrDefault((String) connection.getMetaData().get("UniqueId"), ""))));
-    }
-
-    protected void addTimers(BlockStmt body, StateBasedNode node) {
-        if (node.getTimers() != null) {
-            node.getTimers().forEach((timer, action) -> {
-                DroolsConsequenceAction droolsAction = (DroolsConsequenceAction) action;
-                body.addStatement(getFactoryMethod(getNodeId(node), METHOD_TIMER,
-                        new StringLiteralExpr(timer.getDelay()),
-                        getOrNullExpr(timer.getPeriod()),
-                        new StringLiteralExpr(droolsAction.getDialect()),
-                        new StringLiteralExpr(StringEscapeUtils.escapeJava(droolsAction.getConsequence()))));
-            });
-        }
     }
 }
--- a/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractNodeVisitor.java
+++ b/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractNodeVisitor.java
@@ -48,15 +48,16 @@ import static org.jbpm.ruleflow.core.factory.MappableNodeFactory.METHOD_IN_MAPPI
 import static org.jbpm.ruleflow.core.factory.MappableNodeFactory.METHOD_OUT_MAPPING;
 import static org.jbpm.ruleflow.core.factory.NodeFactory.METHOD_DONE;
 import static org.jbpm.ruleflow.core.factory.NodeFactory.METHOD_NAME;
+import static org.jbpm.ruleflow.core.factory.StateBasedNodeFactory.METHOD_TIMER;
 
 public abstract class AbstractNodeVisitor extends AbstractVisitor {
 
-    protected static final String METADATA_TRIGGER_REF = "TriggerRef";
-    protected static final String METADATA_MESSAGE_TYPE = "MessageType";
-    protected static final String METADATA_TRIGGER_TYPE = "TriggerType";
-    protected static final String METADATA_TRIGGER_MAPPING = "TriggerMapping";
-    protected static final String METADATA_MAPPING_VARIABLE = "MappingVariable";
-    protected static final String METADATA_EVENT_TYPE = "EventType";
+    public static final String METADATA_TRIGGER_REF = "TriggerRef";
+    public static final String METADATA_MESSAGE_TYPE = "MessageType";
+    public static final String METADATA_TRIGGER_TYPE = "TriggerType";
+    public static final String METADATA_TRIGGER_MAPPING = "TriggerMapping";
+    public static final String METADATA_MAPPING_VARIABLE = "MappingVariable";
+    public static final String METADATA_EVENT_TYPE = "EventType";
 
     protected static final String EVENT_TYPE_SIGNAL = "signal";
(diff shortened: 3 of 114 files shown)


jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractNodeVisitor.java
```diff
@@ -48,15 +48,16 @@
...
-    protected static final String METADATA_TRIGGER_TYPE = "TriggerType";
-    protected static final String METADATA_TRIGGER_MAPPING = "TriggerMapping";
-    protected static final String METADATA_MAPPING_VARIABLE = "MappingVariable";
-    protected static final String METADATA_EVENT_TYPE = "EventType";
+    public static final String METADATA_TRIGGER_REF = "TriggerRef";
+    public static final String METADATA_MESSAGE_TYPE = "MessageType";
+    public static final String METADATA_TRIGGER_TYPE = "TriggerType";
+    public static final String METADATA_TRIGGER_MAPPING = "TriggerMapping";
+    public static final String METADATA_MAPPING_VARIABLE = "MappingVariable";
+    public static final String METADATA_EVENT_TYPE = "EventType";
```
evacchi, line 60: if these are public maybe we should move them to some more descriptive class name. Like Messages or something like that?

api/kogito-api/src/main/java/org/kie/api/definition/process/Node.java
```diff
@@ -72,9 +72,9 @@
     /**
      * The NodeContainer this Node lives in.
      *
-     * @return the NodeContainer
+     * @return the parent NodeContainer
      */
-    NodeContainer getNodeContainer();
+    NodeContainer getParentContainer();
```
evacchi, line 77: could it be mis-read? I was thinking this was pointing to the "parent" of this node (i.e. another node).

ruromero: I renamed it because some nodes have their own nodeContainer. Besides, whenever this method is used is assigned to a variable called `parent` and the attribute returned is `parentNodeContainer`.

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractNodeVisitor.java
ruromero: Do you think a Metadata interface would be better? I thought putting these constants in the "owner" visitors or in the common ancestor if they are shared.

ruromero:
jenkins retest this

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/impl/NodeImpl.java
```diff
@@ -245,7 +245,7 @@ public Connection getTo() {
         return getOutgoingConnections(org.jbpm.workflow.core.Node.CONNECTION_DEFAULT_TYPE);
     }
 
-    public NodeContainer getNodeContainer() {
+    public NodeContainer getParentContainer() {
         return parentNodeContainer;
```
MarianMacik, line 249: Hmm, what about calling it `getParentNodeContainer()`? This way it will be consistent, i.e. `getNodeContainer()` for getting this node's container and `getParentNodeContainer()` for getting the node container of this node parent :) Wdyt?

kogito-codegen/src/test/java/org/kie/kogito/codegen/process/ProcessGenerationTest.java
```diff
@@ -194,7 +197,7 @@ private static void assertActions(Node eNode, ExtendedNodeImpl expected, Extende
             try {
                 if (expected.getActions(actionType) == null) {
                     assertNull(current.getActions(actionType));
-                } else {
+                } else if (!EVENT_NODE_ENTER.equals(actionType) && !EVENT_NODE_EXIT.equals(actionType)){
```
MarianMacik, line 200: So in this case the actionType key is not present in a map? In this case we should probably assert for null in the previous if condition, right?

kogito-codegen/src/test/resources/org/kie/kogito/codegen/process/process-generation-test.skip.txt
```diff
@@ -1,3 +1,6 @@
+cases/ScriptRoleAssignmentCase.bpmn2
+cases/UserTaskWithStageCase.bpmn2
+cases/UserTaskWithStageCaseAutoStart.bpmn2
```
MarianMacik, line 3: What is preventing us from testing these 3 cases specifically? You added many more and these are (presumably) tested.

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/impl/NodeImpl.java
ruromero: What about just `getContainer` and `getParentContainer` ? FYI not all nodes have their own container, only `CompositeNode` nodes have it.

jbpm/jbpm-flow/src/main/java/org/jbpm/ruleflow/core/factory/BoundaryEventNodeFactory.java
```diff
@@ -118,24 +115,4 @@ public BoundaryEventNodeFactory eventType(String eventTypePrefix, String eventTy
...
-    }
-
-    public BoundaryEventNodeFactory timeDuration(String timeDuration) {
-        eventType(EVENT_TYPE_TIMER, timeDuration);
-        return metaData(METADATA_TIME_DURATION, timeDuration);
-    }
-
-    public BoundaryEventNodeFactory cancelActivity(boolean cancelActivity) {
-        return metaData(METADATA_CANCEL_ACTIVITY, cancelActivity);
-    }
```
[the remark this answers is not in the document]
ruromero: The main reason is that Metadata attributes are copied in the VisitMetadata method so this seems redundant.

kogito-codegen/src/test/resources/org/kie/kogito/codegen/process/process-generation-test.skip.txt
ruromero: The two UserTaskWithStageCase processes depend on custom model definition and are ignored like in other processes.
The ScriptRoleAssignmentCase should be removed once the case model supports caseAssignments or something equivalent and the process is updated.

kogito-codegen/src/test/java/org/kie/kogito/codegen/process/ProcessGenerationTest.java
MarianMacik: As discussed, this is ok.

ruromero:
I have done all the changes

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/MilestoneNodeVisitor.java
```diff
@@ -0,0 +1,53 @@
...
+    private static final String NODE_KEY = "milestoneNode";
+
+    @Override
+    protected String getNodeKey() {
+        return NODE_KEY;
+    }
+
+    @Override
+    public void visitNode(String factoryField, Node node, BlockStmt body, VariableScope variableScope, ProcessMetaData metadata) {
+        MilestoneNode milestoneNode = (MilestoneNode) node;
```
tiagodolphine, line 42: same here, to avoid the cast it could be used generics, maybe on AbstractNodeVisitor<T>

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/Node.java
```diff
@@ -28,7 +28,7 @@
  */
 public interface Node extends org.kie.api.definition.process.Node, Contextable, Serializable {
 
-    static final String CONNECTION_DEFAULT_TYPE = "DROOLS_DEFAULT";
+    String CONNECTION_DEFAULT_TYPE = "DROOLS_DEFAULT";
```
tiagodolphine, line 31: why the `static final` was removed? Anyway, if it is not a constant anymore you can change the name to camel case.

api/kogito-api/src/main/java/org/kie/api/definition/process/Node.java
tiagodolphine: Only `CompositeNode` has its own nodeContainer? I mean it seems it is returning itself. 
Why not just overriding the original `getNodeContainer` on  `CompositeNode`, or whatever node that has a different behavior other than returning the parent.

kogito-codegen/src/test/resources/org/kie/kogito/codegen/process/process-generation-test.skip.txt
MarianMacik: Ah, I see. And about the custom model - is there a plan to support it as well? Just curious.

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/CompositeContextNodeVisitor.java
```diff
@@ -71,6 +73,8 @@ public void visitNode(String factoryField, Node node, BlockStmt body, VariableSc
             visitVariableScope(getNodeId(node), variableScopeNode, body, new HashSet<>());
         }
 
+        visitCustomFields(compositeContextNode).forEach(body::addStatement);
```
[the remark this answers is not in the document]
MarianMacik: Probably the best option would be to pass body as another argument as with other visitors.

api/kogito-api/src/main/java/org/kie/api/definition/process/Node.java
MarianMacik: Ah, I see now. `CompositeNode` has another `NodeContainer` named `nodeContainer` but this, until now, wasn't returned by any of the `CompositeNode` methods. So by calling `getNodeContainer()` the `NodeContainer` from the `NodeImpl` parent was returned and there was no option to obtain the `NodeContainer` of `CompositeNode` using a getter. Now, the `getParentContainer()` will return the one-level-above container (the one from `NodeImpl`) and `getNodeContainer()` will return the one from `CompositeNode`. In practice, these nodes can be nested into each other many times, but now it is possible to obtain just the container with nodes inside ("belonging to") the `CompositeNode` and the container which `CompositeNode` is in separately.

jbpm/jbpm-flow/src/main/java/org/jbpm/ruleflow/core/factory/BoundaryEventNodeFactory.java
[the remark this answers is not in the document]
MarianMacik: I see, so eventTypes and metaData are "copied" when bpmn is traversed anyway.

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/impl/NodeImpl.java
MarianMacik: Yes, +1.

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/Node.java
ruromero: On interfaces, every field is defined as static final and public so this is redundant.

ruromero:
> Only CompositeNode has its own nodeContainer? I mean it seems it is returning itself.
Why not just overriding the original getNodeContainer on CompositeNode, or whatever node that has a different behavior other than returning the parent.

Because it also has to return the parentNode for some generic methods. It's not only the `CompositeNode` but also it's descendants.

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/Node.java
tiagodolphine: ooh this is an interface, yep you are right!

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/CompositeContextNodeVisitor.java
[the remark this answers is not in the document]
ruromero: > why not passing the function,`body::addStatement` as a parameter to the `visitCustomFields` method and apply it on the method implementation, I mean to keep the same approach of other visit methods, or just change the method name to `getCustomFields` ?

That will make it look very different from the other visit methods.

> Probably the best option would be to pass body as another argument as with other visitors.

I wanted to avoid passing an object to be modified as an argument. This is part of some refactoring I did everywhere that didn't require too big changes, like in visitNodes.

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/DynamicNodeVisitor.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Override
+    protected String getDefaultName() {
+        return "Dynamic";
+    }
+
+    @Override
+    public Stream<MethodCallExpr> visitCustomFields(CompositeContextNode node) {
+        Collection<MethodCallExpr> methods = new ArrayList<>();
+        DynamicNode dynamicNode = (DynamicNode) node;
```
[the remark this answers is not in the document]
ruromero: I have refactored all the visitor classes to make them use generics, I think they look much better. Thanks for the suggestion.

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/AbstractCompositeNodeVisitor.java
```diff
@@ -17,21 +17,27 @@
...
-        for (Node node : nodes) {
-            AbstractNodeVisitor visitor = nodesVisitors.get(node.getClass());
+    @Override
+    public void visitNode(T node, BlockStmt body, VariableScope variableScope, ProcessMetaData metadata) {
+        super.visitNode(node, body, variableScope, metadata);
+    }
+
+    protected <U extends Node> void visitNodes(String factoryField, U[] nodes, BlockStmt body, VariableScope variableScope, ProcessMetaData metadata) {
+        for (U node : nodes) {
+            AbstractNodeVisitor<U> visitor = (AbstractNodeVisitor<U>) nodesVisitors.get(node.getClass());
```
MarianMacik, line 40: Generics now works but just because here we use unsafe casting. Without it the types in the call on line 44 `visitor.visitNode(factoryField, node, body, variableScope, metadata);` wouldn't match as `U[] nodes` is always array of Node (passed from the visitNode method) so U is Node but then we are passing this `node` variable of type `Node` to that method call on line 44. But given we did unsafe casting here, compiler is happy and it is ok after type erasure as during runtime the signature of abstractNodeVisitor visitNode() method has `Node`.

In short, I am not sure if this helped the hierarchy :) We also use raw types for Visitors in ProcessVisitor.java file.

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ProcessVisitor.java
```diff
@@ -103,6 +105,9 @@ public ProcessVisitor(ClassLoader contextClassLoader) {
         this.nodesVisitors.put(CompositeContextNode.class, new CompositeContextNodeVisitor(nodesVisitors));
```
MarianMacik, line 105: Here we are using raw types for example.

Same issue with unsafe calls now appeared in `visitNodes` method as well.

tiagodolphine:
@cristianonicolai would you like to take a look at the PR, or are we ok to merge it?
Thanks.

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/EventSubprocessNodeVisitor.java
```diff
@@ -15,16 +15,27 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Map;
+import java.util.stream.Stream;
 
-public class EventSubprocessNodeVisitor extends CompositeContextNodeVisitor {
+import static org.jbpm.ruleflow.core.factory.EventSubProcessNodeFactory.METHOD_EVENT;
+import static org.jbpm.ruleflow.core.factory.EventSubProcessNodeFactory.METHOD_KEEP_ACTIVE;
+
+public class EventSubprocessNodeVisitor extends CompositeContextNodeVisitor<EventSubProcessNode> {
```
cristianonicolai, line 34:
```suggestion
public class EventSubProcessNodeVisitor extends CompositeContextNodeVisitor<EventSubProcessNode> {
```

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/MilestoneNodeVisitor.java
```diff
@@ -0,0 +1,51 @@
...
+    private static final String NODE_KEY = "milestoneNode";
+
+    @Override
+    protected String getNodeKey() {
+        return NODE_KEY;
+    }
+
+    @Override
+    public void visitNode(String factoryField, MilestoneNode node, BlockStmt body, VariableScope variableScope, ProcessMetaData metadata) {
+        body.addStatement(getAssignedFactoryMethod(factoryField, MilestoneNodeFactory.class, getNodeId(node), NODE_KEY,new LongLiteralExpr(node.getId())))
```
cristianonicolai, line 41:
```suggestion
        body.addStatement(getAssignedFactoryMethod(factoryField, MilestoneNodeFactory.class, getNodeId(node), NODE_KEY, new LongLiteralExpr(node.getId())))
```

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
```diff
@@ -84,7 +84,7 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(ProcessCodegen.class);
 
     private static final SemanticModules BPMN_SEMANTIC_MODULES = new SemanticModules();
-    public static final Set<String> SUPPORTED_BPMN_EXTENSIONS = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(".bpmn", ".bpmn2")));
+    public static final Set<String> SUPPORTED_BPMN_EXTENSIONS = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(".bpmn", ".bpmn2", ".bpmn-cm")));
```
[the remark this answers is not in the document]
ruromero: bpmn-cm is a format introduced in jbpm7 for case modeler. It is compatible with bpmn2 and as we are adding functionalities for case management processes I think it's nice to have. Otherwise, processes created with the case modeler jbpm7.x will have to be renamed to work on Kogito.

jbpm/jbpm-flow/src/main/java/org/jbpm/ruleflow/core/factory/DynamicNodeFactory.java
```diff
@@ -25,6 +25,10 @@
 
 public class DynamicNodeFactory extends CompositeContextNodeFactory {
 
+	public static final String METHOD_LANGUAGE = "language";
```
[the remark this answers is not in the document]
ruromero: It's just a matter of taste, I considered the owner of the method should have the constant that will be used by the visitor to make it easier to maintain and remove/update as the methods are defined here.

jbpm/jbpm-flow/src/main/java/org/jbpm/ruleflow/core/factory/EventSubProcessNodeFactory.java
```diff
@@ -16,13 +16,16 @@
...
 
+import org.jbpm.process.core.event.EventTypeFilter;
 import org.jbpm.ruleflow.core.RuleFlowNodeContainerFactory;
 import org.jbpm.workflow.core.NodeContainer;
 import org.jbpm.workflow.core.node.CompositeContextNode;
 import org.jbpm.workflow.core.node.EventSubProcessNode;
 
 public class EventSubProcessNodeFactory extends CompositeContextNodeFactory {
 
+    public static final String METHOD_KEEP_ACTIVE = "keepActive";
```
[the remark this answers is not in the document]
ruromero: same

jbpm/jbpm-flow/src/main/java/org/jbpm/ruleflow/core/factory/EventSubProcessNodeFactory.java
```diff
@@ -32,4 +35,16 @@ public EventSubProcessNodeFactory(RuleFlowNodeContainerFactory nodeContainerFact
     protected CompositeContextNode createNode() {
         return new EventSubProcessNode();
     }
+
+    public EventSubProcessNodeFactory keepActive(boolean keepActive) {
```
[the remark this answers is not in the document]
ruromero: It is used in the generated code. See `EventSubprocessNodeVisitor#visitCustomFields`

jbpm/jbpm-flow/src/main/java/org/jbpm/ruleflow/core/factory/EventSubProcessNodeFactory.java
```diff
@@ -32,4 +35,16 @@ public EventSubProcessNodeFactory(RuleFlowNodeContainerFactory nodeContainerFact
...
+        ((EventSubProcessNode) getCompositeNode()).setKeepActive(keepActive);
+        return this;
+    }
+
+    public EventSubProcessNodeFactory event(String event) {
```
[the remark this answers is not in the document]
ruromero: same

jbpm/jbpm-flow/src/main/java/org/jbpm/ruleflow/core/factory/MilestoneNodeFactory.java
```diff
@@ -23,6 +23,9 @@
 
 public class MilestoneNodeFactory extends StateBasedNodeFactory {
 
+    public static final String METHOD_CONSTRAINT = "constraint";
```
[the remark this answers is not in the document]
ruromero: same as for the `EventSubProcessNodeFactory` I consider factories are the owners of the methods and these literals should be here.

jbpm/jbpm-flow/src/main/java/org/jbpm/ruleflow/core/factory/StateBasedNodeFactory.java
```diff
@@ -24,6 +24,8 @@
 
 public abstract class StateBasedNodeFactory extends ExtendedNodeFactory {
 
+    public static final String METHOD_TIMER = "timer";
```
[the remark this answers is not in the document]
ruromero: They're not used because timers have not yet been added to `StateBasedNodeVisitor` and its hierarchy. However as this method is not new I decided to add it. I can remove it or try to add timers to `WorkItemNode`, `RuleSetNode`, `SubProcessNode`, `MilestoneNode` and `HumanTaskNode`

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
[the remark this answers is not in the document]
cristianonicolai: @ruromero you do have a point but I believe that it should be discussed with the team to understand our overall strategy around it. At least to me, it is not clear if we want to promote that with Kogito.

jbpm/jbpm-flow/src/main/java/org/jbpm/ruleflow/core/factory/StateBasedNodeFactory.java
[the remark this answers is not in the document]
cristianonicolai: np to leave it then

cristianonicolai:
@ruromero not sure if the failures on kogito-examples are related, you might need to rebase your PR, please take a look.

PR #470, opened 2020-04-22
[KOGITO-1644] Move docker-based test cases to maven-failsafe

jstastny-cz:
Moving tests depending on docker images to integration phase using maven failsafe plugin (in case of 'infinispan-persistence-addon' also enabling previously disabled test).

### DIFF
--- a/addons/persistence/infinispan-persistence-addon/src/test/java/org/kie/kogito/infinispan/CacheProcessInstancesIT.java
+++ b/addons/persistence/infinispan-persistence-addon/src/test/java/org/kie/kogito/infinispan/CacheProcessInstancesIT.java
@@ -25,25 +25,41 @@ import java.util.List;
 
 import org.drools.core.io.impl.ClassPathResource;
 import org.infinispan.client.hotrod.RemoteCacheManager;
+import org.infinispan.client.hotrod.configuration.ClientIntelligence;
 import org.infinispan.client.hotrod.configuration.ConfigurationBuilder;
 import org.infinispan.client.hotrod.impl.ConfigurationProperties;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
+import org.kie.kogito.auth.SecurityPolicy;
 import org.kie.kogito.persistence.KogitoProcessInstancesFactory;
 import org.kie.kogito.process.ProcessInstance;
 import org.kie.kogito.process.WorkItem;
 import org.kie.kogito.process.bpmn2.BpmnProcess;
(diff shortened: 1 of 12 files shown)


addons/persistence/infinispan-persistence-addon/src/test/java/org/kie/kogito/infinispan/CacheProcessInstancesIT.java
```diff
@@ -25,25 +25,41 @@
...
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
 
-@Disabled("Requires running Infinispan instance")
-public class CacheProcessInstancesTest {
+@Testcontainers
+public class CacheProcessInstancesIT {
+
+    @Container
+    public InfinispanServerContainer container = new InfinispanServerContainer();
```
cristianonicolai, line 46: @jstastny-cz if you want, this can be simplified by https://github.com/kiegroup/kogito-examples/pull/243/files#diff-8e9ae312cb9d5554bd4e5cd70c566376R47

jstastny-cz:
@evacchi  @cristianonicolai please review the second commit, the idea behind is not to fail maven build on Integration test failure (by removing the verify goal), but collecting failsafe-reports instead (so that jenkins reports them).

cristianonicolai:
@jstastny-cz how about running it locally, it tests wouldnt fail in this case?

jstastny-cz:
@cristianonicolai the tests will fail, but the maven build won't. That's thanks to having only integration-test goal set in maven-failsafe-plugin. Goal verify would fail the actual maven build, which we don't want.

cristianonicolai:
@evacchi after discussing with @jstastny-cz, I still not confident that removing the _verify_ goal will be helpful. A PR can not be green in case we have IT test failures IMO. So I suggest that we bring in the IT tests from this change, but open a separated PR to discuss the _verify_ goal.

cristianonicolai:
@evacchi first step would be to check which result Jenkins would post in github about failing tests.

evacchi:
@jstastny-cz can you also update the requrements on the README so that it says that Docker is required for running the integration tests? Something like

```
- Maven 3.6.2 
- JDK 11
- optional: Docker for integration tests (use `--DskipITs` to skip them)
```

jstastny-cz:
@evacchi I will add the note to readmes.

Other than that I would not touch this PR and leave the 'verify' goal here. Its removal will be done in a separate PR.

Why this is still faling PR build is that there's matching PR in kogito-apps which suffers from dependent tests - will resolve shortly.

PR #474, opened 2020-04-23
KOGITO-2002: Implemented hot-reload test for adding a new unit.

DuncanDoyle:
@mariofusco @evacchi This is just the hot-reload unit test, so this will fail ... intentionally. But figured it would be helpful to implement and PR the test first :-)


DuncanDoyle:
Fixed by @mariofusco in https://github.com/kiegroup/kogito-runtimes/commit/6eb8d46d0e0567e423fbd2a688295019956ff487

PR #477, opened 2020-04-24
 [KOGITO-1551] - Process Instance persistence in Mongo DB

swati-kale:
Signed-off-by: Swati Kale <swkale@redhat.com>

- [ ] You have read the [contributors guide](https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/README.md)
- [ ] Pull Request title is properly formatted: `[KOGITO-XYZ] Subject`
- [ ] Pull Request title contains the target branch if not targeting master: `[0.9.x] [KOGITO-XYZ] Subject`


addons/persistence/mongodb-persistence-addon/pom.xml
```diff
@@ -0,0 +1,93 @@
...
+	<parent>
+		<groupId>org.kie.kogito</groupId>
+		<artifactId>persistence</artifactId>
+		<version>8.0.0-SNAPSHOT</version>
+	</parent>
+	<artifactId>mongodb-persistence-addon</artifactId>
+	<dependencies>
+		<dependency>
+			<groupId>io.quarkus</groupId>
+			<artifactId>quarkus-mongodb-client</artifactId>
```
mswiderski, line 14: This should depend on mongo client snd not quarkus as it should be free of any framework used as runtime. So the  same addon can be properly used in both quarkus and springboot

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/PersistProcessInstances.java
```diff
@@ -0,0 +1,165 @@
...
+	private final MongoCollection<Document> collection;
+	private ProcessInstanceMarshaller marshaller;
+	private static final String KOGITO_STORE = "Kogito_store";
+
+	public PersistProcessInstances(MongoClient mongoClient, org.kie.kogito.process.Process<?> process) {
+
+		this.process = process;
+		this.mongoDatabase = mongoClient.getDatabase(KOGITO_STORE);
+		this.collection = this.mongoDatabase.getCollection(process.id() + "_store");
+		this.marshaller = new ProcessInstanceMarshaller();
```
mswiderski, line 54: Isn’t this just oversimplification? I mean usind mongodb as key value store? I’d suggest to make use of mongo as different storage so yiu could marshal process instance as document and thus take advantage of mongodb as document db

mswiderski: In addition, this will not work in native mode in my opinion because the variables will be serlialized using java serialization which is not supported (or was when last checked) on substratevm

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/persistence/KogitoProcessInstancesFactory.java
```diff
@@ -0,0 +1,56 @@
...
+
+		this.mongoClient = mongoClient;
+	}
+
+	public PersistProcessInstances createProcessInstances(Process<?> process) {
+		return new PersistProcessInstances(this.mongoClient, process);
+	}
+
+	public String proto() {
+		return null;
```
mswiderski, line 46: Are you planning to use protobuf for variables and thus use this methods?

addons/persistence/mongodb-persistence-addon/pom.xml
swati-kale: Ok, I will add separate mongo-java-driver.

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/persistence/KogitoProcessInstancesFactory.java
swati-kale: No, will remove.

swati-kale:
Modified the PR to save the process instance in document format in Mongo DB instead of byte array.

swati-kale:
Below are the changes,
Saving process instance as a document in the mongodb instead byte array
Reused the existing framework involved in process instance marshalling and JBPMMessages
Model variables are saved in the form of JSON string in the DB instead of byte array
Converted JBPMMessages to JSON and then into mongo document.
Below is the screenshot,
ProcessInstanceMongo

cristianonicolai:
jenkins please retest this

cristianonicolai:
@swati-kale can you please rebase your PR, that should help to get Jenkins to build it

api/kogito-api/pom.xml
```diff
@@ -45,7 +45,14 @@
         </exclusion>
       </exclusions>
     </dependency>
-
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
```
evacchi, line 49: I don't think these dependencies should go into kogito-api

swati-kale:
Thanks, I will go through the review comments and make appropriate changes.

api/kogito-api/pom.xml
swati-kale: I removed it, thanks.

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/utils/CommonUtils.java
```diff
@@ -0,0 +1,57 @@
...
+
+    private CommonUtils() {}
+
+    public static MongoCollection<ProcessInstanceDocument> getCollection(MongoClient mongoClient, String processId) {
+        CodecRegistry registry = fromRegistries(MongoClientSettings.getDefaultCodecRegistry(), fromProviders(PojoCodecProvider.builder().automatic(true).build(), new ProcessInstanceDocumentCodecProvider()));
+        MongoDatabase mongoDatabase = mongoClient.getDatabase(KOGITO_STORE).withCodecRegistry(registry);
+        return mongoDatabase.getCollection(processId, ProcessInstanceDocument.class).withCodecRegistry(registry);
+    }
+
+    public static ProcessInstanceDocument convertToProcessInstanceDoument(ProcessInstanceData data) {
```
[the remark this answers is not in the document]
swati-kale: Thanks, added mapper function classes.

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/utils/CommonUtils.java
```diff
@@ -0,0 +1,57 @@
...
+import org.kie.kogito.mongodb.codec.ProcessInstanceDocumentCodecProvider;
+import org.kie.kogito.mongodb.model.ProcessInstanceDocument;
+import org.kie.kogito.mongodb.model.Strategy;
+
+import static org.bson.codecs.configuration.CodecRegistries.fromProviders;
+import static org.bson.codecs.configuration.CodecRegistries.fromRegistries;
+
+public class CommonUtils {
+
+    private static final String KOGITO_STORE = "Kogito_store";
```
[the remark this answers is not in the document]
swati-kale: Database name can be configured at application.properties of client application else "kogito" as a defult

addons/persistence/mongodb-persistence-addon/src/test/java/org/kie/kogito/mongodb/PersistableProcessInstanceTest.java
```diff
@@ -0,0 +1,109 @@
...
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
+import org.testcontainers.containers.wait.strategy.Wait;
+import org.testcontainers.junit.jupiter.Container;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.kie.api.runtime.process.ProcessInstance.STATE_ACTIVE;
+import static org.kie.api.runtime.process.ProcessInstance.STATE_COMPLETED;
+
+class PersistableProcessInstanceTest {
```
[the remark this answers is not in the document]
swati-kale: done.

api/kogito-api/src/main/java/org/kie/api/marshalling/ObjectMarshallingStrategy.java
```diff
@@ -20,8 +20,13 @@
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 
-public interface ObjectMarshallingStrategy {
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 
+public interface ObjectMarshallingStrategy {
+     
+    ObjectMapper MAPPER = new ObjectMapper();
```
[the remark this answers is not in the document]
swati-kale: moved to mongodb-addon.

addons/persistence/mongodb-persistence-addon/src/test/java/org/kie/kogito/mongodb/PersistableProcessInstanceTest.java
```diff
@@ -0,0 +1,109 @@
...
+        MongoClient mongoClient = MongoClients.create();
+        BpmnProcess process = BpmnProcess.from(new ClassPathResource("BPMN2-UserTask.bpmn2"))
+                                         .get(0);
+        process.setProcessInstancesFactory(new PersistProcessInstancesFactory(mongoClient));
+        process.configure();
+        ProcessInstance<BpmnVariables> processInstance = process
+                                                                .createInstance(BpmnVariables.create(Collections.singletonMap("test", "test")));
+
+        processInstance.start();
+        assertThat(processInstance.status()).isEqualTo(STATE_ACTIVE);
```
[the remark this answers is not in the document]
swati-kale: done

addons/persistence/mongodb-persistence-addon/src/test/java/org/kie/kogito/mongodb/PersistableProcessInstanceTest.java
```diff
@@ -0,0 +1,109 @@
...
+import org.testcontainers.containers.FixedHostPortGenericContainer;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
+import org.testcontainers.containers.wait.strategy.Wait;
+import org.testcontainers.junit.jupiter.Container;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.kie.api.runtime.process.ProcessInstance.STATE_ACTIVE;
+import static org.kie.api.runtime.process.ProcessInstance.STATE_COMPLETED;
+
```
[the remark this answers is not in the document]
swati-kale: done

addons/persistence/mongodb-persistence-addon/src/test/java/org/kie/kogito/mongodb/PersistableProcessInstanceTest.java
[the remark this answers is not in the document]
swati-kale: added tests for mentioned  datatypes

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/DocumentMarshallingStrategy.java
```diff
@@ -0,0 +1,84 @@
...
+    public Object unmarshal(String dataType,
+                            Context context,
+                            ObjectInputStream is,
+                            byte[] object,
+                            ClassLoader classloader) throws IOException, ClassNotFoundException {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public String marshalToJson(Object object) {
```
mswiderski, line 67: Why do you need these extra methods and not reuse marshall/unmarshall? Introducing this json would require (or kind of confuse) users to always implement json approach

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/PersistableProcessInstances.java
```diff
@@ -0,0 +1,131 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(PersistableProcessInstances.class);
+    private org.kie.kogito.process.Process<?> process;
+    private final MongoCollection<ProcessInstanceDocument> collection;
+    private ProcessInstanceMarshaller marshaller;
+    private static final String PROCESS_INSTANCE_ID = "processInstance.id";
+
+    public PersistableProcessInstances(MongoClient mongoClient, org.kie.kogito.process.Process<?> process, String dbName) {
+        this.process = process;
+        collection = CommonUtils.getCollection(mongoClient, process.id(), dbName);
+        marshaller = new ProcessInstanceMarshaller(new DocumentMarshallingStrategy());
```
mswiderski, line 51: I find it quite weird to mix two ways of serializing process instance 

- protobuf for process instance
- json/string for variables

Why nit to use document for everything?

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/utils/ProcessInstanceMapper.java
```diff
@@ -0,0 +1,47 @@
...
+
+public class ProcessInstanceMapper {
+
+    private ProcessInstanceMapper() {}
+
+    public static class ProcessInstanceDocumentMapper implements Function<ProcessInstanceData, ProcessInstanceDocument> {
+
+        @Override
+        public ProcessInstanceDocument apply(ProcessInstanceData data) {
+            return new ProcessInstanceDocument(Document.parse(data.getlegacyProcessInstance()), data.getStrategies().entrySet().stream().map(e -> new Strategy(e.getValue(), e.getKey())).collect(Collectors.toList()));
```
mswiderski, line 34: Small typo i think, `getlegacy...` should be `getLegacy`

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/codec/ProcessInstanceDocumentCodec.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Override
+    public Class<ProcessInstanceDocument> getEncoderClass() {
+        return ProcessInstanceDocument.class;
+    }
+
+    @Override
+    public ProcessInstanceDocument generateIdIfAbsentFromDocument(ProcessInstanceDocument document) {
+        if (!documentHasId(document)) {
+            document.setId(UUID.randomUUID().toString());
```
mswiderski, line 67: This should fail instead as storing process instance without id should never happen

jbpm/jbpm-flow/src/main/java/org/jbpm/marshalling/impl/ProtobufProcessMarshaller.java
```diff
@@ -186,11 +187,17 @@ public static Variable marshallVariable(MarshallerWriteContext context,
         if(value != null){
             ObjectMarshallingStrategy strategy = context.objectMarshallingStrategyStore.getStrategyObject( value );
             Integer index = context.getStrategyIndex( strategy );
-            builder.setStrategyIndex( index )
-                   .setDataType(strategy.getType(value.getClass()))
-                   .setValue( ByteString.copyFrom( strategy.marshal( context.strategyContext.get( strategy ),
-                                                                     context,
-                                                                     value ) ) );
+            if (strategy.getName().contains(DOCUMENT_MARSHALLING_STRATEGY)) {
```
mswiderski, line 190: If this would just use normal methods this if is not needed

mswiderski:
> 1)Whether do you want me to implement the steps involved in attribute by attribute marshalling/unmarshalling of processInstance, nodeInstances/ nodeContents etc. similar to the current protobuf implementation?

ideally it would be each attribute defined as document property so you have fully visibility into it not byte arrays anymore.

> 2)Will it return / take plain mongo Documents or new POJOs similar to JBPMMessages.java?

I'd say it depends on the approach taken, maybe you could use protobuf to do the JSON serialisation instead of bytes... I think they had such features in protobuf...

> 3)Since it will depend on Mongo Documents, I assume the implementation will be at mongo-addon module instead jbpm-flow?

I'd say, yes, everything that is mongoldb specific should be in the mongodb addon

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/codec/ProcessInstanceDocumentCodec.java
swati-kale: This is a mongo document ID and not a process instance ID.

cristianonicolai:
Regarding point 2,  from what I recall, protobuf does have support for JSON attributes but it only works in version 3. That means Infinispan wont support it. Please double check this first.

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/codec/ProcessInstanceDocumentCodec.java
mswiderski: why do we need both?

mswiderski:
@cristianonicolai since this is for internal structure of the data to be written and actually only used as intermediate format from process instance to Document then infinispan should not be affected at all, or?

cristianonicolai:
> @cristianonicolai since this is for internal structure of the data to be written and actually only used as intermediate format from process instance to Document then infinispan should not be affected at all, or?

yep that should be fine. I just wanted to bring to attention in case that could also impact the infinispan persistence somehow.

swati-kale:
> > @cristianonicolai since this is for internal structure of the data to be written and actually only used as intermediate format from process instance to Document then infinispan should not be affected at all, or?
> 
> yep that should be fine. I just wanted to bring to attention in case that could also impact the infinispan persistence somehow.
I am doing this as shown below and as discussed with @mswidersk moving this at addon side,
https://github.com/kiegroup/kogito-runtimes/pull/477/files#diff-80a5deb516beed3d5f1f6bfa47b710f1R149.

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/DocumentMarshallingStrategy.java
swati-kale: will take care.

jbpm/jbpm-flow/src/main/java/org/jbpm/marshalling/impl/ProtobufProcessMarshaller.java
swati-kale: removed these changes

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/codec/ProcessInstanceDocumentCodec.java
swati-kale: just saving process instance id as document id.

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/PersistableProcessInstances.java
swati-kale: variables are saved as document in mongodb instead json string.

swati-kale:
> @swati-kale I few more points:
> 
> * _id should be process id attribute
done
> * not sure we need a _processInstance_ object, couldnt we move it all a level up?
done
> * Would be nice if you could increase the test coverage in different areas.
added more tests

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/utils/ProcessInstanceDocumentMapper.java
```diff
@@ -0,0 +1,83 @@
...
+import static org.kie.kogito.mongodb.utils.DocumentUtils.STRATEGIES;
+import static org.kie.kogito.mongodb.utils.DocumentUtils.VALUE;
+import static org.kie.kogito.mongodb.utils.DocumentUtils.VARIABLE;
+import static org.kie.kogito.mongodb.utils.DocumentUtils.VERSION;
+import static org.kie.kogito.mongodb.utils.DocumentUtils.VERSION_MAJOR;
+import static org.kie.kogito.mongodb.utils.DocumentUtils.VERSION_MINOR;
+import static org.kie.kogito.mongodb.utils.DocumentUtils.VERSION_REVISION;
+import static org.kie.kogito.mongodb.utils.DocumentUtils.getObjectMapper;
+
+public class ProcessInstanceDocumentMapper implements Function<ProcessMarshallerWriteContext, Document> {
```
cristianonicolai, line 41: @swati-kale I wonder if would simplify things if you take this approach: https://quarkus.io/guides/mongodb#simplifying-mongodb-client-usage-using-bson-codec
That way the marshall/unmarshall wouldnt need to be done manually. wdyt?

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/utils/DocumentUtils.java
```diff
@@ -0,0 +1,72 @@
...
+        } catch (JsonProcessingException e) {
+            throw new DocumentMarshallingException(e);
+        }
+        return json.getBytes();
+    }
+
+    public static Object fromByteArray(String dataType, byte[] object) {
+        try {
+            Class<?> loadClass = Thread.currentThread().getContextClassLoader().loadClass(dataType);
+            return MAPPER.readValue(new String(object), loadClass);
```
mswiderski, line 67: Will this work with list of objects?

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/persistence/PersistenceGenerator.java
```diff
@@ -324,13 +312,61 @@ protected void fileSystemBasedPersistence(List<GeneratedFile> generatedFiles) {
...
-     
-        generatedFiles.add(new GeneratedFile(GeneratedFile.Type.CLASS,
-                                             clazzName.replace('.', '/') + ".java",
-                                             compilationUnit.toString().getBytes(StandardCharsets.UTF_8))); 
-        
+        generatePersistenceProviderClazz(generatedFiles, persistenceProviderClazz, compilationUnit);
+    }
+    
+    private void mongodbBasedPersistence(List<GeneratedFile> generatedFiles) {
+        String dbName = context.getApplicationProperty("kogito.persistence.dbName").orElse("kogito");
```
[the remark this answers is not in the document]
mswiderski: It coul be prblematic if mongo is used for other logic as well. So I’d stick to kogito scoped property, just to add mongodb in the property name to be consistent with other properties

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/utils/DocumentUtils.java
swati-kale: yes I did test and it works.

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/persistence/PersistenceGenerator.java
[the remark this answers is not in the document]
swati-kale: I will change it to "kogito.persistence.mongodb.dbName" and handle runtime configuration.

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/utils/DocumentUtils.java
swati-kale: > Will this work with list of objects?

yes I did test and it works.

mswiderski: Is there a test case?

I mean to verify that having following variable will marshall and unmarshall properly

`List<Person>`

As with this code I think you will get list of LinkedHashMaps and not list of Persons

swati-kale: Yes, you are right, its a problem and looking into it.

mswiderski: since the JSON is actually internal maybe you could use default typing of the ObjectMapper?

https://www.baeldung.com/jackson-inheritance

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/persistence/PersistenceGenerator.java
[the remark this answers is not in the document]
cristianonicolai: @mswiderski @swati-kale to fully avoid clashing with another config, we would need to use named clients like in [here](https://quarkus.io/guides/mongodb#multiple-mongodb-clients). Just the database name would still share most of the configs which might not be desired. I think we should have a default which simply takes the default mongo client, or allowing users to specify the named mongo client they want to use ( `kogito` can be the default ).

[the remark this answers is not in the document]
mswiderski: this is a good point, how about spring? will it have similar approach so there won't be a need for any new kogito specific propeerty?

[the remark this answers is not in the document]
cristianonicolai: yeah, I would expect that we can take similar approach there. Maybe need to check if [this](https://dzone.com/articles/multiple-mongodb-connectors-with-spring-boot) still works. But the idea seems similar, different named MongoTemplates.

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/utils/DocumentUtils.java
swati-kale: Default typing changes the structure of json which cannot be readily (like normal json) converted into Document. or from document back to that json structure which default typing uses for object creation. So its overall tricky. Either you have to do that conversion here or variable marshal/unmarshal to/from  Bsonvalues using bsonwriter/reader.
1)Can't we use the protostream marshalling for variables and  persist the bytearray + document for that variable so that it can be queried and readable? Also there are no direct applications which refers the runtime persistence storage for direct queries. 
the result will look like this,WDYT?
image

mswiderski: Can you elaborate more on what is the problem with default typing? I think you can configure it to add type info as additional field of the object @class field name if I am not mistaken.

swati-kale: Actually I was using as below
MAPPER.activateDefaultTyping(MAPPER.getPolymorphicTypeValidator(), DefaultTyping.OBJECT_AND_NON_CONCRETE, JsonTypeInfo.As.WRAPPER_OBJECT); I
nstead MAPPER.setDefaultTyping(). My bad, it works with later.

swati-kale: Thanks @mswiderski

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/utils/ProcessInstanceDocumentMapper.java
swati-kale: I am using the existing protobuf marshalling for legacy process instance and converting it to json to Mongo document. I do not want to rewrite marshalling logic as there might be an overhead of maintenance. I wrote a codec for wrapper object which wraps legacy process instance, strategy and versions. The manual conversion over here are mainly to convert the variables from bytes to document and vice versa while saving/fetch it to/from DB.

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/persistence/PersistenceGenerator.java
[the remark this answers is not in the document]
swati-kale: @cristianonicolai, As suggested by you, I removed the extra kogito property and injecting the quarkus/springboot property for dbname at code generation,

drools/drools-core/src/main/java/org/drools/core/marshalling/impl/MarshallerReaderContext.java
```diff
@@ -162,6 +162,46 @@ public MarshallerReaderContext(InputStream stream,
             this.kBase.getProcesses().forEach( p -> this.processes.put(p.getId(), p));
         }
     }
+    
+    public MarshallerReaderContext(Map<String, Process> processes,
```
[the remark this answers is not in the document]
swati-kale: @cristianonicolai , I discussed this with you already, so please suggest.

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/marshalling/DocumentProcessInstanceMarshaller.java
```diff
@@ -0,0 +1,102 @@
...
+        return new RuleFlowProcessInstance();
+    }
+
+    public ProcessInstanceDocument marshalProcessInstance(ProcessInstance<?> processInstance) {
+        try {
+            WorkflowProcessInstance pi = ((AbstractProcessInstance<?>) processInstance).internalGetProcessInstance();
+            ProcessMarshallerWriteContext context = new ProcessMarshallerWriteContext(null, null, null, null, env);
+            context.parameterObject = JsonFormat.printer().print(super.writeProcessInstance(context, pi));
+            ProcessInstanceDocument document = new ProcessInstanceDocumentMapper().apply(context);
+            ((WorkflowProcessInstanceImpl) pi).disconnect();
```
cristianonicolai, line 76:
```suggestion
           pi.disconnect();
```

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/marshalling/DocumentProcessInstanceMarshaller.java
```diff
@@ -0,0 +1,102 @@
...
+        } catch (Exception e) {
+            throw new DocumentMarshallingException(processInstance.id(), e, DOCUMENT_MARSHALLING_ERROR_MSG);
+        }
+    }
+
+    public WorkflowProcessInstance unmarshallWorkflowProcessInstance(ProcessInstanceDocument doc, Process<?> process) {
+        try {
+            MarshallerReaderContext context = new MarshallerReaderContext(Collections.singletonMap(process.id(), ((AbstractProcess<?>) process).process()), null, null, null, env, null);
+            context.parameterObject = doc;
+            JBPMMessages.ProcessInstance instance = new ProcessInstanceMessageMapper().apply(context);
```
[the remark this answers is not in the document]
swati-kale: done.

drools/drools-core/src/main/java/org/drools/core/marshalling/impl/MarshallerReaderContext.java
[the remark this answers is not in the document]
cristianonicolai: how about:
```suggestion
        public MarshallerReaderContext(Map<String, Process> processes, Environment env) throws IOException {
        this(null, processes, null, null, null, env);
    }
```

cristianonicolai:
@tiagodolphine @evacchi @mswiderski please take another look as @swati-kale has done quite some good progress on this PR.

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/utils/ProcessInstanceMessageMapper.java
```diff
@@ -0,0 +1,75 @@
...
+                }
+            }
+
+        } catch (Exception e) {
+            throw new DocumentUnmarshallingException(e);
+        }
+        return builder.build();
+    }
+
+    private void applyVariables(JsonNode parent, String variable) {
```
tiagodolphine, line 59: It seems this method is the same as the one on ProcessInstanceDocumentMapper, does it worth to centralize the implementation to avoid duplication?

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/persistence/KogitoProcessInstancesFactory.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public abstract class KogitoProcessInstancesFactory implements ProcessInstancesFactory {
+
+    protected MongoClient mongoClient;
+
+    public KogitoProcessInstancesFactory(MongoClient mongoClient) {
+        this.mongoClient = mongoClient;
+    }
+
+    public String dbName() {
```
tiagodolphine, line 36: make sense to make the method abstract to force the implementations to override it?

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/PersistableProcessInstances.java
```diff
@@ -0,0 +1,133 @@
...
+
+    public PersistableProcessInstances(MongoClient mongoClient, org.kie.kogito.process.Process<?> process, String dbName) {
+        this.process = process;
+        collection = getCollection(mongoClient, process.id(), dbName);
+        marshaller = new DocumentProcessInstanceMarshaller(new DocumentMarshallingStrategy());
+    }
+
+    @Override
+    public Optional<ProcessInstance<T>> findById(String id) {
+        ProcessInstanceDocument piDoc = collection.find(Filters.eq(DOCUMENT_ID, resolveId(id))).first();
```
tiagodolphine, line 56: the `collection.find(Filters.eq(DOCUMENT_ID, resolvedId)).first()` is used in many places, why not extracting into a method to be reused?

drools/drools-core/src/main/java/org/drools/core/marshalling/impl/MarshallerReaderContext.java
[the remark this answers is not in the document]
swati-kale: > how about:

Passing null for Stream gives NPE, same in MarshallerWriterContext.

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/utils/ProcessInstanceMessageMapper.java
swati-kale: As discussed, they are different.

drools/drools-core/src/main/java/org/drools/core/marshalling/impl/MarshallerWriteContext.java
```diff
@@ -121,6 +121,36 @@ public MarshallerWriteContext(OutputStream stream,
         this.env = env;
 
     }
+    
+    public MarshallerWriteContext(InternalKnowledgeBase kBase,
```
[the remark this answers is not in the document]
swati-kale: as above

swati-kale:
@cristianonicolai Hi, As discussed I pushed the code changes needed for processinstance readmode and kogito-example utils.

cristianonicolai:
jenkins please retest this

MarianMacik:
I will have a look at it on Monday at latest.

cristianonicolai:
jenkins please retest this

pom.xml
```diff
@@ -791,6 +796,13 @@
...
         <artifactId>json-schema-validator</artifactId>
         <version>${version.io.restassured}</version>
+    </dependency>
+      
+      <!-- mongoDB related -->
+      <dependency>
+        <groupId>org.mongodb</groupId>
+        <artifactId>mongodb-driver-sync</artifactId>
+        <version>${version.org.mongo}</version>
       </dependency>
```
MarianMacik, line 806: Formatting is a bit off.

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/marshalling/DocumentProcessInstanceMarshaller.java
```diff
@@ -0,0 +1,128 @@
...
+import org.kie.kogito.mongodb.utils.ProcessInstanceMessageMapper;
+import org.kie.kogito.process.Process;
+import org.kie.kogito.process.ProcessInstance;
+import org.kie.kogito.process.impl.AbstractProcess;
+import org.kie.kogito.process.impl.AbstractProcessInstance;
+
+import static org.kie.kogito.mongodb.utils.DocumentConstants.DOCUMENT_MARSHALLING_ERROR_MSG;
+import static org.kie.kogito.mongodb.utils.DocumentConstants.DOCUMENT_UNMARSHALLING_ERROR_MSG;
+
+public class DocumentProcessInstanceMarshaller extends AbstractProtobufProcessInstanceMarshaller {
```
MarianMacik, line 50: Can we make this composition instead of inheritance? This would be more in sync with ProcessInstanceMarshaller, which obtains instance of AbstractProtobufProcessInstanceMarshaller using:
https://github.com/kiegroup/kogito-runtimes/blob/923d55ee5bf94ad777cee3f751419931b593f203/jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/marshalling/ProcessInstanceMarshaller.java#L82

Also right now we are mixing the `org.kie.kogito.mongodb.marshalling` package with the `org.jbpm.marshalling.impl` package which are two different things.

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/utils/ProcessInstanceDocumentMapper.java
```diff
@@ -0,0 +1,76 @@
...
+    @Override
+    public ProcessInstanceDocument apply(KogitoProcessMarshallerWriteContext context) {
+        String instance = (String) context.parameterObject;
+        ProcessInstanceDocument doc = new ProcessInstanceDocument();
+        try {
+            JsonNode instanceNode = getObjectMapper().readTree(instance);
+            doc.setId(instanceNode.get(PROCESS_INSTANCE_ID).asText());
+            applyVariables(instanceNode, VARIABLE);
+            doc.setProcessInstance(Optional.ofNullable(instanceNode).map(json -> Document.parse(json.toString())).orElse(null));
+            doc.setStrategies(context.usedStrategies.entrySet().stream().collect(Collectors.toMap(e -> e.getKey().getName(), Map.Entry::getValue)));
```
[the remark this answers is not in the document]
swati-kale: stratagy will decide variable marshalling/unmarshalling. So it is saved while marshalling so that can be used while unmarshalling in below
https://github.com/kiegroup/kogito-runtimes/blob/923d55ee5bf94ad777cee3f751419931b593f203/jbpm/jbpm-flow/src/main/java/org/jbpm/marshalling/impl/ProtobufProcessMarshaller.java#L252

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/utils/ProcessInstanceDocumentMapper.java
```diff
@@ -0,0 +1,76 @@
...
+            doc.getVersions().put(VERSION_MAJOR, Drools.getMajorVersion());
+            doc.getVersions().put(VERSION_MINOR, Drools.getMinorVersion());
+            doc.getVersions().put(VERSION_REVISION, Drools.getRevisionVersion());
```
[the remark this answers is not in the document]
swati-kale: It was in sync with  original process instance marshalling whereversions and strategies are persisted as a part of header  as below,
https://github.com/kiegroup/kogito-runtimes/blob/master/jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/marshalling/ProcessInstanceMarshaller.java#L87
https://github.com/kiegroup/drools/blob/master/drools-core/src/main/java/org/drools/core/marshalling/impl/PersisterHelper.java#L195-L201

If versions are not required, @cristianonicolai let me know if you have any other opinion.

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/marshalling/DocumentProcessInstanceMarshaller.java
```diff
@@ -0,0 +1,128 @@
...
+        return new RuleFlowProcessInstance();
+    }
+
+    public ProcessInstanceDocument marshalProcessInstance(ProcessInstance<?> processInstance) {
+        try {
+            WorkflowProcessInstance pi = ((AbstractProcessInstance<?>) processInstance).internalGetProcessInstance();
+            try (ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
+                KogitoProcessMarshallerWriteContext context = new KogitoProcessMarshallerWriteContext(baos, null, null, null, null, env);
+                context.parameterObject = JsonFormat.printer().print(super.writeProcessInstance(context, pi));
+                ProcessInstanceDocument document = new ProcessInstanceDocumentMapper().apply(context);
```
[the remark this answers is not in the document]
swati-kale: ProcessInstanceDocumentMapper now maps fom JBPMMessages.ProcessInstance to ProcessInstanceDocument
ProcessInstanceMessageMapper maps from ProcessInstanceDocument to JBPMMessages.ProcessInstance .

swati-kale:
jenkins please retest this

cristianonicolai:
jenkins please retest this

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/utils/ProcessInstanceDocumentMapper.java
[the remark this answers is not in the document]
MarianMacik: Yes, but they are then not read in the `ProcessInstanceMessageMapper`, so I guess they are not needed at all?

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/utils/ProcessInstanceDocumentMapper.java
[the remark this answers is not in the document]
MarianMacik: I also noticed that with these strategies there are also strategy contexts, do we want to serialize them as well? But maybe this can be added only when needed. Wdyt @cristianonicolai?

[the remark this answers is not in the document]
cristianonicolai: @MarianMacik +1

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/utils/ProcessInstanceDocumentMapper.java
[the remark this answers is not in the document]
cristianonicolai: good point, with the unfork now this will be different, so +1 to remove.

cristianonicolai:
jenkins please retest this

cristianonicolai:
jenkins please retest this

addons/persistence/mongodb-persistence-addon/src/test/java/org/kie/kogito/mongodb/ProcessInstanceDocumentCodecProviderTest.java
```diff
@@ -0,0 +1,77 @@
...
+    @Test
+    void codecTest() throws URISyntaxException, IOException {
+        ProcessInstanceDocumentCodec codec = new ProcessInstanceDocumentCodec();
+        assertThat(codec.getEncoderClass()).isEqualTo(ProcessInstanceDocument.class);
+        ProcessInstanceDocument doc = new ProcessInstanceDocument();
+        doc.setProcessInstance((org.bson.Document) TestHelper.getProcessInstanceDocument().get("processInstance"));
+        assertNull(doc.getId(), "ProcessInstanceDocument is null");
+        codec.generateIdIfAbsentFromDocument(doc);
+        assertThat(codec.documentHasId(doc)).isTrue();
+        assertNotNull(codec.getDocumentId(doc), "ProcessInstanceDocument has document id and not null");
```
MarianMacik, line 64: This is a bug reported by SonarCloud. `codec.getDocumentId(doc)` always returns primitive `true` or `false`, so this assertion always passes. The best would be to remove this assertion.

cristianonicolai:
jenkins please retest this

tiagodolphine:
jenkins please retest this

cristianonicolai:
ok to test

addons/persistence/mongodb-persistence-addon/src/test/java/org/kie/kogito/persistence/KogitoProcessInstancesFactoryTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+            @Override
+            public String dbName() {
+                return DB_NAME;
+            }
+        };
+        assertNotNull(factory);
+        assertThat(factory.dbName()).isEqualTo(DB_NAME);
+        assertNotNull(factory.mongoClient);
+        Process<?> process = new TestProcessImpl();
```
cristianonicolai, line 45: perhaps use mock(Process.class) ?

addons/persistence/mongodb-persistence-addon/src/test/resources/process_instance_document.json
```diff
@@ -0,0 +1,136 @@
...
+                     "taskReferenceName":"Review the deal"
+                  }
+               }
+            },
+            "level":1,
+            "slaCompliance":0,
+            "triggerDate":"1597104562089"
+         }
+      ],
+      "variable":[
```
cristianonicolai, line 88: perhaps `variables` ?

swati-kale: It is "variable" only it gets the label from here, 
https://github.com/rhtevan/kogito-runtimes/blob/master/jbpm/jbpm-flow/src/main/resources/org/jbpm/marshalling/jbpmmessages.proto#L175

cristianonicolai: but that is a single variable, here is are storing all the variables from the process right?

swati-kale: yes its is representing list of Variable objects from JBPMMEssages.java --> java.util.List<org.jbpm.marshalling.impl.JBPMMessages.Variable> getVariableList();

cristianonicolai: hm ok then, thanks for the explanation

cristianonicolai:
nice work @swati-kale !

PR #491, opened 2020-05-06
[KOGITO-2002] fix hot reload when a new rule unit is added

mariofusco:


evacchi:
what was wrong exactly?

mariofusco:
> what was wrong exactly?

yep, sorry I did some refactory (the IncrementalRuleCodegen.generate() was 200+ lines long and became unreadable) and it is not easy to see what changed. Basically 2 things:

1. The AssignableChecker, the class using the "right" classloader to check if a type is a DataSource (I introduced it to workaround the problems caused by Quarkus CL, remember?) is now taken directly from the RuleUnitDescription when possible https://github.com/kiegroup/kogito-runtimes/pull/491/files#diff-aaf0409efde7bd18867eb25de7988419R357 I'm afraid this is still a bit brittle and I don't think this is totally solved now, but anyway if I don't do this once again the incremental codegen is not able to recognize that a variable is a DataSource in Duncan's test case.

2. Before the ProjectRuntime (that contains the definition of all KieBases included the ones inferred from the units) wasn't regenerated in case of an hot reload https://github.com/kiegroup/kogito-runtimes/pull/491/files#diff-aaf0409efde7bd18867eb25de7988419L350 I had to remove that condition (it's unfortunately impossible to understand if a new ruleunit has been added or not) and now the ProjectRuntime is always regenerated.

kogito-quarkus-extension/integration-test-hot-reload/src/test/java/io/quarkus/it/kogito/drools/newunithotreloadtest/NewUnitHotReloadTest.java
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright 2005 JBoss Inc
```
cristianonicolai, line 2: back to the future :P

mariofusco: Thanks for reporting :)

PR #492, opened 2020-05-06
KOGITO-1895 Setup CI Build against Drools SNAPSHOT

evacchi:
https://issues.redhat.com/browse/KOGITO-1895

Add a Jenkinsfile to set-up a periodic build for Drools snapshot

### DIFF
--- a/Jenkinsfile.drools
+++ b/Jenkinsfile.drools
@@ -8,11 +8,15 @@ pipeline {
         maven 'kie-maven-3.6.2'
         jdk 'kie-jdk11'
     }
+    triggers {
+        cron ('H 10 * * *')
+    }
     options {
         buildDiscarder logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '', numToKeepStr: '5')
         timeout(time: 90, unit: 'MINUTES')
     }
     environment {
+        KOGITO_CI_EMAIL_TO = credentials('KOGITO_CI_EMAIL_TO')
         // query mvn to get the latest version
         DROOLS_VERSION = """${sh (
             script: '$(mvn versions:display-property-updates -DincludeProperties=version.org.kie7 -DallowSnapshots -N  | grep version.org.kie7| cut -d" " -f8))'
@@ -44,7 +48,7 @@ pipeline {
             steps {
                 dir("kogito-apps") {
                     script {
-                        githubscm.checkoutIfExists('kogito-apps', "$CHANGE_AUTHOR", "$CHANGE_BRANCH", 'kiegroup', "$CHANGE_TARGET")
+                        checkout([$class: 'GitSCM', branches: [[name: env.BRANCH_NAME]], browser: [$class: 'GithubWeb', repoUrl: 'git@github.com:kiegroup/kogito-apps.git'], doGenerateSubmoduleConfigurations: false, extensions: [[$class: 'RelativeTargetDirectory', relativeTargetDir: 'kogito-apps']], submoduleCfg: [], userRemoteConfigs: [[credentialsId: 'kie-ci-user-key', url: 'git@github.com:kiegroup/kogito-apps.git']]])
                         maven.runMavenWithSubmarineSettings("clean install ${MAVEN_EXTRA_ARGS}", false)
                     }
                 }
@@ -54,14 +58,14 @@ pipeline {
             steps {
                 dir("kogito-examples") {
                     script {
-                        githubscm.checkoutIfExists('kogito-examples', "$CHANGE_AUTHOR", "$CHANGE_BRANCH", 'kiegroup', "$CHANGE_TARGET")
+                        checkout([$class: 'GitSCM', branches: [[name: env.BRANCH_NAME]], browser: [$class: 'GithubWeb', repoUrl: 'git@github.com:kiegroup/kogito-examples.git'], doGenerateSubmoduleConfigurations: false, extensions: [[$class: 'RelativeTargetDirectory', relativeTargetDir: 'kogito-examples']], submoduleCfg: [], userRemoteConfigs: [[credentialsId: 'kie-ci-user-key', url: 'git@github.com:kiegroup/kogito-examples.git']]])
                         maven.runMavenWithSubmarineSettings("clean install ${MAVEN_EXTRA_ARGS}", false)
                     }
                 }
                 // Use a separate dir for persistence to not overwrite the test results
                 dir("kogito-examples-persistence") {
                     script {
-                        githubscm.checkoutIfExists('kogito-examples', "$CHANGE_AUTHOR", "$CHANGE_BRANCH", 'kiegroup', "$CHANGE_TARGET")
+                        checkout([$class: 'GitSCM', branches: [[name: env.BRANCH_NAME]], browser: [$class: 'GithubWeb', repoUrl: 'git@github.com:kiegroup/kogito-examples.git'], doGenerateSubmoduleConfigurations: false, extensions: [[$class: 'RelativeTargetDirectory', relativeTargetDir: 'kogito-examples']], submoduleCfg: [], userRemoteConfigs: [[credentialsId: 'kie-ci-user-key', url: 'git@github.com:kiegroup/kogito-examples.git']]])
                         // Don't run with tests so far, see: https://github.com/quarkusio/quarkus/issues/6885
                         maven.runMavenWithSubmarineSettings("clean install ${MAVEN_EXTRA_ARGS} -Ppersistence", true)
                     }
--- a/Jenkinsfile.drools
+++ b/Jenkinsfile.drools
@@ -56,7 +56,7 @@ pipeline {
                      to: env.KOGITO_CI_EMAIL_TO
         }
         unstable {
-            emailext body: 'Kogito daily snapshot deploy #${BUILD_NUMBER} was: ' + "${currentBuild.currentResult}" +  '\n' +
+            emailext body: 'Kogito daily Drools #${BUILD_NUMBER} was: ' + "${currentBuild.currentResult}" +  '\n' +
                      'Please look here: ${BUILD_URL} ',
                      subject: 'Kogito daily Drools #${BRANCH_NAME}',
                      to: env.KOGITO_CI_EMAIL_TO
--- a/Jenkinsfile.drools
+++ b/Jenkinsfile.drools
@@ -52,7 +52,7 @@ pipeline {
         failure {
             emailext body: 'Kogito daily Drools #${BUILD_NUMBER} was: ' + "${currentBuild.currentResult}" +  '\n' +
                     'Please look here: ${BUILD_URL} ',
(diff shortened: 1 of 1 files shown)


Jenkinsfile.drools
```diff
@@ -0,0 +1,50 @@
...
+            steps {
+                sh 'printenv'
+            }
+        }
+        stage('Build kogito-runtimes') {
+            steps {
+                checkout([$class: 'GitSCM', branches: [[name: 'master']], browser: [$class: 'GithubWeb', repoUrl: 'git@github.com:kiegroup/kogito-runtimes.git'], doGenerateSubmoduleConfigurations: false, extensions: [[$class: 'RelativeTargetDirectory', relativeTargetDir: 'kogito-runtimes']], submoduleCfg: [], userRemoteConfigs: [[credentialsId: 'kie-ci-user-key', url: 'git@github.com:kiegroup/kogito-runtimes.git']]])
+                dir("kogito-runtimes") {
+                    script {
+                        maven.runMavenWithSubmarineSettings('clean install -Dversion.org.kie7=7.37.0-SNAPSHOT', false)
```
radtriste, line 26: just noticed ... we could put that as environment var (between options and stages part)
```
environment {
    MAVEN_EXTRA_ARGS="-Dversion.org.kie7=7.37.0-SNAPSHOT"
}
```
and then modify the command:
```suggestion
                        maven.runMavenWithSubmarineSettings("clean install ${MAVEN_EXTRA_ARGS}", false)
```
*Note the double quotes instead of single quote*
that would avoid to repeat each time the correct version

cristianonicolai: That should do, otherwise, you could look into https://www.mojohaus.org/versions-maven-plugin/examples/advancing-dependency-versions.html
_use-latest-snapshots_ seems that could do the trick to keep the job always up-to-date.

evacchi: @cristianonicolai 
mind

Jenkinsfile.drools
```diff
@@ -0,0 +1,70 @@
...
+                         maven.runMavenWithSubmarineSettings("clean install ${MAVEN_EXTRA_ARGS}", false)
+                    }
+                }
+            }
+        }
+
+    }
+    post {
+        failure {
+            emailext body: 'Kogito daily snapshot deploy #${BUILD_NUMBER} was: ' + "${currentBuild.currentResult}" +  '\n' +
```
cristianonicolai, line 53:
```suggestion
            emailext body: 'Kogito daily Drools #${BUILD_NUMBER} was: ' + "${currentBuild.currentResult}" +  '\n' +
```

Jenkinsfile.drools
```diff
@@ -0,0 +1,70 @@
...
+                    'Please look here: ${BUILD_URL} ',
+                     subject: 'Kogito daily snapshot deploy #${BRANCH_NAME}',
```
cristianonicolai, line 55:
```suggestion
                     subject: 'Kogito daily Drools #${BRANCH_NAME}',
```

Jenkinsfile.drools
```diff
@@ -0,0 +1,70 @@
...
+                     to: env.KOGITO_CI_EMAIL_TO
+        }
+        unstable {
+            emailext body: 'Kogito daily snapshot deploy #${BUILD_NUMBER} was: ' + "${currentBuild.currentResult}" +  '\n' +
```
cristianonicolai, line 59:
```suggestion
            emailext body: 'Kogito daily Drools #${BUILD_NUMBER} was: ' + "${currentBuild.currentResult}" +  '\n' +
```

Jenkinsfile.drools
```diff
@@ -0,0 +1,70 @@
...
+                     'Please look here: ${BUILD_URL} ',
+                     subject: 'Kogito daily snapshot deploy #${BRANCH_NAME}',
```
cristianonicolai, line 61:
```suggestion
                     subject: 'Kogito daily Drools #${BRANCH_NAME}',
```

Jenkinsfile.drools
```diff
@@ -0,0 +1,70 @@
...
+                    returnStdout: true).trim()
+                echo "Latest kie7 verson is: ${DROOLS_VERSION}"
+                environment {
+                    MAVEN_EXTRA_ARGS="-Dversion.org.kie7=${DROOLS_VERSION}"
+                }
+            }
+        }
+        stage('Build kogito-runtimes') {
+            steps {
+                checkout([$class: 'GitSCM', branches: [[name: 'master']], browser: [$class: 'GithubWeb', repoUrl: 'git@github.com:kiegroup/kogito-runtimes.git'], doGenerateSubmoduleConfigurations: false, extensions: [[$class: 'RelativeTargetDirectory', relativeTargetDir: 'kogito-runtimes']], submoduleCfg: [], userRemoteConfigs: [[credentialsId: 'kie-ci-user-key', url: 'git@github.com:kiegroup/kogito-runtimes.git']]])
```
cristianonicolai, line 31: because the Jenkins file is in kogito-runtimes, it will automatically checkout kogito-runtimes already, see https://github.com/kiegroup/kogito-runtimes/blob/master/Jenkinsfile.deploy#L27.

Jenkinsfile.drools
```diff
@@ -0,0 +1,70 @@
...
+                dir("kogito-runtimes") {
+                    script {
+                        maven.runMavenWithSubmarineSettings("clean install ${MAVEN_EXTRA_ARGS}", false)
+                    }
+                }
+            }
+        }
+        stage('Build kogito-examples') {
```
cristianonicolai, line 39: missing kogito-apps?

Jenkinsfile.drools
```diff
@@ -0,0 +1,70 @@
...
+    }
+    options {
+        buildDiscarder logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '', numToKeepStr: '5')
+        timeout(time: 90, unit: 'MINUTES')
+    }
+    stages {
+        stage('Initialize') {
+            steps {
+                sh 'printenv'
+                DROOLS_VERSION = sh (
```
cristianonicolai, line 19: this is nice, I would just suggest merging with _Build kogito_runtimes_ stage

radtriste: This can be done in environment in fact. See https://www.jenkins.io/doc/book/pipeline/jenkinsfile/#setting-environment-variables-dynamically

Jenkinsfile.drools
```diff
@@ -0,0 +1,70 @@
...
+                    // query mvn to get the latest version
+                    script: '$(mvn versions:display-property-updates -DincludeProperties=version.org.kie7 -DallowSnapshots -N  | grep version.org.kie7| cut -d" " -f8))'
+                    returnStdout: true).trim()
+                echo "Latest kie7 verson is: ${DROOLS_VERSION}"
+                environment {
```
radtriste, line 24: This should be put under `stages` parent, not here ...

Jenkinsfile.drools
radtriste: do we need to test kogito-apps against drools ?

cristianonicolai: probably not, but I would keep for the sake of consistency across builds.

evacchi: ok that was intended, but I will bring it back -- it's daily after all

Jenkinsfile.drools
evacchi: @cristianonicolai I think I'd rather keep it in the first stage so that it's easier to read. But we can move it...

Jenkinsfile.drools
evacchi: you mean _right_ under `stages {}` ?

evacchi: it should be look better now (it's under pipelines, like the main Jenkinsfile)

radtriste: yeah I meant under `pipeline` ...

Jenkinsfile.drools
```diff
@@ -0,0 +1,92 @@
...
+                script {
+                    maven.runMavenWithSubmarineSettings("clean install ${MAVEN_EXTRA_ARGS}", false)
+                }
+            }
+        }
+        stage('Build kogito-apps') {
+            steps {
+                dir("kogito-apps") {
+                    script {
+                        githubscm.checkoutIfExists('kogito-apps', "$CHANGE_AUTHOR", "$CHANGE_BRANCH", 'kiegroup', "$CHANGE_TARGET")
```
radtriste, line 47: This is no PR check or ?
we should not have the `CHANGE_*` vars here

Jenkinsfile.drools
```diff
@@ -0,0 +1,96 @@
...
+        MAVEN_EXTRA_ARGS="-Dversion.org.kie7=${DROOLS_VERSION}"
+    }
+    stages {
+        stage('Initialize') {
+            steps {
+                sh 'printenv'
+                echo "Latest kie7 verson is: ${DROOLS_VERSION}"
+            }
+        }
+        stage('build sh script') {
```
cristianonicolai, line 33: we dont need this step. the email notification on zulip should be enough.

Jenkinsfile.drools
```diff
@@ -0,0 +1,96 @@
...
+        // query mvn to get the latest version
+        DROOLS_VERSION = """${sh (
+            script: '$(mvn versions:display-property-updates -DincludeProperties=version.org.kie7 -DallowSnapshots -N  | grep version.org.kie7| cut -d" " -f8))'
+            returnStdout: true).trim()}"""
+        MAVEN_EXTRA_ARGS="-Dversion.org.kie7=${DROOLS_VERSION}"
+    }
+    stages {
+        stage('Initialize') {
+            steps {
+                sh 'printenv'
```
cristianonicolai, line 29: @evacchi I mentioning removing this step because there is no point on doing `sh 'printenv'`, I believe that was only for debugging.

Jenkinsfile.drools
```diff
@@ -0,0 +1,96 @@
...
+                script {
+                    maven.runMavenWithSubmarineSettings("clean install ${MAVEN_EXTRA_ARGS}", false)
+                }
+            }
+        }
+        stage('Build kogito-apps') {
+            steps {
+                dir("kogito-apps") {
+                    script {
+                        checkout([$class: 'GitSCM', branches: [[name: env.BRANCH_NAME]], browser: [$class: 'GithubWeb', repoUrl: 'git@github.com:kiegroup/kogito-apps.git'], doGenerateSubmoduleConfigurations: false, extensions: [[$class: 'RelativeTargetDirectory', relativeTargetDir: 'kogito-apps']], submoduleCfg: [], userRemoteConfigs: [[credentialsId: 'kie-ci-user-key', url: 'git@github.com:kiegroup/kogito-apps.git']]])
```
radtriste, line 51: `checkout` should be outside the `dir`, else the folder might not exists

Jenkinsfile.drools
evacchi: oh ok

PR #502, opened 2020-05-12
[KOGITO-1968] Allow milestones to trigger completion and tests

ruromero:
* Remove legacy Case related classes
* Add lambda to trigger event in EndNodeVisitor 
* MilestoneNodes should be completed when the event is received --> No conditions supported yet [KOGITO-2168](https://issues.redhat.com/browse/KOGITO-2168)
* Small clean-up
* Expose `milestones` method in API


api/kogito-api/src/main/java/org/kie/api/runtime/rule/RuleRuntime.java
```diff
@@ -45,6 +45,13 @@
      */
     Agenda getAgenda();
 
+    /**
+     * @return true if the Runtime supports the Agenda
+     */
+    default boolean hasAgenda() {
```
evacchi, line 51: why is this needed? if we make getAgenda() return null in the dummy implementation then you can just null check `getAgenda()`. Remember we want to get rid of the dummy implementation.

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
```diff
@@ -142,7 +144,7 @@ public void addNodeInstance(final NodeInstance nodeInstance) {
 
     @Override
     public int getLevelForNode(String uniqueID) {
-        if ("true".equalsIgnoreCase(System.getProperty("jbpm.loop.level.disabled"))) {
+        if (Boolean.parseBoolean(System.getProperty("jbpm.loop.level.disabled"))) {
```
evacchi, line 147: memo: I wonder if we should keep these overrides in Kogito

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/MilestoneNodeInstance.java
```diff
@@ -53,10 +53,13 @@ public void internalTrigger(final NodeInstance from, String type) {
...
-                + "-" + getMilestoneNode().getUniqueId();
-        boolean isActive = ((InternalAgenda) getProcessInstance().getKnowledgeRuntime().getAgenda())
-                .isRuleActiveInRuleFlowGroup("DROOLS_SYSTEM", rule, getProcessInstance().getId());
+        boolean isActive = true;
+        if(getProcessInstance().getKnowledgeRuntime().hasAgenda()) {
+            String rule = "RuleFlow-Milestone-" + getProcessInstance().getProcessId()
+                    + "-" + getMilestoneNode().getUniqueId();
+            isActive = ((InternalAgenda) getProcessInstance().getKnowledgeRuntime().getAgenda())
+                    .isRuleActiveInRuleFlowGroup("DROOLS_SYSTEM", rule, getProcessInstance().getId());
+        }
```
evacchi, line 62: I can see now where rules are used, and we should probably **write a short spec for this** and see how we can reproduce it in the best way. Rule flow groups should be considered deprecated for Drools. Maybe we can write a "System RuleUnit" in the libary if these rules are static, or we could synthesize a rule unit to be used internally.

api/kogito-api/src/main/java/org/kie/api/runtime/rule/RuleRuntime.java
tiagodolphine: good point, +1

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
```diff
@@ -193,22 +195,15 @@ public void removeNodeInstance(final NodeInstance nodeInstance) {
...
     @Override
     public NodeInstance getNodeInstance(String nodeInstanceId, boolean recursive) {
-        for (NodeInstance nodeInstance : getNodeInstances(recursive)) {
-            if (nodeInstance.getId().equals(nodeInstanceId)) {
-                return nodeInstance;
-            }
-        }
-        return null;
+        return getNodeInstances(recursive).stream()
+                .filter(nodeInstance -> nodeInstance.getId().equals(nodeInstanceId))
```
tiagodolphine, line 204: suggestion
`.filter(nodeInstance -> Objects.equals(nodeInstance.getId(), nodeInstanceId))`

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/MilestoneNodeInstance.java
```diff
@@ -53,10 +53,13 @@ public void internalTrigger(final NodeInstance from, String type) {
             throw new IllegalArgumentException(
                     "A MilestoneNode only accepts default incoming connections!");
         }
-        String rule = "RuleFlow-Milestone-" + getProcessInstance().getProcessId()
-                + "-" + getMilestoneNode().getUniqueId();
-        boolean isActive = ((InternalAgenda) getProcessInstance().getKnowledgeRuntime().getAgenda())
-                .isRuleActiveInRuleFlowGroup("DROOLS_SYSTEM", rule, getProcessInstance().getId());
+        boolean isActive = true;
```
tiagodolphine, line 56: trying to simplify this block... just sugestion:
```
Optional.ofNullable(getProcessInstance())
                .map(WorkflowProcessInstance::getKnowledgeRuntime)
                .map(InternalKnowledgeRuntime::getAgenda)                
                .filter(InternalAgenda.class::isInstance)
                .map(InternalAgenda.class::cast)
                .map(agenda -> {
                    String rule =
                            "RuleFlow-Milestone-" + getProcessInstance().getProcessId() + "-" + getMilestoneNode().getUniqueId();
                    return agenda.isRuleActiveInRuleFlowGroup("DROOLS_SYSTEM", rule, getProcessInstance().getId());
                })
                .orElse(true);
```

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcessInstance.java
```diff
@@ -163,14 +164,20 @@ public void start(String trigger, String referenceId) {
             this.status = legacyProcessInstance.getState();
         }
     }
+
+    public void complete() {
+        if(((WorkflowProcess)legacyProcessInstance.getProcess()).isDynamic() && this.status.equals(ProcessInstance.STATE_ACTIVE)) {
```
tiagodolphine, line 169: Are the castings safe? WorkflowProcess, WorkflowProcessInstance, Is it worth checking instanceof before casting?

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/MilestoneNodeInstance.java
ruromero: I have filed the following JIRA https://issues.redhat.com/browse/KOGITO-2168

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/MilestoneNodeInstance.java
ruromero: Although your suggestion is fancier I believe the existing code is easier to understand and maintain.

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcessInstance.java
ruromero: If the legacyProcessInstance.getProcess() is always returning what is generated by the `legacyProcess()` method coming from the codegen I'd say the only possibility is that it is a RuleFlowProcess which extends WorkflowProcessImpl

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/MilestoneNodeInstance.java
tiagodolphine: Sure, no problem, go with it, as I said just a suggestion a matter of taste, I really like and think a more functional approach is cleaner :)

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcessInstance.java
tiagodolphine: :+1:

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/MilestoneNodeInstance.java
evacchi: either way, this is just a refactoring of the existing code (@ruromero only added the null check around it and reformatted a bit); it doesn't really matter because I am pretty sure all this will have to be nuked in favor of rule units 🤔

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcessInstance.java
```diff
@@ -163,14 +164,20 @@ public void start(String trigger, String referenceId) {
...
+            ((WorkflowProcessInstance)legacyProcessInstance).setState(STATE_ACTIVE);
+            this.status = ProcessInstance.STATE_COMPLETED;
```
MarianMacik, line 171: So Kogito process instance is completed, but legacy process instance is still active? Is that right? If so, why is it like that? I understand that cases are basically never complete as we discussed it on Zulip, but this can be considered as an inconsistency - one status is completed, the second one is active.

MarianMacik: Also btw - this feature is already present in 7.x? I think that we can there only close or destroy a case instance, right? We cannot explicitly complete it...

ruromero: Ad-hoc processes, i.e. cases, can't autocomplete unless there is a terminate end node but this is not common in cases where knowledge workers might want to add/update information in the case or trigger discretionary tasks. That means there wasn't a mechanism to close a case.

Regarding the naming. In 7.x you could close a case but here we are not explicitly using a Case model naming. We're still talking about processes and processes from the ACTIVE state move to the COMPLETED state, thus this action. Do you agree?

MarianMacik: I agree, but somewhat I missed the explanation why the legacy process is still being kept ACTIVE :)

ruromero: I get your point and that had to be aligned indeed. Thanks for your thorough review.

ruromero:
I did a rebase against master

tiagodolphine:
@cristianonicolai missing your review, would you like to take look?

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
```diff
@@ -65,7 +65,12 @@
      * @param referenceId optional reference id that points to a another  component triggering this instance
      */
     void startFrom(String nodeId, String referenceId);
-    
+
+    /**
+     * Completes a dynamic process that is in the ACTIVE state.
+     */
+    void complete();
```
cristianonicolai, line 72: from a perspective that processes will have at least some degree of structure, would it still make sense to allow it to be completed at any time? Thinking if we should first try to go for having a minial path that demonstrates how the process will be completed, not so much like a case close action. @ruromero @krisv wdyt?

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/MilestoneNodeInstance.java
```diff
@@ -53,10 +53,14 @@ public void internalTrigger(final NodeInstance from, String type) {
...
+        // KOGITO-2168 Conditions not supported
+        if(getProcessInstance().getKnowledgeRuntime().getAgenda() != null) {
```
cristianonicolai, line 58: @ruromero is there any point for us to add this if we wont actually use this approach with the internal agenda? I guess we should simply set active to true.

kogito-codegen/src/test/java/org/kie/kogito/process/impl/MilestoneTest.java
```diff
@@ -0,0 +1,62 @@
...
+import org.kie.kogito.codegen.AbstractCodegenTest;
+import org.kie.kogito.process.Process;
+import org.kie.kogito.process.ProcessInstance;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class MilestoneTest extends AbstractCodegenTest {
+
+    @Test
+    void testSimpleMilestone() throws Exception {
```
cristianonicolai, line 33: @ruromero should the actual milestone state be checked here?

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/MilestoneNodeInstance.java
ruromero: I didn't want to remove the old functionality because other tests still rely on the agenda. Until we don't provide a replacement for conditions I prefer to keep it this way.

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
ruromero: I think this is something we need at this point and we can adapt later on. Dynamic processes need a way to be completed.

cristianonicolai: @ruromero sure, in the example you created the test can be completed once the milestone is achieved, we dont necessarly want to manually have to call complete in a process. We're trying to give some flexibility but not complete freedom, to me the process needs to to handle a expected path to end, otherwise people can still abort the process instance.

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/MilestoneNodeInstance.java
cristianonicolai: so maybe we should remove the old tests as well :) My concern is to bring things that we wont/dont need, removing later is way more complicated to assess.

kogito-codegen/src/test/java/org/kie/kogito/process/impl/MilestoneTest.java
ruromero: Check my latest changes

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
ruromero: Complete only affects dynamic processes that are active. How do you want to limit it's usage? By only allow processes without any active node to be completed for example?

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/MilestoneNodeInstance.java
ruromero: I'm always up for removing stuff. I just didn't dare to do it without a second opinion.

kogito-codegen/src/test/java/org/kie/kogito/process/impl/MilestoneTest.java
```diff
@@ -0,0 +1,75 @@
...
+        processInstance.complete();
+        assertState(processInstance, ProcessInstance.STATE_PENDING);
+
+        processInstance.start();
+        assertState(processInstance, ProcessInstance.STATE_ACTIVE);
+        
+        RuleFlowProcessInstance legacyProcessInstance = (RuleFlowProcessInstance)((AbstractProcessInstance<?>) processInstance).legacyProcessInstance;
+        assertThat(legacyProcessInstance.getNodeInstances()).isEmpty();
+        assertThat(legacyProcessInstance.getNodeIdInError()).isNullOrEmpty();
+        Optional<String> milestoneId = Stream.of(legacyProcessInstance.getNodeContainer().getNodes()).filter(node -> node.getName().equals("Milestone1")).map(n -> (String)n.getMetaData().get(Metadata.UNIQUE_ID)).findFirst();
```
cristianonicolai, line 61: @ruromero I guess the idea would be to add a method into _RuleFlowProcessInstance_ with a list of miliestones, that then we could check using a status like _MilestoneStatus_ from v7. No need to publish it in the ProcessInstance object for now.

kogito-codegen/src/test/java/org/kie/kogito/process/impl/MilestoneTest.java
```diff
@@ -0,0 +1,75 @@
...
+        assertThat(app).isNotNull();
+
+        Process<? extends Model> p = app.processes().processById("TestCase.SimpleMilestone");
+
+        Model m = p.createModel();
+        Map<String, Object> parameters = new HashMap<>();
+        m.fromMap(parameters);
+        ProcessInstance<?> processInstance = p.createInstance(m);
+
+        assertState(processInstance, ProcessInstance.STATE_PENDING);
```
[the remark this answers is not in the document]
ruromero: I want to make sure the state in the processInstance and in the legacyProcessInstance are in sync

kogito-codegen/src/test/java/org/kie/kogito/process/impl/MilestoneTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+        Application app = generateCodeProcessesOnly("cases/SimpleMilestone.bpmn");
+        assertThat(app).isNotNull();
+
+        Process<? extends Model> p = app.processes().processById("TestCase.SimpleMilestone");
+
+        Model m = p.createModel();
+        Map<String, Object> parameters = new HashMap<>();
+        m.fromMap(parameters);
+        ProcessInstance<?> processInstance = p.createInstance(m);
```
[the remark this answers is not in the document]
ruromero: Not exactly this but I see your point

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/MilestoneNodeInstance.java
cristianonicolai: sorry, wrong gitlog :P looks fine as is

kogito-codegen/src/test/java/org/kie/kogito/process/impl/MilestoneTest.java
ruromero: I have added the `milestones` method as suggested

kogito-codegen/src/test/java/org/kie/kogito/process/impl/MilestoneTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+        assertState(processInstance, ProcessInstance.STATE_PENDING);
+        processInstance.complete();
```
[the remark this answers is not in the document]
ruromero: As discussed I have removed the `complete` method. A terminate end node must be present in order for the dynamic process to complete

api/kogito-api/src/main/java/org/kie/kogito/process/CaseNodeInstance.java
```diff
@@ -0,0 +1,50 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.process;
+
+import java.io.Serializable;
+
+public class CaseNodeInstance implements Serializable {
```
cristianonicolai, line 21:
```suggestion
public class MilestoneInstance implements Serializable {
```

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcessInstance.java
```diff
@@ -368,6 +378,35 @@ public void transitionWorkItem(String id, Transition<?> transition) {
...
+    public Collection<Milestone> milestones() {
+        return getCaseNodeInstances(MilestoneNode.class)
+                .map(n -> {
+                    String uid = (String) n.getMetaData().get(UNIQUE_ID);
+                    return new Milestone(uid, n.getName(), getStatus(uid));
+                })
+                .collect(Collectors.toList());
+    }
+
+    private Stream<Node> getCaseNodeInstances(Class<? extends Node> nodeClass) {
```
cristianonicolai, line 391:
```suggestion
    private Stream<Node> getNodeInstances(Class<? extends Node> nodeClass) {
```

ruromero: They're just nodes so I renamed it to `getNodes`

PR #505, opened 2020-05-13
add org.testcontainers:junit-jupiter to pom

mariofusco:
Before we were (I guess mistakenly) taken that dependency from quarkus bom, see https://github.com/quarkusio/quarkus/issues/8694#issuecomment-627772609


geoand:
Thanks @mariofusco!

PR #506, opened 2020-05-14
[KOGITO-1768] allow code scaffolding in kogito-maven-plugin

mariofusco:
@evacchi I had to rebase the old PR so instead I closed the old one, made the changes you suggested and open this one.

It is true that (except a minor thing to exclude generation of the rest service in the unlikely case that also the ws annotation are not available on the classpath) I didn't change much on the plugin itself. In particular I did the change to generate the sources in different folders depending on the type of the source file as we discussed but then I found it inconvenient and I undid it. In fact by doing so the user, in order to use the programmatic API and compile his project has to add as source folders in his project all the different folders where we put our code instead of just one. I'm open to discuss this but I didn't find this reasonable (and now I wonder if other libraries like jOOQ has the same problem and how they solved it). I still want to change the generated code and the programmatic API but this is out of the scope of this task and I opened another jira ( https://issues.redhat.com/browse/KOGITO-2074 ) to work on this.

### DIFF
--- a/kogito-maven-plugin/src/main/java/org/kie/kogito/maven/plugin/AbstractKieMojo.java
+++ b/kogito-maven-plugin/src/main/java/org/kie/kogito/maven/plugin/AbstractKieMojo.java
@@ -45,37 +45,39 @@ public abstract class AbstractKieMojo extends AbstractMojo {
     }
 
     protected DependencyInjectionAnnotator discoverDependencyInjectionAnnotator(boolean dependencyInjection, MavenProject project) {
-        if (!dependencyInjection) {
-            return null;
+        if (dependencyInjection) {
+            if ( hasSpring( project ) ) {
+                return new SpringDependencyInjectionAnnotator();
+            }
+
+            if ( hasQuarkus( project ) ) {
+                return new CDIDependencyInjectionAnnotator();
+            }
         }
 
-        boolean hasSpring = project.getDependencies().stream().anyMatch(d -> d.getArtifactId().contains("spring"));
-        if (hasSpring) {
-            return new SpringDependencyInjectionAnnotator();
-        }
-
-        boolean hasQuarkus = project.getDependencies().stream().anyMatch(d -> d.getArtifactId().contains("quarkus"));
-        if (hasQuarkus) {
-            return new CDIDependencyInjectionAnnotator();
-        }
-
-        throw new IllegalStateException("Unable to find dependency injection annotator");
+        return null;
     }
 
     protected KogitoBuildContext discoverKogitoRuntimeContext(MavenProject project)  {
-        boolean hasSpring = project.getDependencies().stream().anyMatch(d -> d.getArtifactId().contains("spring"));
-        if (hasSpring) {
+        if ( hasSpring( project ) ) {
             return new SpringBootKogitoBuildContext(fqcn -> hasClassOnClasspath(project, fqcn));
         }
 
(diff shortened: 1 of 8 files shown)


evacchi:
I don't think having *different* source folders was discussed; there might have been a misunderstanding

kogito-quarkus-extension/deployment/src/main/java/org/kie/kogito/quarkus/deployment/KogitoAssetsProcessor.java
```diff
@@ -238,31 +233,46 @@ public void generateModel(ArchiveRootBuildItem root,
...
-        });
+        }
     }
 
+    private boolean isCustomizable(GeneratedFile f) {
+        return f.relativePath().endsWith("Resource.java")
+                || (f.relativePath().contains("Query") && f.relativePath().endsWith(".java"))
+                || (f.relativePath().contains("DTO") && f.relativePath().endsWith(".java"));
+    }
+
```
evacchi, line 261: this is still using a "naming convention," and I feel it's going to be quite fragile

kogito-maven-plugin/src/main/java/org/kie/kogito/maven/plugin/AbstractKieMojo.java
```diff
@@ -45,36 +45,39 @@ protected void setSystemProperties(Map<String, String> properties) {
...
     }
-    
+
+    private boolean hasQuarkus( MavenProject project ) {
+        return project.getDependencies().stream().anyMatch( d -> d.getArtifactId().contains( "quarkus" ) );
+    }
+
+    private boolean hasSpring( MavenProject project ) {
+        return project.getDependencies().stream().anyMatch( d -> d.getArtifactId().contains( "spring" ) );
+    }
```
danielezonca, line 79: These checks are really error prone: it is enough to use one of [these](https://mvnrepository.com/search?q=spring&d=io.quarkus) quarkus-spring artifacts to break it.
I would suggest to find a better mechanism or at least looks for a more specific artifact. What about [`quarkus-core`](https://mvnrepository.com/artifact/io.quarkus/quarkus-core) for Quarkus and [`spring-core`](https://mvnrepository.com/artifact/org.springframework/spring-core)?

mariofusco: @danielezonca in general taking something that we discussed on Sunday afternoon and putting it in a github review on Monday morning is not fair imo :) 
Jokes apart I agree on that and will review it.

danielezonca: Actually I told you this same yesterday and now I just added a note to remember it, so I'm safe :P

mariofusco: @danielezonca the dependencies through which we are looking for are only the ones directly present in the pom, not the transitive ones. For this reason I'm afraid it is not safe to look for a specific quarkus or spring dependency, because it could be there only transitively and then we wouldn't find it. The only thing I've done is giving higher precedence to quarkus since, as we discussed, it would be much more common for a quarkus project to also have some spring dependencies that the opposite.

evacchi: let's create a JIRA for this. This PR is not doing much new in this regard, except refactoring some code.

PR #507, opened 2020-05-14
[KOGITO-1968] Implement missing features for DynamicNodes

ruromero:
* ActionNodes will only declare the variables that are mentioned in the script
* CompositeContextNodeVisitor will take the parent context when it is empty or null. Before it was done only when null
* ~~Added stages method that returns a collection of the existing stages in the process along with the adhoc fragments~~
* Added condition to Milestone
* Missing setAutocomplete to visitor
* Small refactoring for `WorkItemHandlerNotFoundException` and some redundant casts have been also removed
* Generate only used variables in actionNodes
* Propagate scope when empty in CompositeContextNodes


jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcessInstance.java
```diff
@@ -381,13 +387,33 @@ public void transitionWorkItem(String id, Transition<?> transition) {
...
                 .collect(Collectors.toList());
     }
 
+    @Override
+    public Collection<Stage> stages() {
+        return getNodes(DynamicNode.class)
+                .map(DynamicNode.class::cast)
+                .map(n -> {
+                    String uid = (String) n.getMetaData().get(UNIQUE_ID);
+                    return new Stage(uid, n.getName(), getStatus(uid), n.getActivationExpression(), n.getCompletionExpression(), n.isAutoComplete(), collectAdHocFragments(n));
```
tiagodolphine, line 404: What about moving the `StageBuilder` from test to src and use here as well?

ruromero: Yes, I'll do it

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/DynamicNodeInstance.java
```diff
@@ -63,43 +64,43 @@ public String getNodeName() {
...
-//    		triggerCompleted(NodeImpl.CONNECTION_DEFAULT_TYPE);
-//    	}
-
-        
-        String rule = "RuleFlow-AdHocComplete-" + getProcessInstance().getProcessId() + "-" + getDynamicNode().getUniqueId();
-        boolean isActive = ((InternalAgenda) getProcessInstance().getKnowledgeRuntime().getAgenda())
-            .isRuleActiveInRuleFlowGroup(getRuleFlowGroupName(), rule, getProcessInstance().getId());
+        boolean isActive = false;
+        // KOGITO-2168 Conditions not supported
+        if(getProcessInstance().getKnowledgeRuntime().getAgenda() != null) {
```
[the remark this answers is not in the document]
ruromero: Well, it is used 3 times and is something that will change or at least be revisited when supporting conditions. As discussed with @evacchi in the previous PR the agenda is something that should be dropped completely. So adding functions specific to the agenda is not worth it.

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ActionNodeVisitor.java
```diff
@@ -48,22 +51,31 @@ public void visitNode(String factoryField, ActionNode node, BlockStmt body, Vari
...
         }
         visitMetaData(node.getMetaData(), body, getNodeId(node));
         body.addStatement(getDoneMethod(getNodeId(node)));
     }
+
+    private boolean isVarPresent(Variable v, DroolsAction action) {
+        if(action == null) {
+            return false;
+        }
+        return action.toString().contains(v.getName());
```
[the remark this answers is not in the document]
ruromero: As the droolsConsequenceAction is the only one implementing the `toString` I check the consequence instead.

ruromero:
@tiagodolphine changes done

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcessInstance.java
```diff
@@ -381,13 +385,37 @@ public void transitionWorkItem(String id, Transition<?> transition) {
     @Override
     public Collection<Milestone> milestones() {
         return getNodes(MilestoneNode.class)
+                .map(MilestoneNode.class::cast)
```
cristianonicolai, line 388: if you change the getNodes method to return a types version, then this map is unnecessay:

```
private <T extends Node> Stream<T> getNodes(Class<T> nodeClass) {
        if (!(legacyProcessInstance instanceof RuleFlowProcessInstance)) {
            return Stream.empty();
        }
        RuleFlowProcessInstance processInstance = ((RuleFlowProcessInstance) legacyProcessInstance);
        return Arrays.stream(processInstance.getNodeContainer().getNodes()).filter(nodeClass::isInstance).map(n -> (T)n);
    }
```

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcessInstance.java
```diff
@@ -381,13 +385,37 @@ public void transitionWorkItem(String id, Transition<?> transition) {
...
+                            .withStatus(getStatus(uid))
+                            .withActivationExpression(n.getActivationExpression())
+                    .withCompletionExpression(n.getCompletionExpression()).withAutoComplete(n.isAutoComplete());
+                    getAdHocNodes(n).forEach(builder::withFragment);
+                    return builder.build();
+                })
+                .collect(Collectors.toList());
+    }
+
+    private Stream<Node> getAdHocNodes(DynamicNode node) {
```
cristianonicolai, line 413: how about adhoc nodes that are not inside of stages, should we also expose it here?

kogito-codegen/src/test/java/org/kie/kogito/process/impl/StageTest.java
```diff
@@ -0,0 +1,269 @@
...
+        Application app = generateCodeProcessesOnly("cases/SimpleAdHoc.bpmn");
+        assertThat(app).isNotNull();
+
+        Process<? extends Model> p = app.processes().processById("TestCase.SimpleAdHoc");
+
+        Model m = p.createModel();
+        Map<String, Object> parameters = new HashMap<>();
+        m.fromMap(parameters);
+
+        ProcessInstance<?> processInstance = p.createInstance(m);
```
cristianonicolai, line 63:
```suggestion
        ProcessInstance<?> processInstance = p.createInstance(p.createModel());
```

kogito-codegen/src/test/java/org/kie/kogito/process/impl/StageTest.java
```diff
@@ -0,0 +1,269 @@
...
+        assertThat(((Model) processInstance.variables()).toMap().get("caseFile_currentStage")).isEqualTo(3);
+        assertThat(processInstance.workItems().size()).isEqualTo(0);
+        assertState(processInstance, ProcessInstance.STATE_COMPLETED);
+        expected = expected.stream().map(s -> new Stage.Builder(s).withStatus(COMPLETED).build()).collect(Collectors.toList());
+        expected.forEach(eStage -> assertContainsStage(eStage, processInstance.stages()));
+    }
+
+    private Collection<Stage> buildExpectedMultipleAdHoc() {
+        Collection<Stage> expected = new ArrayList<>();
+        expected.add(new Stage.Builder("_4A154E74-F085-4ECA-93F1-DE452E624FB1").withName("Stage 1").withStatus(AVAILABLE).withAutoComplete(true).withFragment(new ActionNode(), "Task1").build());
```
cristianonicolai, line 123: can we avoid hard coding the node id, I guess that will need update as soon as the file is changed in the editor, it is probably not relevant for the test anyway.

kogito-codegen/src/test/java/org/kie/kogito/process/impl/StageTest.java
```diff
@@ -0,0 +1,269 @@
...
+        expected.add(new Stage.Builder("_168F4098-8ACA-4E81-9F97-5EAAC0782574").withName("Stage 2").withStatus(AVAILABLE).withAutoComplete(true).withFragment(new ActionNode(), "Task2").build());
+        expected.add(new Stage.Builder("_E2F977AC-205A-4314-A406-00CEA619CDF9").withName("Stage 3").withStatus(AVAILABLE).withAutoComplete(true).withFragment(new ActionNode(), "Task3").build());
+        return expected;
+    }
+
+    @Test
+    void testCaseFile() throws Exception {
```
cristianonicolai, line 130: based on what we discussed, shouldnt this handling of case file be avoided, so we can look into that shared scope approach?

drools/drools-core/src/main/java/org/drools/core/process/instance/impl/DefaultWorkItemManager.java
```diff
@@ -73,8 +76,7 @@ public void internalExecuteWorkItem(WorkItem workItem) {
         WorkItemHandler handler = this.workItemHandlers.get(workItem.getName());
         if (handler != null) {
             handler.executeWorkItem(workItem, this);
-        } else throw new WorkItemHandlerNotFoundException( "Could not find work item handler for " + workItem.getName(),
```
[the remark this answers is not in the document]
ruromero: There's no need. It's just an improvement.

drools/drools-core/src/main/java/org/drools/core/process/instance/impl/WorkItemImpl.java
```diff
@@ -26,15 +25,15 @@
 import org.kie.api.runtime.process.NodeInstance;
 import org.kie.api.runtime.process.ProcessInstance;
 
-public class WorkItemImpl implements WorkItem, Serializable {
+public class WorkItemImpl implements WorkItem {
```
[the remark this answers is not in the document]
ruromero: All WorkItem implementations must be Serializable so why not just moving it to the interface? Why do you think it makes more sense to have it in the Class?

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/VariableScope.java
```diff
@@ -20,13 +20,10 @@
...
 
-import org.jbpm.process.core.datatype.impl.type.ObjectDataType;
 import org.jbpm.process.core.Context;
 import org.jbpm.process.core.context.AbstractContext;
+import org.jbpm.process.core.datatype.impl.type.ObjectDataType;
 
-/**
- * 
- */
 public class VariableScope extends AbstractContext {
```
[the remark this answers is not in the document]
ruromero: Sure, I usually try to avoid too many formatting changes to reduce the "noise". I will re-format this file.

drools/drools-core/src/main/java/org/drools/core/process/instance/impl/WorkItemImpl.java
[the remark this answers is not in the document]
mswiderski: why does every work item must be serializable?

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/VariableScope.java
```diff
@@ -74,7 +69,7 @@ public Variable findVariable(String variableName) {
                 return variable;
             }
         }
-        if (variableName.startsWith(CASE_FILE_PREFIX) && variableName.indexOf(".") == -1) {
+        if (variableName.startsWith(CASE_FILE_PREFIX) && variableName.contains(".")) {
```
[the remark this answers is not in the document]
mswiderski: what is the reason to introduce case file prefix in Kogito?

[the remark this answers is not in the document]
ruromero: Yes

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
```diff
@@ -172,4 +173,10 @@
      * @return All the process milestones with their current status
      */
     Collection<Milestone> milestones();
+
+    /**
+     * Returns the process stages
+     * @return All the process stages with their current status
+     */
+    Collection<Stage> stages();
```
mswiderski, line 181: wouldn't it make more sense to either 
- use new sub class for case related methods
- use more process related names instead 

mixing these two concepts can become really confusing...

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/LightWorkItemManager.java
```diff
@@ -253,7 +251,7 @@ public void retryWorkItem( String workItemID, Map<String, Object> params ) {
         
     }
     
-    private class TransitionToActive implements Transition<Void> {
+    private static class TransitionToActive implements Transition<Void> {
```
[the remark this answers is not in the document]
ruromero: non-static nested classes keep the reference to the parent object which is not needed in this case

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
ruromero: This is a behaviour that is already implicit in a process so sub-classing wouldn't make much sense to me. I am in favour of renaming stages for something else. Any suggestion?

kogito-codegen/src/test/java/org/kie/kogito/process/impl/StageTest.java
ruromero: I'll look into it. Makes sense

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/VariableScope.java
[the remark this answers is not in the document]
ruromero: > what is the reason to introduce case file prefix in Kogito?

This is not new, it was already there. Maybe the question is should this be removed?

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/VariableScope.java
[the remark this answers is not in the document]
cristianonicolai: yeah, I agree to that, but it was little strange to review actually.

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/VariableScope.java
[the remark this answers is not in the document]
cristianonicolai: yeah case file prefix inst something that we would be using in Kogito

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcessInstance.java
ruromero: nice suggestion

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/VariableScope.java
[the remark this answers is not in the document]
ruromero: Removing, then

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcessInstance.java
mswiderski: Wouldn't it be enough to cast to WorkflowProcessInstance instead of RuleFlowProcessInstance?

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
mswiderski: stage is actually a subprocess, why not having this as such? If needed you can narrow it down to only ad hoc subprocesses all all...

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcessInstance.java
ruromero: Yes

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcessInstance.java
tiagodolphine: in fact, it could be NodeInstanceContainer

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/VariableScope.java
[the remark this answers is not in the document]
ruromero: done

kogito-codegen/src/test/java/org/kie/kogito/process/impl/StageTest.java
ruromero: I will remove it for now until we implement the shared variables

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
```diff
@@ -169,7 +187,22 @@
 
     /**
      * Returns the process milestones
+     *
      * @return All the process milestones with their current status
      */
     Collection<Milestone> milestones();
+
+    /**
+     * Returns the AdHocSubProcesses in the proces
```
tiagodolphine, line 196: Does it return only Adhoc subprocesses or any subprocesses?

api/kogito-api/src/main/java/org/kie/kogito/process/casemgmt/SubProcess.java
```diff
@@ -0,0 +1,92 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.process.casemgmt;
+
+import java.util.ArrayList;
+import java.util.Collection;
+
+public class SubProcess extends ItemDescription {
```
tiagodolphine, line 22: Should we name it Subprocess or AdhocSubprocess?

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/VariableScope.java
```diff
@@ -89,36 +72,36 @@ public Context resolveContext(Object param) {
...
+                throw new IllegalArgumentException("Variable '" + name + "' has incorrect data type expected:"
+                        + var.getType().getStringType() + " actual:" + value.getClass().getName());
+            }
+        }
+    }
+
+    /*
+     * mainly for test coverage to easily switch between settings
+     */
+    public static void setVariableStrictOption(boolean turnedOn) {
```
tiagodolphine, line 98: why not just setting the system property on the tests?

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
ruromero: Refactoring leftover, thanks

api/kogito-api/src/main/java/org/kie/kogito/process/casemgmt/SubProcess.java
ruromero: AdHocSubProcesses are a specific type of SubProcesses

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
```diff
@@ -169,7 +187,22 @@
...
+     * @return All the process {@link SubProcess} with their current status
+     */
+    Collection<SubProcess> subProcesses();
+
+    /**
+     * Returns the process adHocFragments
+     *
+     * @return All the {@link AdHocFragment} in the process
+     */
+    Collection<AdHocFragment> getAdHocFragments();
```
cristianonicolai, line 207: I think to align with the other methods naming we would call this: 
```suggestion
    Collection<AdHocFragment> adHocFragments();
```

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/VariableScope.java
cristianonicolai: @ruromero why not simply use the system property instead?

cristianonicolai: just formatting old code, nevermind @ruromero

ruromero: I couldn't just disappoint you but I didn't change everything or else tests would go crazy

PR #517, opened 2020-05-19
[KOGITO-2202] avoid use of quarkus deprecated methods in quarkus exte…

mariofusco:


kogito-codegen/src/main/java/org/kie/kogito/codegen/decision/DecisionContainerGenerator.java
```diff
@@ -55,15 +53,13 @@
     private static final String TEMPLATE_JAVA = "/class-templates/DMNApplicationClassDeclTemplate.java";
 
     private String applicationCanonicalName;
-    private final Path basePath;
-    private final Collection<DMNModel> models;
+    private final List<DMNResource> resources;
     private boolean useTracing = false;
 
-    public DecisionContainerGenerator(String applicationCanonicalName, Path basePath, Collection<DMNModel> models) {
+    public DecisionContainerGenerator(String applicationCanonicalName, List<DMNResource> resources) {
```
evacchi, line 59: what happened to `Path basePath` ? was it unused? Why is the Collection now a list?

kogito-codegen/src/main/java/org/kie/kogito/codegen/decision/DecisionContainerGenerator.java
```diff
@@ -91,9 +87,9 @@ public ClassOrInterfaceDeclaration classDeclaration() {
         ClassOrInterfaceDeclaration typeDeclaration = (ClassOrInterfaceDeclaration) clazz.getTypes().get(0);
         ClassOrInterfaceType applicationClass = StaticJavaParser.parseClassOrInterfaceType(applicationCanonicalName);
         ClassOrInterfaceType inputStreamReaderClass = StaticJavaParser.parseClassOrInterfaceType(java.io.InputStreamReader.class.getCanonicalName());
-        for (DMNModel model : models) {
-            Path sourcePath = Paths.get(model.getResource().getSourcePath());
-            Path relativizedPath = basePath.relativize(sourcePath);
+        for (DMNResource resource : resources) {
+            Path sourcePath = Paths.get(resource.getDmnModel().getResource().getSourcePath());
+            Path relativizedPath = resource.getPath().relativize(sourcePath);
```
evacchi, line 92: are you sure this will always work?

kogito-codegen/src/main/java/org/kie/kogito/codegen/decision/DecisionContainerGenerator.java
mariofusco: @evacchi I discussed this change with @tarilabs. Now we can have multiple paths, so a DMNResource is basically a Pair<Path, DMNModel>. In this way you know from which path each model has been created and this should also answer to the question you made on the other comment.

PR #526, opened 2020-05-21
[KOGITO-2272] Hot reload test stuck if deployment is not available

danielezonca:
https://issues.redhat.com/browse/KOGITO-2272

### DIFF
--- a/kogito-quarkus-extension/integration-test-hot-reload/pom.xml
+++ b/kogito-quarkus-extension/integration-test-hot-reload/pom.xml
@@ -13,7 +13,7 @@
     <artifactId>kogito-quarkus-integration-test-hot-reload</artifactId>
     <name>Kogito - Quarkus - Integration Tests (Hot Reload)</name>
 
-     <dependencies>
+    <dependencies>
         <dependency>
             <groupId>org.kie.kogito</groupId>
             <artifactId>kogito-quarkus</artifactId>
@@ -22,6 +22,15 @@
             <groupId>org.kie.kogito</groupId>
             <artifactId>drools-decisiontables</artifactId>
         </dependency>
+
+        <!-- this is used implicitly by hot reload tests so let's make Maven aware of it -->
+        <dependency>
+            <groupId>org.kie.kogito</groupId>
+            <artifactId>kogito-quarkus-deployment</artifactId>
+            <version>${project.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
 
         <dependency>
             <groupId>io.quarkus</groupId>
@@ -90,7 +99,9 @@
                                 </goals>
                                 <configuration>
                                     <systemProperties>
-                                        <native.image.path>${project.build.directory}/${project.build.finalName}-runner</native.image.path>
+                                        <native.image.path>
+                                            ${project.build.directory}/${project.build.finalName}-runner
+                                        </native.image.path>
                                     </systemProperties>
                                 </configuration>
                             </execution>
--- a/kogito-quarkus-extension/integration-test-hot-reload/pom.xml
+++ b/kogito-quarkus-extension/integration-test-hot-reload/pom.xml
@@ -28,7 +28,6 @@
             <groupId>org.kie.kogito</groupId>
             <artifactId>kogito-quarkus-deployment</artifactId>
             <version>${project.version}</version>
-            <scope>provided</scope>
         </dependency>
 
 
--- a/kogito-codegen/src/test/java/org/kie/kogito/codegen/RuleUnitCompilerTest.java
+++ b/kogito-codegen/src/test/java/org/kie/kogito/codegen/RuleUnitCompilerTest.java
@@ -69,7 +69,7 @@ public class RuleUnitCompilerTest extends AbstractCodegenTest {
 
         sofia.setAge( 22 );
         adults.getPersons().update( dhSofia, sofia );
-        assertEquals( 1, instance.fire() );
+        assertEquals( 10, instance.fire() );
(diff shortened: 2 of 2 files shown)


danielezonca:
@evacchi @mariofusco 
Ok I think we have at least a problem in the configuration of the project and maybe a bug in Quarkus too ([logs](https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/blue/rest/organizations/jenkins/pipelines/KIE/pipelines/kogito/pipelines/kogito-runtimes/branches/PR-526/runs/1/log/?start=0)):
- I just created a simple test failure in `kogito-codegen` on [purpose](https://github.com/kiegroup/kogito-runtimes/blob/1ac51c72364fc3d97030154757e61af4093b3f1b/kogito-codegen/src/test/java/org/kie/kogito/codegen/RuleUnitCompilerTest.java#L72)
- This makes the build of that module to fail (of course :) )
- All maven dependent module are skipped (in this case `kogito-quarkus-deployment`) [1]
- Hot reload test module doesn't have an explicit dependency declaration on `deployment` so it starts
- `kogito-quarkus-deployment` is not available and it makes the test fails [2]
- The runner stuck until the timeout (this is probably the bug in Quarkus I mentioned) [3]
- Run is aborted and you need to find the _real_ failure in the full log (the last error is completely misleading) [4]

[1]
```
[INFO] ------------------------------------------------------------------------
[INFO] Skipping Kogito - Quarkus - Deployment
[INFO] This project has been banned from the build due to previous failures.
[INFO] ------------------------------------------------------------------------
```
[2]
```
Caused by: org.eclipse.aether.transfer.ArtifactNotFoundException: Could not find artifact org.kie.kogito:kogito-quarkus-deployment:jar:8.0.0-SNAPSHOT
	at org.eclipse.aether.internal.impl.DefaultArtifactResolver.resolve(DefaultArtifactResolver.java:414)
	... 58 more
```
[3]
I do believe this is a bug in Quarkus runner but in all the usages of `QuarkusDevModeTest` I [found](https://github.com/quarkusio/quarkus/search?q=QuarkusDevModeTest&unscoped_q=QuarkusDevModeTest), it is used as integration test _inside_ `deployment` module so they probably do not expect this code to fail. I plan to open an issue on they GitHub later tomorrow.
Is there any reason why we put hot reload test in a different module instead of inside `deployment`? @evacchi @mariofusco 
[4]
```
Timeout has been exceeded
Finished: ABORTED
```

Now I will try to add `kogito-quarkus-deployment` as dependency to Hot reload module with scope `provided` just to inform Maven of this "implicit" dependency and I expect the build to fail properly.

danielezonca:
Ok as expected now the build fails in the proper way :)
My plan now is:
- Revert the broken test
- Make the PR ready for review
- Wait a green build/reviews
- Merge this fix and in parallel show this behavior to Quarkus team

kogito-quarkus-extension/integration-test-hot-reload/pom.xml
```diff
@@ -23,6 +23,15 @@
             <artifactId>drools-decisiontables</artifactId>
         </dependency>
 
+        <!-- this is used implicitly by hot reload tests so let's make Maven aware of it -->
+        <dependency>
+            <groupId>org.kie.kogito</groupId>
+            <artifactId>kogito-quarkus-deployment</artifactId>
+            <version>${project.version}</version>
+            <scope>provided</scope>
```
evacchi, line 31: i think the scope provided may be unnecessary; if ITs are usually _inside_ the deployment module, then by definition they _do_ depend on the deployment module.

danielezonca: Done

danielezonca:
@MarianMacik @radtriste 
Can you please review?
The executive summary: the CI build stuck on hotreload test until timeout if there is an error in an upper dependency 
Solution: added a dependency to `deployment` module so Maven is aware of this "implicit dependency" and skip the build of the module if needed

kogito-quarkus-extension/integration-test-hot-reload/pom.xml
```diff
@@ -23,6 +23,14 @@
             <artifactId>drools-decisiontables</artifactId>
         </dependency>
 
+        <!-- this is used implicitly by hot reload tests so let's make Maven aware of it -->
+        <dependency>
+            <groupId>org.kie.kogito</groupId>
+            <artifactId>kogito-quarkus-deployment</artifactId>
+            <version>${project.version}</version>
```
radtriste, line 30: could we make it `provided` ?

danielezonca: Lol it was `provided` and then Edoardo proposed to remove `provided` :) ([see comment](https://github.com/kiegroup/kogito-runtimes/pull/526#discussion_r429100522))
Actually I have mixed feeling about this: it is true that provided should be enough (and this is the reason why I did it at the beginning) but it is also true that all other extensions have hot reload tests inside the `deployment` module so it is more similar now with a proper dependency declaration.

radtriste: ahah oki :)
I just read the executive summary before reviewing sry...
Anyway this is just the integration-tests, so it should be ok.

MarianMacik: Another option would be to create an artificial dependency like we have it for KIE Server:

https://github.com/kiegroup/droolsjbpm-integration/blob/9e089567726f5c1fdffa295a81079d45c17aad43/kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/pom.xml#L246-L259

But if the tests depend on it, I would leave it with the default scope as it is now :)

PR #537, opened 2020-06-01
[KOGITO-2168] Implement Conditions for Milestones

ruromero:
Fix https://issues.redhat.com/browse/KOGITO-2168

Adds a processEventListener to watch variables changes until the condition is met.

It breaks JBPM7.x code intentionally because the Agenda is not implemented in Kogito's DummyKnowledgeRuntime


ruromero:
@evacchi Do you think we can completely remove the constraint from the MilestoneNode?

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/MilestoneNodeVisitor.java
```diff
@@ -37,13 +48,36 @@ protected String getNodeKey() {
     @Override
     public void visitNode(String factoryField, MilestoneNode node, BlockStmt body, VariableScope variableScope, ProcessMetaData metadata) {
         body.addStatement(getAssignedFactoryMethod(factoryField, MilestoneNodeFactory.class, getNodeId(node), getNodeKey(), new LongLiteralExpr(node.getId())))
-                .addStatement(getNameMethod(node, "Milestone"))
-                .addStatement(getFactoryMethod(getNodeId(node), METHOD_CONSTRAINT, new StringLiteralExpr(StringEscapeUtils.escapeJava(node.getConstraint()))));
+                .addStatement(getNameMethod(node, "Milestone"));
+        if(node.getConstraint() != null && !node.getConstraint().trim().isEmpty()) {
+            body.addStatement(getFactoryMethod(getNodeId(node), METHOD_CONSTRAINT, new StringLiteralExpr(StringEscapeUtils.escapeJava(node.getConstraint()))));
```
evacchi, line 53: if this is a constraint _expression_ why is it a string and not an expression?

jbpm/jbpm-flow-builder/src/test/java/org/jbpm/test/util/AbstractBaseTest.java
```diff
@@ -46,7 +46,7 @@ public KieSession createKieSession(KiePackage... pkg) {
         try { 
             return createKieSession(false, pkg);
         } catch(Exception e ) { 
-            String msg = "There's no reason fo an exception to be thrown here (because the kbase is not being serialized)!";
+            String msg = "There's no reason for an exception to be thrown here (because the kbase is not being serialized)!";
             fail( msg );
             throw new RuntimeException(msg, e);
```
evacchi, line 51: not your code, but I am noticing that an exception is thrown after fail(); so, I believe it's actually _never_ thrown. You can probably remove the throw?

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/DynamicNodeInstance.java
```diff
@@ -79,9 +79,8 @@ public void internalTrigger(NodeInstance from, String type) {
                 agenda.getRuleFlowGroup(ruleFlowGroup).setAutoDeactivate(false);
                 agenda.activateRuleFlowGroup(ruleFlowGroup, getProcessInstance().getId(), getUniqueId());
             }
-            String rule =  "RuleFlow-AdHocComplete-" + getProcessInstance().getProcessId() + "-" + getDynamicNode().getUniqueId();
             isActive = ((InternalAgenda) getProcessInstance().getKnowledgeRuntime().getAgenda())
-                    .isRuleActiveInRuleFlowGroup(getRuleFlowGroupName(), rule, getProcessInstance().getId());
+                    .isRuleActiveInRuleFlowGroup(getRuleFlowGroupName(), getActivationEventType(), getProcessInstance().getId());
```
evacchi, line 83: if there is a rule flow group is no longer used, do we still need this line?

ruromero:
@cristianonicolai @mswiderski do you think the removed tests don't make sense any more in the current Kogito implementation?

mswiderski:
> @cristianonicolai @mswiderski do you think the removed tests don't make sense any more in the current Kogito implementation?

in my opinion yes, they don't make sense any more in Kogito, so +1 to remove them

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
```diff
@@ -189,7 +189,7 @@
      *
      * @return All the process milestones with their current status
      */
-    Collection<Milestone> milestones();
```
cristianonicolai, line 192: why not keep as Milestone?

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/node/DynamicNode.java
```diff
@@ -109,4 +119,22 @@ public String getActivationEventName() {
...
+        this.activationPredicate = activationPredicate;
+        return this;
+    }
+
+    public DynamicNode setCompletionExpression(Predicate<ProcessContext> copmletionPredicate) {
+        this.completionPredicate = copmletionPredicate;
+        return this;
+    }
+
+    public boolean testActivation(ProcessContext context) {
```
tiagodolphine, line 133: what about `isActivated(ProcessContext context)`?

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/node/DynamicNode.java
```diff
@@ -109,4 +119,22 @@ public String getActivationEventName() {
...
+        return activationPredicate == null || activationPredicate.test(context);
+    }
+
+    public boolean testCompletion(ProcessContext context) {
```
tiagodolphine, line 137: what about ` isCompleted(ProcessContext context)`?

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
ruromero: reverting

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/node/MilestoneNode.java
```diff
@@ -16,71 +16,56 @@
...
-        return this.constraint;
-    }    
-    
-    public String getMatchVariable() {
-        return matchVariable;
-    }
-    
-    public void setMatchVariable(String matchVariable) {
-        this.matchVariable = matchVariable;
+    public boolean testCondition(ProcessContext context) {
```
tiagodolphine, line 39: `isConditionAchived`? `isConditionMatched`? well no idea what would be a better name, just thinking the "test" prefix is strange...

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/node/CompositeNode.java
```diff
@@ -41,7 +38,7 @@
     private Map<String, CompositeNode.NodeAndType> outConnectionMap = new HashMap<String, CompositeNode.NodeAndType>();
 	private boolean cancelRemainingInstances = true;
 
-	private boolean autoComplete = true;
+	boolean autoComplete = true;
```
tiagodolphine, line 41: one option is to add a constructor to initialize this the field and keep the default as true.

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/node/DynamicNode.java
```diff
@@ -19,20 +19,30 @@
...
 	
 	private String activationExpression;
 	private String completionExpression;
+    private Predicate<ProcessContext> activationPredicate;
+    private Predicate<ProcessContext> completionPredicate;
 	private String language;
-			
+
+    public DynamicNode() {
+        autoComplete = false;
```
tiagodolphine, line 43: one option is to use a constructor on the superclass.

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/node/MilestoneNode.java
```diff
@@ -16,71 +16,56 @@
...
-        this.constraint = constraint.getConstraint();
-    }
-    
-    public void setConstraint(String constraint){
-        this.constraint = constraint;
+    public void setCondition(Predicate<ProcessContext> conditionPredicate) {
+        this.conditionPredicate = conditionPredicate;
     }
 
-    public String getConstraint(){
```
tiagodolphine, line 46: What about keeping the method `getConstraint()` and just returning the `getMetadata("CONDITION")` on it, or change to `getCondition()`, just to simplify for callers, instead of doing the `milestone.getMetadata(CONDITION)`...

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/node/CompositeNode.java
ruromero: that is a mistake, indeed. I already reverted it

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/node/MilestoneNode.java
ruromero: with what purpose?

tiagodolphine: to make the usage better and simpler, in this way the callers don't need to know the value is gotten from the metadata and even the key to be used of the metadata that would be used to get the value...

ruromero: I mean, who needs to know about the condition? Do you think this should be exposed to the Milestone object? Same for the activation/completion conditions in the AdHocSubProcess

tiagodolphine: mmm if the condition value would be accessed outside the class I think it would be great, to simplify getting the information... otherwise yeah.. it does not make sense...

cristianonicolai:
> > @cristianonicolai @mswiderski do you think the removed tests don't make sense any more in the current Kogito implementation?
> 
> in my opinion yes, they don't make sense any more in Kogito, so +1 to remove them

+1 to remove

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/MilestoneNodeInstance.java
```diff
@@ -53,112 +45,42 @@ public void internalTrigger(final NodeInstance from, String type) {
...
         super.addEventListeners();
-        addActivationEventListener();
+        addEventListener();
     }
 
-    private void addActivationEventListener() {
-        getProcessInstance().getKnowledgeRuntime().addEventListener(this);
-        getProcessInstance().addEventListener(getActivationEventType(), this, true);
+    private void addEventListener() {
+        getProcessInstance().getKnowledgeRuntime().getProcessRuntime().addEventListener(ContextAwareEventListener.using(listener -> {
```
cristianonicolai, line 68: @ruromero I think there is some inconsistency on how the events are handled here. See how the events used to be registered and remove. Is this intentional or should we reevaluate to use the same mechanism? For instance on line 79, the listener for _getActivationEventType_ is never added, the the one add here is not removed.

api/kogito-api/src/main/java/org/kie/kogito/process/flexible/Milestone.java
```diff
@@ -18,40 +18,29 @@
...
-        return condition;
     }
 
     @Override
     public String toString() {
-        return "Milestone{" +
-                "condition='" + condition + '\'' +
-                ", {" + super.toString() + "}";
+        return "Milestone{" + super.toString() + "}";
     }
```
cristianonicolai, line 28: maybe add an accessor method for the builder:

`    public static Builder builder(){
        return new Builder();
    }`

api/kogito-api/src/main/java/org/kie/kogito/process/flexible/Milestone.java
```diff
@@ -18,40 +18,29 @@
...
 
     public static class Builder {
 
-        private String id;
-        private String name;
-        private Status status;
-        private String condition;
+        protected String id;
```
cristianonicolai, line 32: should this remain as private?

api/kogito-api/src/main/java/org/kie/kogito/process/flexible/Milestone.java
```diff
@@ -18,40 +18,29 @@
...
+        protected String name;
+        protected Status status;
 
         public Builder(String id) {
```
cristianonicolai, line 36: a builder constructor should be without parameter, then you could add a _id_ or _withId_ method.

api/kogito-api/src/main/java/org/kie/kogito/process/flexible/Milestone.java
```diff
@@ -18,40 +18,29 @@
...
             this.id = id;
         }
 
-        public Builder(Milestone m) {
-            this.id = m.getId();
-            this.name = m.getName();
-            this.status = m.getStatus();
-            this.condition = m.getCondition();
+        public Builder(ItemDescription i) {
```
cristianonicolai, line 40: I'm not sure why we need a builder that takes an instance with all the parameters, this feels more like a clone method.
In the tests, it seems that only the name is reused, so why not create a new builder instance, then _withName_, _withStatus_?

kogito-codegen/src/test/java/org/kie/kogito/process/impl/MilestoneTest.java
```diff
@@ -75,20 +79,32 @@ void testSimpleMilestone() throws Exception {
...
-        expected.add(new Milestone.Builder("").withName("Milestone2").withStatus(AVAILABLE).withCondition("CaseData(data.get(\"dataComplete\") == true)").build());
+        expected.add(new Milestone.Builder("").withName("Milestone").withStatus(AVAILABLE).build());
+        assertMilestones(expected, processInstance.milestones());
+
+        processInstance.start();
+        assertState(processInstance, ProcessInstance.STATE_ACTIVE);
 
-        Collection<Milestone> milestones = processInstance.milestones();
-        assertMilestones(expected, milestones);
+        expected = expected.stream().map(m -> new Milestone.Builder(m).withStatus(AVAILABLE).build()).collect(Collectors.toList());
```
cristianonicolai, line 100: missing `assertMilestones(expected, processInstance.milestones()); `?

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/node/MilestoneNode.java
cristianonicolai: @ruromero I think we're overusing the metadata for this, keeping a specific attribute and method I think is more clear unles there is a specific reason for this change. I would actually prefer to keep `getContraint()` and same for `node.getActivationExpression()` and `node.getCompletionExpression()`

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/DynamicNodeInstance.java
```diff
@@ -66,188 +55,103 @@ public void internalTrigger(NodeInstance from, String type) {
...
+                .setNodeInstance(this);
+        return getNodeInstances(false).isEmpty() && getDynamicNode().testCompletion(context);
     }
-    
+
     private void addActivationListener() {
-        getProcessInstance().getKnowledgeRuntime().addEventListener(this);
-        getProcessInstance().addEventListener(getActivationEventType(), this, true);
+        getProcessInstance().getKnowledgeRuntime().getProcessRuntime().addEventListener(ContextAwareEventListener.using(listener -> {
+            if (canActivate() && getState() == ProcessInstance.STATE_PENDING) {
```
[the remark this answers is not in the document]
ruromero: It's the state of the node instance. As it is a CompositeNode, it has a state and the state constants used are the same as for the ProcessInstance.
What I'm trying to do here is to set the state to pending when it is not ready to be activated to avoid triggering the inner nodes multiple times

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/MilestoneNodeInstance.java
```diff
@@ -53,112 +45,42 @@ public void internalTrigger(final NodeInstance from, String type) {
...
+    private boolean isCompleted() {
+        ProcessContext context = new ProcessContext(getProcessInstance().getKnowledgeRuntime())
+                .setNodeInstance(this);
+        return getMilestoneNode().testCondition(context);
+    }
+
     @Override
     public void addEventListeners() {
         super.addEventListeners();
-        addActivationEventListener();
```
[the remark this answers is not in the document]
ruromero: This is actually about completion. The node is activated once the signal is received but it won't complete until the condition is met. I will rename it

api/kogito-api/src/main/java/org/kie/kogito/process/flexible/Milestone.java
ruromero: That means the id is mandatory. I prefer to keep it like this to make it more evident.

api/kogito-api/src/main/java/org/kie/kogito/process/flexible/Milestone.java
ruromero: removed

api/kogito-api/src/main/java/org/kie/kogito/process/flexible/Milestone.java
ruromero: yes, reverted. It was a refactoring leftover

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/node/MilestoneNode.java
ruromero: The purpose of keeping these expressions in the metadata is only to allow the nodeVisitor access so it can generate the lambdas. What I could do is to remove them in the visitors.
Why would they be needed further this point?

api/kogito-api/src/main/java/org/kie/kogito/process/flexible/Milestone.java
cristianonicolai: I guess that would be a not null check in the constructor then. A builder is just a facilitator to create the object IMHO.

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/DynamicNodeInstance.java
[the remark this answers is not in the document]
cristianonicolai: hm, ok, that is strange to see the process state enum reused in here.

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/StateBasedNodeInstance.java
```diff
@@ -46,6 +46,7 @@
 import org.jbpm.workflow.instance.impl.ExtendedNodeInstanceImpl;
 import org.jbpm.workflow.instance.impl.NodeInstanceResolverFactory;
 import org.jbpm.workflow.instance.impl.WorkflowProcessInstanceImpl;
+import org.kie.api.event.process.ProcessEventListener;
```
cristianonicolai, line 49: please remove unused import

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/xml/processes/MilestoneNodeHandler.java
```diff
@@ -18,6 +18,7 @@
 
 import org.drools.compiler.compiler.xml.XmlDumper;
 import org.drools.core.xml.ExtensibleXmlParser;
+import org.jbpm.ruleflow.core.Metadata;
```
cristianonicolai, line 21: unused import

cristianonicolai:
jenkins please retest this

ruromero:
jenkins please retest this

PR #565, opened 2020-06-17
KOGITO-2480 - Serverless Workflow Specification - Add support for integration type functions

tsurdilo:
allows to define and call any present camel route and get results from it

### DIFF
--- a/jbpm/jbpm-serverless-workflow/src/main/java/org/jbpm/serverless/workflow/parser/ServerlessWorkflowParser.java
+++ b/jbpm/jbpm-serverless-workflow/src/main/java/org/jbpm/serverless/workflow/parser/ServerlessWorkflowParser.java
@@ -56,6 +56,7 @@ public class ServerlessWorkflowParser {
     private static final String SYSOUT_TYPE_PARAM = "message";
     private static final String SERVICE_TYPE = "service";
     private static final String DECISION_TYPE = "decision";
+    private static final String INTEGRATION_TYPE = "integration";
     private static final String RULE_TYPE = "rule";
     private static final String NODE_START_NAME = "Start";
     private static final String NODE_END_NAME = "End";
@@ -429,8 +430,12 @@ public class ServerlessWorkflowParser {
                         current = factory.ruleSetNode(idCounter.getAndIncrement(), action.getFunctionRef().getRefName(), actionFunction, embeddedSubProcess);
                         factory.connect(start.getId(), current.getId(), start.getId() + "_" + current.getId(), embeddedSubProcess);
                         start = current;
+                    } else if (INTEGRATION_TYPE.equals(actionFunction.getType())) {
+                        current = factory.camelRouteServiceNode(idCounter.getAndIncrement(), action.getFunctionRef().getRefName(), actionFunction, embeddedSubProcess);
+                        factory.connect(start.getId(), current.getId(), start.getId() + "_" + current.getId(), embeddedSubProcess);
+                        start = current;
                     } else {
-                        LOGGER.warn("currently unsupported function type, supported types are 'script', 'sysout', 'service', 'decision', 'ruleunit'");
+                        LOGGER.warn("currently unsupported function type, supported types are 'script', 'sysout', 'service', 'decision', 'rule', 'integration'");
                         LOGGER.warn("defaulting to script type");
                         String script = ServerlessWorkflowUtils.scriptFunctionScript("");
                         current = factory.scriptNode(idCounter.getAndIncrement(), action.getFunctionRef().getRefName(), script, embeddedSubProcess);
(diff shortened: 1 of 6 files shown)


tsurdilo:
@ricardozanini we do not need the rest workitemhandler - we cannot do that any more as in we cannot keep updating some custom rest call library that will never be "enough" for everyone that uses it. You have two options here, either use "integration" function which uses this in-line impl or use the "java" based function where it will call your own defined service that can interact with the external camel service

tsurdilo:
@ricardozanini the problem is that with REST you have to deal with different kinds of auth, keystores, configurations, headers, sso, timeouts, sockets, etc etc etc. We have had to update our rest workitem  50 times over the years cause it was constantly asked to add this add that to it and it still is not enough. No matter what we put in place as a "solution" ..if its custom it will not be good for use in the long run and has to be maintained like we did w/ restworkitemhandler. 

Just use https://camel.apache.org/components/latest/http-component.html :)

tsurdilo:
@ricardozanini why does the call to camel-k have to be done explicitly via http call? that call should be really internal imo. All that workflow model should care about is the name of the route to call and the parameters user wants to pass to it.

ricardozanini:
> @ricardozanini why does the call to camel-k have to be done explicitly via http call? that call should be really internal imo. All that workflow model should care about is the name of the route to call and the parameters user wants to pass to it.

We can't make an "internal" call because Camel-K is deployed as a separated service in the namespace. It's provisioned by Knative, think that it's all about a microservice out there waiting to be called. That's why the operator will handle the creation of the route and the deployment of the service. That will put them together. If we are aiming to compose microservices, one of the way of doing this communication flow is using HTTP calls. Also, Camel-K accepts "events", so we can produce an event to react to a Camel-K service.

Camel-K has nothing to do with regular Camel project. We won't call it routes within the JVM, we won't share a JVM at all. It wil be just another function being deployed in the namespace.

tsurdilo:
@ricardozanini yes thanks for explanation. I understand better now whats needed. The "workitem" is a publicly exposed thing that needs to be maintained for customers. We need something else. I'll work on that next then.

mswiderski:
@ricardozanini was reading these comments and something cought my attention - which operator will deploy camel routes? And based on what conditions?

In addition to what was said, i think calling a camel k service is just one use case. Or to put it in other words, calling a service in the same namespace will be rather rare case while calling services outside given namespace or even outside of openshift cluster will be most common. Then needs like auth, header setup, transformation etc will be of great use, plus having that outside of the process/workflow definition would be easier to manage in my opinion.

ricardozanini:
Hi @mswiderski ! thanks for chime in!

We are on draft phase for this support on our operator. The ideia is to have the WorkFlow Functions being processed by the operator and then inject the needed information into the workflow image on runtime. 

So in the use case you’ve mentioned, one can easily define a function with metadata fields pointing to secrets/config maps with information regarding authentication on external services. The operator would then “resolve” these informations on deployment time, injecting in form of java properties all the information about that function to the workflow. So the workflow can call the services seamlessly. More or less the same thing we have today with OnBoarding example, but the engine wouldn’t call any Kubernetes API, but the operator. We’re trying to separate the “infrastructure” domain from the workflow execution. 

The idea is to have the workflow engine ready to run any kind of functions, while the Operator will be responsible to “deploy” or “resolve” them on deployment time. 

We are preparing a PR to the spec to handle scenarios like this, will send you the link as soon as I finish it!

mswiderski:
@ricardozanini thanks for the details

it's ok to push resolve responsibility to operator but in the end you would have to implement something that actually calls the service/functions over the wire. Let's say you have one that requires OAuth authentication while another one requires special headers to be set. The REST work item handler would have to support all of these features... and they are already done in camel. That's my main point, you will end up duplicating exact same features not to mention error handling and data transformation.

> The idea is to have the workflow engine ready to run any kind of functions, while the Operator will be responsible to “deploy” or “resolve” them on deployment time.

and what in case the service you're going to call is not in OpenShift? Like an AWS service/function?

ricardozanini:
> @ricardozanini thanks for the details
> 
> it's ok to push resolve responsibility to operator but in the end you would have to implement something that actually calls the service/functions over the wire. Let's say you have one that requires OAuth authentication while another one requires special headers to be set. The REST work item handler would have to support all of these features... and they are already done in camel. That's my main point, you will end up duplicating exact same features not to mention error handling and data transformation.
> 

We will use an internal Camel route to implement the requirements for rest calls, like the OAuth you mentioned. It's not the responsibility of the operator to call the services, but the engine. The operator would only resolve the k8s information/deploy the resources needed for the service to be called. The call itself you be made by the engine, IDK if it was clear in my last message.

> and what in case the service you're going to call is not in OpenShift? Like an AWS service/function?

Kubernetes services endpoints will be resolved by the Operator and injected in the engine, other functions/services that does not require operator work, will be handled directed by the engine. In the AWS service scenario for instance, I'd say a Camel Route again calling this service. Or if you want to decouple this call from the engine, you can create a Camel-K service to make that call for you.

mswiderski:
So are you saying that kogito operator will read the function definition from the workflow json/yaml file and create camel routes?

What I am really after here is that the need for the engine to call http endpoints as this is where all the complexity - look at the v7 RESTWorkItemHandler to see how much code is needed and yet this impl covers only very basic cases.

ricardozanini:
> So are you saying that kogito operator will read the function definition from the workflow json/yaml file and create camel routes?

Yes. The operator would always read the Functions to figure out how to deploy/create the infra needed for the function to run. See the draft CRD: https://github.com/cncf/wg-serverless-workflow/pull/28

> What I am really after here is that the need for the engine to call http endpoints as this is where all the complexity - look at the v7 RESTWorkItemHandler to see how much code is needed and yet this impl covers only very basic cases.

That's why @tsurdilo is researching a new way of doing that by using an internal Camel route, so we should be able to delegate to the `camel-http` instead.

PR #572, opened 2020-06-19
KOGITO-2472 Codegen does not create handler for ServiceTasks with overloaded methods

evacchi:
https://issues.redhat.com/browse/KOGITO-2472

the problem is actually _subtler_. The handlers that are internally generated are named using the `interfaceName_methodName` template, which causes them to overwrite each other in the case of overloading. The solution is to use a _mangled_ name instead, including type parameters. Now the template is ``SomeClass_doSomethingHandler_my$Type1$$my$Type2`` which should be more reliable


cristianonicolai:
jenkins please retest this

PR #573, opened 2020-06-19
KOGITO-1678 Service tasks should support futures

evacchi:
depends on #572 

It should now be possible for services to be defined as:

```
public class MyService {
   public CompletionStage<T> myMethod(...) {
      return CompletableFuture....
   }
}
```

the completion will be async


jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ServiceTaskDescriptor.java
```diff
@@ -213,29 +211,70 @@ public ClassOrInterfaceDeclaration classDeclaration() {
...
-                    .addVariable(new VariableDeclarator(new ClassOrInterfaceType(null, Object.class.getCanonicalName()), "result", callService));
+            resultVarDecl.setInitializer(new MethodCallExpr(new NameExpr("java.util.Collections"), "singletonMap")
+                                                 .addArgument(new StringLiteralExpr(outAssociations.get(0).getSources().get(0)))
+                                                 .addArgument(new NameExpr("value")));
+            completeWorkItemBody.addStatement(resultVar);
+            completeWorkItemBody.addStatement(completeWorkItem);
+        }
 
-            executeWorkItemBody.addStatement(resultField);
+        if (CompletionStage.class.isAssignableFrom(m.getReturnType())) {
```
danielezonca, line 256: What about also consider `Is Async` flag from the XML to automatically wrap method call with a `CompletableFuture`?
In this way user can specify an async task just returning a completable stage or using the flag

mswiderski:
@evacchi have you tried this with persistence enabled? The thing is that you pass work item and manager to the async completion method which wont be connected to runtime any more. It will be called from anither thread and thus won’t be part of the unit of work so even if it will not throw any exception during completion of work item it won’t be saved into the storage.
I’d say using `process.instances().findById().completeWorkItem()` Should be used instead.

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ServiceTaskDescriptor.java
mswiderski: I’d review if this is async has still significant role in kogito.... unless you plan to introduce transactions

kogito-codegen/src/test/java/org/kie/kogito/codegen/tests/ServiceTaskTest.java
```diff
@@ -164,98 +171,134 @@ public void shouldInferMethodSignatureFromClass() throws Exception {
...
+
+        Process<? extends Model> p = app.processes().processById("ServiceProcessFuture");
+        Model m = p.createModel();
+        m.fromMap(Collections.singletonMap("latch", latch));
+        ProcessInstance<? extends Model> processInstance = p.createInstance(m);
+        processInstance.start();
+
+        assertThat(processInstance.status()).isEqualTo(ProcessInstance.STATE_ACTIVE);
+        latch.await(2, TimeUnit.SECONDS);
+        assertThat(latch.getCount()).isEqualTo(0);
```
mswiderski, line 301: Would be good to do asserts on data to make sure they are set based on async service invocation

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ServiceTaskDescriptor.java
evacchi: @mswiderski I think @danielezonca was thinking about retrofitting the flag with a different sematnics; the idea would be to _not_ requiring to return a CompletionStage when the flag is set. in That case, the engine would wrap the entire method in a CompletableFuture or equivalent (internally). I think that would be nice

evacchi:
>  I’d say using process.instances().findById().completeWorkItem() Should be used instead.

thanks will verify that

kogito-codegen/src/test/java/org/kie/kogito/codegen/tests/ServiceTaskTest.java
evacchi: I have to confess I had a hard time getting synchronization to work, especially with state. The main issue here is that the latch goes to zero _within_ the body of the service; however, before the workitem is actually _completed_, there is still some time. Apparently this time is too short for the unit test to catch up with it, and therefore the test fail; hence why, neither `STATE_COMPLETED` is asserted. I was planning to refine the PR further to fix this.

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ServiceTaskDescriptor.java
danielezonca: @mswiderski 
I think it still makes sense in Kogito but with a different meaning: not related to transaction but to simplify this "pattern".
Said differently if a service node (aka a Java method) takes time to execute and I would like to make REST service returns, it could be enough to just flag the node as async to automatically add [this "wrapper" code](https://github.com/kiegroup/kogito-runtimes/blob/133fbb18d389a62e9a9b48c82dc4e1bcc0a059e2/kogito-codegen/src/test/java/org/kie/kogito/codegen/data/HelloService.java#L77-L94) during codegen.
Wdyt?

mswiderski: it might be bit oversimplified approach, as what will happen when the service takes too much time and then crashes in between? what's the status of the work... thus I think just wrapping in an async execution could be a potential source of issues.

kogito-codegen/src/test/java/org/kie/kogito/codegen/tests/ServiceTaskTest.java
mswiderski: maybe use process event listeners (afterProcessCompleted) to call the latch. Plus you can use UnitOfWork so the listeners are delayed until you call complete on the unit. that way you should have more control over the execution.

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ServiceTaskDescriptor.java
danielezonca: Yes this is true but how different is with the current PR?
User writes a method that returns a CompletableFuture that can runs forever or fail and Kogito has to manage process status accordingly.
I don't get the difference of the CompletableFuture written by the user and a wrapper that mimics the same (aka invoke long running method inside an async lambda)

mswiderski: @danielezonca the main difference is that this service was designed and implemented as async so it is certainly more suited for such invocation from the process. 

About is async and this PR ... there is not that big difference (except above comment) so I'd like to bring the awareness that this is not just to move the execution to another thread but also to take into account that it might have consequences.

With sync execution, in case or error the node will be marked as error and thus can be easily retriggered, skippedetc. With async I don't think this is covered and thus you might end up in a non recoverable state of process instance. So having this risk clearly stated users who prefer to use async methods might implement the logic in the service to make sure it won't leave process instance in non recoverable state. If you just wrap it then there is no way to do so and it's so easy for people to just mark it as async because it will be "more performant"

just saying ... :)

evacchi:
I managed to get this green with commit 8b60a9f, now I am trying to address @mswiderski's comment using `process.instances().findById().completeWorkItem()` however, this cannot be used in every situation because if it is synchronous, then the PI-ID might not be present yet. This is a behavior @danielezonca and I observed while writing [our home automation demo](https://github.com/danielezonca/home-automation-poc) (using signals). At the time we wondered if we could set the pi-id _before_ the process started so that the value would have been available; however I am not sure whether this changes the semantics somewhere else

/cc @cristianonicolai @ricardozanini

mswiderski:
@evacchi not sure what you mean by

> this cannot be used in every situation because if it is synchronous, then the PI-ID might not be present yet.

instance id is assigned before process instance is started so it's there.

evacchi:
@mswiderski maybe the issue is different then. This line fails:

```
app.processes().processById(pid).instances().findById(piid) 
                                       .orElseThrow(() -> new IllegalArgumentException(piid))
                                       .completeWorkItem(piid, result);
```

throwing the IllegalArgumentException in test `org.kie.kogito.codegen.tests.WorkItemParamsTest`; i.e. the pi-id cannot be found. Maybe there is some different initialization issue

mswiderski:
ok, that is slightly different as what it fails here is not the missing process instance id but persistence layer does not have it. Which is sort of expected as this is the call that starting process instance and thus it should never be there until call/unit of work is completed.

So this code should only be invoked from another thread, meaning when async method call has completed.

evacchi:
> ok, that is slightly different as what it fails here is not the missing process instance id but persistence layer does not have it. Which is sort of expected as this is the call that starting process instance and thus it should never be there until call/unit of work is completed.
> 
> So this code should only be invoked from another thread, meaning when async method call has completed.

I suspected so; but now I fear that there is a looming race condition there :/ i.e. the detached thread may occasionally bump into a situation where the ID is not set in either place

mswiderski:
correct, it might be a race condition so maybe you could integrate that with unit of work... so it is called after it was successfully persisted...

evacchi:
> you could integrate that with unit of work...

what do you mean with "that" ? :)

mswiderski:
call to complete work item 

`app.processes().processById(pid).instances().findById(piid) 
                                       .orElseThrow(() -> new IllegalArgumentException(piid))
                                       .completeWorkItem(piid, result);`

evacchi:
so you're suggesting it is not done in the handler anymore?

mswiderski:
handler only registers it to be done not earlier than the completion of this request, same as process event listeners are "delayed"

PR #582, opened 2020-06-24
KOGITO-2207 - Create process without adding to ProcessInstanceManager

cristianonicolai:
…hen loading process

### DIFF
--- a/jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcessInstance.java
+++ b/jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcessInstance.java
@@ -52,17 +52,8 @@ import org.kie.internal.process.CorrelationAwareProcessRuntime;
 import org.kie.internal.process.CorrelationKey;
 import org.kie.internal.process.CorrelationProperty;
 import org.kie.kogito.Model;
-import org.kie.kogito.process.EventDescription;
-import org.kie.kogito.process.MutableProcessInstances;
-import org.kie.kogito.process.NodeInstanceNotFoundException;
-import org.kie.kogito.process.NodeNotFoundException;
 import org.kie.kogito.process.Process;
-import org.kie.kogito.process.ProcessError;
-import org.kie.kogito.process.ProcessInstance;
-import org.kie.kogito.process.ProcessInstanceDuplicatedException;
-import org.kie.kogito.process.ProcessInstanceNotFoundException;
-import org.kie.kogito.process.Signal;
-import org.kie.kogito.process.WorkItem;
+import org.kie.kogito.process.*;
 import org.kie.kogito.process.flexible.AdHocFragment;
 import org.kie.kogito.process.flexible.ItemDescription.Status;
 import org.kie.kogito.process.flexible.Milestone;
@@ -84,7 +75,7 @@ public abstract class AbstractProcessInstance<T extends Model> implements Proces
 
     protected Integer status;
     protected String id;
-    protected String businessKey;
+    protected CorrelationKey correlationKey;
     protected String description;
 
     protected ProcessError processError;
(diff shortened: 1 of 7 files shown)


jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcessInstance.java
```diff
@@ -125,6 +125,7 @@ public void internalSetProcessInstance(org.kie.api.runtime.process.ProcessInstan
         if (this.legacyProcessInstance != null && this.status != ProcessInstance.STATE_PENDING) {
             throw new IllegalStateException("Impossible to override process instance that already exists");
         }
+        ((WorkflowProcessInstanceImpl) this.legacyProcessInstance).disconnect();
```
mswiderski, line 128: why is this needed? Since process instance was loaded from the storage it should not be connected at all. So how does it get into connected state here as the reconnect is few lines below?

cristianonicolai: yep, I'm doing this only so that legacy process instance is removed from the _DefaultProcessInstanceManager_, otherwise it will keep appending more and more instances to it.

mswiderski: why is it added there in the first place?

cristianonicolai: This is done when unmarshalling the process: https://github.com/kiegroup/kogito-runtimes/blob/master/jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/marshalling/ProcessInstanceMarshaller.java#L102
and https://github.com/kiegroup/kogito-runtimes/blob/master/jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcessInstance.java#L112

mswiderski: in that case I'd change it there as it looks very confusing to disconnect and then reconnect.

cristianonicolai: @mswiderski ok so trying a new approach here, so create process instance does not add it to _ProcessInstanceManager_. Some side effects are that now id is only generated when process is started and not created. Same goes for validating business key, id can only be done when starting the process now, not when creating. wdyt?

cristianonicolai:
jenkins please retest this

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcessInstance.java
```diff
@@ -84,7 +75,7 @@
 
     protected Integer status;
     protected String id;
-    protected String businessKey;
+    protected CorrelationKey correlationKey;
```
mswiderski, line 78: why this change is needed? correlation key type is way to complex and kind of legacy too, so keeping string here and use correlation key on legacy process instance would make more sense to me, but not a huge issue I guess.

cristianonicolai: @mswiderski I end up changing this because the the `CorrelationKey` object is needed for when adding the process instance to the `ProcessInstanceManager`. I can revert that but then I would need to recreate this object every time the process is started. See: https://github.com/kiegroup/kogito-runtimes/pull/582/files#diff-c72f318d542b5bbf6d5de58b1b5c2f1fR163

cristianonicolai:
jenkins please retest this

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcessInstance.java
MarianMacik: @cristianonicolai Just a clarification. Being this a process instance, if we want to start two process instances, then we need to first create 2 process instance objects anyway, right? That would mean that the recreation will only be needed for the second time when the process instance is started (so once it will be done in the constructor, second time in the star method). But to me it makes sense to compute this in the constructor and then reuse it in whatever method within the same object.

cristianonicolai: @MarianMacik correct, we need to first create then start, so it makes sense to me to be done at the constructor level.

PR #585, opened 2020-06-25
KOGITO-2546  Refactor process/rules/decision-specific config out of ApplicationConfig

evacchi:
A lot of the logic in ApplicationConfig is really for ProcessConfig. We can move this logic in a generated ProcessConfig class, and generate that as well. Same for other type of configs (rules, decisions)

- Create different config classes depending on the underlying runtime (CDI, Spring, None -- i.e. codegen tests)
- Use templates for different types of runtime for code clarity

The code looks "more" but it's actually being substituted by templates, and I think the clarity has improved a lot. There are less conditional branches that insert framework-specific logic to the generated classes, and instead there are framework-specific templates, so it's easy to see what goes where. 

The framework-agnostic codegen has been preserved -- it has its own template when appropriate.

Expect more of this to come (re: Application.java and Process/Decision.java, etc.)

https://issues.redhat.com/browse/KOGITO-2546


evacchi:
I am adding a bunch of reviewers, not because they're all necessary, but because they might be interested, as this was a conversation we had, or it was something we may have tangentially discussed.

kogito-codegen/src/main/java/org/kie/kogito/codegen/AbstractConfigGenerator.java
```diff
@@ -0,0 +1,74 @@
...
+
+    private final String packageName;
+    private final String sourceFilePath;
+
+    private final String resourceCdi;
+    private final String resourceSpring;
+
+    private DependencyInjectionAnnotator annotator;
+
+    public AbstractConfigGenerator(String packageName, String targetTypeName, String resourceCdi, String resourceSpring) {
```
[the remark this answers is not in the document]
evacchi: it won't be exposed unless it's overridden by the subclass AFAIK. Anyway sure it can be protected..

drools/kogito-dmn/src/main/java/org/kie/kogito/dmn/config/AbstractDecisionConfig.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.util.Collection;
+import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.kie.kogito.decision.DecisionEventListenerConfig;
+
+public class AbstractDecisionConfig implements org.kie.kogito.decision.DecisionConfig {
```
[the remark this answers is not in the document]
evacchi: good question. It's more of an implementation detail that something I'd like to see people to instantiate explicitly, that's why it's not necessary for it to be a "concrete" class.

kogito-codegen/src/main/resources/class-templates/config/SpringApplicationConfigTemplate.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.drools.core.config.StaticRuleConfig;
+import org.kie.kogito.dmn.config.StaticDecisionConfig;
+import org.kie.kogito.process.impl.StaticProcessConfig;
+
+@org.springframework.stereotype.Component
+public class ApplicationConfig implements org.kie.kogito.Config {
```
gitgabrio, line 13: @evacchi 
Hi. I write this comment here but it is valid for the other templates `implements org.kie.kogito.Config `

Those templates implement the interface (that it is statically written in java). So everything needed to generate this code is already known/retrievable from the interface itself. What happen with all those templates is that every time a new model has been added, those templates must be manually modified with copy-and-paste code -that is basically error-prone and time consuming IMO.
I think that everything that is "mechanically" derived from java code should be automatically generated - leaving hard-coded template only for complex cases that are not easily created out from existing sources

drools/drools-core/src/main/java/org/drools/core/config/AbstractRuleConfig.java
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
```
cristianonicolai, line 2:
```suggestion
 * Copyright 2020 Red Hat, Inc. and/or its affiliates.
```

drools/drools-core/src/main/java/org/drools/core/config/AbstractRuleConfig.java
```diff
@@ -0,0 +1,61 @@
...
+public abstract class AbstractRuleConfig implements RuleConfig {
+
+    private final RuleEventListenerConfig ruleEventListenerConfig;
+
+    public AbstractRuleConfig(RuleEventListenerConfig ruleEventListenerConfig) {
+        this.ruleEventListenerConfig = ruleEventListenerConfig;
+    }
+
+    public AbstractRuleConfig(
+            Iterable<org.kie.kogito.rules.RuleEventListenerConfig> ruleEventListenerConfigs,
```
cristianonicolai, line 36: just some general code format, to add imports and review the formmating

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcessConfig.java
```diff
@@ -0,0 +1,110 @@
...
+public abstract class AbstractProcessConfig implements ProcessConfig {
+
+    private final WorkItemHandlerConfig workItemHandlerConfig;
+    private final SignalManagerHub signalManagerHub = new DefaultSignalManagerHub();
+    private final ProcessEventListenerConfig processEventListenerConfig;
+    private final UnitOfWorkManager unitOfWorkManager;
+    private final JobsService jobsService;
+
+    protected AbstractProcessConfig(
+            Iterable<org.kie.kogito.process.WorkItemHandlerConfig> workItemHandlerConfig,
```
cristianonicolai, line 43: same for formatting.

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/StaticProcessConfig.java
```diff
@@ -31,6 +33,14 @@
     private final UnitOfWorkManager unitOfWorkManager;
     private final JobsService jobsService;
 
+    public StaticProcessConfig() {
+        this.workItemHandlerConfig = new DefaultWorkItemHandlerConfig();
```
cristianonicolai, line 37: use `this(new DefaultWorkItemHandlerConfig()...` instead?

kogito-codegen/src/main/resources/class-templates/config/ApplicationConfigTemplate.java
```diff
@@ -1,40 +1,33 @@
...
+import org.drools.core.config.StaticRuleConfig;
+import org.kie.kogito.dmn.config.StaticDecisionConfig;
+import org.kie.kogito.process.impl.StaticProcessConfig;
 
 public class ApplicationConfig implements org.kie.kogito.Config {
 
-    protected ProcessConfig processConfig;
-    protected RuleConfig ruleConfig;
-    protected DecisionConfig decisionConfig;
+    private org.kie.kogito.process.ProcessConfig processConfig = new StaticProcessConfig();
```
cristianonicolai, line 14: keep the imports?

kogito-codegen/src/main/resources/class-templates/config/CdiDecisionConfigTemplate.java
```diff
@@ -0,0 +1,24 @@
+import java.util.Collection;
```
cristianonicolai, line 1: review imports and formatting

kogito-codegen/src/test/java/org/kie/kogito/codegen/tests/BusinessRuleTaskTest.java
```diff
@@ -70,7 +70,7 @@ public void testBasicBusinessRuleTaskWithAgendaListener() throws Exception {
         Application app = generateCode(Collections.singletonList("ruletask/BusinessRuleTask.bpmn2"), Collections.singletonList("ruletask/BusinessRuleTask.drl"));
         assertThat(app).isNotNull();
         final AtomicInteger counter = new AtomicInteger();
-        ((DefaultRuleEventListenerConfig)app.config().rule().ruleEventListeners()).register(new DefaultAgendaEventListener() {
```
cristianonicolai, line 73: @evacchi isnt `register` a little more explicit about the intent here? like registering a listener?

evacchi: it would but it's non-public and we're downcasting to the (possibly wrong) impl now. This alternative works in every situation (well, in this current impl) and it's ok for `kogito-codegen` test cases. In examples, even in tests, we should actually `@annotate` the event listener and let the runtime do the wiring, so this method won't be called at all.

kogito-codegen/src/main/resources/class-templates/config/CdiDecisionConfigTemplate.java
evacchi: I have unqualified imports except for CDI/Spring + org.kie.kogito.process.ProcessConfig, etc because the generated code has the same unqualified name (ProcessConfig, etc) and it may clash

kogito-codegen/src/main/resources/class-templates/config/SpringApplicationConfigTemplate.java
evacchi: we clarified the overall strategy on a voice call :) for all the others:

- refactor all common logic to an abstract class or a method library
- simplify and strip down templates to a bare minimum necessary to enable (constructor) injection and delegate to the library methods/abstract classes

this should make repetitive code only a minor annoyance, while keeping the flow of the codegen procedures clearer to follow. While we both agree that mechanical derivation is the optimum, the idea is to find the sweet spot between full automation and simple readability.

evacchi:
@cristianonicolai @danielezonca @tiagodolphine I am merging this because it's been sitting here and I don't want to risk more conflicts ( had one today :) ). We can further refine with more PRs!

PR #590, opened 2020-06-26
[KOGITO-2567] Restore root classloader in DMNKogito

danielezonca:
https://issues.redhat.com/browse/KOGITO-2567

### DIFF
--- a/drools/kogito-dmn/src/main/java/org/kie/kogito/dmn/DMNKogito.java
+++ b/drools/kogito-dmn/src/main/java/org/kie/kogito/dmn/DMNKogito.java
@@ -53,7 +53,6 @@ public class DMNKogito {
     public static DMNRuntime createGenericDMNRuntime(Reader... readers) {
         List<Resource> resources = Stream.of(readers).map(ReaderResource::new).collect(Collectors.toList());
         DMNRuntime dmnRuntime = DMNRuntimeBuilder.fromDefaults()
-                                                 .setRootClassLoader(null)
                                                  .buildConfiguration()
                                                  .fromResources(resources)
                                                  .getOrElseThrow(e -> new RuntimeException("Error initalizing DMNRuntime", e));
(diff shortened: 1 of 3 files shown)


evacchi:
@tarilabs I am not sure what you mean. The CL was set to `null` already, this is just removing that line. Where's the difference? Anyhow, I don't necessarily want to merge this

tarilabs:
If you remove the line, as this PR is proposing, the DMNRuntime internal WILL try to use the Classloader.
My understanding in native scenario, we cannot use it to load class dynamically.
So my perspective is that this PR should not be merged, as otherwise would just be rugging a problem under the carpet; we should leave master code as-is, setting the classloader as null, so we make SURE there is NO use of the classloader by the dmnruntime internals.

evacchi:
oh ok I was missing that detail. @danielezonca is there any particular reason why you wanted to restore this?

tarilabs:
I synced with @danielezonca on this immediately after I commented on the PR.
Daniele was looking to have the CL which is also used by the DMNRuntime internals for service load, but the solution, imho for Kogito, is not passing through that mechanism.

danielezonca:
Let me add a bit of background: we tried to test [jPMML-DMN integration](https://github.com/kiegroup/kie-jpmml-integration/tree/master/kie-dmn-jpmml) in Kogito (JVM mode) and it failed with a NPE, the exception came from [this code](https://github.com/kiegroup/drools/blob/master/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/DMNCompilerImpl.java#L306) in DMNCompilerImpl during DMN import compilation because the classloader is `null`. Using this patch we were able to run an example ([here](https://github.com/kiegroup/trusty-ai-sandbox/tree/master/end2endPoC/loanEligibilityApp/app)).

We do not aim to make jPMML native compile (actually I'm not sure it can because it performs a lot of XML work at runtime) so we tested only JVM mode and for now this is the only PMML impl we can use because the old PMML impl uses 7.x RuleUnit while the new one is not ready yet.
I expect we will have to change some code in `kie-dmn` to be sure the new PMML impl can native compile and work together with DMN but we will look at that later (probably the line of code that throws the NPE).
The build is green so this change is not breaking native compilation of DMN that's why I proposed to merge it but I'm open to discuss.

Side note: I cannot create a DMN-jPMML integration test here because of license limitation so I would probably need a completely new repo only for that because `kie-jpmml-integration` repo is in 7.x stream so we cannot create kie -> kogito dependency.

evacchi:
>  The build is green so this change is not breaking native compilation of DMN that's why I proposed to merge it but I'm open to discuss.

well the CI build currently _does not_ test native builds; I am not sure I follow you here :)

danielezonca:
@evacchi 
Ok fairpoint, I was not aware that CI is not running native tests. Ok I run them manually and it is still working :)

tarilabs:
please Close this, as it's superseeded by the enseble:
https://github.com/kiegroup/drools/pull/2985
https://github.com/kiegroup/kogito-runtimes/pull/622

PR #596, opened 2020-06-30
[KOGITO-2444] Generating REST endpoints for AdHocFragments

ruromero:

Fix https://issues.redhat.com/browse/KOGITO-2444

* Allow dynamic processes without startNode to be started
* Generate signal-based endpoints for AdHocFragments (including HumanTasks)

### DIFF
--- a/integration-tests/integration-tests-quarkus/src/test/java/org/kie/kogito/integrationtests/quarkus/AdHocFragmentsTest.java
+++ b/integration-tests/integration-tests-quarkus/src/test/java/org/kie/kogito/integrationtests/quarkus/AdHocFragmentsTest.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.integrationtests.quarkus;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import io.quarkus.test.junit.QuarkusTest;
+import io.restassured.RestAssured;
+import io.restassured.http.ContentType;
+import org.junit.jupiter.api.Test;
+
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.CoreMatchers.not;
+import static org.hamcrest.CoreMatchers.notNullValue;
+import static org.hamcrest.Matchers.emptyOrNullString;
+
+@QuarkusTest
+class AdHocFragmentsTest {
+
+    static {
+        RestAssured.enableLoggingOfRequestAndResponseIfValidationFails();
+    }
+
+    @Test
+    void testUserTaskProcess() {
+        Map<String, String> params = new HashMap<>();
+        params.put("var1", "Kermit");
+
+        String pid = given()
+                .contentType(ContentType.JSON)
+            .when()
+                .body(params)
+                .post("/AdHocFragments")
+            .then()
+                .statusCode(200)
+                .body("id", not(emptyOrNullString()))
+                .body("var1", equalTo("Kermit"))
+                .extract().path("id");
+
+        String link = given()
+                .contentType(ContentType.JSON)
+            .when()
+                .post("/AdHocFragments/{pid}/AdHocTask1", pid)
+            .then()
+                .statusCode(200)
+                .header("Link", notNullValue())
+                .extract().header("Link");
+
+        String taskPath = link.substring(link.indexOf("<") + 1, link.indexOf(">"));
+
+        params = new HashMap<>();
+        params.put("newVar1", "Gonzo");
+        given()
+                .contentType(ContentType.JSON)
+            .when()
+                .urlEncodingEnabled(false)
+                .body(params)
+                .post("/AdHocFragments/{taskPath}", taskPath)
+            .then()
+                .body("var1", equalTo("Gonzo"))
+                .statusCode(200);
+    }
+
+    @Test
+    void testServiceTaskProcess() {
+        Map<String, String> params = new HashMap<>();
+        params.put("var1", "Kermit");
+
+        String pid = given()
+                .contentType(ContentType.JSON)
+            .when()
+                .body(params)
+                .post("/AdHocFragments")
+            .then()
+                .statusCode(200)
+                .body("id", not(emptyOrNullString()))
+                .body("var1", equalTo("Kermit"))
+                .extract().path("id");
+
+        given()
+                .contentType(ContentType.JSON)
+            .when()
+                .post("/AdHocFragments/{pid}/Service_Task", pid)
+            .then()
+                .statusCode(200)
+                .body("var1", equalTo("Hello Kermit 5!"));
+    }
+
+    @Test
+    void testNonAdHocUserTaskProcess() {
+        Map<String, String> params = new HashMap<>();
+        params.put("var1", "Kermit");
+
+        String pid = given()
+                .contentType(ContentType.JSON)
+            .when()
+                .body(params)
+                .post("/AdHocFragments")
+            .then()
+                .statusCode(200)
+                .body("id", not(emptyOrNullString()))
+                .body("var1", equalTo("Kermit"))
+                .extract().path("id");
+
+        given()
+                .contentType(ContentType.JSON)
+            .when()
+                .post("/AdHocFragments/{pid}/Task", pid)
+            .then()
+                .statusCode(404);
+    }
+}
--- a/integration-tests/integration-tests-quarkus/src/test/java/org/kie/kogito/integrationtests/quarkus/FlexibleProcessTest.java
+++ b/integration-tests/integration-tests-quarkus/src/test/java/org/kie/kogito/integrationtests/quarkus/FlexibleProcessTest.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.integrationtests.quarkus;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import io.quarkus.test.junit.QuarkusTest;
+import io.restassured.RestAssured;
+import io.restassured.http.ContentType;
+import org.junit.jupiter.api.Test;
+
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.CoreMatchers.not;
+import static org.hamcrest.Matchers.emptyOrNullString;
+
+@QuarkusTest
+class FlexibleProcessTest {
+
+    static {
+        RestAssured.enableLoggingOfRequestAndResponseIfValidationFails();
+    }
+
+    @Test
+    void testInstantiateProcess() {
+        Map<String, String> params = new HashMap<>();
+        params.put("var1", "first");
+        params.put("var2", "second");
+
+        given()
+                .contentType(ContentType.JSON)
+            .when()
+                .body(params)
+                .post("/AdHocProcess")
+            .then()
+                .statusCode(200)
+                .body("id", not(emptyOrNullString()))
+                .body("var1", equalTo("Hello first! Script"))
+                .body("var2", equalTo("second Script 2"));
+    }
+}
--- a/integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-it/src/test/java/org/kie/kogito/integrationtests/springboot/AdHocFragmentsTest.java
+++ b/integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-it/src/test/java/org/kie/kogito/integrationtests/springboot/AdHocFragmentsTest.java
@@ -1,0 +1,130 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.integrationtests.springboot;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import io.restassured.RestAssured;
+import io.restassured.http.ContentType;
+import org.junit.jupiter.api.Test;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.annotation.DirtiesContext;
+
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.CoreMatchers.not;
+import static org.hamcrest.CoreMatchers.notNullValue;
+import static org.hamcrest.Matchers.emptyOrNullString;
+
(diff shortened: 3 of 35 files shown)


kogito-codegen/src/main/resources/class-templates/RestResourceUserTaskTemplate.java
```diff
@@ -3,11 +3,32 @@
...
 
 
 public class $Type$Resource {
 
-    
+    @POST
+    @Path("/{id}/$taskname$")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    public javax.ws.rs.core.Response signal(@PathParam("id") final String id) {
```
tiagodolphine, line 17: isn't better to interpolate the method name `signal` based on the `$taskname$`? if you check https://github.com/kiegroup/kogito-runtimes/blob/e5994ce75c5343023547c85c45048ba3dfe2455b/kogito-codegen/src/main/java/org/kie/kogito/codegen/process/AbstractResourceGenerator.java#L151 when the other methods for signals where generated it was used a "counter" but basically to avoid name conflicts probably.

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/AbstractResourceGenerator.java
```diff
@@ -311,19 +317,18 @@ private void interpolateStrings(StringLiteralExpr vv) {
 
     private void interpolateUserTaskStrings(StringLiteralExpr vv, UserTaskModelMetaData userTask) {
         String s = vv.getValue();
-
-        String interpolated =
-                s.replace("$taskname$", userTask.getName().replaceAll("\\s", "_"));
+        String interpolated = s.replace("$taskname$", userTask.getName().replaceAll("\\s", "_"));
+        interpolated = interpolated.replace("$tasknodename$", userTask.getNodeName());
```
tiagodolphine, line 321: minor, but why not use camelCase for `tasknodename`

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/AbstractResourceGenerator.java
```diff
@@ -208,9 +210,15 @@ public String generate() {
                 });
 
                 template.findAll(StringLiteralExpr.class).forEach(s -> interpolateUserTaskStrings(s, userTask));
-
-                template.findAll(ClassOrInterfaceType.class).forEach(c -> interpolateUserTaskTypes(c, userTask.getInputMoodelClassSimpleName(), userTask.getOutputMoodelClassSimpleName()));
+                template.findAll(ClassOrInterfaceType.class).forEach(c -> interpolateUserTaskTypes(c, userTask.getInputModelClassSimpleName(), userTask.getOutputModelClassSimpleName()));
                 template.findAll(NameExpr.class).forEach(c -> interpolateUserTaskNameExp(c, userTask));
+                if(!userTask.isAdHoc()) {
```
tiagodolphine, line 215: so If I understood correctly method is always added by the template and it is being removed here in case of ad-hoc, right? 
Where the method name is generated based on the `"signal_" + userTask.getId()` ? on the template the method is just called `signal`, I missed this point

tiagodolphine:
@evacchi I added you in case you want to take a look on this PR since there are some changes to the endpoints code generation.

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/AbstractResourceGenerator.java
ruromero: following the same naming pattern

kogito-codegen/src/main/resources/class-templates/RestResourceUserTaskTemplate.java
ruromero: For tasks the nodeId is being used, not the taskName. In order to avoid conflicts between userTask signals and other signals I will reuse the existing index.

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/AbstractResourceGenerator.java
ruromero: I think it's better to add it by default and remove it if the node is not adHoc. That way you have an easier understanding of how the method would look like. This is also how it is done in other parts of the process template.

ruromero:
@tiagodolphine @cristianonicolai I have done the suggested changes and also fixed an issue regarding signaling in adHocSubprocesses

integration-tests/integration-tests-quarkus/src/test/java/org/kie/kogito/integrationtests/quarkus/AdHocFragmentsTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+        String taskPath = link.substring(link.indexOf("<") + 1, link.indexOf(">"));
+
+        params = new HashMap<>();
+        params.put("newVar1", "Gonzo");
+        given()
+                .contentType(ContentType.JSON)
+                .when()
+                .body(params)
+                .post("/AdHocFragments" + taskPath)
```
cristianonicolai, line 73:
```suggestion
                .post("/AdHocFragments/{path}", taskPath)
```

integration-tests/integration-tests-quarkus/src/test/java/org/kie/kogito/integrationtests/quarkus/FlexibleProcessTest.java
```diff
@@ -0,0 +1,56 @@
...
+        given()
+                .contentType(ContentType.JSON)
+                .when()
+                .body(params)
+                .post("/AdHocProcess")
+                .then()
+                .statusCode(200)
+                .body("id", not(emptyOrNullString()))
+                .body("var1", equalTo("Hello first! Script"))
+                .body("var2", equalTo("second Script 2"));
```
cristianonicolai, line 54:
```suggestion
        given()
                .contentType(ContentType.JSON)
        .when()
                .body(params)
                .post("/AdHocProcess")
        .then()
                .statusCode(200)
                .body("id", not(emptyOrNullString()))
                .body("var1", equalTo("Hello first! Script"))
                .body("var2", equalTo("second Script 2"));
```

integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-it/src/test/java/org/kie/kogito/integrationtests/springboot/AdHocFragmentsTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+        String taskPath = link.substring(link.indexOf("<") + 1, link.indexOf(">"));
+
+        params = new HashMap<>();
+        params.put("newVar1", "Gonzo");
+        given()
+                .contentType(ContentType.JSON)
+                .when()
+                .body(params)
+                .post("/AdHocFragments" + taskPath)
```
cristianonicolai, line 75:
```suggestion
                .post("/AdHocFragments/{path}", taskPath)
```

ruromero: It encoded the slashes and it didn't work as expected.

integration-tests/integration-tests-quarkus/src/test/java/org/kie/kogito/integrationtests/quarkus/AdHocFragmentsTest.java
ruromero: Updated with disabled encoding

integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-it/src/test/java/org/kie/kogito/integrationtests/springboot/AdHocFragmentsTest.java
ruromero: Same

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/AbstractResourceGenerator.java
```diff
@@ -208,9 +209,12 @@ public String generate() {
                 });
 
                 template.findAll(StringLiteralExpr.class).forEach(s -> interpolateUserTaskStrings(s, userTask));
-
-                template.findAll(ClassOrInterfaceType.class).forEach(c -> interpolateUserTaskTypes(c, userTask.getInputMoodelClassSimpleName(), userTask.getOutputMoodelClassSimpleName()));
+                template.findAll(ClassOrInterfaceType.class).forEach(c -> interpolateUserTaskTypes(c, userTask.getInputModelClassSimpleName(), userTask.getOutputModelClassSimpleName()));
                 template.findAll(NameExpr.class).forEach(c -> interpolateUserTaskNameExp(c, userTask));
+                template.findAll(MethodDeclaration.class)
+                        .stream()
+                        .filter(md -> md.getNameAsString().equals("signal_" + index.getAndIncrement()))
```
tiagodolphine: I still didn't get where is the place the method name was set with `"signal_" + index.getAndIncrement()`, because to filter here, to be removed from the template, you should have set this before, right? I didn't found where it is being set. 
By the way is there a specific test to check if the method was removed from the generated code?

ruromero: This filter was wrong and that's why it failed during the tests. The name of the task methods are now generated using the taskPath
I am updating the integration tests for this too.

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/AbstractResourceGenerator.java
tiagodolphine: :+1:

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
```diff
@@ -636,17 +638,19 @@ public void signalEvent(String type, Object event) {
                     for (Node node : getWorkflowProcess().getNodes()) {
                         if (type.equals(node.getName()) && node.getIncomingConnections().isEmpty()) {
                             NodeInstance nodeInstance = getNodeInstance(node);
+                            Map<String, Object> dynamicParams = new HashMap<>(getVariables());
                             if (event != null) {
-                                Map<String, Object> dynamicParams = new HashMap<>();
```
MarianMacik, line 640: So in the end this Map can be put back inside the if statement, right? As `dynamicParams` is not used outside of the if.

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/CompositeNodeInstance.java
```diff
@@ -272,22 +272,33 @@ public void signalEvent(String type, Object event) {
...
-						List<NodeInstance> nodeInstances = getNodeInstances(node.getId(), currentView);
-						if (nodeInstances != null && !nodeInstances.isEmpty()) {
-							for (NodeInstance nodeInstance : nodeInstances) {
-								((EventNodeInstanceInterface) nodeInstance)
-										.signalEvent(type, event);
-							}
-						}
-					}
-				}
+			if (node instanceof EventNodeInterface && ((EventNodeInterface) node).acceptsEvent(type, event)) {
```
MarianMacik, line 275: Maybe an additional formatting here would help when reading.

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/CompositeNodeInstance.java
```diff
@@ -272,22 +272,33 @@ public void signalEvent(String type, Object event) {
...
+                        for (NodeInstance nodeInstance : nodeInstances) {
+                            ((EventNodeInstanceInterface) nodeInstance)
+                                    .signalEvent(type, event);
+                        }
+                    }
+                }
 			}
+            if (type.equals(node.getName()) && node.getIncomingConnections().isEmpty()) {
+                NodeInstance nodeInstance = getNodeInstance(node);
+                Map<String, Object> dynamicParams = new HashMap<>(getProcessInstance().getVariables());
```
MarianMacik, line 291: Same here.

kogito-codegen/src/main/resources/class-templates/RestResourceSignalTemplate.java
```diff
@@ -28,18 +28,18 @@
     Process<$Type$> process;
 
     @POST
-    @Path("/{id}/$signalName$")
+    @Path("/{id}/$signalPath$")
```
MarianMacik, line 31: Is there any reason to introduce a new parameter? By looking at the AbstractResourceGenerator, signalPath and signalName are replaced with the same signalName value. So having here the same parameter might be less misleading.

kogito-codegen/src/main/resources/class-templates/RestResourceUserTaskTemplate.java
```diff
@@ -3,13 +3,36 @@
...
+            if (pi == null) {
+                return null;
+            }
+            pi.send(Sig.of("$taskNodeName$", java.util.Collections.emptyMap()));
+            java.util.Optional<WorkItem> task = pi.workItems().stream().filter(wi -> wi.getName().equals("$taskName$")).findFirst();
+            if(task.isPresent()) {
+                return javax.ws.rs.core.Response.ok(pi.variables())
+                        .header("Link", "</" + id + "/$taskName$/" + task.get().getId() + ">; rel='instance'")
+                        .build();
+            }
```
MarianMacik, line 29: Just a question from the performance point of view. Now I think in Business Central on 7.x, when we trigger a fragment we just send a signal and don't check if it has been triggered I think. We just return as soon as the signal has been sent successfully. This additional check may be not needed/wanted by some users maybe.

I am not saying it's bad, maybe we should just think also about the quicker option.

kogito-codegen/src/main/resources/class-templates/spring/SpringRestResourceUserTaskTemplate.java
```diff
@@ -13,7 +16,22 @@
...
+        return org.kie.kogito.services.uow.UnitOfWorkExecutor.executeInUnitOfWork(application.unitOfWorkManager(), () -> {
+            ProcessInstance<$Type$> pi = process.instances().findById(id).orElse(null);
+            if (pi == null) {
+                return null;
+            }
+            pi.send(Sig.of("$taskNodeName$", java.util.Collections.emptyMap()));
+            return org.springframework.http.ResponseEntity.ok()
+                    .header("Link", "</" + id + "/$taskName$/" + task.get().getId() + ">; rel='instance'")
+                    .body(pi.variables());
+        });
```
MarianMacik, line 31: Here the additional check is missing. Is it OK?

kogito-codegen/src/main/resources/class-templates/RestResourceSignalTemplate.java
ruromero: SignalPath is the sanitized name e.g. `Service_Task` whereas signalName is the name of the event that will match the node name e.g. `Service Task`

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
ruromero: yep, thanks

kogito-codegen/src/main/resources/class-templates/RestResourceUserTaskTemplate.java
ruromero: The type of userTasks that can receive a signal are the ones that are adHocFragments meaning that are not activated by the previous node or automatically.
I consider these kind of tasks are most likely to be triggered by web applications that might like to use the created task so actually it will save the external service/webapp from requesting all the tasks and filter out the one that was just created.

An example can be an "Add review" button that creates the task for a buyer to add a review but the application transparently renders the form with the link returned in the previous request.

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/CompositeNodeInstance.java
MarianMacik: Formatting is still a little bit of.

kogito-codegen/src/main/resources/class-templates/RestResourceSignalTemplate.java
MarianMacik: I see, thanks!

kogito-codegen/src/main/resources/class-templates/RestResourceUserTaskTemplate.java
MarianMacik: OK, I see now. I didn't notice that you actually use the found task.

kogito-codegen/src/main/resources/class-templates/spring/SpringRestResourceUserTaskTemplate.java
MarianMacik: Check added to be in sync with `RestResourceUserTaskTemplate.java`.

kogito-codegen/src/main/resources/class-templates/spring/SpringRestResourceSignalTemplate.java
```diff
@@ -30,16 +30,15 @@
...
-        pi.send(Sig.of("$signalName$", data));
-
-        return pi.variables();
+        return org.kie.kogito.services.uow.UnitOfWorkExecutor.executeInUnitOfWork(application.unitOfWorkManager(), () -> {
+            ProcessInstance<$Type$> pi = process.instances().findById(id).orElse(null);
+            if (pi == null) {
+                return null;
+            }
+            pi.send(Sig.of("$signalName$", data));
+            return pi.variables();
```
mswiderski, line 41: And one more wrap with `getModel`

PR #598, opened 2020-06-30
[KOGITO-2449] Generate endpoints for json schema

fjtirado:
Two additional endpoints has been added to retrieve schema and schema plus allowed phases. 
/<processId>/<taskName>/schema returns json schema (which also included input/output metadata information)
/<processId>/<processInstanceId>/<taskName>/<taskInstanceId>/schema return previous json schema plus allowed phases for that task intances. In order to retrieve that information, a new method has been added to ProcessIntance interface


fjtirado:
jenkins retest this

mswiderski:
@fjtirado wouldn’t it make more sense to have these endpoints in separate addon that can be easily added/removed. Plus in case extention to the methods are required can be managed eadier i believe

fjtirado:
> @fjtirado wouldn’t it make more sense to have these endpoints in separate addon that can be easily added/removed. Plus in case extention to the methods are required can be managed eadier i believe

Good point for debate. I considered forms endpoint as integral part of task handling (tighly coupled with them), hence modified the current REST template for tasks. 
Besides, these endpoints returned pure json schema (plus information that is part of core cogito, like input/output and phases) they do not preclude any specific ui information, so I do not foresee any usage scenario where there is need to disable them. 
On the other hand, there could be certainly be some scenarios where these endpoint might not be used at all (for example in a custom user interface where forms are statically defined and hece the metadata information is implicit in the custom ui code). 
But, since there is not extra cost during generation in generating them, there is no extra performance cost in deploying them if they are not invoked and since we need the modification in the core to retrieve allowed phases anway, I vote for simplicity here and leave the two endpoints in the task REST template rather than writting  a new add-on.

fjtirado:
Added some integration test to increase coverage

api/kogito-api/pom.xml
```diff
@@ -30,7 +30,12 @@
       <artifactId>jaxb-api</artifactId>
       <scope>provided</scope>
     </dependency>
-    
+
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+    </dependency>
```
evacchi, line 37: why has this been added to kogito-api?

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
```diff
@@ -130,6 +130,13 @@
      * @throws WorkItemNotFoundException in case work item with given id does not exist
      */
     void transitionWorkItem(String id, Transition<?> transition);
+    
+    /**
+     * Identify phases that are available to be transition to for a particular work item
+     * @param id work item 
+     * @return id of the phases that can be transitioned to
+     */
+    Set<String> allowedPhases (String id);
```
evacchi, line 139: I am not really fond of adding this public method to ProcessInstance, as it is a property of the WorkItem. Shouldn't this be a method on WorkItemManager maybe?

kogito-codegen/src/test/java/org/kie/kogito/codegen/JsonSchemaGeneratorTest.java
```diff
@@ -40,7 +40,7 @@
 
     private enum Color {GREEN, WHITE};
     
-    @UserTask(taskName = "test", processName = "test")
+    @UserTask(taskName = "test", processName = "org.jbpm.test")
```
evacchi, line 43: are we consistent with the way process "id", process "package", process "name" are used? IIRC IDs are unqualified, while we have package+name otherwise; not 100% sure but seeing the package here feels inconsistent @cristianonicolai @tiagobento ?

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
fjtirado: This is also defined in WorkItermManager, but as per other method of this interface (like completeWorkItem or aborWorkItem), it is also exposed in ProcessIntance to allow being called from the generated code. 
I mean, Im following the pre existing policy here ;)
Probably it will ill be better to create a getWorkItemManager in process instance and do not have to duplicate methods in both interface. 
I can do that, but we need to discuss it first

kogito-codegen/src/test/java/org/kie/kogito/codegen/JsonSchemaGeneratorTest.java
fjtirado: I feel including package name will guarantee the uniqueneness of process name.

api/kogito-api/pom.xml
fjtirado: some jackson annotations are used in JsonSchema.  
I put JsonSchema in API for mainly two reasons:
1) It is the object returned by the generated rest method, so in a sense is part of the API
2) I need a common placeholder which is a dependency both for code gen and for runtime (I mean, there are some pieces of code in JsonSchema that are both executed when the REST method is generated and when is invoked) 

An alternative implementation is to remove jsonschema,  make the generated REST method to return Map<String,Object) but then we need to do two extra things: 
1) Documment somehow in the swagger api that there is an extra "phases" field returned by the REST API
2) Find a common placeholder for the code that is both executed during generation and runtime (which will be that place?)

fjtirado:
jenkins retest this

api/kogito-api/pom.xml
fjtirado: I changed JsonSchema by Map<String,Object>, kogito-api is hence not changed and the common code was moved to jbpm-flow

kogito-codegen/src/main/resources/class-templates/RestResourceUserTaskTemplate.java
```diff
@@ -75,7 +75,23 @@
...
+    @Produces(MediaType.APPLICATION_JSON)
+    public Map<String,Object> getSchema() {
+        return JsonSchemaUtil.load(this.getClass().getClassLoader(),process.id(),"$taskName$");
+    }
     
+    @GET()
+    @Path("/{id}/$taskName$/{workItemId}/form")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Map<String,Object> getSchemaAndPhases(@PathParam("id") final String id, @PathParam("workItemId") final String workItemId) {
+        Map<String,Object> jsonSchema = JsonSchemaUtil.load(this.getClass().getClassLoader(),process.id(),"$taskName$");
```
tiagodolphine, line 90: It seems there is too much logic for Rest resource class, wouldn't worth extracting this logic to another class, in this way it could be reused in both springboot and quarkus, anyway just a suggestion.

pom.xml
```diff
@@ -94,6 +94,7 @@
     <version.com.github.haifengl.smile>1.5.2</version.com.github.haifengl.smile>
     <version.com.github.javaparser>3.13.10</version.com.github.javaparser>
     <version.com.github.victools>4.12.1</version.com.github.victools>
+    <version.io.rest-assured>4.3.1</version.io.rest-assured>
```
tiagodolphine, line 97: we were already using the rest-assured in many places, I think you can reuse the same version, check https://github.com/kiegroup/kogito-runtimes/blob/bf14ef847ad15339badbd9c2a648ba5046f7486a/pom.xml#L121

kogito-codegen/src/main/resources/class-templates/RestResourceUserTaskTemplate.java
```diff
@@ -75,7 +75,23 @@
             return null;
         }
     }
+
+    @GET()
+    @Path("$taskName$/form")
```
tiagodolphine, line 80: What about using a different name for the resource path instead of `form`, I mean something like `schema`, or any more generic one, the term **form** recalls me UI, I think a more generic term would be better, WDYT?

kogito-codegen/src/main/resources/class-templates/RestResourceUserTaskTemplate.java
fjtirado: Hmmm, most logic is included in JsonSchemaUtils, here we are just specifying paramas anc calling processIntance to retrieve the allowed phases (and includind there in the map), as it is done in other methods of the template, like aborTask. 
Initially, (see my comment to Edoardo in this PR) I have included a jsonschema class in the api, but it does not reduce substantially the login here (we still need to resolve processintance and call allowedPhases), so I remove it since api project follow a differnet inteface-base approach with no dependency with jakson. 
I think the logic that is in the template is the logic that is needed (there has to be some copy paste between this and spring  template anyway)

pom.xml
fjtirado: Thanks, I have not seen it

integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-it/src/test/java/org/kie/kogito/integrationtests/springboot/JsonSchemaTest.java
```diff
@@ -0,0 +1,42 @@
...
+import io.restassured.http.ContentType;
+import org.junit.jupiter.api.Test;
+
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.annotation.DirtiesContext;
+
+import static io.restassured.RestAssured.given;
+import static io.restassured.module.jsv.JsonSchemaValidator.matchesJsonSchemaInClasspath;
+
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT, classes = KogitoSpringbootApplication.class)
```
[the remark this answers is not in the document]
fjtirado: Changed

kogito-codegen/src/main/resources/class-templates/RestResourceUserTaskTemplate.java
fjtirado: hmmm, I initially proposed schema, but Pere believe it was better to use form. 
Whatever you decide is fine for me ; )

mswiderski: then to make both happy use `@Path("$taskName$/formschema")` :)

fjtirado: I like this salomonic solution ;), so if nobody speak up in the next 24 hours Ill use that ;)

pefernan: @fjtirado @tiagodolphine If you ask me, `form` name is my personal preference because it's going to be the main use in my case, but I agree that it represents the task schema, so I'll be ok with whatever decision we make. Also @mswiderski option works for me :+1:

tiagodolphine: I was just questioning it because from the REST perspective the resource name on the path should represent what is returned/created etc... for when I look at `/form` it seems that a form will be returned, and in fact what is retuned is a schema.

fjtirado: Thiago is right, from rest point of view this is the schema of the task, the fact that is used for the form should not be part of the resource name. 
Thinking about this carefully, my preferred name wil be actually metadata, because what we are really returning is metadata information about a particular task. The fact that we are using json schema to represent the metadata should not be part of the REST name either. 
Since Im kind of new, what should be do when there is not agreement? we should reach one or author decides ?;)

pefernan: @fjtirado to me is ok either name you use, I think that calling it `forms` looks like a name too focused to UI, so changing it to `schema` seems a better option for me.

fjtirado: Ok, lets use schema

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
evacchi: ok makes sense; in fact, I am not fond of those other methods either -- there was no "policy" in place when they were added hehe. So let's keep this around, but let's plan another refactoring maybe

kogito-codegen/src/test/java/org/kie/kogito/codegen/JsonSchemaGeneratorTest.java
evacchi: ok I was trying to make sure we're consistent across the other usages of process ids/names. Let's keep it for now

kogito-codegen/src/main/resources/class-templates/RestResourceUserTaskTemplate.java
```diff
@@ -75,7 +75,23 @@
...
+    public Map<String,Object> getSchema() {
+        return JsonSchemaUtil.load(this.getClass().getClassLoader(),process.id(),"$taskName$");
+    }
     
+    @GET()
+    @Path("/{id}/$taskName$/{workItemId}/schema")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Map<String,Object> getSchemaAndPhases(@PathParam("id") final String id, @PathParam("workItemId") final String workItemId) {
+        Map<String,Object> jsonSchema = JsonSchemaUtil.load(this.getClass().getClassLoader(),process.id(),"$taskName$");
+        process.instances().findById(id).ifPresent(pi ->jsonSchema.put("phases",pi.allowedPhases(workItemId)));
```
cristianonicolai, line 91: @fjtirado instead of doing allowed phases via process instances. You could have access or inject here the `ApplicationConfig`, from there you can get `ProcessConfig.workItemHandlers` and retrieve the WorkItemHandler for human tasks. That way, `allowedPhases` is only added to `WorkItemHandler` wdyt?

kogito-codegen/src/test/java/org/kie/kogito/codegen/JsonSchemaGeneratorTest.java
cristianonicolai: name and package are different properties for a process. It might be valid to cover that the generation won't break because "." are included but I dont think this is a common use.

kogito-codegen/src/test/java/org/kie/kogito/codegen/tests/UserTaskTest.java
```diff
@@ -135,8 +136,12 @@ public void testBasicUserTaskProcessPhases() throws Exception {
         assertEquals(Active.ID, wi.getPhase());
         assertEquals(Active.STATUS, wi.getPhaseStatus());
 
-        processInstance.transitionWorkItem(workItems.get(0).getId(), new HumanTaskTransition(Complete.ID, null, securityPolicy));
+        String workItemId = workItems.get(0).getId();
+        assertFalse(processInstance.allowedPhases(workItemId).isEmpty());
+        processInstance.transitionWorkItem(workItemId, new HumanTaskTransition(Complete.ID, null, securityPolicy));
         assertThat(processInstance.status()).isEqualTo(ProcessInstance.STATE_ACTIVE);
+
```
cristianonicolai, line 143: remove?

fjtirado: hmmm, this check was there for the transitionworkitem, unless wrongly specified initially, I think it should stay.

kogito-codegen/src/main/java/org/kie/kogito/codegen/JsonSchemaGenerator.java
```diff
@@ -107,15 +111,25 @@ private JsonSchemaGenerator(Stream<Class<?>> stream) {
...
                 writer.writeValue(outputStream, merged);
-                files.add(new GeneratedFile(Type.JSON_SCHEMA, entry.getKey() + ".json", outputStream.toByteArray()));
+                files.add(new GeneratedFile(Type.JSON_SCHEMA, JsonSchemaUtil.getFileName(entry.getKey()), outputStream.toByteArray()));
             }
         }
         return files;
     }
+    
+    private CustomPropertyDefinition getInputOutput (FieldScope f, SchemaGenerationContext context)
+    {
```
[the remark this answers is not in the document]
fjtirado: File formatted properly now, thanks

jbpm/jbpm-flow/src/test/java/org/jbpm/util/JsonSchemaUtilTest.java
```diff
@@ -0,0 +1,72 @@
...
+            "        },\n" + 
+            "        \"approved\": {\n" + 
+            "            \"type\": \"boolean\",\n" + 
+            "            \"output\": true\n" + 
+            "        }\n" + 
+            "    }}";
+    
+    @Test
+    void testJsonSchema() throws IOException
+    {
```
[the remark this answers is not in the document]
fjtirado: Done

jbpm/jbpm-flow/src/main/java/org/jbpm/util/JsonSchemaUtil.java
```diff
@@ -0,0 +1,47 @@
+package org.jbpm.util;
```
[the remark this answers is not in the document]
fjtirado: uppsss ;)

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
fjtirado: fair enough, we need to revisit all of these methods

evacchi:
I am merging this first version, then we can further refine in other PRs!

PR #619, opened 2020-07-09
KOGITO-2207 Review AbstractProcessInstance to use WorkflowProcessInst…

cristianonicolai:
…ance interface avoiding multiples casts

Small refactor with a series of improvements to avoid multiple casts in AbstractProcessInstance. As it depends on WorkflowProcessInstance anyway. Moved some method to interface to also avoid dpendency on Impl class.

### DIFF
--- a/jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessInstance.java
+++ b/jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessInstance.java
@@ -44,14 +44,14 @@ public interface ProcessInstance extends org.kie.api.runtime.process.ProcessInst
     void setState(int state, String outcome, Object faultData);
 
     void setErrorState(NodeInstance nodeInstanceInError, Exception e);
-
+    
     void setKnowledgeRuntime(InternalKnowledgeRuntime kruntime);
 
     InternalKnowledgeRuntime getKnowledgeRuntime();
 
     void start();
 
-    void start(String tigger);
+    void start(String trigger);
 
     String getOutcome();
 
@@ -60,6 +60,8 @@ public interface ProcessInstance extends org.kie.api.runtime.process.ProcessInst
     void setRootProcessInstanceId(String parentId);
 
     Map<String, Object> getMetaData();
+
+    void setMetaData(String name, Object data);
 
     Object getFaultData();
 
@@ -82,4 +84,8 @@ public interface ProcessInstance extends org.kie.api.runtime.process.ProcessInst
     void configureSLA();
 
     void setReferenceId(String referenceId);
+
+    void disconnect();
+
+    void reconnect();
 }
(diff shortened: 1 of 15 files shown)


jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessInstance.java
```diff
@@ -44,14 +44,14 @@
...
-
+    
     void setKnowledgeRuntime(InternalKnowledgeRuntime kruntime);
 
     InternalKnowledgeRuntime getKnowledgeRuntime();
 
     void start();
 
-    void start(String tigger);
+    void start(String trigger);
```
ruromero, line 54: I prefer `tiger` :rofl:

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/NodeInstanceContainer.java
```diff
@@ -35,21 +37,39 @@
...
-    
+
     int getCurrentLevel();
-    
+
+    void setCurrentLevel(int level);
+
     NodeInstance getNodeInstance(String nodeInstanceId, boolean recursive);
 
+    default NodeInstance getNodeInstanceByNodeDefinitionId(final String nodeDefinitionId, NodeContainer nodeContainer) {
```
ruromero, line 57: why not just `getByDefinitionId` as it is the `NodeInstanceContainer` it can be assumed that the `get` will return a `NodeInstance`
The same could apply to the method above and call it just `get`

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
```diff
@@ -380,6 +359,10 @@ public void setState(final int state, String outcome, Object faultData) {
     @Override
     public void setState(final int state, String outcome) {
         // TODO move most of this to ProcessInstanceImpl
+        if (state != ProcessInstance.STATE_ERROR) {
```
ruromero, line 362: Why is this needed? Can you add a unit test to support it?

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/NodeInstanceContainer.java
cristianonicolai: @ruromero renamed the new method to `getByDefinitionId` as suggested, but keeping old methods as is to avoid more unrelated changes.

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
cristianonicolai: rollback change to not cause any behaviour change.

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessInstance.java
mswiderski: same here tigger (double g to make it stronger) makes a lot of sense :)

PR #623, opened 2020-07-10
KOGITO-2730 Refactoring AbstractCodegenTest

gitgabrio:
@evacchi @danielezonca @mariofusco @tarilabs 

See https://issues.redhat.com/browse/KOGITO-2730

Scope of the PR is to 

1. limit as much as possible needed changes when new type of resource are added inside AbstractCodeGenTest
2. simplify/generify methods, to avoid long list of parameters and invocations with empty/null parameters

Modifications done

1. definition of TYPE enum with managed types (process, rules, decision, ....)
2. Map<TYPE, Function<Codegen> definition to map a given TYPE with its specific codegen invocation
3. replaced AbstractCodeGenTest.generateCode parameter to receive a Map<TYPE, List of resources>)
4. inside the above, iterating over the ordered sequence of TYPES (important!!) and if eventually manage the reources of the type if presents in the retrieved map


kogito-codegen/src/main/java/org/kie/kogito/codegen/decision/DecisionCodegen.java
```diff
@@ -104,19 +104,19 @@ public static DecisionCodegen ofPath(Path... paths) throws IOException {
         return ofDecisions(resources);
     }
 
-    public static DecisionCodegen ofFiles(Path basePath, List<File> files) throws IOException {
```
tarilabs, line 107: If you want to remove it, for me it's fine, but my previous understanding is that this was "by design" of codegenerators? possibly @evacchi and @mariofusco can better advise here.

evacchi: if it never throws, I guess it's ok to drop that. I don't have a strong opinion there :)

kogito-codegen/src/test/java/org/kie/kogito/codegen/AbstractCodegenTest.java
```diff
@@ -52,10 +52,45 @@
...
     private static final Logger logger = LoggerFactory.getLogger(AbstractCodegenTest.class);
-    
+
+    protected enum TYPE {
+        PROCESS,
+        RULES,
+        DECISION,
+        JAVA,
+        PREDICTION
+    }
```
danielezonca, line 62: As you mentioned, order is important so can you please add a comment here and also (if possible) a link to the code where this order is defined?

kogito-codegen/src/main/java/org/kie/kogito/codegen/decision/DecisionCodegen.java
gitgabrio: @evacchi @tarilabs 
I think the exception could be dropped.

kogito-codegen/src/test/java/org/kie/kogito/codegen/AbstractCodegenTest.java
gitgabrio: @danielezonca 
done

gitgabrio:
@danielezonca @tarilabs @mariofusco 
Could you please look at this again and eventually approve/merge it? Thanks!

PR #624, opened 2020-07-10
KOGITO-2726: added kogito-build-parent and refactored bom into kogito-bom

mareknovotny:
https://issues.redhat.com/browse/KOGITO-2726

This is dividing the kogito-bom as end user BOM from build requirements in parent

Merge together with https://github.com/kiegroup/kogito-apps/pull/378


pom.xml
```diff
@@ -245,568 +104,11 @@
...
-      
-    </dependencies>
-  </dependencyManagement>
-
   <profiles>
     <profile>
       <id>snapshot</id>
       <properties>
-        <version.org.kie7>7.33.0-SNAPSHOT</version.org.kie7>
+        <version.org.kie7>7.41.0-SNAPSHOT</version.org.kie7>
```
danielezonca, line 111: This should not be used anymore ( https://github.com/kiegroup/kogito-runtimes/pull/627 )

pom.xml
```diff
@@ -4,28 +4,19 @@
 
   <modelVersion>4.0.0</modelVersion>
 
-  <groupId>org.kie.kogito</groupId>
+  <parent>
+    <groupId>org.kie.kogito</groupId>
+    <artifactId>build-bom</artifactId>
+    <version>8.0.0-SNAPSHOT</version>
+    <relativePath>build-bom/pom.xml</relativePath>
```
danielezonca, line 11: I don't like to have a pom inheriting from a child module. I think it could be better to make all modules directly inherit from  `build-bom` instead of `kogito-runtimes`. Wdyt?

mareknovotny: Could you explain the reason for not liking it?

danielezonca: I think it is misleading to have a parent module that is not coherent with module hierarchy expecially when you have to add/remove properties or dependency.
But probably it is mainly because I find strange that `build-bom` is both _my parent_  and _my child_ :)

mareknovotny: well, this is not something new or too uncommon to maven project/world.
And build-bom is not actually child as it doesn't inherit the top level pom.xml just it is executed/called out from it.

pom.xml
```diff
@@ -216,21 +87,9 @@
...
-  <developers>
-    <developer>
-      <name>All developers are listed on the team website</name>
-      <url>http://kogito.kie.org/community/team.html</url>
-    </developer>
-  </developers>
-
   <modules>
-    <module>bom</module>
+    <module>build-bom</module>
```
mareknovotny, line 91: it is good practice to have the same artifact Id synced up with filesystem until there is a real aim to have this different.

pom.xml
```diff
@@ -216,21 +87,9 @@
...
+    <module>kogito-bom</module>
```
mareknovotny, line 92: I synced up the kogito-bom also with filesystem as it is better to have it the same until there is some special need to name it differently.

pom.xml
danielezonca: Yes with child I meant child from a "file system" perspective.
Personally I prefer Quarkus (and OptaPlanner) setup that is a bit different:
- [build-parent](https://github.com/quarkusio/quarkus/blob/master/build-parent/pom.xml#L5-L9): inherit from root pom
- [root pom](https://github.com/quarkusio/quarkus/blob/master/pom.xml#L6-L10): it is inheriting from jboss-parent while we want to be agnostic so no parent here
- all other modules (i.e. [core](https://github.com/quarkusio/quarkus/blob/master/core/pom.xml#L5-L10): inherit directly or not from `build-parent`

Btw no strong opinion and thanks for the PR because it is a step forward with project configuration :)

mareknovotny: > Yes with child I meant child from a "file system" perspective.
> Personally I prefer Quarkus (and OptaPlanner) setup that is a bit different:
> 
>     * [build-parent](https://github.com/quarkusio/quarkus/blob/master/build-parent/pom.xml#L5-L9): inherit from root pom
> 
>     * [root pom](https://github.com/quarkusio/quarkus/blob/master/pom.xml#L6-L10): it is inheriting from jboss-parent while we want to be agnostic so no parent here

Do we really keep this not inheriting? I think this will be required for later, so I can bring this here too, thanks for pointing it up

> 
>     * all other modules (i.e. [core](https://github.com/quarkusio/quarkus/blob/master/core/pom.xml#L5-L10)): inherit directly or not from `build-parent`
> 
maybe they are not inheriting directly, but this is probably for a reason, it is not part of the build chain, just built as a leaf artifact in the repository?
> 
> Btw no strong opinion and thanks for the PR because it is a step forward with project configuration :)

That is fine to object with explaining why ;)

mareknovotny:
it seems the CI relies a lot on remote repository instead of the local built artifacts

build-bom/pom.xml
```diff
@@ -0,0 +1,1239 @@
...
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.kie.kogito</groupId>
+    <artifactId>kogito-runtimes</artifactId>
+    <version>8.0.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>build-parent</artifactId>
```
cristianonicolai, line 13: perhaps _kogito-parent-bom_ or _kogito-build-bom_ ?

mareknovotny: if we avoid BOM term as that now cotains not only depMgmt then it should be build-parent, thinking about it's not publicly consumable we can omit kogito suffix there, but that is just keep it simple. I am not prefering it one or another, just the artifactId should be the same with directory name.
So I would like to change it on top of your comment to:

- kogito-build-parent
or
- build-parent

WDYT? @cristianonicolai @danielezonca @ge0ffrey

cristianonicolai: +1 to keep consistency between folder name and artifactId.

I would go for `kogito-build-parent`.

danielezonca: I have no strong opinion, fine for me `kogito-build-parent` and +1 to keep folder and artifact name in sync

pom.xml
```diff
@@ -2,8 +2,15 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
 
-  <modelVersion>4.0.0</modelVersion>
 
+  <parent>
+    <groupId>org.jboss</groupId>
+    <artifactId>jboss-parent</artifactId>
```
danielezonca, line 8: What is the reason of this change? Do we expect any benefit using `jboss-parent` as parent?

build-bom/pom.xml
ge0ffrey: I agree

pom.xml
ge0ffrey: The jboss-parent sets all plugin versions. All jboss projects using it, including kie 7, optaplanner, quarkus, wildfly, ...
Forgetting to set one of the plugin versions causes non-reproducible builds.
Settings them ourselves is a PITA to keep them up to date. The jboss-parent guys do a pretty good job of keeping them up to date soon enough and without upgrading them too fast either.

danielezonca: Ok thanks for the clarification :)

mareknovotny:
I think I included all comments about renaming or keeping the same naming for folder and artifactId for new kogito-bom and kogito-build-parent.
Right now biggest work is done, even I would like to polish later also ordering of version properties and some of shared version properties for plugins with dependencies. This would be a follow up task as every change now into parent/bom requires careful rebase I would like to merge this ASAP when it is reviewed and ok.

mareknovotny:
4 tests failed:
 
-     it.pkg.GreetingsTest.testHelloEndpoint
-     org.kie.kogito.integrationtests.quarkus.SignalProcessTest.testProcessSignals
-     org.jbpm.bpmn2.IntermediateEventTest.testTimerMultipleInstances
-     io.quarkus.it.kogito.drools.DroolsTest.testAdult

Any idea why?

mareknovotny:
Jenkins retest this

mareknovotny:
Jenkins retest this

mareknovotny:
Jenkins retest this please

mareknovotny:
Jenkins retest this please

kogito-build-parent/pom.xml
```diff
@@ -0,0 +1,1237 @@
...
+    <version.source.plugin>3.0.1</version.source.plugin>
+    <version.surefire.plugin>2.22.2</version.surefire.plugin> <!-- minimum required by JUnit 5 -->
+    <version.surefire.report.plugin>2.6</version.surefire.report.plugin>
+    <version.swagger.plugin>2.1.0</version.swagger.plugin>
+    <version.taglist.plugin>2.4</version.taglist.plugin>
+    <version.versions.plugin>2.1</version.versions.plugin>
+    <version.jayway.jsonpath>2.4.0</version.jayway.jsonpath>
+
+    <!-- this version property is used in plugins but also in dependencies too -->
+    <version.io.quarkus>1.6.0.Final</version.io.quarkus>
```
evacchi, line 65: please make sure these are all aligned with the changes in this diff 

https://github.com/kiegroup/kogito-runtimes/compare/4204fd29053e2d74948898532c3196af7497bf86..master#diff-600376dffeb79835ede4a0b285078036

evacchi:
@mareknovotny sent you a PR, see if that's good enough https://github.com/mareknovotny/kogito-runtimes/pull/1

kogito-build-parent/pom.xml
mareknovotny: yes, master changes too fast for me

evacchi:
@mareknovotny still red :/

mareknovotny:
Jenkins retest this please

PR #630, opened 2020-07-13
KOGITO-2737 - refactoring injection of addons configuration

r00ta:
Hi,

I'd like to propose the following refactoring of the injection of the addons configurations. Since atm the addons are injected one by one with a simple boolean, I thought that using an object would make it much more easier to extend and maintain. 

Just as a side note: I'm working on a component that needs to know if the `tracing` and the `monitoring` addons are activated, and I would have had to inject both of them with some redundant code for the reason above. 

Let me know what do you think :) 

### DIFF
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/ApplicationGenerator.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/ApplicationGenerator.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.stream.Collectors;
+
 import javax.lang.model.SourceVersion;
 
 import com.github.javaparser.ast.CompilationUnit;
@@ -72,7 +73,6 @@ public class ApplicationGenerator {
     private List<Labeler> labelers = new ArrayList<>();
 
     private GeneratorContext context;
-    private boolean persistence;       
 
     public ApplicationGenerator(String packageName, File targetDirectory) {
         if (packageName == null) {
@@ -192,17 +192,12 @@ public class ApplicationGenerator {
         return this;
     }
 
-   public ApplicationGenerator withPersistence(boolean persistence) {
-       this.persistence = persistence;
-       return this;
-   }
-
-   public ApplicationGenerator withMonitoring(boolean monitoring) {
-       if (monitoring) {
-           this.labelers.add(new PrometheusLabeler());
-       }
-       return this;
-   }
+    public ApplicationGenerator withAddons(AddonsConfig addonsConfig) {
+        if (addonsConfig.useMonitoring()) {
+            this.labelers.add(new PrometheusLabeler());
+        }
+        return this;
+    }
 
     public Collection<GeneratedFile> generate() {
         List<GeneratedFile> generatedFiles = generateComponents();
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/ApplicationGenerator.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/ApplicationGenerator.java
@@ -20,7 +20,9 @@ import java.nio.charset.StandardCharsets;
 import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.stream.Collectors;
@@ -70,7 +72,7 @@ public class ApplicationGenerator {
     private final List<BodyDeclaration<?>> factoryMethods;
     private ConfigGenerator configGenerator;
     private List<Generator> generators = new ArrayList<>();
-    private List<Labeler> labelers = new ArrayList<>();
+    private Map<Class, Labeler> labelers = new HashMap<>();
 
     private GeneratorContext context;
 
@@ -194,7 +196,7 @@ public class ApplicationGenerator {
 
     public ApplicationGenerator withAddons(AddonsConfig addonsConfig) {
         if (addonsConfig.useMonitoring()) {
-            this.labelers.add(new PrometheusLabeler());
+            this.labelers.put(PrometheusLabeler.class, new PrometheusLabeler());
         }
         return this;
     }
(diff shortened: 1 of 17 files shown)


kogito-codegen/src/main/java/org/kie/kogito/codegen/ApplicationGenerator.java
```diff
@@ -192,17 +192,12 @@ public ApplicationGenerator withRuleUnits(boolean hasRuleUnits) {
...
-   public ApplicationGenerator withMonitoring(boolean monitoring) {
-       if (monitoring) {
-           this.labelers.add(new PrometheusLabeler());
-       }
-       return this;
-   }
+    public ApplicationGenerator withAddons(AddonsConfig addonsConfig) {
+        if (addonsConfig.useMonitoring()) {
+            this.labelers.add(new PrometheusLabeler());
+        }
```
danielezonca, line 198: Side note because it is not related to this PR. This code is quite error prone, if `withAddons` is invoked multiple times, multiple `PrometheusLabeler`s are created

kogito-codegen/src/main/java/org/kie/kogito/codegen/decision/DMNRestResourceGenerator.java
```diff
@@ -56,7 +57,7 @@
     private final String resourceClazzName;
     private final String appCanonicalName;
     private DependencyInjectionAnnotator annotator;
-    private boolean useMonitoring;
+    private AddonsConfig addonsConfig = new AddonsConfig();
```
danielezonca, line 60: What about create a `DEFAULT` instance as public static final field in `AddonsConfig` to use as default instance in the codebase? In this way we can also define a different default value in a single place

kogito-codegen/src/main/java/org/kie/kogito/codegen/decision/DecisionCodegen.java
```diff
@@ -136,7 +136,7 @@ private static DecisionCodegen ofDecisions(List<DMNResource> resources) {
 
     private final List<DMNResource> resources;
     private final List<GeneratedFile> generatedFiles = new ArrayList<>();
-    private boolean useMonitoring = false;
+    private AddonsConfig addonsConfig = new AddonsConfig();
```
danielezonca, line 139: Same comment about use a common default instance

kogito-codegen/src/main/java/org/kie/kogito/codegen/decision/DecisionContainerGenerator.java
```diff
@@ -55,16 +56,16 @@
 
     private String applicationCanonicalName;
     private final List<DMNResource> resources;
-    private boolean useTracing = false;
+    private AddonsConfig addonsConfig = new AddonsConfig();
```
danielezonca, line 59: Same as above

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
```diff
@@ -204,7 +205,7 @@ private static Process parseWorkflowFile(Resource r, String parser) {
     private final Map<String, WorkflowProcess> processes;
     private final List<GeneratedFile> generatedFiles = new ArrayList<>();
 
-    private boolean persistence;
+    private AddonsConfig addonsConfig = new AddonsConfig();
```
danielezonca, line 208: Same as above

kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/IncrementalRuleCodegen.java
```diff
@@ -184,7 +184,7 @@ private static ResourceType typeOf(FileSystemResource r) {
 
     private KieModuleModel kieModuleModel;
     private boolean hotReloadMode = false;
-    private boolean useMonitoring = false;
+    private AddonsConfig addonsConfig = new AddonsConfig();
```
danielezonca, line 187: Same as above

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessGenerator.java
```diff
@@ -83,7 +84,7 @@
     private final String appCanonicalName;
     private String targetTypeName;
     private DependencyInjectionAnnotator annotator;
-    private boolean persistence;
+    private AddonsConfig addonsConfig = new AddonsConfig();
```
danielezonca, line 87: Same as above

kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/RuleUnitGenerator.java
```diff
@@ -56,7 +56,7 @@
     private DependencyInjectionAnnotator annotator;
     private Collection<QueryModel> queries;
     private String applicationPackageName;
-    private boolean useMonitoring;
+    private AddonsConfig addonsConfig = new AddonsConfig();
```
danielezonca, line 59: Same as above

kogito-codegen/src/main/java/org/kie/kogito/codegen/decision/DMNRestResourceGenerator.java
r00ta: Sounds good to me

r00ta: Done

kogito-codegen/src/main/java/org/kie/kogito/codegen/ApplicationGenerator.java
r00ta: I agree with you @danielezonca . I've changed it to a `Map<Class, Labeler>` to mitigate it, let me know wdyt

evacchi:
paging @cristianonicolai to take a look

evacchi:
I don't think there's much to be tested here @jiripetrlik ; this is basically a refactoring; the `AddonsConfig` class is just a "carrier" class with no logic, if the CI is green the tests have already covered all expected behavior. @r00ta WDYT?

r00ta:
> I don't think there's much to be tested here @jiripetrlik ; this is basically a refactoring; the `AddonsConfig` class is just a "carrier" class with no logic, if the CI is green the tests have already covered all expected behavior. @r00ta WDYT?

Yes, I agree

jiripetrlik:
@evacchi 
Yes, you are right that it is a carrier. However, I would suggest to add some basic unit tests which check at least whether "withSomething" methods works correctly. It will keep or increase code coverage in this case.

r00ta:
@jiripetrlik I've added a couple of tests for the `AddonsConfig` class, checking that all the addons are disabled by default and that the values are properly changed using `withXYZ`. 
For the `ApplicationGenerator` class, I think we have already covered the cases as @evacchi pointed out. 
lmkwyt

PR #631, opened 2020-07-13
[KOGITO-2736] Removing allowedPhases from ProcessInstance

fjtirado:
Creating as Draft since this requires user and group to be passed in the API (so we should not commit till this information is available at client side

### DIFF
--- a/api/kogito-api/src/main/java/org/kie/api/runtime/process/WorkItemHandler.java
+++ b/api/kogito-api/src/main/java/org/kie/api/runtime/process/WorkItemHandler.java
@@ -16,8 +16,9 @@
 
 package org.kie.api.runtime.process;
 
-import java.util.Set;
+import java.util.stream.Stream;
 
+import org.kie.kogito.process.workitem.LifeCyclePhase;
 import org.kie.kogito.process.workitem.Transition;
 
 /**
@@ -80,8 +81,13 @@ public interface WorkItemHandler {
     default void transitionToPhase(WorkItem workItem, WorkItemManager manager, Transition<?> transition) {
         throw new UnsupportedOperationException();
     }
-
-    default Set<String> allowedPhases(WorkItem workItem) {
-        throw new UnsupportedOperationException();
+    
+    /**
+     * Returns the set of phases the provided phase is able to be transitioned to
+     * @param phaseId  the phase we want to obtain which phases can be transitioned to
+     * @return stream containing all phases that can be transitioned from the provided phase
+     */
+    default Stream<LifeCyclePhase> allowedPhases(String phaseId) {
+        return Stream.empty();
     }
 }
--- a/jbpm/jbpm-flow/src/test/java/org/jbpm/util/JsonSchemaUtilTest.java
+++ b/jbpm/jbpm-flow/src/test/java/org/jbpm/util/JsonSchemaUtilTest.java
@@ -3,12 +3,28 @@ package org.jbpm.util;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Collection;
 import java.util.Map;
+import java.util.Optional;
 
+import org.jbpm.process.instance.impl.humantask.HumanTaskWorkItemHandler;
 import org.junit.jupiter.api.Test;
+import org.kie.api.runtime.process.WorkItemHandler;
+import org.kie.kogito.Application;
+import org.kie.kogito.Config;
+import org.kie.kogito.process.Process;
+import org.kie.kogito.process.ProcessConfig;
+import org.kie.kogito.process.ProcessInstance;
+import org.kie.kogito.process.ProcessInstances;
+import org.kie.kogito.process.WorkItem;
+import org.kie.kogito.process.WorkItemHandlerConfig;
+import org.kie.kogito.process.workitem.Policy;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
 public class JsonSchemaUtilTest {
     
@@ -66,5 +82,31 @@ public class JsonSchemaUtilTest {
         assertEquals(2,properties.size());
         assertTrue((Boolean)((Map)properties.get("approved")).get("output"));
         assertTrue((Boolean)((Map)properties.get("traveller")).get("input"));
+    }
+
+    @Test
+    <T> void testJsonSchemaPhases() throws IOException {
+        InputStream in = new ByteArrayInputStream(example.getBytes());
+        Policy<T>[] policies = new Policy[0];
+        Map<String, Object> schemaMap = JsonSchemaUtil.load(in);
+        Process<T> process = mock(Process.class);
+        ProcessInstances<T> processInstances = mock(ProcessInstances.class);
+        when(process.instances()).thenReturn(processInstances);
+        ProcessInstance<T> processInstance = mock(ProcessInstance.class);
+        when(processInstances.findById("pepe")).thenReturn((Optional) Optional.of(processInstance));
+        WorkItem task = mock(WorkItem.class);
+        when(processInstance.workItem("task", policies)).thenReturn(task);
+        when(task.getPhase()).thenReturn("active");
+        Application application = mock(Application.class);
+        Config config = mock(Config.class);
+        ProcessConfig processConfig = mock(ProcessConfig.class);
+        when(application.config()).thenReturn(config);
+        when(config.process()).thenReturn(processConfig);
+        WorkItemHandlerConfig workItemHandlerConfig = mock(WorkItemHandlerConfig.class);
+        when(processConfig.workItemHandlers()).thenReturn(workItemHandlerConfig);
+        WorkItemHandler workItemHandler = new HumanTaskWorkItemHandler();
+        when(workItemHandlerConfig.forName("Human Task")).thenReturn(workItemHandler);
+        schemaMap = JsonSchemaUtil.addPhases(process, application, "pepe", "task", policies, schemaMap);
+        assertFalse(((Collection) schemaMap.get("phases")).isEmpty());
     }
 }
--- a/kogito-codegen/src/main/resources/class-templates/RestResourceUserTaskTemplate.java
+++ b/kogito-codegen/src/main/resources/class-templates/RestResourceUserTaskTemplate.java
@@ -3,6 +3,7 @@ package com.myspace.demo;
 import java.util.List;
 
 import org.drools.core.WorkItemNotFoundException;
+import org.jbpm.util.JsonSchemaUtil;
 import org.kie.kogito.process.ProcessInstance;
 import org.kie.kogito.process.WorkItem;
 import org.kie.kogito.process.impl.Sig;
@@ -79,17 +80,18 @@ public class $Type$Resource {
     @GET()
     @Path("$taskName$/schema")
     @Produces(MediaType.APPLICATION_JSON)
-    public Map<String,Object> getSchema() {
-        return JsonSchemaUtil.load(this.getClass().getClassLoader(),process.id(),"$taskName$");
+    public Map<String, Object> getSchema() {
+        return JsonSchemaUtil.load(this.getClass().getClassLoader(), process.id(), "$taskName$");
     }
     
     @GET()
     @Path("/{id}/$taskName$/{workItemId}/schema")
     @Produces(MediaType.APPLICATION_JSON)
-    public Map<String,Object> getSchemaAndPhases(@PathParam("id") final String id, @PathParam("workItemId") final String workItemId) {
-        Map<String,Object> jsonSchema = JsonSchemaUtil.load(this.getClass().getClassLoader(),process.id(),"$taskName$");
-        process.instances().findById(id).ifPresent(pi ->jsonSchema.put("phases",pi.allowedPhases(workItemId)));
-        return jsonSchema;
+    public Map<String, Object> getSchemaAndPhases(@PathParam("id") final String id,
+                                                  @PathParam("workItemId") final String workItemId,
+                                                  @QueryParam("user") final String user,
+                                                  @QueryParam("group") final List<String> groups) {
+        return JsonSchemaUtil.addPhases(process, application, id, workItemId, policies(user, groups), JsonSchemaUtil.load(this.getClass().getClassLoader(), process.id(), "$taskName$"));
     }
 
     @DELETE()
--- a/api/kogito-api/src/main/java/org/kie/api/runtime/process/WorkItemManager.java
+++ b/api/kogito-api/src/main/java/org/kie/api/runtime/process/WorkItemManager.java
@@ -17,7 +17,6 @@
 package org.kie.api.runtime.process;
 
 import java.util.Map;
-import java.util.Set;
 
 import org.kie.kogito.process.workitem.Policy;
 import org.kie.kogito.process.workitem.Transition;
@@ -69,11 +68,4 @@ public interface WorkItemManager {
     default void transitionWorkItem(String id, Transition<?> transition) {
         
     }
-    
-    default Set<String> allowedPhases (String id) {
-        throw new UnsupportedOperationException();
-    }
-    
-    
-
 }
--- a/api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
+++ b/api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
@@ -130,14 +130,6 @@ public interface ProcessInstance<T> {
      * @throws WorkItemNotFoundException in case work item with given id does not exist
      */
     void transitionWorkItem(String id, Transition<?> transition);
-    
-    /**
-     * Identify phases that are available to be transition to for a particular work item
-     * @param id work item 
-     * @return id of the phases that can be transitioned to
-     */
-    Set<String> allowedPhases (String id);
-
     /**
      * Returns work item identified by given id if found
      *
--- a/api/kogito-api/src/main/java/org/kie/kogito/process/workitem/LifeCycle.java
+++ b/api/kogito-api/src/main/java/org/kie/kogito/process/workitem/LifeCycle.java
@@ -16,6 +16,7 @@
 package org.kie.kogito.process.workitem;
 
 import java.util.Collection;
+import java.util.stream.Stream;
 
 import org.kie.api.runtime.process.WorkItem;
 import org.kie.api.runtime.process.WorkItemManager;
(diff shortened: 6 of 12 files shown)


mswiderski:
@fjturado have you thought about moving this logic to a generated named/qualified bean? Since the phases are specific to the handler/lifeCycle maybe allowing it to be decoupled from the engine api would make sense?

fjtirado:
> @fjturado have you thought about moving this logic to a generated named/qualified bean? Since the phases are specific to the handler/lifeCycle maybe allowing it to be decoupled from the engine api would make sense?

Im not sure I understood what you are proposing. 
The result of allowedPhases depends on the current active phase and LifeCycle associated with the human task.
In order to retrieve active phase we need to access procesInstance and workItem. 
In order to obtain LifeCycle, we need the handler reference. 
Still not convinced of the improvement accessing the handler throug ApplicationConfig  rather than adding allowedPhases to PRocessInstance interface as done originally. Specially when I realized the other methods related with human task  (transition, complete and abort) needs WorkItemManager, so they can not be removed from ProcessInstance. 
What I though is to create an utility method somewhere (probably in flow project) that given Application, ProcessIntanceId, TaskName  and TaskIntanceId returns the allowedPhases, so I avoid code duplication between SpringBoot and Quarkus templates

mswiderski:
@fjtirado lifecycle phases are bound to the handler that is then registered for given work item. Main usage of it right now is user tasks so it is bound to HumanTadkWorkItem and HumanTaskWorkItemHandler.

What I suggested to assess if it makes sense to make the handler a CDI bean to some extent so you can inject it, with named injection where the name is the name of the work item it is registered with. It might be actually the Process itself that becomes a producer of such bean and thus you would get hold of LifeCycle associated with given handler and by that to all the phases it supports. Does it make it any clearer? :)

fjtirado:
> @fjtirado lifecycle phases are bound to the handler that is then registered for given work item. Main usage of it right now is user tasks so it is bound to HumanTadkWorkItem and HumanTaskWorkItemHandler.
> 
> What I suggested to assess if it makes sense to make the handler a CDI bean to some extent so you can inject it, with named injection where the name is the name of the work item it is registered with. It might be actually the Process itself that becomes a producer of such bean and thus you would get hold of LifeCycle associated with given handler and by that to all the phases it supports. Does it make it any clearer? :)

Cristal clear ;). I liked the idea, let me try that

evacchi:
@fjtirado if the process "owns" (it "produces") the bean as @mswiderski is suggesting (please correct me if I am mistaken) then maybe you can further simplify that ?

e.g. maybe it could become something like:

```
firstTask.allowedPhases(pi)
```
?

just thinking out loud here...

fjtirado:
> @fjtirado if the process "owns" (it "produces") the bean as @mswiderski is suggesting (please correct me if I am mistaken) then maybe you can further simplify that ?
> 
> e.g. maybe it could become something like:
> 
> ```
> firstTask.allowedPhases(pi)
> ```
> 
> ?
> 
> just thinking out loud here...

Hmmmm, will be moving the workItem  retrieval logic from the REST controller to the WorkItemHandler. Also, besides pi parametes, we will have to provide workItemId anyway, so basically, if we later change the method resolve workItem from ProcessInstnaces (by adding an aditional paameter besides policies), we will have to modify all handlers.

kogito-codegen/src/main/resources/class-templates/RestResourceUserTaskTemplate.java
```diff
@@ -2,6 +2,7 @@
 
 import java.util.List;
 
+import org.dmg.pmml.Application;
```
cristianonicolai, line 5: right import?

mswiderski:
> > @fjtirado if the process "owns" (it "produces") the bean as @mswiderski is suggesting (please correct me if I am mistaken) then maybe you can further simplify that ?
> > e.g. maybe it could become something like:
> > ```
> > firstTask.allowedPhases(pi)
> > ```
> > 
> > 
> > ?
> > just thinking out loud here...
> 
> Hmmmm, will be moving the workItem retrieval logic from the REST controller to the WorkItemHandler. Also, besides pi parametes, we will have to provide workItemId anyway, so basically, if we later change the method resolve workItem from ProcessInstnaces (by adding an aditional paameter besides policies), we will have to modify all handlers.

A word of clarification, life cycle phases are not specific to process instance or work item instance. They are bound to handler as handler defines what life cycle to use. This life cycle the applies to all tasks/work item instances that the handler deals with. In this case all user tasks so there is no need to pass either process instance or work item to know the phases. Only the handler provides this

kogito-codegen/src/main/resources/class-templates/RestResourceUserTaskTemplate.java
fjtirado: This should not be there (it is not really used), I guess it was automatically introduced by the IDE while trying different approaches. I removed it. really good catch!!!

fjtirado:
> > > @fjtirado if the process "owns" (it "produces") the bean as @mswiderski is suggesting (please correct me if I am mistaken) then maybe you can further simplify that ?
> > > e.g. maybe it could become something like:
> > > ```
> > > firstTask.allowedPhases(pi)
> > > ```
> > > 
> > > 
> > > ?
> > > just thinking out loud here...
> > 
> > 
> > Hmmmm, will be moving the workItem retrieval logic from the REST controller to the WorkItemHandler. Also, besides pi parametes, we will have to provide workItemId anyway, so basically, if we later change the method resolve workItem from ProcessInstnaces (by adding an aditional paameter besides policies), we will have to modify all handlers.
> 
> A word of clarification, life cycle phases are not specific to process instance or work item instance. They are bound to handler as handler defines what life cycle to use. This life cycle the applies to all tasks/work item instances that the handler deals with. In this case all user tasks so there is no need to pass either process instance or work item to know the phases. Only the handler provides this

LifeCycle is needed  to know avaialbe transitions and WorkItem.phase is needed to know which phase transition we need to check.

mswiderski:
@fjtirado instead of using `canTransition` why not exposing a new method on `LifeCyclePhase` that returns allowed phases for given phase?

That way lifecycle is the only thing you need to obtain/inject and then constantly reuse it as for the same type of tasks (work item) it won't change. 

And yes, you need the current work item phase to know what is possible next but that is irrelevant to obtaining the lifecycle itself from the handler. And whenever querying for task schema you should already have work item and process instance id so that is needed anyway. What I mean these two are not needed to get the life cycle ...

fjtirado:
@mswiderski 
I think allowedPhased probably belongs to HumanTaskItemHandler, The division of responsability I foresee is, LifeCylcle implementation specifies if one phase can be transitioned to another and WorItemHandler adds some logic which is independent of the particular LyfeCycle implemantation (actually AllowedPhases is a handy example of that, from the information specified by the LifeCycle implementation, it builds another representation of that info, keeping the custom interface to be implemented by developer smaller). 
Injecting WorkItemHandler is fine, although we save little code compared with addition of  workItemHandler(String handlerName) method in Process interface ( (as I tried to llusstrate here  https://github.com/kiegroup/kogito-runtimes/pull/631#issuecomment-658276034), which I believe we will need to eventually add anywway. 
What Edoardo is trying to achieve is to reduce verbosity of gernated code, and what Im pointing out is that most of the generated code is dealing with ProcessIntance and WorkItem resolution (Im not saying that is nof fine, just stating a fact ;))

mswiderski:
> I think allowedPhased probably belongs to HumanTaskItemHandler

not sure why, allowed phases are purely what current phase case transition to and that is already implemented on the LifeCyclePhase anyway, so it is just another view of the same data - instead of returning true or false it gives you the list of possible next phases.

Handler itself does not really add much, it just uses the lifecycle implementation which then relies on defined all possible phases in that life cycle but relationship between phases are declared on the phase itself.

fjtirado:
> > I think allowedPhased probably belongs to HumanTaskItemHandler
> 
> not sure why, allowed phases are purely what current phase case transition to and that is already implemented on the LifeCyclePhase anyway, so it is just another view of the same data - instead of returning true or false it gives you the list of possible next phases.
> 
> Handler itself does not really add much, it just uses the lifecycle implementation which then relies on defined all possible phases in that life cycle but relationship between phases are declared on the phase itself.

I see LifeCycle as an information provider and Handler as an information transformer. So LifeCycle is kept small and handler will continously grow. Im assuming we are expecting Kogito users to implement LifeCycle frequently. Since allowedPhases is just a view built from canTransitionTo (and needs to be consistent with CanTransitionTo) I do not think it belongs to the inteface.

mswiderski:
> So LifeCycle is kept small and handler will continously grow.

not really, quite the opposite - handler when you look at it now - `HumanTaskWorkItemHandler` is really tiny and it delegates to the life cycle implementation.

> Im assuming we are expecting Kogito users to implement LifeCycle frequently

It depends if there will be needs for it but I'd say it might happen, there is even a QuickStart showing how to put in custom life cycle for user tasks. So it might be but does not have to. Anyway, LifeCycle defines all the phases so this is what is important.

> Since allowedPhases is just a view built from canTransitionTo (and needs to be consistent with CanTransitionTo) I do not think it belongs to the interface.

which interface you have in mind? LifeCycle or LifeCyclePhase? I meant it should be part of LifeCyclePhase as that phase defines what it can be transitioned to.

> Another way of looking at this is to use hypotethical situation, lets imagine that canTransitionTo did not exist in LifeCycle interface and instead we had allowedPhases. The canTransition logic will be allowedPhases(sourcePhase).contains(targetPhase), in such case, will be we adding canTrasitionTo method to the LifeCycle or to the handler?

None of these as it belongs to LifeCyclePhase interface. Same LifeCyclePhase can be reused across LifeCycle implementations and thus phase should decide where it can transition.

fjtirado:
Hmmm, ok, lets take benefit of "traits" support in java add a new method allowedPhases (String phaseId) to LifeCycle interface, which default implementation will be 

@override
public Set<String> allowedPhases(String phaseId) {
LifeCyclePhase activePhase =lifeCycle.phaseById(phaseId);
return lifeCycle.phases().stream().filter(phase -> phase.canTransition(activePhase)).map(LifeCyclePhase::id).collect(Collectors.toSet());
}

So, this way we avoid adding logic in the HumanWorkItemHandler, which will kept tiny, and implementor of LifeCycle are not forced to implement this (that will rely onLifeCyclePhase. canTranstion)
Rather than injecting the handler, we will be injecting the LifeCycle.

mswiderski:
Overall looks fine to me to go with LifeCycle extension. One comment though, maybe return LifeCyclePhase from ´allowedPhases´ as it can give you more info not just the id. But that is minor...

fjtirado:
> Overall looks fine to me to go with LifeCycle extension. One comment though, maybe return LifeCyclePhase from ´allowedPhases´ as it can give you more info not just the id. But that is minor...
What about this alternative? 
   default Stream<LifeCyclePhase> allowedPhases(String phaseId) {
        LifeCyclePhase activePhase = phaseById(phaseId);
        return phases().stream().filter(phase -> phase.canTransition(activePhase));
    }
So, we return stream of LifeCyclePhase and the caller can perform the proper  conversio, Since this is not remote interface, it is probably better than returning a collection over which is very likely the caller will invoke stream method and this way a double conversion stream->collection (internally in the method) and collectio->stream (externally by the caller) is avoided.

mswiderski:
@fjtirado makes perfect sense!

api/kogito-api/src/main/java/org/kie/api/runtime/process/WorkItemHandler.java
```diff
@@ -81,7 +81,7 @@ default void transitionToPhase(WorkItem workItem, WorkItemManager manager, Trans
         throw new UnsupportedOperationException();
     }
 
-    default Set<String> allowedPhases(WorkItem workItem) {
+    default Set<String> allowedPhases(String phaseId) {
```
evacchi, line 84: wishlist for the future: I'd like it better to have two different interfaces, one with this method, the other without.

- WorkItemHandler
- CustomPhaseWorkItemHandler extends WorkItemHandler

or something along those lines

mswiderski: haven't we agree that this method will be in `LifeCycle` and not on `WorkItemHandler`, @fjtirado?

fjtirado: yes, thats the idea, but I have not uploaded changes related with it yet.
Im not sure it is feasible to inject LifeCycle, without changing current customtask example. Im currently studying the code

fjtirado: Anyway, independentyl on the issue of injecting LifeCycle, I have a foundamental doubt, WorkItemHandler inerface already have these methods
 void executeWorkItem(WorkItem workItem,
                         WorkItemManager manager);

    /**
     * The given work item should be aborted.
     * @param workItem the work item that should be aborted
     * @param manager the manager that requested the work item to be aborted
     */
    void abortWorkItem(WorkItem workItem,
                       WorkItemManager manager);
    
    /**
     * Returns name that it should be registered with, by default simple class name;
     * @return name that should this handler be registered with
     */
    default String getName() {
        return getClass().getSimpleName();
    };
    
    /**
     * Allows to transition given work item into another life cycle phase as defined in
     * the given transition object.
     * @param workItem work item to be transitioned
     * @param transition life cycle phase to be transition to
     */
    default void transitionToPhase(WorkItem workItem, WorkItemManager manager, Transition<?> transition) {
        throw new UnsupportedOperationException();
    }

so, there is actually one existing method that has the word phase, which is is the issue of having allowedPhases added to it, I need to understand because I felt Im missing something.

mswiderski: Remove the ‘ToPhase’ from the method name and there is no more reference to phases any longer. It just the name that is maybe not named in the best way. Nothing more I’d say.

mswiderski:
The potential issue with injecting life cycle is how do you know which handler it should be bound do? Handler has a name that is the connection point between work items/tasks and the impl. Lifecycle is internal to it right now and this can be different for different type of tasks and handlers. 

When it comes to ConditionalOnMissingBean this is done based on jandex index - no impl found means default should be used. All compile/build time

fjtirado:
If we use injection for WorkItemHandler and LifeCycle, this will impliy changes to section 10.7.3 of current documentation?
Basically rather than calling register, use will just have to write a LifeCycle implementation and annotated it with ApplicationScoped. 
Are we fine with this?

evacchi:
I think that's alright @fjtirado, it's just one tiny example.

However, if we do go this route (injecting Lifecycle), if I have understood correctly, that also means that the default implementations of User Task (also Log, but that's a detail) should be injected and overridable.

Now the only concern is that if we rely on injection in this code path, could it be made a bit more fragile? Especially for built-ins? What I mean is that, if we rely on the same mechanism for both built-ins and user-provided classes, wouldn't that make possibly easier for users to inadvertently "break" built-in tasks?

Just my 2c.

mswiderski:
my problem with injecting LifeCycle is how to you distinguish which LifeCycle go into what handler?

fjtirado:
> my problem with injecting LifeCycle is how to you distinguish which LifeCycle go into what handler?

yes, mine too ;)
I was going to assume that there is only one Human Task Handler per process and that this handler uses the managed bean that implements LifeCicle

mswiderski:
Keep in mind that LifeCycle is not exclusively attached to human tasks but can be used for any work item including service tasks. So that is what is the issue here - different life cycle impala will be used by different handlers and thus different type of tasks/work items. 
That’s why I think relying on the handler itself might be the most feasible approach

fjtirado:
Ok, so I guess  handler rather than life cycle should be injected. 
If my understanding is true, then the question that remains open is if we should add allowedHandler method to WorkItemHandler or maybe add a getLifeCycle method to WorkItemHandler, wdyt?

mswiderski:
I'd go with `lifeCycle` on the handler with default method on the interface that returns null.

fjtirado:
Finally I have decided to use application config to retrieve work item handler reference
Injecting workitemhandler will require changes in the engine (change the way register is used in WorkItemHandlerConfig) and I honestly did not see any advantage on doing that (and potentially some unexpected side effects), soI think we should leave that for a different PR. 
In order to reduce verbosity in generated code and also to avoid duplication between quarkus and spring template, I have moved the logic to constant (aka not generated) class JsonSchemaUtil

kogito-codegen/src/main/resources/class-templates/RestResourceUserTaskTemplate.java
```diff
@@ -86,10 +87,8 @@
     @GET()
     @Path("/{id}/$taskName$/{workItemId}/schema")
     @Produces(MediaType.APPLICATION_JSON)
-    public Map<String,Object> getSchemaAndPhases(@PathParam("id") final String id, @PathParam("workItemId") final String workItemId) {
-        Map<String,Object> jsonSchema = JsonSchemaUtil.load(this.getClass().getClassLoader(),process.id(),"$taskName$");
-        process.instances().findById(id).ifPresent(pi ->jsonSchema.put("phases",pi.allowedPhases(workItemId)));
-        return jsonSchema;
+    public Map<String,Object> getSchemaAndPhases(@PathParam("id") final String id, @PathParam("workItemId") final String workItemId , @QueryParam("user") final String user, @QueryParam("group") final List<String> groups) {
+        return JsonSchemaUtil.addPhases(process,application,id,workItemId, policies(user,groups),JsonSchemaUtil.load(this.getClass().getClassLoader(),process.id(),"$taskName$"));
```
cristianonicolai, line 91: please review code style, its missing spaces between vars, see https://github.com/kiegroup/droolsjbpm-build-bootstrap/tree/master/ide-configuration

jbpm/jbpm-flow/src/test/java/org/jbpm/util/JsonSchemaUtilTest.java
```diff
@@ -67,4 +82,30 @@ void testJsonSchema() throws IOException {
...
     }
+
+    @Test
+    <T> void testJsonSchemaPhases() throws IOException {
+        InputStream in = new ByteArrayInputStream(example.getBytes());
+        Policy<T>[] policies = new Policy[0];
+        Map<String, Object> schemaMap = JsonSchemaUtil.load(in);
+        Process<T> process = Mockito.mock(Process.class);
+        ProcessInstances<T> processInstances = Mockito.mock(ProcessInstances.class);
+        Mockito.when(process.instances()).thenReturn(processInstances);
```
cristianonicolai, line 93:
```suggestion
        when(process.instances()).thenReturn(processInstances);
```

kogito-codegen/src/main/resources/class-templates/RestResourceUserTaskTemplate.java
fjtirado: I have it configured, but it was not appliead automatically to template file because is not considered source, I applied it manually now

jbpm/jbpm-flow/src/test/java/org/jbpm/util/JsonSchemaUtilTest.java
fjtirado: Using static import now (Im not fan of them, but in this case I agree save a lot of code)

PR #635, opened 2020-07-15
KOGITO-2729 Better error message if DRL is compiled at runtime and drools-core-static is in the classpath

evacchi:
Add a descriptive error message to the UnsupportedOperationException that gets thrown by the Dummy classloader implementation.


evacchi:
Jenkins please retest this

PR #636, opened 2020-07-15
KOGITO-2597  Some tests fail when testing with GraalVM because of JS support

evacchi:
Add assumption so that the test is skipped when GraalJS is detected -- currently exact match on Class name. May break in the future.


radtriste:
@evacchi small issue it seems on SonarCloud ...

evacchi:
@radtriste this must be the commit with the highest bug-to-code ratio I have ever written :D

PR #646, opened 2020-07-22
[KOGITO-2756] Hide Implementation details of WorkItemHandler

fjtirado:
Reducing generated code size by using map and orElse ( I think orElse(null) should be orElseThrow (()-> new ProcessInstanceNotFound(id)), I have opened https://issues.redhat.com/browse/KOGITO-2875 to change that) 
ToMap() method call cannot be removed unless xxx_TaskOuput implements Model (method fromMap of Model interface does not exist for xxx_TaskOuput, the same than toMap does not exist for xxx_TaskInput)
A set of static initializers has been added for HumanTaskTransition (constructors has been left for backward compatibility)
xxx_TaskInput generated fromMap method now accepts a WorkItem rather than three parameters
The JIRA was asking for return pi.workItem(Travels_6_TaskInput.class, policies(user, groups)); rather than Travels_6_TaskInput.fromMap(pi.workItem(workItemId, policies(user, groups))), but doing that will imply to make ProcessInstance dependant on generated code, which is not a good idea. I mean, one thing is parametrized ProcessInstance with generated model class (as it is currently done) and another to deal will all generated xxx_TaskInput class

### DIFF
--- a/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/UserTaskModelMetaData.java
+++ b/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/UserTaskModelMetaData.java
@@ -63,6 +63,8 @@ public class UserTaskModelMetaData {
     private static final String TASK_INTPUT_CLASS_SUFFIX = "TaskInput";
     private static final String TASK_OUTTPUT_CLASS_SUFFIX = "TaskOutput";
     private static final String TASK_NAME = "TaskName";
+    private static final String WORK_ITEM = "workItem";
+    private static final String PARAMS = "params";
     
     protected static final List<String> INTERNAL_FIELDS = Arrays.asList(TASK_NAME, "NodeName", "ActorId", "GroupId", "Priority", "Comment", "Skippable", "Content", "Locale");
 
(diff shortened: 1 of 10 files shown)


mswiderski:
Not a big fan of this as it seems like putting some utility into the ProcessInstance interface. I am ok in making the template code smaller but not by moving it to the interface

fjtirado:
> Not a big fan of this as it seems like putting some utility into the ProcessInstance interface. I am ok in making the template code smaller but not by moving it to the interface

Lets analyze for a second currently exting ProcessInstances interface
There are two methods, One that returns all  process  intances as a collection (values)  and a method that returns one optional given an id (findById)
findById method  can be implemented using values method, a trivial implementacion being values().stream().findAny(pi -> pi.getId().equals(id))
This will be inneficient since you have to build the whole collection and then find by id,  but it will have been avoided if initially  the interface consist in just one method that returns a Stream. 
Looking at the generated code, I realize that we will be saving a lot of plaiterboile code  if we have a method that when the process instance  id is found applys a function to it or else returns null (or even better throw an excepcion that is handled by JAX/RS layer and generates a 404 response rather than returning null) and the natural place to add this method is ProcessInstances interface. 
An alternative is to add a class ProcessInstancesHelper, but this will be more verborse and we need to find a proper place for it API itself or another place?

fjtirado:
jenkins retest this

fjtirado:
jenkins retest this

evacchi:
>  Looking at the generated code, I realize that we will be saving a lot of code if we have a method that when process instance is found applys a function to it or else returns null (or, even better, throw an excepcion that will be eventually handled by JAX/RS layer and generates a 404 response rather than returning null and generating a 204 response) and the natural place to add this method for me was ProcessInstances interface.

isn't that the right job for an Optional?

PR #656, opened 2020-07-28
[KOGITO-2893] Detach Business Key from ID

ruromero:
Fix: https://issues.redhat.com/browse/KOGITO-2893

### Business Key

- Business Key is kept as is but when provided it will be used as the ID.
- IDs will always be generated upon resource creation.

### Rest API

- Use PUT instead of POST when data is submitted using the id. i.e. resource updates 
- When a resource is created a 201 status code will be returned including a Location header with the url to the resource. This affects
  - Process instances created (custom or generated ID)
  - New Task instantiation (for adhoc tasks)
- When a resource is not found a NotFoundException is thrown instead of returning 204 No Content

### Bonus 

- Updates were only updating the internal variables (legacyProcessInstance), not the variables model.
- Use random port for SpringBoot tests


mswiderski:
@ruromero retrieving instances by business key is already there - it can be used instead of the generated id.

this actually enables filtering by business key on entire list of active instances which I believe should be covered by data index.

When it comes to business key in the model, since that is provided I don't see a need for it to be added to the model as it does not really is part of the model in my opinion.

ruromero:
> @ruromero retrieving instances by business key is already there - it can be used instead of the generated id.
I didn't notice the UUID trick. Thanks for pointing it out.
> 
> this actually enables filtering by business key on entire list of active instances which I believe should be covered by data index.
> 
> When it comes to business key in the model, since that is provided I don't see a need for it to be added to the model as it does not really is part of the model in my opinion.
I do think it should be part of the model because otherwise API consumers won't be able to access this value any more. Let's say I only want to use the businessKey and forget about the ID.

mswiderski:
> I do think it should be part of the model because otherwise API consumers won't be able to access this value any more. Let's say I only want to use the businessKey and forget about the ID.

then go for it but not as extra property, reuse id field to set business key when available otherwise use generated id. Don't expose both as it is confusing and makes it less domain specific.

ruromero:
> > I do think it should be part of the model because otherwise API consumers won't be able to access this value any more. Let's say I only want to use the businessKey and forget about the ID.
> 
> then go for it but not as extra property, reuse id field to set business key when available otherwise use generated id. Don't expose both as it is confusing and makes it less domain specific.

Actually they are 2 different things. One is a correlation ID that is provided by the API consumer. If the user provides such information it means it is useful for the business logic, i.e. the domain. It is totally different from the logical ID and I consider they should be completely independent.
That said, we shouldn't even enforce the businessKey uniqueness. The same key could be used multiple times for retries, for example. It's not Kogito's responsibility to deal with businessKeys uniqueness.

I have therefore decided to remove UUID generation from the businessKey as they're conceptually independent and should not be related at all.

ruromero:
Or maybe use it as an HTTP Header like X-Correlation-ID or X-KOGITO-BusinessKey

tiagodolphine:
Regarding the businessKey and ID, I agree they are independent, the ID identifies the resource in kogito and the businessKey is more a correlation id that is from the client responsibility (how kogito could, in fact, be integrated with other services running on the customer environment) kogito cannot guarantee its uniqueness. 

Now the way the businessKey is sent to be saved with the resource, IMHO it should not be using a query param or even an HTTP Header, it should be in the body of the POST, query param and header are more used to query/filter/select, etc but not sending values to be persisted.

Let's use an example, imagine a customer that has an e-commerce application with multiple microservices and at some point, they want to use a kogito service to control the order payment flow, in this way they have the orderId(that is the businessKey) and they need to send the POST to create an order payment in kogito, so sending the orderId to be correlated it is part of the domain, it is clear how the integration could be done, so why not this businessKey is part of the body request either for POST and GET? From the user of the API is clear, he ID from kogito identifies the order there and the correlationId idenfify the order from the external domain, I mean this common in 
 the microservices world.

ruromero:
Then @tiagodolphine from your example we can say that the businessKey (_orderID_) is not part of the _order payment_ domain but of the higher level e-commerce application scope.
From that I'd say the key should not be part of this specific sub-domain and thus my preference of using HTTP header both in the request and the response.

Besides, I'd avoid using the term correlationID here because, even though it can be used for that I think that conceptually it is still more a business key.

kogito-codegen/src/main/resources/class-templates/ReactiveRestResourceTemplate.java
```diff
@@ -104,18 +98,20 @@
...
+                                                                    @PathParam("id") String id,
+                                                                    $Type$ resource) {
         return CompletableFuture.supplyAsync(() -> {
             return org.kie.kogito.services.uow.UnitOfWorkExecutor.executeInUnitOfWork(application.unitOfWorkManager(), () -> {
                 ProcessInstance<$Type$> pi = process.instances()
                         .findById(id)
                         .orElse(null);
                 if (pi == null) {
-                    return null;
+                    return createProcessInstance(httpHeaders, id, resource);
```
tiagodolphine, line 114: In general, I agree the PUT method should be used to update the resource, in his case the variables, so it should not be a POST as it was, but on the other hand, creating a new resource using a PUT is not proper behavior, what about creating the resource in the POST method but sending the ID inside the body (the resource with the ID) or one other position would be the same as it is here with the ID on the path but using a POST instead of a PUT just to create and let the PUT exclusively to do the update operation. WDYT?

tiagodolphine: Well, looking at the RFC https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.6
> PUT
 ... If the Request-URI does not point to an existing resource, and that URI is capable of being defined as a new resource by the requesting user agent, the origin server can create the resource with that URI. If a new resource is created, the origin server MUST inform the user agent via the 201 (Created) response...
>
So using the PUT can be a solution as you did, and in case the resource does not exist the HTTP code on the response should be 201 instead of 200.

ruromero: Yes, I've done it this way.

ruromero:
jenkins retest this

ruromero:
jenkins please retest this

kogito-codegen/src/main/resources/class-templates/RestResourceUserTaskTemplate.java
```diff
@@ -25,18 +25,15 @@
                                       pi.workItems().stream().filter(wi -> wi.getName().equals("$taskName$"))
                                         .findFirst();
                               if (task.isPresent()) {
-                                  return Response.ok(pi.variables().toOutput())
-                                                 .header("Link", "</" + id + "/$taskName$/" + task.get().getId() +
-                                                                 ">; rel='instance'")
-                                                 .build();
+                                  return Response.created(uriInfo.getAbsolutePathBuilder().path(task.get().getId()).build()).build();
```
tiagodolphine, line 28: Is it necessary to update the documentation regarding this modification? probably for the other points as well where the REST APIs are being changed, maybe create a JIRA for it, WDYT?

ruromero: Done https://issues.redhat.com/browse/KOGITO-3068

ruromero:
@tiagodolphine The build is failing because of kogito-apps and kogito-examples dependencies. If you agree on the changes I will create and link the PR for those 2 projects and once everything is set we can merge all.
Thanks

tiagodolphine:
> @tiagodolphine The build is failing because of kogito-apps and kogito-examples dependencies. If you agree on the changes I will create and link the PR for those 2 projects and once everything is set we can merge all.
> Thanks

You can open PRs to fix the kogito-apps and kogito-examples with the same branch name to fix it. But let's wait for other's approvals before doing it, in case there will be any modification.

@cristianonicolai ^

kogito-codegen/src/main/resources/class-templates/RestResourceUserTaskTemplate.java
cristianonicolai: In my opinion docs changes should be part of the JIRA that is causing the changes.

Also, I think we should continue to return the output variables of task as processes that were created.

cristianonicolai:
@ruromero I think most changes are welcome except for not returning the model on creating the resource. My only feeling is that these changes are beyond the scope of this JIRA now. Would be best to split this work in my opinion.

kogito-codegen/src/main/resources/class-templates/RestResourceTemplate.java
```diff
@@ -35,54 +40,56 @@
...
             String startFromNode = httpHeaders.getHeaderString("X-KOGITO-StartFromNode");
+
             if (startFromNode != null) {
                 pi.startFrom(startFromNode);
             } else {
                 pi.start();
             }
-            return pi.checkError().variables().toOutput();
+            UriBuilder uriBuilder = uriInfo.getAbsolutePathBuilder().path(pi.id());
+            return Response.created(uriBuilder.build())
```
tiagodolphine, line 68: Is seems the same changes are missing on the Spring templates, like SpringRestResourceTemplate.java

ruromero:
jenkins retest this

tiagodolphine:
jenkins retest this

ruromero:
Depends on https://github.com/kiegroup/kogito-apps/pull/409

ruromero:
jenkins retest this

ruromero:
jenkins retest this

cristianonicolai (approved):

PR #660, opened 2020-07-30
KOGITO-2911 Refactor Codegen for "RuleUnits" class in Application.java 

evacchi:
follow up to https://github.com/kiegroup/kogito-runtimes/pull/657 

must merge with: https://github.com/kiegroup/kogito-examples/pull/350

- this mimicks the work done for Application.java and Processes.java but for RuleUnits.java
- RuleUnits is now a factory in non-DI contexts (i.e. kogito-codegen) 
- otherwise it collects and provides all the RuleUnit beans and it can be injected with `@Inject/@Autowire RuleUnits` 

The change is that currently `RuleUnits` is _always_ a factory; i.e. it always returns _new beans_. While in DI contexts, it should delegate generation to the DI container.

This PR also makes less necessary `ProjectModel`/`ProjectRuntime`, which are currently _always_ being used to construct rule units. This PR makes every RuleUnit<T> class self-contained, it no longer requires an external generated factory, _unless_ it is a `SessionRuleUnit`, i.e. a RuleUnit that mimicks the old Session behavior. Follow up to clean up: https://issues.redhat.com/browse/KOGITO-2929

A minor API change is also introduced; RuleUnit has now an id() method that returns a String representing the fully qualified name of the rule unit; in case the rule unit is a `SessionRuleUnit`  then ID is the name of the KieSession. This is required so that the generated `RuleUnits` class can cache and return the bean instances (on CDI/Spring), as it is currently done for processes.

### DIFF
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/IncrementalRuleCodegen.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/IncrementalRuleCodegen.java
@@ -206,7 +206,7 @@ public class IncrementalRuleCodegen extends AbstractGenerator {
     private boolean hotReloadMode = false;
     private AddonsConfig addonsConfig = AddonsConfig.DEFAULT;
     private boolean useRestServices = true;
-    private String packageName = "defaultpkg";
+    private String packageName = KnowledgeBuilderConfigurationImpl.DEFAULT_PACKAGE;
     private final boolean decisionTableSupported;
     private final Map<String, RuleUnitConfig> configs;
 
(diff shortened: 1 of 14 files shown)


evacchi:
jenkins please retest this

evacchi:
jenkins please retest this

evacchi:
jenkins please retest this

https://media.giphy.com/media/3ov9jKovUMHqtDrAuQ/giphy.gif

kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/IncrementalRuleCodegen.java
```diff
@@ -206,7 +206,7 @@ private static ResourceType typeOf(FileSystemResource r) {
     private boolean hotReloadMode = false;
     private AddonsConfig addonsConfig = AddonsConfig.DEFAULT;
     private boolean useRestServices = true;
-    private String packageName;
+    private String packageName = "defaultpkg";
```
mariofusco, line 209: Possibly use constant `org.drools.compiler.builder.impl.KnowledgeBuilderConfigurationImpl.DEFAULT_PACKAGE`

mariofusco (approved):

evacchi:
I would like to merge this as it's been sitting here for a while
/cc @tarilabs may be interested in these changes (if you recall, I told you I was not happy about the way ProjectModel was used; this changes it back the way I wanted)

evacchi:
(why is it rebuilding it now lol)

PR #664, opened 2020-07-31
exclude not used infinispan transitive dependencies no longer availab…

mariofusco:


mariofusco:
Following suggestions of infinispan team I replaced this PR with https://github.com/kiegroup/kogito-runtimes/pull/669

PR #674, opened 2020-08-04
[KOGITO-2753] Processes: Remove model transformations from REST endpoint

fjtirado:
- Model interface extends MapOutput and MapInput interfaces
- xxxTaskOutput classes implements MapOutput (generated code does not
longer need to invoke toMap when building HumanTaskTransition)
- New methods toModel on InputModel and toOutput on Model
- Add chechError to ProcessInstance
- Modification on templates to use them (removing mapInput, mapOutput and
getModel)

### DIFF
--- a/api/kogito-api/src/main/java/org/kie/kogito/MapInput.java
+++ b/api/kogito-api/src/main/java/org/kie/kogito/MapInput.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,9 +15,17 @@
 
 package org.kie.kogito;
 
+import java.util.Map;
+
 /**
- * Represents data model type of objects that are usually descriptor of data holders.
- *
+ * To be implemented by classes which can be populated from a Map
  */
-public interface Model extends MapInput, MapOutput {
+public interface MapInput {
+
+    /**
+     * Fills the class with information retrieved from the map
+     * @param params Map containing keys which matches names of fields
+     * in the class*/
+    void fromMap(Map<String, Object> params);
+
 }
--- a/api/kogito-api/src/main/java/org/kie/kogito/MapOutput.java
+++ b/api/kogito-api/src/main/java/org/kie/kogito/MapOutput.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,9 +15,16 @@
 
 package org.kie.kogito;
 
+import java.util.Map;
+
 /**
- * Represents data model type of objects that are usually descriptor of data holders.
- *
+ * To be implemented by classes which can express its internal information as a Map
  */
-public interface Model extends MapInput, MapOutput {
+public interface MapOutput {
+
+    /**
+     * Returns class representation as map  
+     * @return non null map of data extracted from the class
+     */
+    Map<String, Object> toMap();
 }
--- a/api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
+++ b/api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
@@ -94,7 +94,7 @@ public interface ProcessInstance<T> {
     /**
      * Updates process variables of this process instance
      */
-    void updateVariables(T updates);
+    T updateVariables(T updates);
 
     /**
      * Returns current status of this process instance
@@ -181,6 +181,14 @@ public interface ProcessInstance<T> {
      * @return returns process error
      */
     Optional<ProcessError> error();
+
+    default ProcessInstance<T> checkError() {
+        Optional<ProcessError> error = error();
+        if (error.isPresent()) {
+            throw new ProcessInstanceExecutionException(id(), error.get().failedNodeId(), error.get().errorMessage());
+        }
+        return this;
+    }
 
     void triggerNode(String nodeId);
 
--- a/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ProcessToExecModelGenerator.java
+++ b/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ProcessToExecModelGenerator.java
@@ -20,12 +20,25 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Optional;
+import java.util.function.Consumer;
 
 import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.Modifier;
+import com.github.javaparser.ast.NodeList;
 import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
 import com.github.javaparser.ast.body.MethodDeclaration;
+import com.github.javaparser.ast.expr.AssignExpr;
+import com.github.javaparser.ast.expr.MethodCallExpr;
+import com.github.javaparser.ast.expr.NameExpr;
+import com.github.javaparser.ast.expr.ObjectCreationExpr;
+import com.github.javaparser.ast.expr.VariableDeclarationExpr;
 import com.github.javaparser.ast.nodeTypes.NodeWithName;
+import com.github.javaparser.ast.stmt.BlockStmt;
+import com.github.javaparser.ast.stmt.ReturnStmt;
+import com.github.javaparser.ast.type.ClassOrInterfaceType;
+import org.drools.core.util.StringUtils;
 import org.jbpm.process.core.ContextContainer;
+import org.jbpm.process.core.context.variable.Variable;
 import org.jbpm.process.core.context.variable.VariableScope;
 import org.jbpm.workflow.core.impl.WorkflowProcessImpl;
 import org.jbpm.workflow.core.node.HumanTaskNode;
@@ -33,11 +46,13 @@ import org.kie.api.definition.process.Node;
 import org.kie.api.definition.process.WorkflowProcess;
 
 import static com.github.javaparser.StaticJavaParser.parse;
+import static com.github.javaparser.StaticJavaParser.parseClassOrInterfaceType;
 import static org.drools.core.util.StringUtils.ucFirst;
 
 public class ProcessToExecModelGenerator {
 
-    public static final ProcessToExecModelGenerator INSTANCE = new ProcessToExecModelGenerator(ProcessToExecModelGenerator.class.getClassLoader());
+    public static final ProcessToExecModelGenerator INSTANCE = new ProcessToExecModelGenerator(
+                                                                                               ProcessToExecModelGenerator.class.getClassLoader());
 
     private static final String PROCESS_CLASS_SUFFIX = "Process";
     private static final String MODEL_CLASS_SUFFIX = "Model";
@@ -52,7 +67,9 @@ public class ProcessToExecModelGenerator {
     public ProcessMetaData generate(WorkflowProcess process) {
         CompilationUnit parsedClazzFile = parse(this.getClass().getResourceAsStream(PROCESS_TEMPLATE_FILE));
         parsedClazzFile.setPackageDeclaration(process.getPackageName());
-        Optional<ClassOrInterfaceDeclaration> processClazzOptional = parsedClazzFile.findFirst(ClassOrInterfaceDeclaration.class, sl -> true);
+        Optional<ClassOrInterfaceDeclaration> processClazzOptional = parsedClazzFile.findFirst(
+                                                                                               ClassOrInterfaceDeclaration.class,
+                                                                                               sl -> true);
 
         String extractedProcessId = extractProcessId(process.getId());
 
@@ -62,14 +79,14 @@ public class ProcessToExecModelGenerator {
         ClassOrInterfaceDeclaration processClazz = processClazzOptional.get();
         processClazz.setName(ucFirst(extractedProcessId + PROCESS_CLASS_SUFFIX));
         String packageName = parsedClazzFile.getPackageDeclaration().map(NodeWithName::getNameAsString).orElse(null);
-        ProcessMetaData metadata = new ProcessMetaData(process.getId(),
-                extractedProcessId,
-                process.getName(),
-                process.getVersion(),
-                packageName,
-                processClazz.getNameAsString());
+        ProcessMetaData metadata =
+                new ProcessMetaData(process.getId(), extractedProcessId, process.getName(), process.getVersion(),
+                                    packageName, processClazz.getNameAsString());
 
-        Optional<MethodDeclaration> processMethod = parsedClazzFile.findFirst(MethodDeclaration.class, sl -> sl.getName().asString().equals("process"));
+        Optional<MethodDeclaration> processMethod = parsedClazzFile.findFirst(MethodDeclaration.class, sl -> sl
+                                                                                                               .getName()
+                                                                                                               .asString()
+                                                                                                               .equals("process"));
 
         processVisitor.visitProcess(process, processMethod.get(), metadata);
 
@@ -85,13 +102,9 @@ public class ProcessToExecModelGenerator {
         String extractedProcessId = extractProcessId(process.getId());
 
         String packageName = clazz.getPackageDeclaration().map(NodeWithName::getNameAsString).orElse(null);
-        ProcessMetaData metadata = new ProcessMetaData(process.getId(),
-                extractedProcessId,
-                process.getName(),
-                process.getVersion(),
-                packageName,
-                "process");
-
+        ProcessMetaData metadata =
+                new ProcessMetaData(process.getId(), extractedProcessId, process.getName(), process.getVersion(),
+                                    packageName, "process");
         MethodDeclaration processMethod = new MethodDeclaration();
         processVisitor.visitProcess(process, processMethod, metadata);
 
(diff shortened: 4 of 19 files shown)


api/kogito-api/src/main/java/org/kie/kogito/MapOutput.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package org.kie.kogito;
+
+import java.util.Map;
+
+/**
+ * A model which can express its internal information as a Map
+ */
+public interface MapOutput {
```
evacchi, line 23: `OutputModel`

api/kogito-api/src/main/java/org/kie/kogito/MapInput.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package org.kie.kogito;
+
+import java.util.Map;
+
+/**
+ * A model which can be populated from a Map
+ */
+public interface MapInput {
```
evacchi, line 23: I'd propose to call it `InputModel`

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
```diff
@@ -47,30 +47,30 @@
     /**
      * Starts process instance
      */
-    void start();
+    ProcessInstance<T> start();
```
evacchi, line 50: I am not sure I like making all these methods fluent. The reason is that if we wanted at some point in the future to make any of those to return a value then we won't be able. We can make the code just as concise by using `{ pi.start(); pi.something() }` etc. (notice the semicolon)

I know my original bug report included _some_ fluent interfaces :P but not sure I want it for _all_. I am not sure I want it for _some_ either at this point, because it would make the API asymmetrical.

kogito-codegen/src/main/resources/class-templates/ReactiveRestResourceTemplate.java
```diff
@@ -37,134 +38,75 @@
...
-                }
-                return getModel(pi);
-            });
-        });
+
+        return CompletableFuture.supplyAsync(() -> UnitOfWorkExecutor.executeInUnitOfWork(application.unitOfWorkManager(), () -> {
+            $Type$Input inputModel = resource != null ? resource : new $Type$Input();
+            ProcessInstance<$Type$> pi = process.createInstance(businessKey, inputModel.toModel());
+            String startFromNode = httpHeaders.getHeaderString("X-KOGITO-StartFromNode");
+            return (startFromNode != null ? pi.startFrom(startFromNode) : pi.start()).checkError().variables().toOutput();
```
evacchi, line 53: - I don't like the ternary operator here
- what if we let `startFromNode` accept null => invoke start() internally?.
- let's put checkError() on its line, and make it `void` for now (see comment above)

kogito-codegen/src/main/resources/class-templates/ReactiveRestResourceTemplate.java
```diff
@@ -37,134 +38,75 @@
...
-                        .orElse(null);
-                if (pi == null) {
-                    return null;
-                } else {
-                    pi.updateVariables(resource);
-                    return mapOutput(new $Type$Output(), pi.variables());
-                }
-            });
-        });
+        return CompletableFuture.supplyAsync(() -> UnitOfWorkExecutor.executeInUnitOfWork(application.unitOfWorkManager(), () -> process.instances().findById(id).map(pi -> pi.updateVariables(resource).toOutput()).orElse(null));
```
evacchi, line 88: please format this on more lines -- even it's a oneliner

kogito-codegen/src/main/resources/class-templates/ReactiveRestResourceTemplate.java
```diff
@@ -37,134 +38,75 @@
...
-                        .orElse(null);
-                if (pi == null) {
-                    return null;
-                } else {
-                    pi.abort();
-                    return getModel(pi);
-                }
-            });
-        });
+        return CompletableFuture.supplyAsync(() -> UnitOfWorkExecutor.executeInUnitOfWork(application.unitOfWorkManager(), () -> process.instances().findById(id).map(pi -> pi.abort().variables().toOutput()).orElse(
```
evacchi, line 79: please format this on more lines -- even it's a oneliner

kogito-codegen/src/main/resources/class-templates/RestResourceTemplate.java
```diff
@@ -44,29 +49,19 @@
...
+            ProcessInstance<$Type$> pi = process.createInstance(businessKey, inputModel.toModel());
             String startFromNode = httpHeaders.getHeaderString("X-KOGITO-StartFromNode");
-
-            if (startFromNode != null) {
-                pi.startFrom(startFromNode);
-            } else {
-                pi.start();
-            }
-            return getModel(pi);
+            return (startFromNode != null ? pi.startFrom(startFromNode) : pi.start()).checkError().variables().toOutput();
```
evacchi, line 56: same comment about ternary operator and making checkError() void

kogito-codegen/src/main/resources/class-templates/RestResourceTemplate.java
```diff
@@ -76,29 +71,23 @@
...
 
     @DELETE()
     @Path("/{id}")
     @Produces(MediaType.APPLICATION_JSON)
     public $Type$Output deleteResource_$name$(@PathParam("id") final String id) {
-        return org.kie.kogito.services.uow.UnitOfWorkExecutor.executeInUnitOfWork(application.unitOfWorkManager(), () -> process.instances().findById(id).map(pi -> {
-            pi.abort();
-            return getModel(pi);
-        }).orElse(null));
+        return UnitOfWorkExecutor.executeInUnitOfWork(application.unitOfWorkManager(), () -> process.instances().findById(id).map(pi -> pi.abort().variables().toOutput()).orElse(null));
```
evacchi, line 82: please format this on more lines -- even it's a oneliner

kogito-codegen/src/main/resources/class-templates/RestResourceTemplate.java
```diff
@@ -76,29 +71,23 @@
...
     @POST()
     @Path("/{id}")
     @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
     public $Type$Output updateModel_$name$(@PathParam("id") String id, $Type$ resource) {
-        return org.kie.kogito.services.uow.UnitOfWorkExecutor.executeInUnitOfWork(application.unitOfWorkManager(), () -> process.instances().findById(id).map(pi -> {
-            pi.updateVariables(resource);
-            return mapOutput(new $Type$Output(), pi.variables());
-        }).orElse(null));
+        return UnitOfWorkExecutor.executeInUnitOfWork(application.unitOfWorkManager(), () -> process.instances().findById(id).map(pi -> pi.updateVariables(resource).toOutput()).orElse(null));
```
evacchi, line 90: please format this on more lines -- even it's a oneliner

kogito-codegen/src/main/resources/class-templates/RestResourceUserTaskTemplate.java
```diff
@@ -76,9 +78,10 @@
                                   @QueryParam("phase") @DefaultValue("abort") final String phase,
                                   @QueryParam("user") final String user,
                                   @QueryParam("group") final List<String> groups) {
-        return org.kie.kogito.services.uow.UnitOfWorkExecutor.executeInUnitOfWork(application.unitOfWorkManager(), () -> process.instances().findById(id).map(pi -> {
-            pi.transitionWorkItem(workItemId, org.jbpm.process.instance.impl.humantask.HumanTaskTransition.withoutModel(phase, policies(user, groups)));
-            return getModel(pi);
-        }).orElse(null));
+        return UnitOfWorkExecutor.executeInUnitOfWork(application.unitOfWorkManager(), () -> process.instances().findById(id).map(pi -> pi.transitionWorkItem(workItemId,
```
evacchi, line 81: check the formatting here

kogito-codegen/src/main/resources/class-templates/spring/SpringRestResourceTemplate.java
```diff
@@ -51,100 +58,65 @@
...
+            ProcessInstance<$Type$> pi = process.createInstance(businessKey, inputModel.toModel());
             String startFromNode = httpHeaders.getHeaderString("X-KOGITO-StartFromNode");
-
-            if (startFromNode != null) {
-                pi.startFrom(startFromNode);
-            } else {
-                pi.start();
-            }
-            return getModel(pi);
+            return (startFromNode != null ? pi.startFrom(startFromNode) : pi.start()).checkError().variables().toOutput();
```
evacchi, line 69: same comment on ternary + void checkError()

kogito-codegen/src/main/resources/class-templates/spring/SpringRestResourceUserTaskTemplate.java
```diff
@@ -18,41 +18,45 @@
...
         return org.kie.kogito.services.uow.UnitOfWorkExecutor.executeInUnitOfWork(application.unitOfWorkManager(), () -> {
             return process.instances().findById(id).map(pi -> {
                 pi.send(Sig.of("$taskNodeName$", java.util.Collections.emptyMap()));
                 java.util.Optional<WorkItem> task = pi.workItems().stream().filter(wi -> wi.getName().equals("$taskName$")).findFirst();
                 if (task.isPresent()) {
-                    return javax.ws.rs.core.Response.ok(getModel(pi))
-                                                    .header("Link", "</" + id + "/$taskName$/" + task.get().getId() + ">; rel='instance'")
-                                                    .build();
+                    return ResponseEntity.status(HttpStatus.OK)
+                                         .header("Link", "</" + id + "/$taskName$/" + task.get().getId() + ">; rel='instance'").body(pi.variables().toOutput())
```
evacchi, line 29: add newline before `.body`

kogito-codegen/src/main/resources/class-templates/spring/SpringRestResourceUserTaskTemplate.java
```diff
@@ -18,41 +18,45 @@
...
                                      @PathVariable("workItemId") final String workItemId,
                                      @RequestParam(value = "phase", defaultValue = "complete") final String phase,
                                      @RequestParam(value = "user", required = false) final String user,
                                      @RequestParam(value = "group", required = false) final List<String> groups,
                                      @RequestBody final $TaskOutput$ model) {
-        return org.kie.kogito.services.uow.UnitOfWorkExecutor.executeInUnitOfWork(application.unitOfWorkManager(), () -> process.instances().findById(id).map(pi -> {
-            pi.transitionWorkItem(workItemId, org.jbpm.process.instance.impl.humantask.HumanTaskTransition.withModel(phase, model.toMap(), policies(user, groups)));
-            return getModel(pi);
-        }).orElse(null));
+        return UnitOfWorkExecutor.executeInUnitOfWork(application.unitOfWorkManager(), () -> process.instances().findById(id).map(pi -> pi.transitionWorkItem(workItemId,
```
evacchi, line 45: check formatting

kogito-codegen/src/main/resources/class-templates/spring/SpringRestResourceUserTaskTemplate.java
```diff
@@ -71,9 +75,10 @@ public JsonSchema getSchemaAndPhases(@PathParam("id") final String id, @PathPara
                                   @RequestParam(value = "phase", defaultValue = "abort") final String phase,
                                   @RequestParam(value = "user", required = false) final String user,
                                   @RequestParam(value = "group", required = false) final List<String> groups) {
-        return org.kie.kogito.services.uow.UnitOfWorkExecutor.executeInUnitOfWork(application.unitOfWorkManager(), () -> process.instances().findById(id).map(pi -> {
-            pi.transitionWorkItem(workItemId, org.jbpm.process.instance.impl.humantask.HumanTaskTransition.withoutModel(phase, policies(user, groups)));
-            return getModel(pi);
-        }).orElse(null));
+        return UnitOfWorkExecutor.executeInUnitOfWork(application.unitOfWorkManager(), () -> process.instances().findById(id).map(pi -> pi.transitionWorkItem(workItemId,
```
evacchi, line 78: check formatting

api/kogito-api/src/main/java/org/kie/kogito/MapOutput.java
fjtirado: That was the original name, but then I realize that this interface is just adding a toMap method to any class that implements it, model or not, so it is far more generic and I think clear about the purpose. 
I have to update the comment though, wdyt?

api/kogito-api/src/main/java/org/kie/kogito/MapInput.java
fjtirado: Same rationale than in previous comment

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
fjtirado: Hmmmmmm, 
Two statements in the same line is not a good idea, too confusing. 
I agree on rationale to not make start methods fluent, but then we have to assume a longer generated code (which is probably fine anyway)
Regarding methods in processIntances that I change butt are not start or abort, I think they can be fluent without compromising api consistency. lets analyze them indivdiaully:
 updateVariables can return updated variables and checkError only makes sense if it returns ProcessInstance, if not, the  logic inside it (checking if there is any eeor)  should be make part of start method itself

kogito-codegen/src/main/resources/class-templates/ReactiveRestResourceTemplate.java
fjtirado: Agree, excepto for checkError part (see coment abovce)

kogito-codegen/src/main/resources/class-templates/RestResourceUserTaskTemplate.java
fjtirado: see comment above

kogito-codegen/src/main/resources/class-templates/spring/SpringRestResourceUserTaskTemplate.java
fjtirado: same

kogito-codegen/src/main/resources/class-templates/spring/SpringRestResourceUserTaskTemplate.java
fjtirado: same

kogito-codegen/src/main/resources/class-templates/spring/SpringRestResourceUserTaskTemplate.java
fjtirado: same

api/kogito-api/src/main/java/org/kie/kogito/MapOutput.java
fjtirado: documentation updated

api/kogito-api/src/main/java/org/kie/kogito/MapInput.java
fjtirado: documentation updated

api/kogito-api/src/main/java/org/kie/kogito/MapOutput.java
evacchi: it really boils down to whether this method will be provided by anything that exposes toMap, or only by Output models or Models. Anyway we can always update

MarianMacik: Here it sounds to me like a class implementing this interface can be an output of a map, not that it provides this method. I would suggest something like `MapOutputAware` or similar, but we can always update as this is just a name...

kogito-codegen/src/main/resources/class-templates/RestResourceSignalTemplate.java
```diff
@@ -32,10 +32,10 @@
     @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
     public $Type$Output signal(@PathParam("id") final String id, final $signalType$ data) {
-        return org.kie.kogito.services.uow.UnitOfWorkExecutor.executeInUnitOfWork(application.unitOfWorkManager(), () -> {
+        return UnitOfWorkExecutor.executeInUnitOfWork(application.unitOfWorkManager(), () -> {
             return process.instances().findById(id).map(pi -> {
                 pi.send(Sig.of("$signalName$", data));
-                return getModel(pi);
+                return pi.variables().toOutput();
```
MarianMacik, line 38: Missing `checkError()`.

kogito-codegen/src/main/resources/class-templates/RestResourceTemplate.java
```diff
@@ -77,68 +79,61 @@
...
-            return getModel(pi);
-        }).orElse(null));
+        return UnitOfWorkExecutor.executeInUnitOfWork(
+                                                      application.unitOfWorkManager(),
+                                                      () -> process
+                                                                   .instances()
+                                                                   .findById(id)
+                                                                   .map(pi -> {
+                                                                       pi.abort();
+                                                                       return pi.variables().toOutput();
```
MarianMacik, line 97: Probably missing `checkError()` also here?

jbpm/jbpm-flow-builder/src/main/resources/class-templates/TaskInputTemplate.java
```diff
@@ -24,7 +24,7 @@ public String getName() {
         return this._name;
     }
 
-    public static XXXTaskInput fromMap(org.kie.kogito.process.WorkItem workItem) {
+    public static XXXTaskInput from(org.kie.kogito.process.WorkItem workItem) {
```
ricardozanini, line 27: Just one question: aren't we using `$token$` format? Just asking because `XXX` seems something different.

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ProcessToExecModelGenerator.java
```diff
@@ -101,28 +108,78 @@ public MethodDeclaration generateMethod(WorkflowProcess process) {
...
+                                                                                                         "getId")));
+            }
+            for (Variable var : vars.getTypes().values()) {
+                final String fieldName = StringUtils.ucFirst(var.getSanitizedName());
+                body.addStatement(new MethodCallExpr(returnName, "set" + fieldName).addArgument(new MethodCallExpr(null,
+                                                                                                                   "get" +
+                                                                                                                         fieldName)));
+            }
+            body.addStatement(new ReturnStmt(returnName));
+            method.setBody(body);
```
MarianMacik, line 181: Why not to leverage already existing `fromMap()` and `toMap()` methods on all 3 models (ModelInput, Model, ModelOutput).

Then in ModelInput it would look something like:
```java
public OrdersModel toModel() {
        OrdersModel result = new OrdersModel();
        result.fromMap(toMap());
        return result;
}
```
Of course in case some variables are not available, they will be just set to null in the `result.fromMap()` call, but that's OK.

And for Model:
```java
public OrdersModelOutput toOutput() {
        OrdersModelOutput result = new OrdersModelOutput();
        result.fromMap(getId(), toMap());
        return result;
}
```
Again, ModelOutput will take only what it needs from the Model map so it should be OK.

My main point is to have as little code generation code as possible as it is much less readable in general than the actual runtime code. Otherwise the method duo (`fromMap()`, `toMap()`) will be now just used to bind and unbind variables I guess.

Also in case of a lot of variable, the code can be less verbose. I know it is a generated code but I think everybody has to look at it from time to time :)

fjtirado: I think that reading a map and then using than map to fill the parameters of the class being created is very inneficient comparent with direct attribute assignment. 
The rule Im following to write generated is what code I will write if not generated at all, and if Im creating a outputmodel from a model (a subset) I would use getter an setters rather than toMap-fromMap. 
You are right, fromMap-toMap are not longer use for bind-unbiding any longer. And I think for good...

jbpm/jbpm-flow-builder/src/main/resources/class-templates/TaskInputTemplate.java
fjtirado: I guess because the replacement here can be done using setClassName method and there is not need to seach for a particular name

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ProcessToExecModelGenerator.java
```diff
@@ -101,28 +108,78 @@ public MethodDeclaration generateMethod(WorkflowProcess process) {
...
     public ModelMetaData generateInputModel(WorkflowProcess process) {
         String packageName = process.getPackageName();
-        String name = extractModelClassName(process.getId()) + "Input";
-
-        return new ModelMetaData(process.getId(), packageName, name, process.getVisibility(),
-                VariableDeclarations.ofInput((VariableScope) ((org.jbpm.process.core.Process) process).getDefaultContext(VariableScope.VARIABLE_SCOPE)),
-                true, "/class-templates/ModelNoIDTemplate.java");
+        String modelName = extractModelClassName(process.getId());
+        String name = modelName + "Input";
+        VariableDeclarations inputVars = VariableDeclarations.ofInput(getVariableScope(process));
```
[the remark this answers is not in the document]
fjtirado: Here Im using inputVars local variable because it is used as parameter twice and I do not want to generate two instances. 
In generateModel Im reusing the scope. 
I changed the format to make it more readable, hopefully....

kogito-codegen/src/main/resources/class-templates/ReactiveRestResourceTemplate.java
```diff
@@ -38,134 +39,108 @@
...
+                                                                                                      httpHeaders.getHeaderString("X-KOGITO-StartFromNode");
+                                                                                              if (startFromNode != null) {
+                                                                                                  pi.startFrom(startFromNode);
+                                                                                              } else {
+                                                                                                  pi.start();
+                                                                                              }
+                                                                                              return pi.checkError()
+                                                                                                       .variables()
+                                                                                                       .toOutput();
+                                                                                          }));
```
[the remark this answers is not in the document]
fjtirado: Ill try, but the code style is tricky in that line

kogito-codegen/src/main/resources/class-templates/ReactiveRestResourceTemplate.java
```diff
@@ -38,134 +39,108 @@
...
-    
-    protected $Type$Output getModel(ProcessInstance<$Type$> pi) {
-        if (pi.status() == ProcessInstance.STATE_ERROR && pi.error().isPresent()) {
-            throw new ProcessInstanceExecutionException(pi.id(), pi.error().get().failedNodeId(), pi.error().get().errorMessage());
-        }
-        
-        return mapOutput(new $Type$Output(), pi.variables());
+    public CompletionStage<Map<String, String>> getTasks_$name$(@PathParam("id") String id,
+                                                                @QueryParam("user") final String user,
+                                                                @QueryParam("group") final List<String> groups) {
```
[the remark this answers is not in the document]
fjtirado: Ill review that, but any change in format from previous version is probably performed by the IDE automatically according to the code style globally defined for KIE group

kogito-codegen/src/main/resources/class-templates/RestResourceSignalTemplate.java
fjtirado: good catch

kogito-codegen/src/main/resources/class-templates/RestResourceTemplate.java
fjtirado: hmm, it will not harm

kogito-codegen/src/main/resources/class-templates/ReactiveRestResourceTemplate.java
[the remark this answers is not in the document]
fjtirado: After trying during a while, I formatted it manuall, 
Hope the result is more read friendly

kogito-codegen/src/main/resources/class-templates/ReactiveRestResourceTemplate.java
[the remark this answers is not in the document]
fjtirado: After trying during a while, I formatted it manually, hope the result is more read friendly

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/humantask/HumanTaskTransition.java
```diff
@@ -36,8 +37,8 @@
     private Map<String, Object> data;
     private List<Policy<?>> policies = new ArrayList<>();
     
-    public static HumanTaskTransition withModel(String phase, Map<String, Object> data, Policy<?>... policies) {
-        return new HumanTaskTransition(phase, data, policies);
+    public static <T extends MapOutput> HumanTaskTransition withModel(String phase, T data, Policy<?>... policies) {
+        return new HumanTaskTransition(phase, data.toMap(), policies);
```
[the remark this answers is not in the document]
fjtirado: I think it was cooler that way, but you are right, uneeded and more readeable with MapOutput

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ProcessToExecModelGenerator.java
fjtirado: toMap method is still used for human task transition. 
fromMap can be probably removed, but I think this should be done in a different JIRA

fjtirado:
jenkins retest this

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ProcessToExecModelGenerator.java
MarianMacik: I think that reading and writing to a HashMap is basically a constant operation, but I also see your point. Bear in mind that toMap and fromMap are still used in generated bind/unbind methods of process instance, so we cannot remove them.

fjtirado:
jenkins retest this

PR #642, opened 2020-07-20
[KOGITO-2745] - Introducing CloudEvent processing through HTTP

ricardozanini:
**Jul 29th Updates**
First integration tests are OK. We can send CloudEvent messages to a given process through HTTP, having the correct headers. Outcoming is a problem right now since smallrye-http does not add `Content-Type`. Also the producer are ignoring `type` and `source`, unfortunately they are required by the spec. The target service are rejecting our produced events.

**Jul 30th Updates**
Finished the decoration work for HTTP CloudEvents messages outcoming from smallrye infra. I'm not 100% happy with this implementation since we are lacking some refactoring in the code-gen side. Should do the trick now, and we can come back to this later once @evacchi's work is finished. The decoration worked and now the outcoming messages have the correct content-type. The messages are being sent without `source` and `type` attributes, being rejected by Knative Serving. I'll assume `/process/< processinstance>` for source and `type` being the channel name.

**Jul 31th Updates**
Finished the CE SDK migration, but the Resteasy integration on Quarkus is not working, need more work and a clean reproducer to figure out why. I guess that the providers are not being registered, plus an extension might be need. With the smallrye work upcoming, I'd rather to wait and go with the approach I took in this PR: manually call their converters/builders. Is working fine.

Now I need to clear the clutter, figure why CI is failing, fix Sonar issues and write the technical docs for my friends to review this PR.

This is how an event processed by our engine looks like on Knative:

```
☁️  cloudevents.Event
Validation: valid
Context Attributes,
  specversion: 1.0
  type: process.travelers.processedtravellers
  source: /process/Travelers/f4533dca-f09d-4701-a61f-7f096fadd694
  id: 3cec97b9-78c0-4aa3-b8fd-c9506fa29b3f
  time: 2020-07-31T23:21:34.874877Z
Extensions,
  kogitoprocessid: Travelers
  kogitoprocessinstanceid: f4533dca-f09d-4701-a61f-7f096fadd694
  kogitoprocessinstancestate: 1
Data,
  {"firstName":"Hello World!!","lastName":null,"email":null,"nationality":null,"processed":true}
```

Now we have context on `type` and `source`. Also a custom extension. :)
I believe that in the future refactoring work, we could trim the attributes we have in this extension and use the ones provided by the spec instead.

**Aug 3rd Updates**
Clean up Sonar issues, need to increase unit test a little bit since some files I've changed there's no test at all.

TODO: 

[...]

PR #676, opened 2020-08-05
KOGITO-2959 Remove redundant smallrye dependency from runtimes

evacchi:


evacchi:
re: https://github.com/kiegroup/kogito-runtimes/pull/642/files/b94b82e1a354895f2816f937644f7b8d2834745a#r465776360

a fix may be that we do not put it in <dependenycManagement> but we explicitly depend on it using the version property. This way importing the bom won't conflict with other boms. WDYT? @radtriste @ricardozanini 

see also https://issues.redhat.com/browse/KOGITO-2726

pom.xml
```diff
@@ -120,7 +120,6 @@
     <version.io.quarkus>1.6.0.Final</version.io.quarkus>
     <version.io.restassured>4.3.0</version.io.restassured>
     <version.io.restassured.springboot>3.3.0</version.io.restassured.springboot>    <!-- Manually override to 3.3.0 because Spring boot uses rest-assured-common@3.3.0, otherwise error: The type io.restassured.common.mapper.TypeRef cannot be resolved. It is indirectly referenced from required .class files -->
-    <version.io.smallrye.reactive>2.1.1</version.io.smallrye.reactive>
```
ricardozanini, line 123: Just a reminder to add it back! #642

ricardozanini:
We need the version since `smallrye-reactive-messaging-http` is not available on Quarkus.

evacchi:
folded this inside #682

PR #677, opened 2020-08-06
KOGITO-2968: Duplicate WorkItemHandlerNotFoundException causes native-image to break

evacchi:
https://issues.redhat.com/browse/KOGITO-2968

We delete the exception and create a new exception that extends the one in v7.
Future work: remove all dependency from v7-age APIs from processes. https://issues.redhat.com/browse/KOGITO-2972


evacchi:
jenkins please retest this

PR #681, opened 2020-08-06
KOGITO-2979 Use jar processing in quarkus extension

evacchi:
https://issues.redhat.com/browse/KOGITO-2979

hot fix for an issue when processing serverless workflow causing classes not to be generated when issuing maven package with quarkus

PR #688, opened 2020-08-10
KOGITO-3055  Refactor Quarkus Extension

evacchi:
I'll be away until end of August, so I invite you to take a look, and if something looks off, I would still encourage you to merge __and then__ amend with further PRs, because I believe this refactor is long overdue -- unless there are larger issues. Alternatively, whoever has right may push on top of this. 

- I have moved and renamed a lot of code into methods and/or their own classes. It is a bit tidier, although it can be further improved. E.g. I have created a Quarkus-specific InMemoryCompiler that delegates to the Drools' Java Compiler wrapper hiding some internal flags.

- I have also removed usages of the `XGenerator#fromJar()` method; instead I have added an utility method to read a <Path, Resource> pair (called CollectedResource for lack of a better name) from Paths. If a Path is actually a Jar, then the the resource is read from the Jar directly into a Resource, transparently.

   The new usage pattern is `XGenerator.fromCollectedResources(CollectedResources.fromPaths(Path...))`. This is cleaner and more transparent.
 
   We can probably deprecate most of the `XGenerator.fromY()` methods in favor of this one.

- I have also added a new ResourceType SW (serverless workflow). Because SW uses a double extension (sw.yaml, sw.json etc.) I have removed the check that raised an exception if the extension contained a dot (because now it does).
Incidentally, this also means that the method to match a file name to a resource type is a bit less cheap because you cannot just do lastIndexOf('.').

### DIFF
--- a/api/kogito-api/src/main/java/org/kie/api/io/ResourceType.java
+++ b/api/kogito-api/src/main/java/org/kie/api/io/ResourceType.java
@@ -83,9 +83,6 @@ public class ResourceType
 
         CACHE.put( resourceType, resource );
         resource.getAllExtensions().forEach( ext -> {
-            if (ext.contains( "." )) {
-                throw new IllegalArgumentException( "A resource extension cannot contain a dot. Found: " + ext );
-            }
             CACHE.put( "." + ext, resource );
         } );
         return resource;
@@ -264,6 +261,14 @@ public class ResourceType
                                                                      "src/main/resources",
                                                                      "feel");
 
+    /** Serverless Workflow language */
+    public static final ResourceType SW = addResourceTypeToRegistry("SW",
+                                                                      false,
+                                                                      "Serverless Workflow",
+                                                                      "src/main/resources",
+                                                                      "sw.yml", "sw.yaml", "sw.json");
+
+
     /** NO-Operation ResourceType - used for example to dynamically disable a given AssemblerService */
     public static final ResourceType NOOP = addResourceTypeToRegistry("NOOP",
                                                                       false,
@@ -281,8 +286,10 @@ public class ResourceType
     }
 
     public static ResourceType determineResourceType(final String resourceName) {
-        int dotPos = resourceName.lastIndexOf( '.' );
-        return dotPos < 0 ? null : CACHE.get( resourceName.substring( dotPos ) );
+        return CACHE.values().stream()
+                .filter(r -> r.matchesExtension(resourceName))
+                .findFirst()
+                .orElse(null);
     }
 
     public boolean matchesExtension(String resourceName) {
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/io/CollectedResource.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/io/CollectedResource.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.kie.kogito.codegen.io;
 
 import java.io.IOException;
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/io/CollectedResource.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/io/CollectedResource.java
@@ -35,8 +35,15 @@ import org.kie.api.io.ResourceType;
 import static org.drools.core.util.IoUtils.readBytesFromInputStream;
 import static org.kie.api.io.ResourceType.determineResourceType;
 
+/**
+ * A (Path basePath, Resource resource) pair
+ */
 public class CollectedResource {
 
+    /**
+     * Returns a collection of CollectedResource from the given paths.
+     * If a path is a jar, then walks inside the jar.
+     */
     public static Collection<CollectedResource> fromPaths(Path... paths) {
         Collection<CollectedResource> resources = new ArrayList<>();
 
@@ -55,6 +62,9 @@ public class CollectedResource {
         return resources;
     }
 
+    /**
+     * Returns a collection of CollectedResource from the given jar file.
+     */
     public static Collection<CollectedResource> fromJarFile(Path jarPath) {
         Collection<CollectedResource> resources = new ArrayList<>();
         try (ZipFile zipFile = new ZipFile(jarPath.toFile())) {
@@ -76,6 +86,9 @@ public class CollectedResource {
         }
     }
 
+    /**
+     * Returns a collection of CollectedResource from the given directory.
+     */
     public static Collection<CollectedResource> fromDirectory(Path path) {
         Collection<CollectedResource> resources = new ArrayList<>();
         try {
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/io/CollectedResource.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/io/CollectedResource.java
@@ -1,0 +1,91 @@
+package org.kie.kogito.codegen.io;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Enumeration;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+
+import org.drools.core.io.impl.ByteArrayResource;
+import org.drools.core.io.impl.FileSystemResource;
+import org.drools.core.io.internal.InternalResource;
+import org.kie.api.io.Resource;
+import org.kie.api.io.ResourceType;
+
+import static org.drools.core.util.IoUtils.readBytesFromInputStream;
+import static org.kie.api.io.ResourceType.determineResourceType;
+
+public class CollectedResource {
+
+    public static Collection<CollectedResource> fromPaths(Path... paths) {
+        Collection<CollectedResource> resources = new ArrayList<>();
+
+        for (Path path : paths) {
+            if (path.toFile().isDirectory()) {
+                Collection<CollectedResource> res = fromDirectory(path);
+                resources.addAll(res);
+            } else if (path.getFileName().toString().endsWith(".jar")) {
+                Collection<CollectedResource> res = fromJarFile(path);
+                resources.addAll(res);
+            } else {
+                throw new IllegalArgumentException("Expected directory or archive, file given: " + path);
+            }
+        }
+
+        return resources;
+    }
+
+    public static Collection<CollectedResource> fromJarFile(Path jarPath) {
+        Collection<CollectedResource> resources = new ArrayList<>();
+        try (ZipFile zipFile = new ZipFile(jarPath.toFile())) {
+            Enumeration<? extends ZipEntry> entries = zipFile.entries();
+            while (entries.hasMoreElements()) {
+                ZipEntry entry = entries.nextElement();
+                ResourceType resourceType = determineResourceType(entry.getName());
+                if (resourceType != null) {
+                    InternalResource resource = new ByteArrayResource(readBytesFromInputStream(zipFile.getInputStream(entry)));
+                    resource.setResourceType(resourceType);
+                    resource.setSourcePath(entry.getName());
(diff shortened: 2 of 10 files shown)


kogito-quarkus-extension/deployment/src/main/java/org/kie/kogito/quarkus/deployment/AppPaths.java
```diff
@@ -0,0 +1,172 @@
+package org.kie.kogito.quarkus.deployment;
```
evacchi, line 1: moved to its own class, instead of a static private class

api/kogito-api/src/main/java/org/kie/api/io/ResourceType.java
```diff
@@ -83,9 +83,6 @@ public static ResourceType addResourceTypeToRegistry(final String resourceType,
 
         CACHE.put( resourceType, resource );
         resource.getAllExtensions().forEach( ext -> {
-            if (ext.contains( "." )) {
```
evacchi, line 86: .sw.json, sw.yaml, etc. need the dot

api/kogito-api/src/main/java/org/kie/api/io/ResourceType.java
```diff
@@ -281,8 +286,10 @@ public static ResourceType getResourceType(final String resourceType) {
     }
 
     public static ResourceType determineResourceType(final String resourceName) {
-        int dotPos = resourceName.lastIndexOf( '.' );
-        return dotPos < 0 ? null : CACHE.get( resourceName.substring( dotPos ) );
+        return CACHE.values().stream()
```
evacchi, line 289: therefore, the simple lastIndexOf() check does not work

kogito-quarkus-extension/deployment/src/main/java/org/kie/kogito/quarkus/deployment/GeneratedFileWriter.java
```diff
@@ -0,0 +1,105 @@
...
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Collection;
+
+import org.kie.kogito.codegen.GeneratedFile;
+
+public class GeneratedFileWriter {
```
evacchi, line 27: writes a GeneratedFile to the right path, looking at its GeneratedFile.Type

evacchi:
I am adding a bunch of reviewers, mostly so that we're all on the same page.

@mariofusco take a look at the refactoring
@tarilabs make sure that all the DMN-specific sections are still there and they still work as intended
@gitgabrio I think the fixes that I have applied should make it possible to simplify some of the exposed methods in PredictionCodegen (probably also in the other `XCodegen`s)
@ricardozanini + @tsurdilo this should still work as intended w.r.t this fix https://github.com/kiegroup/kogito-runtimes/pull/681
@fjtirado make sure your fix to native image has not gone lost (I have added it, but I don't want to miss something)

kogito-codegen/src/main/java/org/kie/kogito/codegen/io/CollectedResource.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+        return resources;
+    }
+
+    private final Path basePath;
+    private final Resource resource;
+
+    public CollectedResource(Path basePath, Resource resource) {
+        this.basePath = basePath;
+        this.resource = resource;
```
tarilabs, line 110: sorry for the stupid question, but what is the difference between this new `CollectedResource` and a Kie api `Resource`? Or to better frame the question, why this class is not a specialization of the Kie api `Resource`?

evacchi: Try not to be misled by the name. I have used `CollectedResource` for lack of a better name; `PairPathResource` would be just as good, to me. it's a decorator so that it can be any Resource; I didn't extend `Resource` itself because I didn't think having a basePath would make sense all the time (and I wouldn't know what to provide if it's null). 

But above all, I observed the pattern of pairing Resources/Models with a base Path from DMN and PMML and extracted it to a common class e.g. see `parseDecisions(Path path, List<Resource> resources)` https://github.com/kiegroup/submarine-runtimes/blob/53043bb8fd2f1e786010998158774530ece48ebc/kogito-codegen/src/main/java/org/kie/kogito/codegen/decision/DecisionCodegen.java#L135

I figured that if the Resource was enough, then we wouldn't have both arguments.

kogito-codegen/src/main/java/org/kie/kogito/codegen/io/CollectedResource.java
```diff
@@ -0,0 +1,120 @@
...
+import org.kie.api.io.Resource;
+import org.kie.api.io.ResourceType;
+
+import static org.drools.core.util.IoUtils.readBytesFromInputStream;
+import static org.kie.api.io.ResourceType.determineResourceType;
+
+/**
+ * A (Path basePath, Resource resource) pair
+ */
+public class CollectedResource {
```
evacchi, line 41: I could not find a better name for this <Path, Resource> pair. feel free to propose a better one.

kogito-codegen/src/main/java/org/kie/kogito/codegen/prediction/PredictionCodegen.java
```diff
@@ -86,6 +88,14 @@ public PredictionCodegen(List<PMMLResource> resources) {
...
 
+    public static PredictionCodegen ofCollectedResources(Collection<CollectedResource> resources) {
+        List<PMMLResource> dmnResources = resources.stream()
+                .filter(r -> r.resource().getResourceType() == ResourceType.PMML)
+                .flatMap(r -> parsePredictions(r.basePath(), Collections.singletonList(r.resource())).stream())
+                .collect(toList());
+        return ofPredictions(dmnResources);
+    }
+
     public static PredictionCodegen ofJar(Path... jarPaths) throws IOException {
```
ricardozanini, line 99: I think this is the same as `PredictionCodegen.ofCollectedResource(CollectedResource.fromJar(...))`

We could refactor these functions and annotate with `@Deprecate`

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
```diff
@@ -108,6 +111,24 @@
...
+                                .filter(e -> resource.getSourcePath().endsWith(e.getKey()))
+                                .map(e -> parseWorkflowFile(resource, e.getValue()));
+                    }
+                })
+                .collect(toList());
+
+        return ofProcesses(processes);
+    }
+
     public static ProcessCodegen ofJar(Path... jarPaths) {
```
ricardozanini, line 132: Same here: ProcessCodegen.ofCollectedResource(CollectedResource.fromJar(...))

kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/IncrementalRuleCodegen.java
```diff
@@ -89,6 +90,14 @@
...
 
+    public static IncrementalRuleCodegen ofCollectedResources(Collection<CollectedResource> resources) {
+        List<Resource> dmnResources = resources.stream()
+                .map(CollectedResource::resource)
+                .filter(IncrementalRuleCodegen::isSupportedResourceType)
+                .collect(toList());
+        return ofResources(dmnResources);
+    }
+
     public static IncrementalRuleCodegen ofJar(Path... jarPaths) {
```
ricardozanini, line 101: Same here :D `@Deprecate` and CollectedResource new API.

kogito-codegen/src/main/java/org/kie/kogito/codegen/io/CollectedResource.java
ricardozanini: I like this name.

kogito-codegen/src/main/java/org/kie/kogito/codegen/prediction/PredictionCodegen.java
evacchi: maybe let's file a JIRA for that

evacchi:
I can't believe it's green :green_heart:

kogito-codegen/src/main/java/org/kie/kogito/codegen/io/CollectedResource.java
```diff
@@ -0,0 +1,120 @@
...
+     * If a path is a jar, then walks inside the jar.
+     */
+    public static Collection<CollectedResource> fromPaths(Path... paths) {
+        Collection<CollectedResource> resources = new ArrayList<>();
+
+        for (Path path : paths) {
+            if (path.toFile().isDirectory()) {
+                Collection<CollectedResource> res = fromDirectory(path);
+                resources.addAll(res);
+            } else if (path.getFileName().toString().endsWith(".jar")) {
```
mariofusco, line 54: I suspect this wouldn't be completely correct since you're missing the case of the file ending with ".jar.original" as covered by the AppPaths at line 133 ( https://github.com/kiegroup/kogito-runtimes/pull/688/files#diff-02ab48c8cb4aa88ed360c563c29777b1R133 ). More in general wouldn't be better to use the enumeration already defined there to avoid duplicating that logic?

danielezonca: This is one of the two jar related points I have added to this ticket
https://issues.redhat.com/browse/KOGITO-3112
Fine for me to verify it (aka test uber-jar native mode with this code) during this PR but this means to wait Edoardo to be back.
What do you prefer?

fjtirado: I vote for merging this now and work out pending improvements throug JIRAs. 
I have some refactor to do while Edoardo is away and it colides with these changes, so please lets merge as soon as possible

fjtirado (approved):

mariofusco (approved):

kogito-quarkus-extension/deployment/src/main/java/org/kie/kogito/quarkus/deployment/KogitoAssetsProcessor.java
```diff
@@ -193,259 +386,96 @@ public RuntimeInitializedClassBuildItem runtimeInitializedClass() {
...
+                new ReflectiveClassBuildItem(true, true, "org.kie.kogito.services.event.impl.UserTaskInstanceEventBody"));
+
+        // not sure there is any generated class directly inheriting from AbstractDataEvent, keeping just in case
+        index.getAllKnownSubclasses(DotName.createSimple(org.kie.kogito.event.AbstractDataEvent.class.getCanonicalName()))
+                .forEach(c -> reflectiveClass.produce(
+                        new ReflectiveClassBuildItem(true, true, c.name().toString())));
+
+        index.getAllKnownSubclasses(DotName.createSimple(org.kie.kogito.services.event.AbstractProcessDataEvent.class.getCanonicalName()))
+                .forEach(c -> reflectiveClass.produce(
+                        new ReflectiveClassBuildItem(true, true, c.name().toString())));
```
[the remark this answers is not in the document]
danielezonca: Done

ricardozanini:
Guys I believe we can merge this? @mariofusco @danielezonca

danielezonca:
@ricardozanini 
I'm fine with this PR but we should wait a QE approval (/cc @radtriste @MarianMacik ).
If you are blocked by this PR we can decide to proceed considering that should be low risky (it is mainly an internal refactor)

kogito-quarkus-extension/deployment/src/main/java/org/kie/kogito/quarkus/deployment/AppPaths.java
```diff
@@ -0,0 +1,145 @@
...
+                    classesPaths.add(path);
+                    projectPaths.add(path.getParent().getParent());
+                    break;
+                }
+                case TEST_CLASSES: {
+                    projectPaths.add(path.getParent().getParent());
+                    break;
+                }
+                case JAR: {
+                    isJar = true;
```
MarianMacik, line 65: Just a side note, if there are more paths in the PathsCollection and only one of them is a JAR, then this flag is set for all paths, which may break the getPath() method as it will either treat all paths as jars or as classes paths. Maybe it doesn't cause any issues now, but it may in future... For sure it behaves differently, as without a JAR it will return `src/main/resources` directories, but with at least one JAR it will return JAR file paths and classes in `target\classes` - i.e. just entries stored in the `classesPaths` list.

MarianMacik (approved):
Looks good! Just one side note which can be answered once @evacchi is back.

kogito-quarkus-extension/deployment/src/main/java/org/kie/kogito/quarkus/deployment/AppPaths.java
danielezonca: @MarianMacik 
I have created this ticket, I think it covers your question too https://issues.redhat.com/browse/KOGITO-3112
Wdyt?

MarianMacik: Yes, thanks @danielezonca!

PR #678, opened 2020-08-06
KOGITO-2973: Set default container images in TestResources

Sgitario:
JIRA: https://issues.redhat.com/browse/KOGITO-2973
Description: The images for containers are set at pom level. However, when running the tests via IDE, these images are not injected. This task is about to default the images.
When running the tests via maven, it won't make any change. But when running the tests via IDE, it will continue working as expected.

Many thanks for submitting your Pull Request :heart:! 

Please make sure that your PR meets the following requirements:

- [x] You have read the [contributors guide](CONTRIBUTING.md)
- [x] Pull Request title is properly formatted: `KOGITO-XYZ Subject`
- [x] Pull Request title contains the target branch if not targeting master: `[0.9.x] KOGITO-XYZ Subject`
- [x] Pull Request contains link to the JIRA issue
- [ ] Pull Request contains link to any dependent or related Pull Request
- [x] Pull Request contains description of the issue
- [x] Pull Request does not include fixes for issues other than the main ticket

PR #692, opened 2020-08-11
KOGITO-3009 Consolidate Docker image tests to use kogito-test-utils

cristianonicolai:


cristianonicolai:
jenkins please retest this

kogito-test-utils/src/main/java/org/kie/kogito/testcontainers/quarkus/KeycloakQuarkusTestResource.java
```diff
@@ -40,12 +57,21 @@ protected String getKogitoPropertyValue() {
         return String.format("http://localhost:%s/auth/realms/kogito", getTestResource().getMappedPort());
     }
 
+    @Override
+    public Map<String, String> start() {
+        final Map<String, String> start = super.start();
```
Sgitario, line 62: There is an issue here. If the container is disabled by configuration, the start will be called, but the inner container won't be started. 
A possible solution to this could be:
- Make the start method "final" in ConditionalQuarkusTestResource and ConditionalSpringBootTestResource, so we won't allow to override it.
- Add a new method preStart in ConditionalQuarkusTestResource and ConditionalSpringBootTestResource which is executed always when the test resource is enabled. 
- The new preStart method could return the map and the start method could join the maps.

What do you think?

pom.xml
```diff
@@ -154,8 +154,7 @@
 
     <!-- container images for testing -->
     <container.image.infinispan>quay.io/infinispan/server:${version.org.infinispan.image}</container.image.infinispan>
-    <container.image.keycloak.version>8.0.1</container.image.keycloak.version>
```
Sgitario, line 157: I was following the approach of having the arguments "container.image.[COMPONENT].version" (see [this PR](https://github.com/kiegroup/kogito-runtimes/pull/678/files) to consolidate the one in Infinispan).
I don't mind to use either "version.org.[COMPONENT]" or "container.image.[COMPONENT].version" but let's use the same approach. 
If you prefer using "version.org.[COMPONENT]", I would need to make some changes in some PRs (as it implies changes also in kogito-apps and kogito-examples).

kogito-test-utils/src/main/java/org/kie/kogito/testcontainers/KafkaContainer.java
```diff
@@ -18,26 +18,24 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.testcontainers.containers.KafkaContainer;
 import org.testcontainers.containers.output.Slf4jLogConsumer;
 
 /**
  * Kafka Container for Kogito examples.
  */
-public class KogitoKafkaContainer extends KafkaContainer implements TestResource {
+public class KafkaContainer extends org.testcontainers.containers.KafkaContainer implements TestResource {
```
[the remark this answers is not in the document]
cristianonicolai: so I think we should rename Infinispan and Keyckloack to Kogito* ? Its just that the naming is not consistent between these. I dont really have a strong preference for one way or another.

pom.xml
cristianonicolai: As discussed, we will prefer to use "version.org.[COMPONENT]"

Sgitario: +1

kogito-test-utils/src/main/java/org/kie/kogito/testcontainers/KafkaContainer.java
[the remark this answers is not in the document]
Sgitario: I would rather to append Kogito to Infinispan and Keycloak test containers instead of having a name conflict with KafkaContainer. But that's only my feeling :)

kogito-test-utils/src/main/java/org/kie/kogito/testcontainers/quarkus/KeycloakQuarkusTestResource.java
cristianonicolai: In that case start map would be empty so the entire logic here would be skipped. Am I missing something?

Sgitario: Well, we're relying on an underlying implementation detail: start() returns an empty map. Tomorrow, it might return an inmmutable map and this code would return an exception. Or maybe we could decide to inject the parameters in the start() method.

kogito-test-utils/src/main/java/org/kie/kogito/testcontainers/KafkaContainer.java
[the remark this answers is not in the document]
cristianonicolai: np, added Kogito* as prefix for container names.

kogito-test-utils/src/main/java/org/kie/kogito/testcontainers/quarkus/KeycloakQuarkusTestResource.java
cristianonicolai: ok, for now I changed the implementation so that it doesnt matter if the returned map is immutable or not. Regarding future improvements, we can look into that as needed :)

cristianonicolai:
jenkins please retest this

PR #702, opened 2020-08-18
KOGITO-2720 Apps and Examples checked in parallel

radtriste:
https://issues.redhat.com/browse/KOGITO-2720

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -37,33 +37,43 @@ pipeline {
         }
         stage('Build kogito-runtimes') {
             steps {
-                mavenCleanInstall("kogito-runtimes", false, ["run-code-coverage"])
-                runMaven("validate", "kogito-runtimes", false, ["sonarcloud-analysis"], "-e -nsu")
+                // TODO enable tests
+                mavenCleanInstall("kogito-runtimes", true, ["run-code-coverage"])
+                runMaven("validate", "kogito-runtimes", true, ["sonarcloud-analysis"], "-e -nsu")
             }
         }
-        stage('Build integration-tests with persistence') {
-            steps {
-                mavenCleanInstall("integration-tests", false, ["persistence"])
-            }
-        }
-        stage('Build kogito-apps') {
-            steps {
-                mavenCleanInstall("kogito-apps")
-            }
-        }
-        stage('Build kogito-examples') {
-            steps {
-                mavenCleanInstall("kogito-examples")
-            }
-        }
-        stage('Build kogito-examples with persistence') {
-            steps {
-                mavenCleanInstall("kogito-examples-persistence", false, ["persistence"])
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -41,29 +41,33 @@ pipeline {
...
+        stage('Build apps and examples') {
+            parallel {
+                stage('Build integration-tests with persistence') {
+                    steps {
+                        mavenCleanInstall("integration-tests", false, ["persistence"])
+                    }
+                }
+                stage('Build kogito-apps') {
+                    steps {
+                        mavenCleanInstall("kogito-apps")
```
danielezonca, line 53: Are you sure it is possible to build apps and example in parallel?
As far as I remember some examples are using data-index and probably other so it should not work.
Can you please verify?

cristianonicolai: it should be possible now.... the dependency for kogito-apps in the examples should be at runtime only. Any integration/test should be in kogito-apps itself as @tiagodolphine showed.
of course there might be a challange to get things stable in regards to random ports/conflicts

radtriste: issue is part of epic https://issues.redhat.com/browse/KOGITO-2715 and it was decided to go for that way. but indeed, there may be some work to be done on port conflicts ...

radtriste: not sure I have the time though to handle the port conflicts ... (leaving on holidays on Friday)

radtriste:
Jenkins retest this

Jenkinsfile
danielezonca: Actually I was thinking at test automation for some end to end example like travel agency. Btw you are probably right that now with integration tests module in kogito-apps we can verify the integration there :+1:

radtriste:
Almost there but getting `java.io.FileNotFoundException: Too many open files` :(

radtriste:
Jenkins retest this

radtriste:
The parallel checks are taking much more time than sequential. It won't be improvement unless we really raise the memory capacity (which is already 16 GB ...)

danielezonca:
Ok last build took almost than 4h 30m to run and 1 test failed too so we have not enough resources to run in parallel.
I think it is better to close this PR and stay with the current setup. Wdyt?

Sgitario:
> Ok last build took almost than 4h 30m to run and 1 test failed too so we have not enough resources to run in parallel.
> I think it is better to close this PR and stay with the current setup. Wdyt?

I have fixed the test failing in https://github.com/kiegroup/kogito-examples/pull/348. 
About the build time, I will try to do some improvements and if it's not better than 2h 30m, I think it makes sense to close this PR.

Sgitario:
Let's close this PR as no improvements have been achieved.

PR #730, opened 2020-08-27
[KOGITO-2893] Detach Business Key from ID - follow-up cleaning

MarianMacik:
Follow-up of [KOGITO-2893](https://issues.redhat.com/browse/KOGITO-2893) (as it got merged before I had a chance to review it)

As we discussed with @ruromero business key is now just set and used only when querying using data index. This means that methods removed in this PR are no longer needed.

### DIFF
--- a/jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/LightProcessRuntime.java
+++ b/jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/LightProcessRuntime.java
@@ -130,7 +130,7 @@ public class LightProcessRuntime implements InternalProcessRuntime {
     public ProcessInstance startProcess(String processId, Map<String, Object> parameters, String trigger) {
         ProcessInstance processInstance = createProcessInstance(processId, parameters);
         if (processInstance != null) {
-            processInstanceManager.addProcessInstance(processInstance, null);
+            processInstanceManager.addProcessInstance(processInstance);
             return startProcessInstance(processInstance.getId(), trigger);
         }
         return null;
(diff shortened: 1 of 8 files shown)


jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/LightProcessRuntime.java
```diff
@@ -187,7 +187,7 @@ public ProcessInstance createProcessInstance(String processId, CorrelationKey co
 
     @Override
     public ProcessInstance getProcessInstance(CorrelationKey correlationKey) {
```
ruromero, line 189: Is it too much refactoring to remove the whole method? Is it even used?
I think you just have to remove it from the CorrelationAwareProcessRuntime interface

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessRuntimeImpl.java
```diff
@@ -229,7 +229,7 @@ public ProcessInstance createProcessInstance(String processId, CorrelationKey co
 
     @Override
     public ProcessInstance getProcessInstance(CorrelationKey correlationKey) {
-        return processInstanceManager.getProcessInstance(correlationKey);
+        return null; // Querying of correlation keys is no longer supported
```
ruromero, line 232: Same, I think it can be removed

ruromero (changes requested):
I agree with the changes so far

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/LightProcessRuntime.java
MarianMacik: Yep, was considering it. But then the `CorrelationAwareProcessRuntime` interface would have methods startProcess and createProcessInstance but no query methods. If that's OK, I will remove also those.

MarianMacik:
Tests were unable to pull Infinispan docker image. Retrying.

jenkins retest this

ruromero (approved):

cristianonicolai (approved):

tiagodolphine (approved):

PR #731, opened 2020-08-28
[KOGITO-1031] Add Maven local deploy

Kevin-Mok:
See: https://issues.redhat.com/browse/KOGITO-1031

This is an accompanying PR to [the one opened in kogito-pipelines](https://github.com/kiegroup/kogito-pipelines/pull/46). 

## Changes
- deploys all artifacts locally before zipping them altogether to push to Nexus
  * added to provide authentication to push artifacts to PR artifact repository
- replace hardcoded "kogito-runtimes" string with environment variable

## Testing
I was able to [deploy the artifacts](https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/custom/job/kmok/job/KOGITO-1031-full/job/kogito-runtimes-deploy/79/execution/node/112/log/) to the PR test repository using the code from this PR.

## Requirements

### DIFF
--- a/Jenkinsfile.deploy
+++ b/Jenkinsfile.deploy
@@ -34,6 +34,7 @@ pipeline {
 
         // Deploy information
         string(name: 'MAVEN_DEPLOY_REPOSITORY', defaultValue: '', description: 'Specify a Maven repository to deploy the artifacts.')
+        string(name: 'MAVEN_REPO_CREDS_ID', defaultValue: '', description: 'Credentials for Maven repository to deploy the artifacts.')
 
         // Release information
         booleanParam(name: 'RELEASE', defaultValue: false, description: 'Is this build for a release?')
@@ -51,6 +52,8 @@ pipeline {
     }
 
     environment {
+        REPO_NAME = 'kogito-runtimes'
+
         MAVEN_OPTS = '-Xms1024m -Xmx4g'
 
         BOT_BRANCH_HASH = "${util.generateHash(10)}"
@@ -73,7 +76,7 @@ pipeline {
                         assert getProjectVersion() != ''
                     }
 
-                    checkoutRepo('kogito-runtimes')
+                    checkoutRepo(getRepoName())
                 }
             }
             post {
@@ -92,7 +95,7 @@ pipeline {
                 expression { return isRelease() }
             }
             steps {
-                prepareForPR('kogito-runtimes')
+                prepareForPR(getRepoName())
             }
         }
         stage('Update project version'){
@@ -138,12 +141,17 @@ pipeline {
             steps {
                 script {
                     // Usually `params.MAVEN_DEPLOY_REPOSITORY` is used for testing purpose
-                    if(!isRelease() || params.MAVEN_DEPLOY_REPOSITORY) {
+                    if (params.MAVEN_DEPLOY_REPOSITORY && params.MAVEN_REPO_CREDS_ID) {
+                        // Deploy to specific repository with credentials
+                        runMavenDeployLocally(getMavenCommand())
+                        maven.uploadLocalArtifacts(params.MAVEN_REPO_CREDS_ID, getLocalDeploymentFolder(), getMavenRepoZipUrl())
+                    } else if(!isRelease() || params.MAVEN_DEPLOY_REPOSITORY){
+                        // Normal deploy
                         runMavenDeploy(getMavenCommand())
                     } else {
                          // Deploy locally and then to staging
-                        runMavenDeployLocally(getMavenCommand(), 'runtimes')
-                        runMavenStage(getMavenCommand(), 'runtimes')
+                        runMavenDeployLocally(getMavenCommand())
+                        runMavenStage(getMavenCommand())
                     }
                 }
             }
@@ -153,16 +161,15 @@ pipeline {
                 expression { return isRelease() }
             }
             steps {
-                commitAndCreatePR('kogito-runtimes')
+                commitAndCreatePR(getRepoName())
             }
             post {
                 success {
                     script {
-                        // Runtimes
-                        setDeployPropertyIfNeeded('kogito-runtimes.pr.source.uri', "https://github.com/${getBotAuthor()}/kogito-runtimes")
-                        setDeployPropertyIfNeeded('kogito-runtimes.pr.source.ref', getBotBranch())
-                        setDeployPropertyIfNeeded('kogito-runtimes.pr.target.uri', "https://github.com/${getGitAuthor()}/kogito-runtimes")
-                        setDeployPropertyIfNeeded('kogito-runtimes.pr.target.ref', getBuildBranch())
+                        setDeployPropertyIfNeeded("${getRepoName()}.pr.source.uri", "https://github.com/${getBotAuthor()}/${getRepoName()}")
+                        setDeployPropertyIfNeeded("${getRepoName()}.pr.source.ref", getBotBranch())
+                        setDeployPropertyIfNeeded("${getRepoName()}.pr.target.uri", "https://github.com/${getGitAuthor()}/${getRepoName()}")
+                        setDeployPropertyIfNeeded("${getRepoName()}.pr.target.ref", getBuildBranch())
                     }
                 }
             }
@@ -213,6 +220,10 @@ boolean isRelease() {
     return params.RELEASE
 }
 
+String getRepoName(){
+    return env.REPO_NAME
+}
+
 String getGitAuthor(){
     return params.GIT_AUTHOR
 }
@@ -243,10 +254,10 @@ void setDeployPropertyIfNeeded(String key, def value){
     }
 }
 
-MavenCommand getMavenCommand(String dir = 'kogito-runtimes'){
+MavenCommand getMavenCommand(){
     return new MavenCommand(this, ['-fae'])
                 .withSettingsXmlId(params.MAVEN_SETTINGS_CONFIG_FILE_ID)
-                .inDirectory(dir)
+                .inDirectory(getRepoName())
                 .withProperty('full')
 }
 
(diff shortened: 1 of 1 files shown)


Jenkinsfile.deploy
```diff
@@ -176,15 +183,28 @@ pipeline {
...
+        }
+        stage('Push zip of artifacts to Maven repository') {
             steps {
-                mavenDeploy('kogito-runtimes')
-                mavenDeploy('kogito-apps')
-                mavenDeploy('kogito-examples')
+                withCredentials([usernameColonPassword(credentialsId: 'unpacks-zip-on-qa-nexus', variable: 'kieUnpack')]) {
+                    // unpack zip to QA Nexus
+                    sh """
+                    cd ${env.LOCAL_DEPLOY_DIR}
```
Ginxo, line 202: you should try to avoid browsing folder this way, you can do

```
dir(${env.LOCAL_DEPLOY_DIR}) {
    sh """
    zip -r kiegroup .
    curl --silent --upload-file kiegroup.zip -u ${kieUnpack} -v ${params.MAVEN_DEPLOY_REPOSITORY}/content-compressed
    """
}
```

Jenkinsfile.deploy
```diff
@@ -312,6 +332,20 @@ String getBuildBranch(){
     return params.BUILD_BRANCH_NAME
 }
 
+String getDefaultBranch(String repo) {
+    // if not triggered by PR
+    if (params.PR_TARGET_BRANCH == '') {
```
Ginxo, line 337: `if (!params.PR_TARGET_BRANCH)`

Jenkinsfile.deploy
```diff
@@ -312,6 +332,20 @@ String getBuildBranch(){
...
+        return 'master'
+    }
+    // if PR target branch exists
+    def repositoryScm = githubscm.getRepositoryScm(repo, 'kiegroup', params.PR_TARGET_BRANCH)
+    if (repositoryScm != null) {
+        return params.PR_TARGET_BRANCH
+    }
+    // if target branch doesn't exist in repos other than PR repo
+    return 'master'
```
Ginxo, line 346: I would suggest to replace 340-346 by
`return repositoryScm ? params.PR_TARGET_BRANCH : 'master'`

jenkins-tests/src/test/vars/JenkinsfileDeploy.groovy
```diff
@@ -0,0 +1,39 @@
+import com.homeaway.devtools.jenkins.testing.JenkinsPipelineSpecification
+
+class JenkinsfileDeploy extends JenkinsPipelineSpecification {
```
Ginxo, line 3: fix indentation problems please

Ginxo (changes requested):
Please check.
I can't tell about business here, only my technical POV

jenkins-tests/src/test/vars/JenkinsfileDeploy.groovy
Kevin-Mok: Same as [comment in pipelines PR](https://github.com/kiegroup/kogito-pipelines/pull/46#discussion_r494627253).

Jenkinsfile.deploy
```diff
@@ -334,11 +363,18 @@ void mavenCleanInstall(String directory, boolean skipTests = false, List profile
...
+void mavenLocalDeploy(String directory) {
+    extraArgs = "-DaltDeploymentRepository=local::default::file://${env.LOCAL_DEPLOY_DIR}"
     runMaven('clean deploy', directory, true, [], extraArgs)
 }
 
+String getMavenRepoZipUrl() {
+    String repoUrl = params.MAVEN_DEPLOY_REPOSITORY
+    // ensure URL has trailing slash
+    repoUrl += (!repoUrl.endsWith('/')) ? '/' : ''
+    return repoUrl.replaceAll('/content/', '/service/local/') + 'content-compressed'
```
Ginxo, line 375: I would do
```
String getMavenRepoZipUrl() {
     return "${params.MAVEN_DEPLOY_REPOSITORY.replaceAll('/content/', '/service/local/').replaceFirst('/*$', '')}/content-compressed"
}
```
the $ matches the end of a string and `*` in case there are multiple slashes, so it will remove it in case it exists and you always concatenate `/` after url.
This is just a suggestion.

Ginxo (approved):
LGTM, please check my suggestion.

Kaitou786 (approved):

Jenkinsfile.deploy
```diff
@@ -334,11 +363,18 @@ void mavenCleanInstall(String directory, boolean skipTests = false, List profile
...
 
-void mavenDeploy(String directory) {
-    extraArgs = params.MAVEN_DEPLOY_REPOSITORY != '' ? "-DaltDeploymentRepository=runtimes-artifacts::default::${params.MAVEN_DEPLOY_REPOSITORY}" : ''
+void mavenLocalDeploy(String directory) {
+    extraArgs = "-DaltDeploymentRepository=local::default::file://${env.LOCAL_DEPLOY_DIR}"
     runMaven('clean deploy', directory, true, [], extraArgs)
 }
 
+String getMavenRepoZipUrl() {
+    String repoUrl = params.MAVEN_DEPLOY_REPOSITORY
```
radtriste, line 372: what happens here if no `MAVEN_DEPLOY_REPOSITORY` is given ?
Usually, if none mentioned, we are using the default one defined into the pom.xml.

radtriste (changes requested):

Jenkinsfile.deploy
Kevin-Mok: I'm not sure how to go about deployment now. I tried just using the [current `mavenDeploy()`](https://github.com/kiegroup/kogito-runtimes/blob/03d1fbecd3cfdb6fccad2114359485479e3f7dd7/Jenkinsfile.deploy#L273-L276) with the parameter of the PR check repository, but this results in a [401 error](https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/custom/job/kmok/job/KOGITO-1031-full/job/kogito-runtimes-deploy/23/execution/node/162/log/) when deploying to that repo. 

[This zip and curl method](https://github.com/Kevin-Mok/kogito-runtimes/blob/3b49b28fc332bc1a5c96be31fee9d0799766a59d/Jenkinsfile.deploy#L197-L209) I was using previously uses the `unpacks-zip-on-qa-nexus` credential to authenticate itself to the repo, I'm not sure how to deploy to the default repo in the `pom.xml` in that case.  

I could create 2 different branches of deployment depending on whether or not there is a `MAVEN_DEPLOY_REPOSITORY` parameter, but that seems kind of janky. Also, would the `unpacks-zip-on-qa-nexus` credential even work universally for that?

How would you suggest I go about this?

radtriste: I guess we would also need to retrieve the user/pwd from settings.xml to upload to usual maven as it is a different one...
or have different credentials with this information to do the curl. and so I would have the creds name as optional parameter of the pipeline (if empty get from settings.xml) and the main pipeline would send directly this information (here it would send `unpacks-zip-on-qa-nexus` for this use case).

wdyt ?

Kevin-Mok: Using the zip and curl method if there are artifact repository and credentials parameters passed in requires using a `withCredentials` block, but authenticating with the retrieved credentials from `settings.xml` wouldn't. In this case, would it just be simpler to create two different branches depending on whether artifact repository and credentials parameters are passed in? That is, using a `withCredentials` block if there are and the current `mavenDeploy()` with no repository argument if there isn't. I initially thought this was not ideal, but it may be the simplest implementation after all?

radtriste: yep you could have an if statement and have 2 different behaviors. I have no problem with that ;)

Jenkinsfile.deploy
```diff
@@ -35,10 +37,14 @@ pipeline {
         // Bot author information. Set as params for easy testing.
         string(name: 'GIT_AUTHOR_BOT', defaultValue: 'bsig-gh-bot', description: 'From which author should the PR be created ?')
         string(name: 'BOT_CREDENTIALS_ID', defaultValue: 'bsig-gh-bot', description: 'Credentials for PR creation')
+
+        // PR parameter to be filled out by GitHub Pull Request Builder for PR checks
+        string(name: 'PR_TARGET_BRANCH', defaultValue: '', description: 'What is the target branch of the PR? Will be automatically filled out by GitHub Pull Request Builder.')
```
radtriste, line 42: is that still used ? as far as I can see https://github.com/kiegroup/kogito-pipelines/pull/46, you are checking before calling if the branch exist (which I prefer :))

Jenkinsfile.deploy
```diff
@@ -192,10 +214,19 @@ void saveReports(boolean allowEmpty=false){
...
 
+String getDefaultBranch(String repo) {
+    // if not triggered by PR
+    if (!params.PR_TARGET_BRANCH) {
+        return 'master'
+    }
+    def repositoryScm = githubscm.getRepositoryScm(repo, 'kiegroup', params.PR_TARGET_BRANCH)
+    return repositoryScm ? params.PR_TARGET_BRANCH : 'master'
+}
+
```
radtriste, line 229: I guess that one could then be deleted

Jenkinsfile.deploy
```diff
@@ -271,10 +302,18 @@ void mavenCleanInstall(String directory, boolean skipTests = false, List profile
 }
 
 void mavenDeploy(String directory) {
-    extraArgs = params.MAVEN_DEPLOY_REPOSITORY != '' ? "-DaltDeploymentRepository=runtimes-artifacts::default::${params.MAVEN_DEPLOY_REPOSITORY} -Denforcer.skip=true" : ''
```
radtriste, line 274: This should not be deleted because in case of `MAVEN_DEPLOY_CREDS_ID` not defined we may still need to deploy to `MAVEN_DEPLOY_REPOSITORY`

radtriste (changes requested):

Jenkinsfile.deploy
Kevin-Mok: This is still needed because there are 2 repositories being checked out in runtimes. The original pipeline only checks if the branch exists for one of them (the one the PR is for). The other repository will need to fall back to checking out the target branch.

radtriste: ah I see, you want to make use of the `checkoutIfExist` and the merge capability in there.
In that case, I think that would need to be added to each deploy pipeline (not only runtimes) and added as param for all call (as you did )
also, 2 small things:
- can you change the comment on that param ? This is confusing as it is not automatic, the main pipeline is setting it
- could you move that param to the `Git information` part ?

Jenkinsfile.deploy
```diff
@@ -227,10 +238,13 @@ void setDeployPropertyIfNeeded(String key, def value){
     }
 }
 
-MavenCommand getMavenCommand(){
+MavenCommand getMavenCommand(String directory = ''){
+    if(!directory) {
```
radtriste, line 242: when do you need to set another directory ?

radtriste (approved):
one minor comment

Jenkinsfile.deploy
Kevin-Mok: We don't need to. I'll remove it from the other repositories as well.

radtriste (approved):

radtriste (approved):

PR #738, opened 2020-09-01
KOGITO-3231 GraalVM 20.2 breaks native image

evacchi:
https://issues.redhat.com/browse/KOGITO-3231 

MVEL-related code is now causing graalvm native image to blow up. 
This PR patches processes by checking a magic flag that is set only when native image is compiling; otherwise, falls back to the regular behavior.

For Drools-related code, we are awaiting for a larger patch that will make MVEL optional, so that we can actually remove it -- in kogito it's basically dead code; so for that code path, I am just using native-image's capability of live patching code using "substitutions". 

Again in Kogito all MVEL-related code is basically dead code, so I am substituting the original methods with nops (actually, throw exceptions, but they will be unreachable and they are just to make the compiler happy)

please kindly double check:

      mvn -Dversion.io.quarkus=999-SNAPSHOT -Dquarkus.version=999-SNAPSHOT -Dnative -Dquarkus.native.container-build=true install  -DskipTests


cristianonicolai (approved):
minor comment to remove comment out code.

mariofusco (approved):

PR #741, opened 2020-09-03
KOGITO-3110 Refactor ofCollectedResources methods in *Codegen

evacchi:
https://issues.redhat.com/browse/KOGITO-3110

The method ofCollectedResources is almost exactly the same in DecisionCodegen, PredictionCodegen, ProcessCodegen and IncrementalRuleCodegen.
Remove all other variations of the ofPath, ofFiles etc. and use ofCollectedResources everywhere (where applicable)

### DIFF
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/decision/DecisionContainerGenerator.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/decision/DecisionContainerGenerator.java
@@ -16,6 +16,8 @@
 package org.kie.kogito.codegen.decision;
 
 import java.io.File;
+import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.List;
@@ -91,9 +93,13 @@ public class DecisionContainerGenerator extends AbstractApplicationSection {
 
     private String getDecisionModelRelativeResourcePath(CollectedResource resource) {
         String source = getDecisionModelJarResourcePath(resource);
-        Path sourcePath = Paths.get(source).toAbsolutePath();
-        Path relativizedPath = resource.basePath().toAbsolutePath().relativize(sourcePath);
-        return "/" + relativizedPath.toString().replace(File.separatorChar, '/');
+        try {
+            Path sourcePath = Paths.get(source).toAbsolutePath().toRealPath();
+            Path relativizedPath = resource.basePath().toAbsolutePath().toRealPath().relativize(sourcePath);
+            return "/" + relativizedPath.toString().replace(File.separatorChar, '/');
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
     }
 
     private void setupExecIdSupplierVariable(ClassOrInterfaceDeclaration typeDeclaration) {
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/decision/DecisionContainerGenerator.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/decision/DecisionContainerGenerator.java
@@ -91,7 +91,8 @@ public class DecisionContainerGenerator extends AbstractApplicationSection {
 
     private String getDecisionModelRelativeResourcePath(CollectedResource resource) {
         String source = getDecisionModelJarResourcePath(resource);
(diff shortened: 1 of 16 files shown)


evacchi:
jenkins please retest this

works on my machine

kogito-codegen/src/main/java/org/kie/kogito/codegen/io/CollectedResource.java
```diff
@@ -102,6 +107,22 @@
...
+    public static Collection<CollectedResource> fromFiles(Path basePath, File... files) {
+        Collection<CollectedResource> resources = new ArrayList<>();
+        try (Stream<File> paths = Arrays.stream(files)) {
+            paths.filter(File::isFile)
+                    .map(f -> new FileSystemResource(f)
+                            .setResourceType(determineResourceType(f.getName())))
+                    .map(f -> new CollectedResource(basePath, f))
+                    .forEach(resources::add);
+        }
+        return resources;
```
tarilabs, line 122: not sure where this method is used, but I would consider if meaningful for this method to check `basePath` is in fact a valid parent of each `files`?

kogito-codegen/src/main/java/org/kie/kogito/codegen/decision/DecisionContainerGenerator.java
```diff
@@ -91,8 +93,13 @@ private String getDecisionModelJarResourcePath(CollectedResource resource) {
 
     private String getDecisionModelRelativeResourcePath(CollectedResource resource) {
         String source = getDecisionModelJarResourcePath(resource);
-        Path relativizedPath = resource.basePath().relativize(Paths.get(source));
-        return "/" + relativizedPath.toString().replace(File.separatorChar, '/');
+        try {
+            Path sourcePath = Paths.get(source).toAbsolutePath().toRealPath();
+            Path relativizedPath = resource.basePath().toAbsolutePath().toRealPath().relativize(sourcePath);
+            return "/" + relativizedPath.toString().replace(File.separatorChar, '/');
+        } catch (IOException e) {
```
tarilabs, line 100: difficult to understand why this change is required, but if all the tests are passing I guess it's super fine for me

tarilabs (approved):
approved with 2 minor comments which can be dealt separately if preferred

kogito-codegen/src/main/java/org/kie/kogito/codegen/decision/DecisionContainerGenerator.java
evacchi: toRealPath() throws an checked IOExcepetion, the reason for the change is that issue with symlinks :)

kogito-codegen/src/main/java/org/kie/kogito/codegen/io/CollectedResource.java
evacchi: good point. added as an invariant check on the costructor 2eb4ad1d09

gitgabrio (approved):
@evacchi 
LGTM. Thanks!!!

kogito-codegen/src/main/java/org/kie/kogito/codegen/prediction/PredictionCodegen.java
```diff
@@ -96,55 +83,10 @@ public static PredictionCodegen ofCollectedResources(Collection<CollectedResourc
         return ofPredictions(dmnResources);
     }
 
-    public static PredictionCodegen ofJar(Path... jarPaths) throws IOException {
```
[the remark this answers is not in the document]
evacchi: I knew you would have liked this PR :P

mariofusco (approved):
Great job, but please double check that this is not breaking integration tests in Quarkus platform before merging: those ofJar() methods were used especially for that and I wanted to make sure that this refactor is covering everything it did before.

evacchi:
@mariofusco unless I did something wrong I just tried locally and it looks like it's working. FYI: the `CollectedResources#fromPaths()` method is checking whether the path is a jar and acting accordingly; `ofJar(...)` is then useless because `ofCollectedResources(...)` gets already the right `Resource` subtype. 

A recent change in Quarkus makes it apparently _always_ use the jar-based path, including `mvn test` and `mvn package`, so basically that is often _the_ code path that is being executed, even outside platform :sweat_smile:

PR #748, opened 2020-09-08
KOGITO-3292 KNative eventing is not enabled with Quarkus Extension

evacchi:
https://issues.redhat.com/browse/KOGITO-3292

KNative eventing was erroneously disabled when
KOGITO-3055 Refactor Quarkus Extension (#688)
was merged

PR #750, opened 2020-09-08
[KOGITO-3297] Bump CloudEvents api to 2.x

ruromero:
Fix https://issues.redhat.com/browse/KOGITO-3297

Upgrade CloudEvents dependencyManagement to 2.0.0-milestone2 that was already being used by the event addon.
Refactor tracing addon to use the new version
Refactor event addon to simplify the use


ricardozanini:
Depends on: https://github.com/kiegroup/kogito-runtimes/pull/748

ricardozanini (approved):
LGTM, the example https://github.com/kiegroup/kogito-examples/tree/stable/process-knative-quickstart-quarkus needs to be reviewed.

I'll do the changes as part of https://issues.redhat.com/browse/KOGITO-3294

ruromero:
I am making sure that any library depending on this addon  doesn't require any additional configuration

ruromero:
now it should be ready

ricardozanini (changes requested):
@ruromero since you removed the `HttpRequestConverter` class we need another one to verify if the addon is there:
https://github.com/kiegroup/kogito-runtimes/pull/748/files#diff-b31d1dc9b637fba5701e7c942847842fR107

ruromero:
done @ricardozanini

kogito-quarkus-extension/deployment/src/main/java/org/kie/kogito/quarkus/deployment/KogitoAssetsProcessor.java
```diff
@@ -104,7 +104,7 @@
     private static final DotName persistenceFactoryClass = DotName.createSimple("org.kie.kogito.persistence.KogitoProcessInstancesFactory");
     private static final DotName metricsClass = DotName.createSimple("org.kie.kogito.monitoring.rest.MetricsResource");
     private static final DotName tracingClass = DotName.createSimple("org.kie.kogito.tracing.decision.DecisionTracingListener");
-    private static final DotName knativeEventingClass = DotName.createSimple("org.kie.kogito.events.knative.ce.http.HttpRequestConverter");
+    private static final DotName knativeEventingClass = DotName.createSimple("org.kie.kogito.events.knative.ce.extensions.KogitoProcessExtension");
```
ricardozanini, line 107: I believe there are one more place:

```shell
kogito-maven-plugin/src/main/java/org/kie/kogito/maven/plugin/GenerateModelMojo.java:235:        boolean useKnativeEventing = hasClassOnClasspath(project, "org.kie.kogito.events.knative.ce.http.HttpRequestConverter");
```
Dammit we need to refactor this ASAP @evacchi

ricardozanini (changes requested):
Sorry, there's one more place to replace the deleted file.

addons/events/knative-eventing-addon/src/main/java/org/kie/kogito/events/knative/ce/extensions/KogitoProcessExtension.java
```diff
@@ -53,35 +53,35 @@
                     PROCESS_INSTANCE_STATE, REF_ID));
 
     private final Map<String, Object> innerValues;
-    private String kogitoProcessinstanceId;
-    private String kogitoRootProcessinstanceId;
+    private String kogitoProcessInstanceId;
```
evacchi, line 56: this is only fixing capitalization, right?

addons/events/knative-eventing-addon/src/test/java/org/kie/kogito/events/knative/ce/CloudEventConverterTest.java
```diff
@@ -47,14 +48,12 @@ void verifyBasicCloudEventConversion() {
                         .withData(payload.getBytes())
                         .build();
 
-        final String ceJson = CloudEventConverter.toJson(cloudEvent);
-        assertThat(ceJson).isNotEmpty().contains("Oi Mundo!");
+        final String ceJson = new String(cloudEvent.getData());
```
evacchi, line 51: new String ?

evacchi (approved):
I think overall it looks good, just left a couple of comments

addons/events/knative-eventing-addon/src/main/java/org/kie/kogito/events/knative/ce/extensions/KogitoProcessExtension.java
ruromero: yes

addons/events/knative-eventing-addon/src/test/java/org/kie/kogito/events/knative/ce/CloudEventConverterTest.java
ruromero: Yes, the `getData()` method is a byte[] representing the String

kogito-quarkus-extension/deployment/src/main/java/org/kie/kogito/quarkus/deployment/KogitoAssetsProcessor.java
ruromero: done

mareknovotny:
Jenkins retest this please

mareknovotny:
@ricardozanini I triggered CI here to test the new maven mirror settings.xml ;)

winklerm:
Jenkins retest this please

winklerm:
Jenkins retest this please

addons/events/knative-eventing-addon/src/test/java/org/kie/kogito/events/knative/ce/CloudEventConverterTest.java
evacchi: oh ok. you may want to pass in a charset encoding then

ruromero: Actually this doesn't test anything. I have replaced it by a simple Json object

addons/tracing/tracing-decision-api/src/main/java/org/kie/kogito/tracing/decision/event/CloudEventUtils.java
```diff
@@ -21,31 +21,53 @@
...
+        } catch (JsonProcessingException e) {
+            LOG.error("Unable to serialize CloudEvent data", e);
+        }
+        return CloudEventBuilder.v1()
                 .withType(dataType.getName())
                 .withId(id)
                 .withSource(source)
-                .withData(data)
+                .withData(bytes)
                 .build();
```
danielezonca, line 52: I don't think we should "hide" a serialization exception in an empty result. What about propagate the exception?

addons/tracing/tracing-decision-api/src/main/java/org/kie/kogito/tracing/decision/event/CloudEventUtils.java
```diff
@@ -21,31 +21,53 @@
...
 
-    public static <E> CloudEventImpl<E> decode(String json, TypeReference<CloudEventImpl<E>> ref) {
-        return Json.decodeValue(json, ref);
+    public static CloudEvent decode(String json) {
+        try {
+            return OBJECT_MAPPER.readValue(json, CloudEvent.class);
+        } catch (JsonProcessingException e) {
+            LOG.error("Unable to decode CloudEvent", e);
+            return null;
+        }
```
danielezonca, line 70: Same as above

addons/tracing/tracing-decision-common/pom.xml
```diff
@@ -43,6 +43,16 @@
       <artifactId>mockito-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.datatype</groupId>
+      <artifactId>jackson-datatype-jdk8</artifactId>
```
danielezonca, line 48: What is this module used for?

addons/tracing/tracing-decision-api/src/main/java/org/kie/kogito/tracing/decision/event/CloudEventUtils.java
ruromero: Propagate the exception up to where? I need advise here

addons/tracing/tracing-decision-api/src/main/java/org/kie/kogito/tracing/decision/event/CloudEventUtils.java
ruromero: I tried to propagate it but I didn't know up to where or where to really deal with it. Any advise?

addons/tracing/tracing-decision-common/pom.xml
ruromero: It's not needed. I added it by mistake. Thanks

addons/events/knative-eventing-addon/src/test/resources/application.properties
```diff
@@ -1 +1,4 @@
-quarkus.log.level=INFO
\ No newline at end of file
+quarkus.log.level=INFO
+
+quarkus.index-dependency.cloudevents.group-id=io.cloudevents
+quarkus.index-dependency.cloudevents.artifact-id=cloudevents-http-restful-ws
```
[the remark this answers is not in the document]
ruromero: No, it's not needed. With the service file it's enough to have the provider injected and available in other projects depending on this addon

addons/events/knative-eventing-addon/src/main/resources/META-INF/services/javax.ws.rs.ext.Providers
```diff
@@ -0,0 +1 @@
+io.cloudevents.http.restful.ws.CloudEventsProvider
```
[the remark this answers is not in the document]
ruromero: Yes, this code works in native mode. Actually it will allow any other library using the addon to have the CloudEventsProvider injected without any additional configuration.

addons/tracing/tracing-decision-api/src/main/java/org/kie/kogito/tracing/decision/event/CloudEventUtils.java
danielezonca: @kostola 
Can you please support? What if an event is impossible to be serialized? (same question for comment below)

addons/events/knative-eventing-addon/src/test/resources/application.properties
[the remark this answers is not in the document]
danielezonca: Ok so it is not clear to me why you had to add this configuration only for the test. Shouldn't be possible to use service mechanism here too?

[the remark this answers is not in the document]
ruromero: Yes, I deleted it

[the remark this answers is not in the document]
ruromero: I temporarily added it back and can be removed in cloudevents-2.0.0-milestone3
see https://github.com/cloudevents/sdk-java/pull/233

addons/tracing/tracing-decision-api/src/main/java/org/kie/kogito/tracing/decision/event/CloudEventUtils.java
kostola: With the previous API the `data` field was typed, so there's no particular error handling for serialization errors.
Imho sending a `CloudEvent` with `null` data, as I see here, is not very useful. I would avoid that.
I would rather log a more detailed error (possibly with the payload that triggered the error somehow) so that in a production environment it's possible to investigate by grabbing the logs of the services.
Wdyt @danielezonca @ruromero ?

danielezonca: Fine for me a more detailed log but what else? Error propagation until where?

kostola: I would catch the exception, log the message without sending a CloudEvent with `null` data and then that's it in the code. It seems to me a very rare error but correct me if I'm wrong.
Then the best way imho to catch those errors in a prod environment would be to setup a log management system (e.g. [Graylog](https://www.graylog.org/)) that ingests all the logs and allows custom triggers on specific messages to be configured.
Does this seem reasonable to you?

kostola: Probably the best solution for this method would be to return a `Optional<CloudEvent>` and let the caller decide how to handle the serialization failure with something like [ifPresentOrElse](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Optional.html#ifPresentOrElse(java.util.function.Consumer,java.lang.Runnable)) (which in my case would be to simply log the error without sending anything as I wrote above).

danielezonca: @ruromero 
Can you please verify if the proposed approach can be implemented? Return Optional<CloudEvent> and log error in caller if empty

ruromero: @kostola logging the payload won't be of much help as it is an array of bytes representing anything (a Json object, a String, a serialized object in any language, etc.). In this case we expect an object of type E but if decoding failed you don't know how to represent the data content.

I think the log is fine as it is. It's an error saying the CloudEvent data cannot be deserialized with the stacktrace.

Regarding the Optional propagation. There are 2 chains of invocations:
1) BaseModelEventEmitter#publishDecisionModels. In this case it's simple, we decide whether to emit the event or not.
2) DefaultAggregator#aggregate: There are 2 invocations and for these cases the event is always created right before the invocation so it doesn't make sense to expect an error. If for some reason such error occurs I will throw an unchecked exception.

ruromero:
jenkins retest this

ruromero:
jenkins retest this

ricardozanini (approved):

evacchi:
Let's move this forward for now, we can refine in further PRs @ricardozanini

PR #756, opened 2020-09-11
KOGITO-3318 Messaging test using Kafka

cristianonicolai:


integration-tests/integration-tests-springboot-kafka/src/it/integration-tests-springboot-it/src/test/java/org/kie/kogito/integrationtests/springboot/PingPongMessageTest.java
```diff
@@ -0,0 +1,73 @@
...
+import org.springframework.test.context.ContextConfiguration;
+
+import static io.restassured.RestAssured.given;
+import static org.awaitility.Awaitility.await;
+import static org.hamcrest.CoreMatchers.equalTo;
+
+@ExtendWith(SpringExtension.class)
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT, classes = KogitoSpringbootApplication.class)
+@ContextConfiguration(initializers =  { KafkaSpringBootTestResource.class, InfinispanSpringBootTestResource.Conditional.class })
+public class PingPongMessageTest extends BaseRestTest {
```
evacchi, line 38: mmm I think we'll have to remove this since it is already in the other suite

evacchi: misread the path forget it

evacchi (approved):

danielezonca (commented):
Can you please add 
```xml
<failIfNoProjects>true</failIfNoProjects>
```
in all the configurations of `maven-invoker-plugin` ( [link](https://github.com/kiegroup/kogito-runtimes/blob/master/integration-tests/integration-tests-springboot/pom.xml#L28) and also in the new `integration-tests-springboot-kafka` module ) to make the system more reliable?

Without that if the `it` project doesn't contain a `pom.xml` it will just skip it without any error

jbpm/jbpm-flow/src/main/java/org/kie/kogito/event/impl/DefaultEventMarshaller.java
```diff
@@ -28,8 +28,9 @@
     private static final Logger logger = LoggerFactory.getLogger(DefaultEventMarshaller.class);
 
     @Override
-    public <T, P extends AbstractProcessDataEvent<T>> String marshall(T dataEvent, Function<T, P> cloudFunction,
-            Optional<Boolean> isCloudEvent) {
+    public <T, P extends AbstractProcessDataEvent<T>> String marshall(T dataEvent,
```
fjtirado, line 31: is this really needed? I mean, I know formatting is critical (although apparently not enough to align Intellij and Eclipse one) , but in this case I think we can save a change

cristianonicolai:
> Can you please add
> 
> ```
> <failIfNoProjects>true</failIfNoProjects>
> ```
> 
> in all the configurations of `maven-invoker-plugin` ( [link](https://github.com/kiegroup/kogito-runtimes/blob/master/integration-tests/integration-tests-springboot/pom.xml#L28) and also in the new `integration-tests-springboot-kafka` module ) to make the system more reliable?
> 
> Without that if the `it` project doesn't contain a `pom.xml` it will just skip it without any error

@danielezonca added to the plugin management so it applies everywhere its used.

jbpm/jbpm-flow/src/main/java/org/kie/kogito/event/impl/DefaultEventMarshaller.java
cristianonicolai: just a minor tweak to make more readable in my opinion... we can check the reason why both formatters are not aligned later on.

PR #761, opened 2020-09-14
KOGITO-3330 Hotfix: Conditional eager reading of resources on native-image

evacchi:
https://issues.redhat.com/browse/KOGITO-3330

In GraalVM 20.2 under certain conditions, it is not possible to hold onto an input stream reader that has been generated from a classpath resource -- see https://github.com/quarkusio/quarkus/issues/8694#issuecomment-691828808

The hotfix consists in eagerly reading the entire resource when native mode is detected at build time (detected via special System property).

in other words, instead of:

```
new InputStreamReader(...getResourceAsStream())
```

for native image we do 

```
bytes = getResourceAsStream().readAllBytes()
new InputStreamReader(new ByteArrayInputStream(bytes))
```

We may have to revisit the issue to see if there are "better" ways to fix it (e.g. Quarkus has most probably a "proper" way to handle resources); in the meantime, we need to merge this to proceed with the release today.

(tested locally with native-image. You can double check integration-tests/integration-tests-quarkus  if you want)

### DIFF
--- a/kogito-codegen/src/main/resources/class-templates/DecisionContainerTemplate.java
+++ b/kogito-codegen/src/main/resources/class-templates/DecisionContainerTemplate.java
@@ -1,6 +1,6 @@
 public class DecisionModels implements org.kie.kogito.decision.DecisionModels {
 
-    private final static boolean IS_JDK = System.getProperty("org.graalvm.nativeimage.imagecode") == null;
+    private final static boolean IS_JDK = System.getProperty("org.graalvm.nativeimage.imagecode") != null;
 
     private final static java.util.function.Function<java.lang.String, org.kie.api.runtime.KieRuntimeFactory> kieRuntimeFactoryFunction = PredictionModels.kieRuntimeFactoryFunction;
     private final static org.kie.dmn.api.core.DMNRuntime dmnRuntime = org.kie.kogito.dmn.DMNKogito.createGenericDMNRuntime(kieRuntimeFactoryFunction);
@@ -15,7 +15,7 @@ public class DecisionModels implements org.kie.kogito.decision.DecisionModels {
     }
 
     private static java.io.InputStreamReader readResource(java.io.InputStream stream) {
-        if (IS_JDK) {
+        if (!IS_NATIVE_IMAGE) {
             return new java.io.InputStreamReader(stream);
         }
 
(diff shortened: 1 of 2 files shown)


kogito-codegen/src/main/resources/class-templates/DecisionContainerTemplate.java
```diff
@@ -1,9 +1,7 @@
-import java.util.function.Function;
-
-import org.kie.api.runtime.KieRuntimeFactory;
-
 public class DecisionModels implements org.kie.kogito.decision.DecisionModels {
 
+    private final static boolean IS_JDK = System.getProperty("org.graalvm.nativeimage.imagecode") == null;
```
geoand, line 3: I would personally invert this and call it `IS_GRAALVM` :)

evacchi: let's call it `IS_NATIVE_IMAGE` then

geoand (approved):

mariofusco (approved):

kogito-codegen/src/main/resources/class-templates/DecisionContainerTemplate.java
```diff
@@ -1,9 +1,7 @@
-import java.util.function.Function;
-
-import org.kie.api.runtime.KieRuntimeFactory;
-
 public class DecisionModels implements org.kie.kogito.decision.DecisionModels {
 
+    private final static boolean IS_NATIVE_IMAGE = System.getProperty("org.graalvm.nativeimage.imagecode") != null;
```
tarilabs, line 3: Good for me, I had similar to 
https://github.com/kiegroup/drools/blob/d72e9594cbb7f71a70f1bc13cf84b168ffa52e69/kie-dmn/kie-dmn-feel/src/main/java/org/kie/dmn/feel/util/ClassLoaderUtil.java#L21
eventually I can consolidate

tarilabs (approved):

kogito-codegen/src/main/resources/class-templates/DecisionContainerTemplate.java
evacchi: ok let's open a JIRA

evacchi: https://issues.redhat.com/browse/KOGITO-3331

danielezonca (approved):
LGTM

PR #763, opened 2020-09-14
[KOGITO-2766] Refactoring Process<T> initialization code

fjtirado:
Summary of changes (please read JIRA initial description for more information):

Generated handlers are now passed to constructor explicitly (handlers
field has been removed), which registers them, both for injection and non injection
paths 

Additionally,  when there is any handler present, additional non injected constructor that only accepts Application reference is provided, Itautomatically creates and registers the handlers (user can still pass handler instances using the other constructor in non injection)

Configure and registerListeners methods are only generated if they have
content (which happen when there are listeners)

Activate method will be invoked from generated constructors.

Generated code example 
```
    public TravellersProcess(org.kie.kogito.app.Application app) {
        this(
            app,
            new org.kie.kogito.handlers.GreetingTravellerService_greetTraveller_3_Handler(),
            new org.kie.kogito.handlers.TravellerAuditService_auditTraveller_7_Handler(),
            new org.kie.kogito.handlers.StoreTravellerService_storeTraveller_2_Handler());
    }

    @javax.inject.Inject()
    public TravellersProcess(org.kie.kogito.app.Application app,
                             org.kie.kogito.handlers.GreetingTravellerService_greetTraveller_3_Handler greetingTravellerService_greetTraveller_3_Handler,
                             org.kie.kogito.handlers.TravellerAuditService_auditTraveller_7_Handler travellerAuditService_auditTraveller_7_Handler,
                             org.kie.kogito.handlers.StoreTravellerService_storeTraveller_2_Handler storeTravellerService_storeTraveller_2_Handler) {
        super(
            app,
            java.util.Arrays
                .asList(
                        greetingTravellerService_greetTraveller_3_Handler,
                        travellerAuditService_auditTraveller_7_Handler,
                        storeTravellerService_storeTraveller_2_Handler));
        activate();
    }

```
Handler classes now have two constructors

 ```
public GreetingTravellerService_greetTraveller_3_Handler() {
        this.service = new org.acme.travels.services.GreetingTravellerService();
    }

    @javax.inject.Inject()
    public GreetingTravellerService_greetTraveller_3_Handler(org.acme.travels.services.GreetingTravellerService service) {
        this.service = service;
    }



```

### DIFF
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/di/DependencyInjectionAnnotator.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/di/DependencyInjectionAnnotator.java
@@ -19,7 +19,6 @@ import java.util.ArrayList;
 import java.util.List;
 
 import com.github.javaparser.ast.NodeList;
-import com.github.javaparser.ast.body.MethodDeclaration;
 import com.github.javaparser.ast.expr.ArrayInitializerExpr;
 import com.github.javaparser.ast.expr.Expression;
 import com.github.javaparser.ast.expr.MethodCallExpr;
@@ -86,8 +85,18 @@ public interface DependencyInjectionAnnotator {
      * Annotates given node with injection annotations e.g. Inject, Autowire
      *
      * @param node node to be annotated
+     * @boolean forceLazyInit use lazy initialization (for those container that applies)
      */
-    <T extends NodeWithAnnotations<?>> T withInjection(T node);
+    <T extends NodeWithAnnotations<?>> T withInjection(T node, boolean forceLazyInit);
+
+    /**
+     * Annotates given node with injection annotations e.g. Inject, Autowire
+     *
+     * @param node node to be annotated
+     */
+    default <T extends NodeWithAnnotations<?>> T withInjection(T node) {
+        return withInjection(node,false);
+    }
 
     /**
      * Annotates given node with injection annotations e.g. Inject, Autowire
@@ -165,14 +174,6 @@ public interface DependencyInjectionAnnotator {
         }
         return node;
     }
-
-    /**
-     * Create initialization method with given expression as body
-     *
-     * @param expression body of the init method
-     * @return complete initialization method declaration
-     */
-    MethodDeclaration withInitMethod(Expression... expression);
 
     /**
      * Returns type that allows to inject optional instances of the same type
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessGenerator.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessGenerator.java
@@ -18,8 +18,10 @@ package org.kie.kogito.codegen.process;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Map.Entry;
 import java.util.NoSuchElementException;
+import java.util.Optional;
 
 import javax.lang.model.SourceVersion;
 
@@ -37,7 +39,6 @@ import com.github.javaparser.ast.expr.AssignExpr;
 import com.github.javaparser.ast.expr.AssignExpr.Operator;
 import com.github.javaparser.ast.expr.CastExpr;
 import com.github.javaparser.ast.expr.FieldAccessExpr;
-import com.github.javaparser.ast.expr.LambdaExpr;
 import com.github.javaparser.ast.expr.MethodCallExpr;
 import com.github.javaparser.ast.expr.NameExpr;
 import com.github.javaparser.ast.expr.ObjectCreationExpr;
@@ -48,14 +49,15 @@ import com.github.javaparser.ast.expr.ThisExpr;
 import com.github.javaparser.ast.stmt.BlockStmt;
 import com.github.javaparser.ast.stmt.ReturnStmt;
 import com.github.javaparser.ast.type.ClassOrInterfaceType;
-import com.github.javaparser.ast.type.UnknownType;
 import org.drools.compiler.compiler.io.memory.MemoryFileSystem;
+import org.drools.core.util.KieFunctions;
 import org.drools.core.util.StringUtils;
 import org.jbpm.compiler.canonical.ProcessMetaData;
 import org.jbpm.compiler.canonical.TriggerMetaData;
+import org.jbpm.process.core.datatype.impl.type.ObjectDataType;
+import org.jbpm.ruleflow.core.RuleFlowProcessFactory;
 import org.kie.api.definition.process.Process;
 import org.kie.api.definition.process.WorkflowProcess;
-import org.kie.api.runtime.process.WorkItemHandler;
 import org.kie.api.runtime.process.WorkflowProcessInstance;
 import org.kie.kogito.Model;
 import org.kie.kogito.codegen.AddonsConfig;
@@ -63,6 +65,8 @@ import org.kie.kogito.codegen.BodyDeclarationComparator;
 import org.kie.kogito.codegen.di.DependencyInjectionAnnotator;
 import org.kie.kogito.process.ProcessInstancesFactory;
 import org.kie.kogito.process.impl.AbstractProcess;
+
+import static com.github.javaparser.StaticJavaParser.parseClassOrInterfaceType;
 
 /**
  * Generates the Process&lt;T&gt; container
@@ -74,11 +78,12 @@ public class ProcessGenerator {
     
     private static final String BUSINESS_KEY = "businessKey";
     private static final String CREATE_MODEL = "createModel";
+    private static final String APPLICATION = "app";
     private static final String WPI = "wpi";
 
     private final String packageName;
     private final WorkflowProcess process;
-    private final ProcessExecutableModelGenerator processGenerator;
+    private final ProcessExecutableModelGenerator processExecutable;
     private final String typeName;
     private final String modelTypeName;
     private final String generatedFilePath;
@@ -102,7 +107,7 @@ public class ProcessGenerator {
 
         this.packageName = process.getPackageName();
         this.process = process;
-        this.processGenerator = processGenerator;
+        this.processExecutable = processGenerator;
         this.typeName = typeName;
         this.modelTypeName = modelTypeName;        
         this.targetTypeName = typeName + "Process";
@@ -134,9 +139,9 @@ public class ProcessGenerator {
 
     public CompilationUnit compilationUnit() {
         CompilationUnit compilationUnit = new CompilationUnit(packageName);
-        compilationUnit.addImport("org.jbpm.process.core.datatype.impl.type.ObjectDataType");
-        compilationUnit.addImport("org.jbpm.ruleflow.core.RuleFlowProcessFactory");
-        compilationUnit.addImport("org.drools.core.util.KieFunctions");
+        compilationUnit.addImport(ObjectDataType.class);
+        compilationUnit.addImport(RuleFlowProcessFactory.class);
+        compilationUnit.addImport(KieFunctions.class);
         compilationUnit.getTypes().add(classDeclaration());
         return compilationUnit;
     }
@@ -156,6 +161,7 @@ public class ProcessGenerator {
                 .addModifier(Modifier.Keyword.PUBLIC)
                 .addParameter(modelTypeName, "value")
                 .setType(processInstanceFQCN)
+                .addAnnotation(Override.class)
                 .setBody(new BlockStmt()
                                  .addStatement(returnStmt));
         return methodDeclaration;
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: kogito-codegen/src/main/java/org/kie/kogito/codegen/di/DependencyInjectionAnnotator.java [hunk 1]
```diff
@@ -137,6 +137,8 @@
      * @param defaultValue value to be used in case there is no config parameter defined
      */
     <T extends NodeWithAnnotations<?>> T withConfigInjection(T node, String configKey, String defaultValue);
+ 
+    default <T extends NodeWithAnnotations<?>> T witLazyInit(T node) { return node; }
```
evacchi, line 141: `with`

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessGenerator.java
```diff
@@ -382,76 +343,118 @@ public ClassOrInterfaceDeclaration classDeclaration() {
...
-                                 .addStatement(
-                                         new AssignExpr(new FieldAccessExpr(new ThisExpr(), "app"), new NameExpr("app"), AssignExpr.Operator.ASSIGN)));
-        
-        ConstructorDeclaration emptyConstructorDeclaration = new ConstructorDeclaration()
-                .setName(targetTypeName)
-                .addModifier(Modifier.Keyword.PUBLIC);
-        
+        ConstructorDeclaration constructorDeclaration = getConstructorDeclaration().addParameter(appCanonicalName, APPLICATION);
+        BlockStmt constructorBody = new BlockStmt()
+                // super(module.config().process())
```
evacchi, line 350: remove comment

evacchi (changes requested):
can you paste here an example of the new code that is being generated now? :)
also, I was hoping we were moving away from the `annotator ` mechanism in lieu of a template-based approach. We can do that in a subsequent refactoring, but any particular reason why you can't do that? How much code is still framework-specific?

kogito-codegen/src/main/java/org/kie/kogito/codegen/di/DependencyInjectionAnnotator.java
fjtirado: Thanks

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessGenerator.java
fjtirado: This comment is not actually code, but displaying what the code below is generating, are we sure we want to remove it?

fjtirado:
> ay from the `annotator ` mechanism in lieu of a template-based approach. We can do that in a subsequent refactoring, but any particular reason why you can't do that? How much code is still framework-specific?

We still need to use CDI annotations for Constructor and for initialization.

kogito-codegen/src/main/java/org/kie/kogito/codegen/di/DependencyInjectionAnnotator.java
```diff
@@ -137,6 +136,8 @@
      * @param defaultValue value to be used in case there is no config parameter defined
      */
     <T extends NodeWithAnnotations<?>> T withConfigInjection(T node, String configKey, String defaultValue);
+ 
+    default <T extends NodeWithAnnotations<?>> T withLazyInit(T node) { return node; }
```
danielezonca, line 140: As far as I can see this concept only exists in `Spring` and not in `CDI` (even if quarkus impl should be lazy by default).
Why do you need it?

Minor note, can you add a javadoc?

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessGenerator.java
```diff
@@ -268,105 +276,58 @@ private MethodCallExpr createProcessRuntime() {
...
                 MethodCallExpr signalManager = new MethodCallExpr(new NameExpr("services"), "getSignalManager");
                 MethodCallExpr registerListener = new MethodCallExpr(signalManager, "addEventListener").addArgument(new StringLiteralExpr(subProcess.getValue())).addArgument(new NameExpr("completionEventListener"));
-                
                 body.addStatement(registerListener);
             }
+            return internalRegisterListeners;
         }
-        
-        return internalRegisterListeners;
+        return null;
```
danielezonca, line 330: What about return `Optional<MethodDeclaration>`?

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessGenerator.java
```diff
@@ -268,105 +276,58 @@ private MethodCallExpr createProcessRuntime() {
...
+                            NodeList.nodeList()));
                 
                 body.addStatement(registerListener);
                 
                 additionalClasses.add(listener);
             });
+            body.addStatement(new ReturnStmt(new ThisExpr()));
+            return internalConfigure;
         }
+        return null;
```
danielezonca, line 309: What about return `Optional<MethodDeclaration>`?

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessGenerator.java
```diff
@@ -548,6 +535,30 @@ public ClassOrInterfaceDeclaration classDeclaration() {
...
 
+    
+    private ConstructorDeclaration getConstructorDeclaration() {
+        return new ConstructorDeclaration()
+            .setName(targetTypeName)
+            .addModifier(Modifier.Keyword.PUBLIC);
+    }
+
+    
+    private void forEachField (ClassOrInterfaceDeclaration clazz, FieldDeclaration fd) {
```
danielezonca, line 546: Can you please rename the method to better explain its scope? Something like `initializeField` or similar

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessGenerator.java
```diff
@@ -382,76 +343,113 @@ public ClassOrInterfaceDeclaration classDeclaration() {
...
+
         
+        MethodDeclaration configure = internalConfigure(processMetaData);
+        if (configure != null) {
+            cls.addMember(configure);
+        }
+        MethodDeclaration registerListener = internalRegisterListeners(processMetaData);
+        if (registerListener != null) {
+            cls.addMember(registerListener);
+        }
```
danielezonca, line 451: If you make them return `Optional<MethodDeclaration>` you can replace with something like 
```suggestion
        internalConfigure(processMetaData).ifPresent(cls::addMember);
        internalRegisterListeners(processMetaData)::ifPresent(cls::addMember);
```

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessGenerator.java
```diff
@@ -268,105 +276,58 @@ private MethodCallExpr createProcessRuntime() {
...
-                MethodCallExpr eventSupport = new MethodCallExpr(new NameExpr("services"), "getEventSupport");            
-                MethodCallExpr registerListener = new MethodCallExpr(eventSupport, "addEventListener").addArgument(new ObjectCreationExpr(null, new ClassOrInterfaceType(null, packageName + "." + clazzName), NodeList.nodeList()));
+                MethodCallExpr eventSupport = new MethodCallExpr(new NameExpr("services"), "getEventSupport");
+                MethodCallExpr registerListener = new MethodCallExpr(eventSupport, "addEventListener")
+                    .addArgument(
+                        new ObjectCreationExpr(
+                            null,
+                            new ClassOrInterfaceType(
+                                null,
+                                listener.getPackageDeclaration().map(pd -> pd.getName().toString()).orElse("") + "." +clazz.getName()),
```
danielezonca, line 299: If `getPackageDeclaration()` returns an empty `Optional` this will produce an invalid string: `.ClassName`
Can you please verify if `getPackageDeclaration()` can be empty (i.e. default package)?
If not I think it is better to throw an exception like "This should never happen" instead of a empty string.

In general we can probably replace the code with this alternative
```suggestion
                                listener.getPackageDeclaration().map(pd -> pd.getName().toString() + "." + clazz.getName()).orElse(clazz.getName()),
```
Wdyt?

kogito-codegen/src/main/java/org/kie/kogito/codegen/di/DependencyInjectionAnnotator.java
fjtirado: As you correctly detect, for SpringBoot we need to add lazy annotation to handlers passed as parameters in the constructor. 
SpringBoot has issues with circular dependencies when using construction injection. If, as happens in one of the examples, one of the beans injected into the handler, depends on the Process, SpringBoot wont start unless we include lazy annotation (this issue does not appear in Quarkus)
I use a descriptive name, but we can use a more abstract one for the new method something as constructorParam. wdyt?

fjtirado: After thinking a while, I think for consistency with the existing interface, the word lazy needs to be there

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessGenerator.java
fjtirado: Ill do it

kogito-codegen/src/main/java/org/kie/kogito/codegen/di/DependencyInjectionAnnotator.java
danielezonca: :thinking: I'm fine with the name, I have more concern with the need of this annotation: I'm worried that sooner or later this problem (circular dependencies) will appear again if we continue to use construction injection.
What if we generate the constructor with all the parameters (to allow manual instantiation) but then we inject on fields? Does it solve the need to have lazy init?

@evacchi wdyt? If this works I would prefer to do this change instead of have the risk to add `Lazy` annotation "everywhere"

fjtirado: I think constructor injection is what this JIRA was all about (this allow us to remove fields and remove separate initialization). I do not see a big trouble with lazy if used where it should be (just the constructor parameters). Also, please notice this is a workaroung for SpringBoot, Quakus resolve circular dependencies without any issue.

fjtirado: What Im trying to say is that annotating something with Lazy is not an issue at all. Which is an issue is to use construction injection on springboot without the lazy. If we use construction injection somewhere else, we need to have this into account (and the annotator will be ready).
Does it make sense to use constructor injection in this particular scenario, I think the answer is yes, so I vote going for it

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessGenerator.java
evacchi: oh ok it looked like older code

kogito-codegen/src/main/java/org/kie/kogito/codegen/di/DependencyInjectionAnnotator.java
evacchi: ok, here's what I think

1) I don't love the "Annotator" at all really :P I believe that we should actually move to templates everywhere.
2) We don't necessarily have to introduce templates with this PR but then I would at least wish for a follow-up JIRA where we try; [I am trying to do it for Rules -- the code there is arguably less, though](https://github.com/kiegroup/kogito-runtimes/pull/660)
3) However, the following code could be substituted for the injection of Instance/Collection 
(and in fact IIRC we were injecting a collection, before)


```java
    @javax.inject.Inject()
    public TravellersProcess(org.kie.kogito.app.Application app, org.kie.kogito.handlers.GreetingTravellerService_greetTraveller_3_Handler greetingTravellerService_greetTraveller_3_Handler, org.kie.kogito.handlers.TravellerAuditService_auditTraveller_7_Handler travellerAuditService_auditTraveller_7_Handler, org.kie.kogito.handlers.StoreTravellerService_storeTraveller_2_Handler storeTravellerService_storeTraveller_2_Handler) {
        super(app.config().process());
        this.app = app;
        org.kie.api.runtime.process.WorkItemManager workItemManager = services.getWorkItemManager();
        workItemManager.registerWorkItemHandler("org.acme.travels.services.GreetingTravellerService_greetTraveller_3_Handler", greetingTravellerService_greetTraveller_3_Handler);
        workItemManager.registerWorkItemHandler("org.acme.travels.services.TravellerAuditService_auditTraveller_7_Handler", travellerAuditService_auditTraveller_7_Handler);
        workItemManager.registerWorkItemHandler("org.acme.travels.services.StoreTravellerService_storeTraveller_2_Handler", storeTravellerService_storeTraveller_2_Handler);
        this.activate();
    }
```

in other words:

```java
    @javax.inject.Inject()
    public TravellersProcess(
        org.kie.kogito.app.Application app, 
        // java.util.Collection in the case of Spring
        javax.enterprise.inject.Instance<org.kie.api.runtime.process.WorkItemHandler> handlers) {
        super(app.config().process());
        this.app = app;
        org.kie.api.runtime.process.WorkItemManager workItemManager = services.getWorkItemManager();
        for (WorkItemHandler wih : handlers) {
            workItemManager.registerWorkItemHandler(wih.getClass().getCanonicalName(), wih);
        }
        this.activate();
    }
```

to the point where we can actually move the logic to a super constructor:

```java
protected AbstractProcess(
    org.kie.kogito.app.Application app, 
    // iterable is a common superclass between Instance (CDI) and Collection (Spring)
    Iterable<org.kie.api.runtime.process.WorkItemHandler> handlers) {
        this(app.config().process());
        this.app = app;
        org.kie.api.runtime.process.WorkItemManager workItemManager = services.getWorkItemManager();
        for (WorkItemHandler wih : handlers) {
            workItemManager.registerWorkItemHandler(wih.getClass().getCanonicalName(), wih);
        }
        this.activate();
}
```

thus transforming the code as such:


```java
    @javax.inject.Inject()
    public TravellersProcess(
        org.kie.kogito.app.Application app, 
        // java.util.Collection in the case of Spring
        javax.enterprise.inject.Instance<org.kie.api.runtime.process.WorkItemHandler> handlers) {
        super(app.config().process(), handlers);
    }
```

as you can see, this way we trim down the generated code considerably, we get a much more meaningful and simpler template

4) Now, for the `@Lazy` annotation, first of all I would like to know what's causing the issue, and rather try to solve the issue at its root if possible; IIRC Application gets injected ApplicationProcesses, which contains all of the available processes; this in a CDI/Spring context is _not_ a dependency of this class; however, this could be still cyclic; to resolve the cycle, it may be just enough to annotate the class `kogito-codegen/src/main/resources/class-templates/SpringProcessContainerTemplate.java`with `@Lazy`

5) if there is not way to break the cycle, then I'd rather annotate the constructor rather than the single parameters 

hope this answers all of your questions

evacchi (changes requested):
see comment on constructor/templates

kogito-codegen/src/main/java/org/kie/kogito/codegen/di/DependencyInjectionAnnotator.java
fjtirado: Let me try to reply
1) and 2) I think are matter of a different JIRA which completely change the approach we are using and use two different templates rather than annotator (I confess Im not a fan of that approach and do not see the advantage, we are going to duplicate a lot of code in the templates as already happen with resources) 
In any case, place where the annotator is used are now minimized in codegen., 
3) Since we are using a generator, I think is better to restrict the handler to the ones that really makes sense for a particular process, thats why Im using concrete handlers rather than a generic list, so I ensure only those handlers that are part of the process definition are registered. This was my humble aportation to the JIRA description and I think is an improvement.
Also please note that first parameter of register handler is not really equal to class cannonical name
Name is org.acme.travels.services.GreetingTravellerService_greetTraveller_3_Handler
Class name is org.kie.kogito.handlers.GreetingTravellerService_greetTraveller_3_Handler 
If they matched the code I would generate is
```
    @javax.inject.Inject()
    public TravellersProcess(org.kie.kogito.app.Application app, org.kie.kogito.handlers.GreetingTravellerService_greetTraveller_3_Handler greetingTravellerService_greetTraveller_3_Handler, org.kie.kogito.handlers.TravellerAuditService_auditTraveller_7_Handler travellerAuditService_auditTraveller_7_Handler, org.kie.kogito.handlers.StoreTravellerService_storeTraveller_2_Handler storeTravellerService_storeTraveller_2_Handler) {
        super(app.config().process());   registerHandler(greetingTravellerService_greetTraveller_3_Handler,travellerAuditService_auditTraveller_7_Handler,storeTravellerService_storeTraveller_2_Handler);
}
```
But this can simply not be done, neither what you are proposing. 
3a) BTW, do we need app to be a field at all?
4) Circular dependency happens here https://github.com/kiegroup/kogito-examples/blob/stable/process-optaplanner-quarkus/src/main/java/org/kie/kogito/examples/service/FlightSeatingSolveService.java. This is a service injected by the handler that is injecting the process. Since there is not way to prevent that circular depedency, lazy is the solution for spring boot. 
5) I applied a fine grane approach, only annotate with lazy what really needs to be annotated (please note that currently Application bean is not annotated with lazy because we know for sure that there is not circular dependency here), but of course we can switch to coarse approach and lazy initialize all constructor parameters

fjtirado: @evacchi I updated code according to your point 3) proposal, please review

evacchi: 2) ok let's file a jira for that
3) we are using the Collection approach for ApplicationConfig and ProcessConfig. I think this is a simpler approach, and although I agree we could be more specific in the way we generate code, I have also come to the conclusion it is better to lean on the side of simplicity to reduce the complexity and amount of code we generate. In this case, I'd rather uniform the codebase with that approach; otherwise we would have to modify ProcessConfig et al. for consistency.

>  Also please note that first parameter of registerhandler method is not really equal to class cannonical name
Name is org.acme.travels.services.GreetingTravellerService_greetTraveller_3_Handler

weird why is that so? How was that solved before? I am pretty sure we were injecting a collection

3a) the app is used inside process definition, that's why it's passed on (Strictly speaking we may not necessarily always need it, it depends on how the process is defined)


4) I'd argue we should make lazy the `Process` field of that service in the example instead, wdyt ? :)

fjtirado: regarding 3). I think this is simple enough
```
  public TravellersProcess(org.kie.kogito.app.Application app) {
        this(app, new org.kie.kogito.handlers.GreetingTravellerService_greetTraveller_3_Handler(), new org.kie.kogito.handlers.TravellerAuditService_auditTraveller_7_Handler(), new org.kie.kogito.handlers.StoreTravellerService_storeTraveller_2_Handler());
    }

    @javax.inject.Inject()
    public TravellersProcess(org.kie.kogito.app.Application app, org.kie.kogito.handlers.GreetingTravellerService_greetTraveller_3_Handler greetingTravellerService_greetTraveller_3_Handler, org.kie.kogito.handlers.TravellerAuditService_auditTraveller_7_Handler travellerAuditService_auditTraveller_7_Handler, org.kie.kogito.handlers.StoreTravellerService_storeTraveller_2_Handler storeTravellerService_storeTraveller_2_Handler) {
        super(app, java.util.Arrays.asList(greetingTravellerService_greetTraveller_3_Handler, travellerAuditService_auditTraveller_7_Handler, storeTravellerService_storeTraveller_2_Handler));
    }

    public TravellersProcess() {
    }
```

fjtirado: 4) I think we cannot guarantee what the user code will do. I think is better to prevent this issue on runtime side

fjtirado: > weird why is that so? How was that solved before? I am pretty sure we were injecting a collection
handler.getName

fjtirado: Regarding lazy, I removed from process parameters and add to handler constructor.

fjtirado: Please note that activate needs to be invoked from child class constructor, since it relies on abtract method process and this is only avaiable when super() has finished

fjtirado: Rather than new mthod withLazy, a new parameter lazy has been added to withInjection method, default value being false

evacchi (approved):
alright let's move this forward

kogito-codegen/src/main/java/org/kie/kogito/codegen/di/DependencyInjectionAnnotator.java
```diff
@@ -86,8 +85,19 @@
      * Annotates given node with injection annotations e.g. Inject, Autowire
      *
      * @param node node to be annotated
+     * @boolean lazy use lazy initialization
      */
-    <T extends NodeWithAnnotations<?>> T withInjection(T node);
+    <T extends NodeWithAnnotations<?>> T withInjection(T node, boolean lazy);
```
danielezonca, line 90: What about rename `lazy` to something like `forceLazyInit` or similar? Mainly because lazy=false doesn't mean eager but means "use framework default"

danielezonca (approved):
Thanks for the PR, adding just a minor commend but LGTM :+1:

kogito-codegen/src/main/java/org/kie/kogito/codegen/di/DependencyInjectionAnnotator.java
fjtirado: good point, done

cristianonicolai (approved):

PR #766, opened 2020-09-14
KOGITO-3318 Consolidate SpringBoot IT test in same project

cristianonicolai:


evacchi (approved):
I don't understand how you managed to make this work and what we did wrong last time :D

PR #762, opened 2020-09-14
KOGITO-2277 fix OpenAPI schema for temporals

tarilabs:
part of an ensemble of PRs

https://issues.redhat.com/browse/KOGITO-2277

Requires PRs:

* https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1458
* https://github.com/kiegroup/drools/pull/3097

PR #770, opened 2020-09-16
[KOGITO-3078] DMN strongly typed: output definition

tkobayas:
https://issues.redhat.com/browse/KOGITO-3078

This PR is to have `OutputSet` as a return type of DMN REST endpoint when stronglytyped is enabled. So swagger-ui can provide type information and proper payload example.

- [x ] You have read the [contributors guide](CONTRIBUTING.md)
- [x ] Pull Request title is properly formatted: `KOGITO-XYZ Subject`
- [x ] Pull Request title contains the target branch if not targeting master: `[0.9.x] KOGITO-XYZ Subject`
- [x ] Pull Request contains link to the JIRA issue
- [x ] Pull Request contains link to any dependent or related Pull Request
- [x ] Pull Request contains description of the issue
- [x ] Pull Request does not include fixes for issues other than the main ticket

### DIFF
--- a/kogito-codegen/src/test/java/org/kie/kogito/codegen/decision/DecisionCodegenTest.java
+++ b/kogito-codegen/src/test/java/org/kie/kogito/codegen/decision/DecisionCodegenTest.java
@@ -163,7 +163,7 @@ public class DecisionCodegenTest {
         List<GeneratedFile> generatedFiles = codeGenerator.generate();
         assertThat(generatedFiles).anyMatch(x -> x.relativePath().endsWith("InputSet.java"));
         GeneratedFile inputSetFile = generatedFiles.stream().filter(x -> x.relativePath().endsWith("InputSet.java")).findFirst().get();
-        assertThat(new String(inputSetFile.contents())).contains("@org.eclipse.microprofile.openapi.annotations.media.Schema(", "enumeration");
+        assertThat(new String(inputSetFile.contents())).containsPattern("@org\\.eclipse\\.microprofile\\.openapi\\.annotations\\.media\\.Schema\\(.*enumeration");
     }
 
     @Test
@@ -175,6 +175,6 @@ public class DecisionCodegenTest {
         List<GeneratedFile> generatedFiles = codeGenerator.generate();
         assertThat(generatedFiles).anyMatch(x -> x.relativePath().endsWith("InputSet.java"));
         GeneratedFile inputSetFile = generatedFiles.stream().filter(x -> x.relativePath().endsWith("InputSet.java")).findFirst().get();
-        assertThat(new String(inputSetFile.contents())).doesNotContain("@org.eclipse.microprofile.openapi.annotations.media.Schema(enumeration");
+        assertThat(new String(inputSetFile.contents())).doesNotContain("@org.eclipse.microprofile.openapi.annotations.media.Schema");
     }
 }
--- a/kogito-codegen/src/test/java/org/kie/kogito/codegen/decision/DecisionCodegenTest.java
+++ b/kogito-codegen/src/test/java/org/kie/kogito/codegen/decision/DecisionCodegenTest.java
@@ -48,6 +48,7 @@ public class DecisionCodegenTest {
         List<GeneratedFile> generatedFiles = codeGenerator.generate();
         assertThat(generatedFiles.size()).isGreaterThanOrEqualTo(6);
         assertThat(fileNames(generatedFiles)).containsAll(Arrays.asList("decision/InputSet.java",
+                                                                        "decision/OutputSet.java",
                                                                         "decision/TEmployee.java",
                                                                         "decision/TAddress.java",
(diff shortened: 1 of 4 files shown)


tkobayas:
@tarilabs @lucamolteni @hellowdan 
This is a draft PR of KOGITO-3078. Any feedback would be welcome.

Especially, regarding testing. I manually tested dmn-quarkus-example and dmn-springboot-example. Then I confirmed that swagger-ui gives expected type information. But should I add some more unit tests? Feel free to share your thoughts. Thanks!

tkobayas:
Jenkins retest this

kogito-codegen/src/test/java/org/kie/kogito/codegen/decision/DecisionCodegenTest.java
```diff
@@ -161,7 +163,7 @@ public void testNSEW_positive() throws Exception {
         List<GeneratedFile> generatedFiles = codeGenerator.generate();
         assertThat(generatedFiles).anyMatch(x -> x.relativePath().endsWith("InputSet.java"));
         GeneratedFile inputSetFile = generatedFiles.stream().filter(x -> x.relativePath().endsWith("InputSet.java")).findFirst().get();
-        assertThat(new String(inputSetFile.contents())).contains("@org.eclipse.microprofile.openapi.annotations.media.Schema(enumeration");
+        assertThat(new String(inputSetFile.contents())).contains("@org.eclipse.microprofile.openapi.annotations.media.Schema(", "enumeration");
```
tarilabs, line 166: agreed, or you can also consider this approach:
https://github.com/kiegroup/kogito-runtimes/pull/762/files#diff-59003bc182a20b006a260bae9299342aR164

what do you think? 
That way, it should not detect `enumeration` somewhere else in the source code (for instance a field name) since the `.*` should not match line breaking character IIRC ?

tarilabs (approved):
This is good, and with regards to general test, I would raise it as a separate JIRA.

We might want to consider isolated new modules in `kogito-runtimes/integration-tests` where we isolate the stronglytyped flag activated, then you can REST GET the `openapi.json` and check the referenced responde payload is not a generic `object {}` type, but a named reference

What do you think?

hellowdan (approved):

tkobayas:
> We might want to consider isolated new modules in `kogito-runtimes/integration-tests` where we isolate the stronglytyped flag activated, then you can REST GET the `openapi.json` and check the referenced responde payload is not a generic `object {}` type, but a named reference

Thank you for the suggestion, I filed https://issues.redhat.com/browse/KOGITO-3368

kogito-codegen/src/test/java/org/kie/kogito/codegen/decision/DecisionCodegenTest.java
tkobayas: Thank you! I'll follow your approach.

tkobayas:
Hi @tarilabs  , here are some points.

1. `io.quarkus.it.kogito.dmn.DMNTest` and `io.quarkus.it.kogito.decision.SimpleModifyHotReloadTest` failed with drools 7.43.1.Final because "Using `OutputSet` for the input `DMNContextFPAImpl`" causes the "undefined vs null" problem which was fixed by DROOLS-5517. So I use `InputSet` for the input `DMNContextFPAImpl` for this PR. I think it's not a big difference unless a user wants to start scaffolding. Then I can raise a follow-up JIRA/PR to use `OutputSet` for the input `DMNContextFPAImpl` once drools 7.44.0 is available. Is it okay?

2. This PR returns `OutputSet` but it doesn't recreate a `DMNContextFPAImpl` which contains the `OutputSet` (See `extractStronglyTypedContextIfSucceded` method) because there is no use of the `DMNContextFPAImpl` at the moment. Is it fine now? Probably we will revisit when we will consider scaffolding use cases further.

3. I noticed that we haven't yet "strongly typed" `DMNResultImpl.decisionResults`. So `extractSingletonDSIfSucceded()` still returns Object in this PR. Shall I raise a JIRA to be able to return strongly typed class for `DMNResultImpl.decisionResults`?

Thanks!

tarilabs:
Hi @tkobayas 

1. sounds good to me!

2. I think on the way to returning from the REST method, is just fine that an instance of `OutputSet` will be returned, without having to create `DMNContextFPAImpl`, but I agree, we can revisit this later if you think there is some value

3. yes definitely a separate JIRA, but also as very low priority.. to be honest I don't have an opinion on this use-case specifically yet, since I'm not sure if we will need a creation of an OutputSet specialised for the given decision service?

Good to merge for me!

tkobayas:
Thanks! I filed https://issues.redhat.com/browse/KOGITO-3381 for "OutputSet for the input DMNContextFPAImpl". and https://issues.redhat.com/browse/DROOLS-5654 for "DMN Strongly typed DMNResultImpl.decisionResults".

mariofusco (approved):

PR #369, opened 2020-03-15
KOGITO-1448 - Upgrade Quarkus to version 1.3.0

cristianonicolai:

PR #768, opened 2020-09-16
KOGITO-3211 - SW update to latest spec and runtime updates

tsurdilo:
Signed-off-by: Tihomir Surdilovic <tsurdilo@redhat.com>

Many thanks for submitting your Pull Request :heart:! 

@ricardozanini 
This PR aligns the kogito SW parsing with the latest of the specification
Updated parsing and generation code to work with the latest Switch state
Updated all examples and tests so the json and yaml use the latest spec definitions

I will do the associated PR for kogito-examples tomorrow.

PR #778, opened 2020-09-18
[KOGITO-3158] Adding RestTaskHandler

fjtirado:
Changes for adding a RestWorkItemHandler

Please check example at [#369](https://github.com/kiegroup/kogito-examples/pull/369)

Summary of changes:

-  Using vert.x both for Quarkus and Springboot (jax-rs client is not working on native mode, check https://github.com/quarkusio/quarkus/issues/6151). It is used in async mode, so do not block the thread while the rest invocation is being performed and handler is completed when the response arrive
- New work item handler for Rest invocations (RestTaskHandler) has been added. Required parameters are endpoint (read from function defintion), method (read from function metadata) and ResultHandler (this one is auto generated by serverless parser). User and password for basic authentication are optional. 
-  Since input/output parameters can be Json objects (when serverless) or Pojo (when bpmn), rather than putting that knowledge in the handler, I have tried to encapsulate the details using Function, Supplier, Bifunction generated by code generator, but where the runtime part is not visible to RestTaskHandler class: 
       - Supplier is used when generating code to fill the work iterm parameter with objects (either function or bifunction)
       - Function is used to calculate the objects to be used in different parts of the request from the input model
       - Bifunction is used to format the ouputs of the task (in case of jsonobject a non deep merge is performed)

### DIFF
--- a/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/RestTaskDescriptor.java
+++ b/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/RestTaskDescriptor.java
@@ -1,0 +1,41 @@
+/*
+ *  Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.jbpm.compiler.canonical;
+
+import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
+import com.github.javaparser.ast.body.ConstructorDeclaration;
+
+import static com.github.javaparser.StaticJavaParser.parse;
+
+public class RestTaskDescriptor {
+
+    private RestTaskDescriptor() {}
+
+    public static String getClassName(ProcessMetaData processMetadata) {
+        return processMetadata.getProcessId() + "RestWorkItemHandler";
+    }
+
+    public static CompilationUnit generateHandlerClassForService(String className) {
+        CompilationUnit compilationUnit =
+                parse(RestTaskDescriptor.class.getResourceAsStream("/class-templates/RestWorkItemHandlerTemplate.java"));
+        compilationUnit.setPackageDeclaration("org.kie.kogito.handlers");
+        compilationUnit.findFirst(ClassOrInterfaceDeclaration.class).ifPresent(c -> c.setName(className));
+        compilationUnit.findAll(ConstructorDeclaration.class).forEach(c -> c.setName(className));
+        return compilationUnit;
+    }
+}
--- a/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ServiceTaskDescriptor.java
+++ b/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ServiceTaskDescriptor.java
@@ -28,6 +28,7 @@ import java.util.Map;
 import com.github.javaparser.StaticJavaParser;
 import com.github.javaparser.ast.CompilationUnit;
 import com.github.javaparser.ast.Modifier;
+import com.github.javaparser.ast.Modifier.Keyword;
 import com.github.javaparser.ast.NodeList;
 import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
 import com.github.javaparser.ast.body.FieldDeclaration;
@@ -37,12 +38,14 @@ import com.github.javaparser.ast.body.VariableDeclarator;
 import com.github.javaparser.ast.expr.AssignExpr;
 import com.github.javaparser.ast.expr.CastExpr;
 import com.github.javaparser.ast.expr.Expression;
+import com.github.javaparser.ast.expr.FieldAccessExpr;
 import com.github.javaparser.ast.expr.MethodCallExpr;
 import com.github.javaparser.ast.expr.NameExpr;
 import com.github.javaparser.ast.expr.NullLiteralExpr;
 import com.github.javaparser.ast.expr.ObjectCreationExpr;
 import com.github.javaparser.ast.expr.SimpleName;
 import com.github.javaparser.ast.expr.StringLiteralExpr;
+import com.github.javaparser.ast.expr.ThisExpr;
 import com.github.javaparser.ast.expr.VariableDeclarationExpr;
 import com.github.javaparser.ast.stmt.BlockStmt;
 import com.github.javaparser.ast.stmt.CatchClause;
@@ -74,7 +77,7 @@ public class ServiceTaskDescriptor {
     private final WorkItemNode workItemNode;
     private final String mangledName;
     private Class<?> cls;
-
+    
     ServiceTaskDescriptor(WorkItemNode workItemNode, ClassLoader contextClassLoader) {
         this.workItemNode = workItemNode;
         interfaceName = (String) workItemNode.getWork().getParameter("Interface");
@@ -170,9 +173,27 @@ public class ServiceTaskDescriptor {
                 .addImplementedType(WorkItemHandler.class.getCanonicalName());
         ClassOrInterfaceType serviceType = new ClassOrInterfaceType(null, interfaceName);
 
+        final String serviceName = "service";
         FieldDeclaration serviceField = new FieldDeclaration()
-                .addVariable(new VariableDeclarator(serviceType, "service"));
+                .addVariable(new VariableDeclarator(serviceType, serviceName));
         cls.addMember(serviceField);
+        cls
+            .addConstructor(Keyword.PUBLIC)
+            .setBody(
+                     new BlockStmt()
+                         .addStatement(new MethodCallExpr("this", new ObjectCreationExpr().setType(serviceType))));
+        cls
+            .addConstructor(Keyword.PUBLIC)
+            .addParameter(serviceType, serviceName)
+            .setBody(
+                     new BlockStmt()
+                         .addStatement(
+                                       new AssignExpr(
+                                           new FieldAccessExpr(
+                                               new ThisExpr(),
+                                               serviceName),
+                                           new NameExpr(serviceName),
+                                           AssignExpr.Operator.ASSIGN)));
 
         // executeWorkItem method
         BlockStmt executeWorkItemBody = new BlockStmt();
--- a/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/WorkItemNodeVisitor.java
+++ b/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/WorkItemNodeVisitor.java
@@ -18,6 +18,7 @@ package org.jbpm.compiler.canonical;
 
 import java.util.Map.Entry;
 import java.util.Objects;
+import java.util.function.Supplier;
 
 import com.github.javaparser.ast.CompilationUnit;
 import com.github.javaparser.ast.expr.BooleanLiteralExpr;
@@ -26,6 +27,7 @@ import com.github.javaparser.ast.expr.IntegerLiteralExpr;
 import com.github.javaparser.ast.expr.LongLiteralExpr;
 import com.github.javaparser.ast.expr.MethodCallExpr;
 import com.github.javaparser.ast.expr.NameExpr;
+import com.github.javaparser.ast.expr.NullLiteralExpr;
 import com.github.javaparser.ast.expr.StringLiteralExpr;
 import com.github.javaparser.ast.stmt.BlockStmt;
 import org.jbpm.process.core.Work;
@@ -85,6 +87,11 @@ public class WorkItemNodeVisitor<T extends WorkItemNode> extends AbstractNodeVis
             CompilationUnit generatedHandler = d.generateHandlerClassForService();
             metadata.getGeneratedHandlers().put(mangledName, generatedHandler);
             workName = mangledName;
+        } else if (workName.equals("Rest Task")) {
+            workName = RestTaskDescriptor.getClassName(metadata);
+            metadata
+                .getGeneratedHandlers()
+                .computeIfAbsent(workName, RestTaskDescriptor::generateHandlerClassForService);
         }
 
         body.addStatement(getAssignedFactoryMethod(factoryField, WorkItemNodeFactory.class, getNodeId(node), getNodeKey(), new LongLiteralExpr(node.getId())))
@@ -110,30 +117,45 @@ public class WorkItemNodeVisitor<T extends WorkItemNode> extends AbstractNodeVis
             if(work.getParameterDefinition(entry.getKey()) != null) {
                 paramType = work.getParameterDefinition(entry.getKey()).getType().getStringType();
             }
-            body.addStatement(getFactoryMethod(variableName, METHOD_WORK_PARAMETER, new StringLiteralExpr(entry.getKey()), getParameterExpr(paramType, entry.getValue().toString())));
+            body.addStatement(getFactoryMethod(variableName, METHOD_WORK_PARAMETER, new StringLiteralExpr(entry.getKey()), getParameterExpr(paramType, entry.getValue())));
         }
     }
+    
+  
 
-    private Expression getParameterExpr(String type, String value) {
+    private Expression getParameterExpr(String type, Object value) {
+        if (value == null) {
+            return new NullLiteralExpr();
+        }
         ParamType pType = ParamType.fromString(type);
         if (pType == null) {
-            return new StringLiteralExpr(value);
+            if (value instanceof Supplier) {
+                return ((Supplier<Expression>) value).get();
+            } else {
+                return new StringLiteralExpr(value.toString());
+            }
+
         }
         switch (pType) {
             case BOOLEAN:
-                return new BooleanLiteralExpr(Boolean.parseBoolean(value));
+                return new BooleanLiteralExpr(asBoolean(value));
             case FLOAT:
                 return new MethodCallExpr()
-                        .setScope(new NameExpr(Float.class.getName()))
-                        .setName("parseFloat")
-                        .addArgument(new StringLiteralExpr(value));
+                    .setScope(new NameExpr(Float.class.getName()))
+                    .setName("parseFloat")
+                    .addArgument(new StringLiteralExpr(value.toString()));
             case INTEGER:
-                return new IntegerLiteralExpr(Integer.parseInt(value));
+                return new IntegerLiteralExpr(asInteger(value));
             default:
-                return new StringLiteralExpr(value);
+                return new StringLiteralExpr(value.toString());
         }
     }
 
+    private Boolean asBoolean(Object value) {
+        return value instanceof Boolean ? (Boolean) value : Boolean.parseBoolean(value.toString());
+    }
 
-
+    private Integer asInteger(Object value) {
+        return value instanceof Number ? ((Number) value).intValue() : Integer.parseInt(value.toString());
+    }
 }
--- a/jbpm/jbpm-serverless-workflow/src/main/java/org/jbpm/serverless/workflow/parser/ServerlessWorkflowParser.java
+++ b/jbpm/jbpm-serverless-workflow/src/main/java/org/jbpm/serverless/workflow/parser/ServerlessWorkflowParser.java
@@ -23,13 +23,22 @@ import java.util.concurrent.atomic.AtomicLong;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import org.jbpm.ruleflow.core.RuleFlowProcess;
+import org.jbpm.serverless.workflow.api.Workflow;
+import org.jbpm.serverless.workflow.api.actions.Action;
 import org.jbpm.serverless.workflow.api.branches.Branch;
 import org.jbpm.serverless.workflow.api.end.End;
 import org.jbpm.serverless.workflow.api.events.EventDefinition;
 import org.jbpm.serverless.workflow.api.functions.Function;
 import org.jbpm.serverless.workflow.api.interfaces.State;
 import org.jbpm.serverless.workflow.api.mapper.BaseObjectMapper;
-import org.jbpm.serverless.workflow.api.states.*;
+import org.jbpm.serverless.workflow.api.states.DefaultState.Type;
+import org.jbpm.serverless.workflow.api.states.DelayState;
+import org.jbpm.serverless.workflow.api.states.EventState;
+import org.jbpm.serverless.workflow.api.states.InjectState;
+import org.jbpm.serverless.workflow.api.states.OperationState;
+import org.jbpm.serverless.workflow.api.states.ParallelState;
+import org.jbpm.serverless.workflow.api.states.SubflowState;
+import org.jbpm.serverless.workflow.api.states.SwitchState;
 import org.jbpm.serverless.workflow.api.switchconditions.DataCondition;
 import org.jbpm.serverless.workflow.api.switchconditions.EventCondition;
 import org.jbpm.serverless.workflow.api.transitions.Transition;
@@ -38,12 +47,17 @@ import org.jbpm.serverless.workflow.parser.util.ServerlessWorkflowUtils;
 import org.jbpm.serverless.workflow.parser.util.WorkflowAppContext;
 import org.jbpm.workflow.core.impl.ConnectionRef;
 import org.jbpm.workflow.core.impl.ConstraintImpl;
-import org.jbpm.workflow.core.node.*;
+import org.jbpm.workflow.core.node.ActionNode;
+import org.jbpm.workflow.core.node.CompositeContextNode;
+import org.jbpm.workflow.core.node.EndNode;
+import org.jbpm.workflow.core.node.EventNode;
+import org.jbpm.workflow.core.node.Join;
+import org.jbpm.workflow.core.node.Split;
+import org.jbpm.workflow.core.node.StartNode;
+import org.jbpm.workflow.core.node.SubProcessNode;
+import org.jbpm.workflow.core.node.TimerNode;
 import org.kie.api.definition.process.Node;
 import org.kie.api.definition.process.Process;
-import org.jbpm.serverless.workflow.api.Workflow;
-import org.jbpm.serverless.workflow.api.actions.Action;
-import org.jbpm.serverless.workflow.api.states.DefaultState.Type;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -51,6 +65,7 @@ public class ServerlessWorkflowParser {
     private static final Logger LOGGER = LoggerFactory.getLogger(ServerlessWorkflowParser.class);
 
     private static final String SCRIPT_TYPE = "script";
+    private static final String REST_TYPE = "rest";
     private static final String SCRIPT_TYPE_PARAM = "script";
     private static final String SYSOUT_TYPE = "sysout";
     private static final String SYSOUT_TYPE_PARAM = "message";
@@ -393,65 +408,101 @@ public class ServerlessWorkflowParser {
         }
     }
 
-    protected void handleActions(List<Function> workflowFunctions, List<Action> actions, RuleFlowProcess process, CompositeContextNode embeddedSubProcess) {
+    protected void handleActions(List<Function> workflowFunctions,
+                                 List<Action> actions,
+                                 RuleFlowProcess process,
+                                 CompositeContextNode embeddedSubProcess) {
         if (actions != null && !actions.isEmpty()) {
-            StartNode embeddedStartNode = factory.startNode(idCounter.getAndIncrement(), "EmbeddedStart", embeddedSubProcess);
+            StartNode embeddedStartNode =
+                    factory.startNode(idCounter.getAndIncrement(), "EmbeddedStart", embeddedSubProcess);
             Node start = embeddedStartNode;
             Node current = null;
 
             for (Action action : actions) {
-                Function actionFunction = workflowFunctions.stream()
-                        .filter(wf -> wf.getName().equals(action.getFunctionRef().getRefName()))
-                        .findFirst()
-                        .get();
-
+                Function actionFunction = workflowFunctions
+                    .stream()
+                    .filter(wf -> wf.getName().equals(action.getFunctionRef().getRefName()))
+                    .findFirst()
+                    .orElseThrow(
+                                 () -> new IllegalArgumentException(
+                                     "cannot find function " + action.getFunctionRef().getRefName()));
                 if (actionFunction.getType() != null) {
-                    if (SCRIPT_TYPE.equalsIgnoreCase(actionFunction.getType())) {
-                        String script = ServerlessWorkflowUtils.scriptFunctionScript(action.getFunctionRef().getParameters().get(SCRIPT_TYPE_PARAM));
-                        current = factory.scriptNode(idCounter.getAndIncrement(), action.getFunctionRef().getRefName(), script, embeddedSubProcess);
-
-                        factory.connect(start.getId(), current.getId(), start.getId() + "_" + current.getId(), embeddedSubProcess);
-                        start = current;
-                    } else if (SYSOUT_TYPE.equalsIgnoreCase(actionFunction.getType())) {
-                        String script = ServerlessWorkflowUtils.sysOutFunctionScript(action.getFunctionRef().getParameters().get(SYSOUT_TYPE_PARAM));
-                        current = factory.scriptNode(idCounter.getAndIncrement(), action.getFunctionRef().getRefName(), script, embeddedSubProcess);
-
-                        factory.connect(start.getId(), current.getId(), start.getId() + "_" + current.getId(), embeddedSubProcess);
-                        start = current;
-                    } else if (SERVICE_TYPE.equalsIgnoreCase(actionFunction.getType())) {
-                        current = factory.serviceNode(idCounter.getAndIncrement(), action.getFunctionRef().getRefName(), actionFunction, embeddedSubProcess);
-                        factory.connect(start.getId(), current.getId(), start.getId() + "_" + current.getId(), embeddedSubProcess);
-                        start = current;
-                    } else if (DECISION_TYPE.equals(actionFunction.getType())) {
-                        current = factory.humanTaskNode(idCounter.getAndIncrement(), action.getFunctionRef().getRefName(), actionFunction, process, embeddedSubProcess);
-                        factory.connect(start.getId(), current.getId(), start.getId() + "_" + current.getId(), embeddedSubProcess);
-                        start = current;
-                    } else if (RULE_TYPE.equals(actionFunction.getType())) {
-                        current = factory.ruleSetNode(idCounter.getAndIncrement(), action.getFunctionRef().getRefName(), actionFunction, embeddedSubProcess);
-                        factory.connect(start.getId(), current.getId(), start.getId() + "_" + current.getId(), embeddedSubProcess);
-                        start = current;
-                    } else if (INTEGRATION_TYPE.equals(actionFunction.getType())) {
-                        current = factory.camelRouteServiceNode(idCounter.getAndIncrement(), action.getFunctionRef().getRefName(), actionFunction, embeddedSubProcess);
-                        factory.connect(start.getId(), current.getId(), start.getId() + "_" + current.getId(), embeddedSubProcess);
-                        start = current;
-                    } else {
-                        LOGGER.warn("currently unsupported function type, supported types are 'script', 'sysout', 'service', 'decision', 'rule', 'integration'");
-                        LOGGER.warn("defaulting to script type");
-                        String script = ServerlessWorkflowUtils.scriptFunctionScript("");
-                        current = factory.scriptNode(idCounter.getAndIncrement(), action.getFunctionRef().getRefName(), script, embeddedSubProcess);
-
-                        factory.connect(start.getId(), current.getId(), start.getId() + "_" + current.getId(), embeddedSubProcess);
-                        start = current;
+                    switch (actionFunction.getType().toLowerCase()) {
+                        case SCRIPT_TYPE:
+                            current = factory
+                                .scriptNode(
+                                            idCounter.getAndIncrement(),
+                                            action.getFunctionRef().getRefName(),
+                                            ServerlessWorkflowUtils
+                                                .scriptFunctionScript(
+                                                                      action
+                                                                          .getFunctionRef()
+                                                                          .getParameters()
+                                                                          .get(SCRIPT_TYPE_PARAM)),
+                                            embeddedSubProcess);
+                            break;
+                        case SYSOUT_TYPE:
+                            current = factory
+                                .scriptNode(
+                                            idCounter.getAndIncrement(),
+                                            action.getFunctionRef().getRefName(),
+                                            ServerlessWorkflowUtils
+                                                .sysOutFunctionScript(
+                                                                      action
+                                                                          .getFunctionRef()
+                                                                          .getParameters()
+                                                                          .get(SYSOUT_TYPE_PARAM)),
+                                            embeddedSubProcess);
+                            break;
+                        case SERVICE_TYPE:
+                            current = factory
+                                .serviceNode(
+                                             idCounter.getAndIncrement(),
+                                             action.getFunctionRef().getRefName(),
+                                             actionFunction,
+                                             embeddedSubProcess);
+                            break;
+                        case DECISION_TYPE:
+                            current = factory
+                                .humanTaskNode(
+                                               idCounter.getAndIncrement(),
+                                               action.getFunctionRef().getRefName(),
+                                               actionFunction,
+                                               process,
+                                               embeddedSubProcess);
+                            break;
+                        case RULE_TYPE:
+                            current = factory
+                                .ruleSetNode(
+                                             idCounter.getAndIncrement(),
+                                             action.getFunctionRef().getRefName(),
+                                             actionFunction,
+                                             embeddedSubProcess);
+                            break;
+                        case INTEGRATION_TYPE:
+                            current = factory
+                                .camelRouteServiceNode(
+                                                       idCounter.getAndIncrement(),
+                                                       action.getFunctionRef().getRefName(),
+                                                       actionFunction,
+                                                       embeddedSubProcess);
+                            break;
+                        case REST_TYPE:
+                            current = factory
+                                .restServiceNode(
+                                                 idCounter.getAndIncrement(),
+                                                 action.getFunctionRef(),
+                                                 actionFunction,
+                                                 embeddedSubProcess);
+                            break;
+                        default:
+                            current = unsupportedNode(action, embeddedSubProcess);
                     }
                 } else {
-                    LOGGER.warn("invalid function type. supported types are 'script', 'sysout', 'service', 'decision', 'ruleunit'");
-                    LOGGER.warn("defaulting to script type");
-                    String script = ServerlessWorkflowUtils.scriptFunctionScript("");
-                    current = factory.scriptNode(idCounter.getAndIncrement(), action.getFunctionRef().getRefName(), script, embeddedSubProcess);
-
-                    factory.connect(start.getId(), current.getId(), start.getId() + "_" + current.getId(), embeddedSubProcess);
-                    start = current;
+                    current = unsupportedNode(action, embeddedSubProcess);
                 }
+                factory.connect(start.getId(), current.getId(), start.getId() + "_" + current.getId(), embeddedSubProcess);
+                start = current;
             }
             EndNode embeddedEndNode = factory.endNode(idCounter.getAndIncrement(), "EmbeddedEnd", true, embeddedSubProcess);
             try {
@@ -460,6 +511,18 @@ public class ServerlessWorkflowParser {
                 LOGGER.warn("unable to connect current node to embedded end node");
             }
         }
+    }
+
+    private Node unsupportedNode(Action action, CompositeContextNode embeddedSubProcess) {
+        LOGGER
+            .warn(
+                  "currently unsupported function type, supported types are 'script', 'sysout', 'service', 'decision', 'rule', 'integration', 'rest'");
+        LOGGER.warn("defaulting to script type");
+        return factory.scriptNode(
+                        idCounter.getAndIncrement(),
+                        action.getFunctionRef().getRefName(),
+                        ServerlessWorkflowUtils.scriptFunctionScript(""),
+                        embeddedSubProcess);
     }
 
 }
--- a/jbpm/jbpm-serverless-workflow/src/main/java/org/jbpm/serverless/workflow/parser/core/ServerlessWorkflowFactory.java
+++ b/jbpm/jbpm-serverless-workflow/src/main/java/org/jbpm/serverless/workflow/parser/core/ServerlessWorkflowFactory.java
@@ -14,6 +14,12 @@
  */
 package org.jbpm.serverless.workflow.parser.core;
 
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+
 import com.fasterxml.jackson.databind.JsonNode;
 import org.drools.compiler.rule.builder.dialect.java.JavaDialect;
 import org.jbpm.process.core.Work;
@@ -31,6 +37,7 @@ import org.jbpm.serverless.workflow.api.Workflow;
 import org.jbpm.serverless.workflow.api.end.End;
 import org.jbpm.serverless.workflow.api.events.EventDefinition;
 import org.jbpm.serverless.workflow.api.functions.Function;
+import org.jbpm.serverless.workflow.api.functions.FunctionRef;
 import org.jbpm.serverless.workflow.parser.util.ServerlessWorkflowUtils;
 import org.jbpm.serverless.workflow.parser.util.WorkflowAppContext;
 import org.jbpm.workflow.core.DroolsAction;
@@ -39,12 +46,25 @@ import org.jbpm.workflow.core.impl.ConnectionImpl;
 import org.jbpm.workflow.core.impl.ConstraintImpl;
 import org.jbpm.workflow.core.impl.DroolsConsequenceAction;
 import org.jbpm.workflow.core.impl.ExtendedNodeImpl;
-import org.jbpm.workflow.core.node.*;
+import org.jbpm.workflow.core.node.ActionNode;
+import org.jbpm.workflow.core.node.CompositeContextNode;
+import org.jbpm.workflow.core.node.EndNode;
+import org.jbpm.workflow.core.node.EventNode;
+import org.jbpm.workflow.core.node.EventTrigger;
+import org.jbpm.workflow.core.node.HumanTaskNode;
+import org.jbpm.workflow.core.node.Join;
+import org.jbpm.workflow.core.node.RuleSetNode;
+import org.jbpm.workflow.core.node.Split;
+import org.jbpm.workflow.core.node.StartNode;
+import org.jbpm.workflow.core.node.SubProcessNode;
+import org.jbpm.workflow.core.node.TimerNode;
+import org.jbpm.workflow.core.node.WorkItemNode;
 import org.kie.api.definition.process.Node;
+import org.kogito.workitem.rest.RestWorkItemHandler;
+import org.kogito.workitem.rest.jsonpath.suppliers.JsonPathExprSupplier;
+import org.kogito.workitem.rest.jsonpath.suppliers.JsonPathResultExprSupplier;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import java.util.*;
 
 public class ServerlessWorkflowFactory {
     private static final Logger LOGGER = LoggerFactory.getLogger(ServerlessWorkflowFactory.class);
@@ -73,7 +93,7 @@ public class ServerlessWorkflowFactory {
     public static final String HT_ACTORID = "actorid";
     public static final String RF_GROUP = "ruleflowgroup";
     public static final String SERVICE_TASK_TYPE = "Service Task";
-
+   
     private WorkflowAppContext workflowAppContext;
 
     public ServerlessWorkflowFactory(WorkflowAppContext workflowAppContext) {
@@ -322,6 +342,39 @@ public class ServerlessWorkflowFactory {
 
         return workItemNode;
     }
+    
+    public Node restServiceNode(long id,
+                                FunctionRef functionRef,
+                                Function function,
+                                NodeContainer nodeContainer) {
+        WorkItemNode workItemNode = new WorkItemNode();
+        workItemNode.setId(id);
+        workItemNode.setName(function.getName());
+        workItemNode.setMetaData("Type", RestWorkItemHandler.REST_TASK_TYPE);
+
+        Work work = new WorkImpl();
+        workItemNode.setWork(work);
+        work.setName(RestWorkItemHandler.REST_TASK_TYPE);
+        work.setParameter(RestWorkItemHandler.ENDPOINT, function.getResource());
+        work.setParameter(RestWorkItemHandler.METHOD, ServerlessWorkflowUtils.resolveFunctionMetadata(function, RestWorkItemHandler.METHOD, workflowAppContext));
+        work.setParameter(RestWorkItemHandler.USER, ServerlessWorkflowUtils.resolveFunctionMetadata(function, RestWorkItemHandler.USER, workflowAppContext));
+        work.setParameter(RestWorkItemHandler.PASSWORD, ServerlessWorkflowUtils.resolveFunctionMetadata(function, RestWorkItemHandler.PASSWORD, workflowAppContext));
+        work.setParameter(RestWorkItemHandler.HOST, ServerlessWorkflowUtils.resolveFunctionMetadata(function, RestWorkItemHandler.HOST, workflowAppContext));
+        work.setParameter(RestWorkItemHandler.PORT, ServerlessWorkflowUtils.resolveFunctionMetadata(function, RestWorkItemHandler.PORT, workflowAppContext));
+        
+        if (functionRef.getParameters() != null) {
+            for (Entry<String, String> param : functionRef.getParameters().entrySet()) {
+                // assuming param value is json string path
+                work.setParameter(param.getKey(), new JsonPathExprSupplier(param.getValue()));
+            }
+        }
+        work.setParameter(RestWorkItemHandler.RESULT_HANDLER, new JsonPathResultExprSupplier());
+        workItemNode.addInMapping(RestWorkItemHandler.PARAMETER, DEFAULT_WORKFLOW_VAR);
+        workItemNode.addOutMapping(RestWorkItemHandler.RESULT, DEFAULT_WORKFLOW_VAR);
+      
+        nodeContainer.addNode(workItemNode);
+        return workItemNode;
+    }
 
     public WorkItemNode serviceNode(long id, String name, Function function, NodeContainer nodeContainer) {
         WorkItemNode workItemNode = new WorkItemNode();
@@ -486,5 +539,4 @@ public class ServerlessWorkflowFactory {
             throw new RuntimeException("Workflow could not be validated !");
         }
     }
-
 }
--- a/kogito-bom/pom.xml
+++ b/kogito-bom/pom.xml
@@ -910,6 +910,17 @@
         <version>${project.version}</version>
         <classifier>sources</classifier>
       </dependency>
+       <dependency>
+        <groupId>org.kie.kogito</groupId>
+        <artifactId>kogito-rest-workitem</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.kie.kogito</groupId>
+        <artifactId>kogito-rest-workitem</artifactId>
+        <version>${project.version}</version>
+        <classifier>sources</classifier>
+      </dependency>
       <dependency>
         <groupId>org.kie.kogito</groupId>
         <artifactId>jbpm-serverless-workflow</artifactId>
--- a/kogito-build-parent/pom.xml
+++ b/kogito-build-parent/pom.xml
@@ -75,6 +75,7 @@
     <version.taglist.plugin>2.4</version.taglist.plugin>
     <version.versions.plugin>2.1</version.versions.plugin>
     <version.jayway.jsonpath>2.4.0</version.jayway.jsonpath>
+    <version.vertx.web.client>3.9.2</version.vertx.web.client>
 
     <!-- this version property is used in plugins but also in dependencies too -->
     <version.io.quarkus>1.8.0.Final</version.io.quarkus>
@@ -349,6 +350,12 @@
         <groupId>com.jayway.jsonpath</groupId>
         <artifactId>json-path-assert</artifactId>
         <version>${version.jayway.jsonpath}</version>
+      </dependency>
+      
+      <dependency>
+        <groupId>io.vertx</groupId>
+        <artifactId>vertx-web-client</artifactId>
+        <version>${version.vertx.web.client}</version>
       </dependency>
 
       <!-- http communication -->
@@ -712,7 +719,6 @@
         <artifactId>mongodb-driver-sync</artifactId>
         <version>${version.org.mongo}</version>
       </dependency>
-
     </dependencies>
   </dependencyManagement>
 
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessGenerator.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessGenerator.java
@@ -388,38 +388,19 @@ public class ProcessGenerator {
                 Parameter parameter = new Parameter(clazzNameType, varName);
                 if (useInjection()) {
                     annotator.withApplicationComponent(handlerClazz);
+                    annotator
+                        .withInjection(
+                                       handlerClazz
+                                           .getConstructors()
+                                           .stream()
+                                           .filter(c -> !c.getParameters().isEmpty())
+                                           .findFirst()
+                                           .orElseThrow(
+                                                        () -> new IllegalStateException(
+                                                            "Cannot find a non empty constructor to annotate in handler class " +
+                                                                                        handlerClazz)),true);
                 }
-                // generate constructor with handler parameters
-                for (FieldDeclaration fd : handler.getValue().findAll(FieldDeclaration.class)) {
-                    String fieldName = fd.getVariable(0).getNameAsString();
-                    handlerClazz
-                        .addConstructor(Keyword.PUBLIC)
-                        .setBody(
-                                 new BlockStmt()
-                                     .addStatement(
-                                                   new AssignExpr(
-                                                       new FieldAccessExpr(
-                                                           new ThisExpr(),
-                                                           fd.getVariable(0).getNameAsString()),
-                                                       new ObjectCreationExpr()
-                                                           .setType(fd.getVariable(0).getType().toString()),
-                                                       AssignExpr.Operator.ASSIGN)));
-                    ConstructorDeclaration handlerConstructor = handlerClazz
-                        .addConstructor(Keyword.PUBLIC)
-                        .addParameter(fd.getVariable(0).getType(), fieldName)
-                        .setBody(
-                                 new BlockStmt()
-                                     .addStatement(
-                                                   new AssignExpr(
-                                                       new FieldAccessExpr(
-                                                           new ThisExpr(),
-                                                           fieldName),
-                                                       new NameExpr(fieldName),
-                                                       AssignExpr.Operator.ASSIGN)));
-                    if (useInjection()) {
-                        annotator.withInjection(handlerConstructor, true);
-                    }
-                }
+             
                 initMethodCall
                     .addArgument(
                                  new ObjectCreationExpr(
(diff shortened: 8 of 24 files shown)


fjtirado:
Currently working on replacing jax-rs by vert.x mutiny client (https://quarkus.io/guides/vertx#using-vert-x-clients). Since vertx uses injection, this requires few changes on codegen, so a similar approach to current service handlers is followed (partially generated, partyally constant)

jbpm/jbpm-flow/pom.xml
```diff
@@ -53,14 +53,23 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
+    </dependency>      
+    <dependency>
+        <groupId>com.jayway.jsonpath</groupId>
+        <artifactId>json-path</artifactId>
```
ricardozanini, line 59: @tsurdilo don't we already have this dependency in your PR? #768

kogito-build-parent/pom.xml
```diff
@@ -119,7 +120,8 @@
     <version.org.infinispan>11.0.1.Final</version.org.infinispan>
     <version.org.infinispan.protostream>4.3.3.Final</version.org.infinispan.protostream>
     <version.org.infinispan.starter>2.2.3.Final</version.org.infinispan.starter>
-    <version.org.jboss.spec.javax.ws.rs.jboss-jaxrs-api_2.1_spec>1.0.1.Final</version.org.jboss.spec.javax.ws.rs.jboss-jaxrs-api_2.1_spec>
+    <version.org.jboss.spec.javax.ws.rs.jboss-jaxrs-api_2.1_spec>2.0.1.Final</version.org.jboss.spec.javax.ws.rs.jboss-jaxrs-api_2.1_spec>
```
ricardozanini, line 123: This is a big name for a property! :sweat_smile:

ricardozanini (commented):
I did not see nothing wrong if this implementation, but I think a demo would be a nice to have. Maybe a call tomorrow morning showing case your example, @fjtirado 
@evacchi @cristianonicolai @tsurdilo what u think?

kogito-build-parent/pom.xml
fjtirado: this needs to be removed, since is not longer used

jbpm/jbpm-flow/pom.xml
fjtirado: > I did not see nothing wrong if this implementation, but I think a demo would be a nice to have. Maybe a call tomorrow morning showing case your example, @fjtirado
> @evacchi @cristianonicolai @tsurdilo what u think?

Please check example in this PR https://github.com/kiegroup/kogito-examples/pull/369

fjtirado:
> I did not see nothing wrong if this implementation, but I think a demo would be a nice to have. Maybe a call tomorrow morning showing case your example, @fjtirado
> @evacchi @cristianonicolai @tsurdilo what u think?

Please check serverless workflow example in https://github.com/kiegroup/kogito-examples/pull/369

tsurdilo:
Initial reaction is to ask -> How is this better or different than what we have in v7? The jBPM rest workitem handler is pretty much same but more powerful - https://github.com/kiegroup/jbpm/blob/master/jbpm-workitems/jbpm-workitems-rest/src/main/java/org/jbpm/process/workitem/rest/RESTWorkItemHandler.java? Could we align the two so at least they have the same functionalities? 

Also the "type" check on SW markup function def is most likely going away. We will remove the type param from the specification prob soon, so that will need to be all updated. It would be nice if the impl is just concerned with the workitem and not too much worry about the Impl with SW atm. WDYT?

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/JSonPathResultHandler.java
```diff
@@ -0,0 +1,111 @@
...
+
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class JSonPathResultHandler implements BiFunction<Object, JsonObject, Object> {
```
tsurdilo, line 27: If you are addressing results of a JsonPath query, it will/should always be an array.

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/JsonPathResolver.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.function.UnaryOperator;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.jayway.jsonpath.Configuration;
+import com.jayway.jsonpath.JsonPath;
+import com.jayway.jsonpath.spi.json.JacksonJsonNodeJsonProvider;
+import com.jayway.jsonpath.spi.mapper.JacksonMappingProvider;
+
+public class JsonPathResolver implements UnaryOperator<Object> {
```
tsurdilo, line 29: Are we exposing this to users in public api? If so it could also be used insides script expressions which is nice.
I would change it tho so it can be used easier, for example JsonPathResolver.resolve(....). much easier to deal w/ in expressions as such

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/RestTaskHandler.java
```diff
@@ -0,0 +1,154 @@
...
+
+    @Override
+    public void abortWorkItem(WorkItem workItem, WorkItemManager manager) {
+        // rest item handler does not support abort
+    }
+
+    //  package scoped to allow unit test
+    static String resolvePathParams(String endPoint, Map<String, Object> parameters, UnaryOperator<Object> resolver) {
+        Set<String> toRemove = new HashSet<>();
+        int start = endPoint.indexOf('{');
```
tsurdilo, line 121: Don't get this:
1) It seems very workflow specific , but then again in new version the queries are enclosed in '{{' and '}}'.
2) Data passed to functions even in SW are not always expressions. It can be custom JSON objects as well. 
3) This seems not useful for BPMN2 scenarios no?
'

jbpm/jbpm-serverless-workflow/src/main/java/org/jbpm/serverless/workflow/parser/core/ServerlessWorkflowFactory.java
```diff
@@ -322,6 +340,37 @@ public WorkItemNode camelRouteServiceNode(long id, String name, Function functio
 
         return workItemNode;
     }
+    
+    public Node restServiceNode(long id,
```
tsurdilo, line 344: Wonder since this defines workitem we can have a generic getWorkitemNode(....) method that can handle any future workitems that we might expose in kogito like this one.

jbpm/jbpm-serverless-workflow/src/main/java/org/jbpm/serverless/workflow/parser/core/ServerlessWorkflowFactory.java
```diff
@@ -322,6 +340,37 @@ public WorkItemNode camelRouteServiceNode(long id, String name, Function functio
...
+        workItemNode.setWork(work);
+        work.setName(RestTaskHandler.REST_TASK_TYPE);
+        work.setParameter(RestTaskHandler.ENDPOINT, function.getResource());
+        work.setParameter(RestTaskHandler.METHOD, ServerlessWorkflowUtils.resolveFunctionMetadata(function, RestTaskHandler.METHOD, workflowAppContext));
+        work.setParameter(RestTaskHandler.USER, ServerlessWorkflowUtils.resolveFunctionMetadata(function, RestTaskHandler.USER, workflowAppContext));
+        work.setParameter(RestTaskHandler.PASSWORD, ServerlessWorkflowUtils.resolveFunctionMetadata(function, RestTaskHandler.PASSWORD, workflowAppContext));
+        
+        if (functionRef.getParameters() != null) {
+            for (Entry<String, String> param : functionRef.getParameters().entrySet()) {
+                // assuming param value is json string path
```
tsurdilo, line 363: Yeah, not a very correct assumption :) It can also be custom JSON data.

tsurdilo:
@ricardozanini I left some small comments, and an overall question :) Will look into more details laster today.

I think that this pr addresses a future demo needs maybe that you are trying to create which is nice.
The thing I was expecting is for a PR like this to only include RestTaskHandler. I think it is pretty obvious that even if we ported the v7 workitem handler to kogito that "rest calls would work" for both bpmn2 and workflow. 
The addition of the SW specific stuff is nice, but again the workflow backend is really moving fast, with spec changes especially that for kogito-runtiomes I would only look at the workitem itself and not the other parts added.

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/JSonPathResultHandler.java
fjtirado: The resulf of the rest invocation needs to be merged into the json tree, so json queries can be performed over it. 
The point is that defailt of the conversion are hidden into that object, so no serverless no handler needs to be changes, just this object (or invoke a different one in generator)

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/JsonPathResolver.java
fjtirado: No, this is not a public API. And even if it was, it is implementin a well know interface defined in jdk8, user will expect an apply as publis method (but it does not matter since it is not public)

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/RestTaskHandler.java
fjtirado: This code is replacing parameters in the endpoint. JAX-RS convention is to include the path parameters to be replaced between {}. 
This applies both for POJO/Json parameters, both for BPMN/SW.

jbpm/jbpm-serverless-workflow/src/main/java/org/jbpm/serverless/workflow/parser/core/ServerlessWorkflowFactory.java
fjtirado: I think parameters will be different depending on the work item type, probably some of this code can be moved to a common method, but most parameters names will differ.

jbpm/jbpm-serverless-workflow/src/main/java/org/jbpm/serverless/workflow/parser/core/ServerlessWorkflowFactory.java
fjtirado: Yes, I introduced that comment to make possible that, when JSON pojo support is added to SW, a new Supplier that deals with POJO can be added here

fjtirado:
> Also the "type" check on SW markup function def is most likely going away. We will remove the type param from the specification prob soon, so that will need to be all updated. It would be nice if the impl is just concerned with the workitem and not too much worry about the Impl with SW atm. WDYT?

I needed to check that the example worked, also I need to understand the whole scenario to define the interface of the handler (the parameters requested). I do not think any of the changes performed in the SW parser jeopardize any evolution of it, I do not pretend them to be definitive, Im just trying to make the scenario work end to end. As specified in the doc that I took as "requirements" type equal to "rest" is the way to identifiy a function that is going to call a rest interface.

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/JsonPathResolver.java
fjtirado: Please note that this class is for now specific to SW, since it is dealing with json path resolution

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/JSonPathResultHandler.java
tsurdilo: Ok, what I'm thinking is that it would be better for SW to deal with this via a script task that does this, than having this code in /jbpm/process/instance. Reason being is that SW should be able to do this independently of the bpmn2 processes for example. Or allow for SW to overwrite this class if possible, so it can evolve this independently

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/RestTaskHandler.java
tsurdilo: Ok. There are things missing here like accept headers, read timeouts, content type etc etc. Also expecting a hard-coded password in the workflow markup is something noone is going to really use as its pretty much only for demos I assume. 
I hardly think that even for a demo anyone out there that still uses hard-coded passwords in their apps is even thinking about moving to a serverless architectures. wdyt?

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/JsonPathResolver.java
tsurdilo: I really do not care if it uses a well or not well known interface or not :) My concern is about usability.

tsurdilo:
> As specified in the doc that I took as "requirements" , type equal to "rest" is the current way to identify a function that is going to call a rest interface.

Yeah that's why I keep repeating myself to try to do this as generic as possible. I am updating the specification alot so its a moving target. When we started the integration with kogito the type param was the most natural way to distinguish but its really not portable at all. The spec is going to soon remove the type param and replace it completely most likely with forcing an openapi def. Runtimes that want to do something else can use the metadata to define this but it will be noted that its not portable. So we will have to look at this impl anyways alot more and change things in the future.

ruromero (changes requested):

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/JSonPathResultHandler.java
fjtirado: SW is specifying that this class should be used when populating the workitem with parameters 
I think this is cleaner that forcing the sw writer to introduce a script task

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/JsonPathResolver.java
fjtirado: I first put this class in sw project, but it turns out that classes there are not used at runtime, thats the only reason this class was add to flow, but I agree is not the better place. The goal of doing so is that this class should be in classpath at runtime

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/RestTaskHandler.java
fjtirado: RestTaskHadled needs evolution (as the SW flow ;)). I think at this stage we should care about the contract (aka how the serverles parser fills the work node item)
Content type is assumed to be json (is going to be different in any case?)
Headers can be specified as a parameter of the workitem (as method, user and password)
I agree that password for basic authentication needs to be at least encrypted (which encryption?)

fjtirado:
> > As specified in the doc that I took as "requirements" , type equal to "rest" is the current way to identify a function that is going to call a rest interface.
> 
> Yeah that's why I keep repeating myself to try to do this as generic as possible. I am updating the specification alot so its a moving target. When we started the integration with kogito the type param was the most natural way to distinguish but its really not portable at all. The spec is going to soon remove the type param and replace it completely most likely with forcing an openapi def. Runtimes that want to do something else can use the metadata to define this but it will be noted that its not portable. So we will have to look at this impl anyways alot more and change things in the future.

When you update the specification, you will change the way the node factories are invoked by the parser, you can do the same with the new one

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/JSonPathResultHandler.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.function.BiFunction;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+
+public class JSonPathResultHandler implements BiFunction<Object, JsonNode, Object> {
+    @Override
+    public Object apply(Object inputParameter, JsonNode node) {
+        ObjectNode result = (ObjectNode) inputParameter;
```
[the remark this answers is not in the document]
fjtirado: Null check is already performed in the Handler (inputModel is taking from parameter, which a mandatory attribute)
If this class in being invoked is because SW fill ResultHander parameter and it did that because the Parameter object is an ObjectNode.

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/JSonPathResultHandler.java
```diff
@@ -0,0 +1,111 @@
...
+            ((JsonArray)value).forEach(v -> addValue(array,v));
+            result.set(key,array);
+        }
+        else {
+            logger.warn("Unrecognized data type for object {} class {}",value,value.getClass());
+        }
+    }
+    
+    
+    private void addValue (ArrayNode result, Object value) {
```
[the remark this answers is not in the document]
fjtirado: Thanks, I have the formatter disable by default

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/JSonPathResultHandler.java
```diff
@@ -0,0 +1,111 @@
...
+public class JSonPathResultHandler implements BiFunction<Object, JsonObject, Object> {
+    
+    private static final Logger logger = LoggerFactory.getLogger(JSonPathResultHandler.class);
+    
+    @Override
+    public Object apply(Object inputParameter, JsonObject node) {
+        return transform(node,(ObjectNode) inputParameter);
+    }
+    
+    private ObjectNode transform (JsonObject src, ObjectNode target) {
```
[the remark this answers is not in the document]
fjtirado: Since I have to merge the input with the response, I prefer to avoid encode/decode step and perform the merge programatically

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/RestTaskHandler.java
```diff
@@ -0,0 +1,154 @@
...
+    @Override
+    public void executeWorkItem(WorkItem workItem, WorkItemManager manager) {
+        // retrieving parameters
+        Map<String, Object> parameters = new HashMap<>(workItem.getParameters());
+        String endPoint = (String) getParam(parameters, ENDPOINT);
+        String method = (String) getParam(parameters, METHOD);
+        Object inputModel = getParam(parameters, PARAMETER);
+        String user = (String) parameters.remove(USER);
+        String password = (String) parameters.remove(PASSWORD);
+        @SuppressWarnings("unchecked")
```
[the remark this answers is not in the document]
fjtirado: yes, but as a generic rule the annotation should in lowest possible scope to prevent other unchecked casting to be unnoticed

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/RestTaskHandler.java
```diff
@@ -0,0 +1,154 @@
...
+        UnaryOperator<Object> resolver = new RestUnaryOperator(inputModel);
+        endPoint = resolvePathParams(endPoint, parameters, resolver);
+        URI uri = URI.create(endPoint);
+        HttpRequest<Buffer> request = client
+            .request(HttpMethod.valueOf(method.toUpperCase()), uri.getPort(), uri.getHost(), uri.getPath());
+
+        if (user != null && !user.trim().isEmpty() && password != null && !password.trim().isEmpty()) {
+            request.basicAuthentication(user, password);
+        }
+        HttpResponse<Buffer> response = method.equalsIgnoreCase("POST") || method.equalsIgnoreCase("PUT")
```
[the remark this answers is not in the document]
fjtirado: sure, thanks

jbpm/jbpm-flow/pom.xml
evacchi: I wonder if we can make this available as an addon so that we don't pollute the core with external deps (maybe another separate jira)

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/RestTaskHandler.java
ricardozanini: Security constraints should be handled by injected configuration properties, please see: https://docs.google.com/document/d/1GgAEAAoiBWsQ3zDKsW_w_Uml4cg5C5dnG6tlGQZlY0g/edit#heading=h.9qnfgrgjxwzl

ricardozanini: There's a JIRA for that: https://issues.redhat.com/browse/KOGITO-3160

evacchi (commented):
In general, I am not really fond of the usage of Unary/Function/BiFunction as classes; for instance, they are not easily searchable; e.g. if you're looking for all impls that may be used you'll find a ton of them. 

 Wherever there is a specific usage -- i.e., non-lambda -- I'd rather see a named interface with a descriptive name; otherwise I'll find myself asking "what is BiFunction" doing ? 
 
 For the Supplier<T> I probably don't care that much, but for Function/BiFunciton, if I am understanding correctly, the purpose is to customize the WorkItem implementation; so why not e.g. defining an (possibly abstract, if there are no meaningful defaults) rest workitem impl, and override those at use-sites?

In this case we would probably have at least 2 meaningful sub-classes 1 for plain processes and 1 for serverless workflows.

WDYT?

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/RestTaskDescriptor.java
```diff
@@ -0,0 +1,75 @@
...
+import com.github.javaparser.ast.stmt.BlockStmt;
+import io.vertx.mutiny.core.Vertx;
+import io.vertx.mutiny.ext.web.client.WebClient;
+import org.jbpm.process.instance.impl.rest.RestTaskHandler;
+
+public class RestTaskDescriptor {
+
+    private RestTaskDescriptor() {}
+
+    public static String getClassName(ProcessMetaData processMetadata) {
```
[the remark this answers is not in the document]
ricardozanini: I asked myself the same thing.

ricardozanini:
> @ricardozanini I left some small comments, and an overall question :) Will look into more details laster today.
> 
> I think that this pr addresses a future demo needs maybe that you are trying to create which is nice.
> The thing I was expecting is for a PR like this to only include RestTaskHandler. I think it is pretty obvious that even if we ported the v7 workitem handler to kogito that "rest calls would work" for both bpmn2 and workflow.
> The addition of the SW specific stuff is nice, but again the workflow backend is really moving fast, with spec changes especially that for kogito-runtimes I would only look at the workitem itself and not the other parts added.
> 
> The rest of stuff like the integrations with the workflow maybe can be done inside a separate branch you use to create your demos from, until we figure out what exactly to change for specification. Would that work for you ?

Well, I won't run any demos requiring this feature for now, so I'd say that we should agree with an implementation and move forward. After finishing the work in #768, I'll finish the example https://github.com/kiegroup/kogito-examples/pull/351. And then I will circle back to this and update that demo to use `rest` instead of `interface`, manually wrapping the REST calls within the project.

So, whatever you guys decide. I will get back to it to add Kubernetes SVC discovery to the endpoints anyway, so I'm on your hands.

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/RestTaskDescriptor.java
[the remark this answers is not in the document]
fjtirado: because the exact same thing is donde for ServiceTaskHander. I can use template, but the code is going to be the same

fjtirado:
> In general, I am not really fond of the usage of Unary/Function/BiFunction as classes; for instance, they are not easily searchable; e.g. if you're looking for all impls that may be used you'll find a ton of them.
> 
> Wherever there is a specific usage -- i.e., non-lambda -- I'd rather see a named interface with a descriptive name; otherwise I'll find myself asking "what is BiFunction" doing ?
> 
> For the Supplier I probably don't care that much, but for Function/BiFunciton, if I am understanding correctly, the purpose is to customize the WorkItem implementation; so why not e.g. defining an (possibly abstract, if there are no meaningful defaults) rest workitem impl, and override those at use-sites?
> 
> In this case we would probably have at least 2 meaningful sub-classes 1 for plain processes and 1 for serverless workflows.
> 
> WDYT?

I was trying to avoid having specific subclasses for sw or bpmn there, thats why Im using Function. I think the usage is pretty clear, but what we can do is to define a specific interface with a well define name that extends BiFunction and the issue of searching the implementation will not longer be a problem

Regarding the jsonpath dependency, I think is prery clear that JsonPathResolver and that dependency should be outside of flow, but in a place where they are included in runtime classpath (serverles workflow is used only at compilation)

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/RestTaskDescriptor.java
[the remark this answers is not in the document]
evacchi: I'd prefer to move as much code as possible to templates, especially new code :)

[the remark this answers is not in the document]
fjtirado: yes, Ill move to template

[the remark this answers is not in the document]
fjtirado: done

fjtirado:
All comments incorporated. 
Since the functional is creating some controversy, let me try to explain my view and why Im using it rather than an OOP approach. 
Im trying to implement a Strategy pattern, the objective is that RestTaskHandler is not aware of the different way to resolve a parameter to be included in the rest invocation. So rather than introduce an if (POJO) then or if (JSONTREE) in the handler itslef (which implies that the Handler is aware of the different invokers and also implies that the handler needs to be modified if a third actor enters into the picture) , the different parameter are evaluated against workitem input using a Function provided by the creator of the work item node (in this case SW parser) 
Thats the overall idea and I think all agree on that. 
Now the details. 
Currently the two function objects, the one to be used for parameter resolution and the one to be used for output merging are located in flow project. This classed depends on a json path library. 
This is not the inteded final location, but the currently simplest one in order to achieve the purpose of these classes being added to classpath on runtime (if located in serverless wf project, they wont be in runtime classpath and an exception will be thrown). These classes, together with the library dependency, should be moved to a module, when available. So, take the current one as a temporary location. The important thing is that they are not directly referenced from the handler, so, when the final location is available, they will be moved there without having to change anything else. 
The other quesiton is why functional interface rather than subclassing the evaluate and the merge abstract functions. The answer, a very subttle one, is to give us more flexibility in case some parameters are resolved from JsonTree and anothers from Pojo (that probably will hardly happen). Since these function objects are instantiated by the generators, we can achieve that flexibility without compromising complexity, since the generator, rather than deciding the name of the subclass base on a global approach, can dedice which function to use to resolve a parameter based on each parameter (it is true that currently, since WF uses only Json Tree, all functions instantiated are of the Json tree type, but the point is that we change that just in the generator, without affecting the RestTaskHandler. 
Hope that now the code is clearer to everyone.

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/RestTaskHandler.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    @Override
+    public void executeWorkItem(WorkItem workItem, WorkItemManager manager) {
+        // retrieving parameters
+        Map<String, Object> parameters = new HashMap<>(workItem.getParameters());
+        String endPoint = getParam(parameters, ENDPOINT, String.class);
+        HttpMethod method = HttpMethod.valueOf(getParam(parameters, METHOD, String.class).toUpperCase());
+        Object inputModel = getParam(parameters, PARAMETER, Object.class);
+        String user = (String) parameters.remove(USER);
```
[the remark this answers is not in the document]
fjtirado: me neither, but this is what is already happening in v7 version of the rest handler.

fjtirado:
https://github.com/kiegroup/kogito-runtimes/pull/778#discussion_r493140674
As I said in my last comment, JsonPath will be moved out of flow when we have a better place, but mutiny dependency must stay if we want RestHandler to be part of flow. They has to be together.

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/JsonPathResolver.java
```diff
@@ -0,0 +1,83 @@
...
+    @Override
+    public Object apply(Object context) {
+        JsonNode node = JsonPath
+            .using(jsonPathConfig)
+            .parse(context)
+            .read(jsonPathExpr, JsonNode.class);
+        return readValue(node);
+    }
+    
+    private Object readValue (JsonNode node) {
```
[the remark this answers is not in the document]
fjtirado: formatted

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/WorkItemNodeVisitor.java
```diff
@@ -110,30 +117,43 @@ protected void addWorkItemParameters(Work work, BlockStmt body, String variableN
...
             default:
-                return new StringLiteralExpr(value);
+                return new StringLiteralExpr(value.toString());
         }
     }
-
-
-
+    
+    private Boolean asBoolean (Object value) {
```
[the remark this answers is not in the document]
fjtirado: WorkItemNodeVisitor changes has been formatted now

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/WorkItemNodeVisitor.java
```diff
@@ -110,30 +117,43 @@ protected void addWorkItemParameters(Work work, BlockStmt body, String variableN
...
+        return value instanceof Boolean ? (Boolean) value : Boolean.parseBoolean(value.toString());
+    }
+    
+    private Integer asInteger (Object value) {
```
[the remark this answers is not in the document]
fjtirado: formatted

fjtirado:
Added unit test for increasing coverage

fjtirado:
PR has been updated to solve the dependency issue. 
Rather than in jbpm-flow, RestTaskHandler has been included in a project called jbpm-rest-handler and SW runtime classes (JSonPathResolver and JSonPathResultHandler) has been included in a project called jbpm-serverless-worflow-runtime. 
This way, SW developers can update jbpm-serverless-workflow-runtime when the SE specification evolves and jbpm-flow does not depend on vert.x

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/JsonPathResolver.java
fjtirado: Class has been moved to serverless-workflow-runtime project (so can be edited by SW developers). It now implementes a well defined interface rather than the generic one

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/rest/JSonPathResultHandler.java
fjtirado: Please read https://github.com/kiegroup/kogito-runtimes/pull/778#issuecomment-697583850

jbpm/jbpm-flow/pom.xml
fjtirado: Classes and dependency moved to a different project than flow

jbpm/jbpm-rest-handler/src/main/java/org/jbpm/process/instance/impl/rest/RestTaskHandler.java
```diff
@@ -0,0 +1,155 @@
...
+            Object body = parameters.isEmpty() ? inputModel : parameters
+                .entrySet()
+                .stream()
+                .collect(
+                         Collectors
+                             .toMap(Entry::getKey, e -> resolver.apply(e.getValue())));
+           
+            request.sendJson(body, handler);
+        } else {
+            request.send(handler);
```
tsurdilo, line 108: Any reason we are not sending json here as well? For SW at least the content type should always be application/json.

tsurdilo:
Would it be possible to get tests that call methods from https://petstore.swagger.io/ or something real?

tsurdilo:
> PR has been updated to solve the dependency issue.
> Rather than in jbpm-flow, RestTaskHandler has been included in a project called jbpm-rest-handler and SW runtime classes (JSonPathResolver and JSonPathResultHandler) has been included in a project called jbpm-serverless-worflow-runtime.
> This way, SW developers can update jbpm-serverless-workflow-runtime when the SE specification evolves and jbpm-flow does not depend on vert.x

@cristianonicolai based on this comment, wouldnt it make more sense to just have this whole thing as an addon then? This seems to be a little bit in-between. WDYT?

jbpm/jbpm-rest-handler/src/main/java/org/jbpm/process/instance/impl/rest/RestTaskHandler.java
fjtirado: Line 108 is a GET or DELETE call, so there is not body, unlike POST or PUT call. 
Json output is handled by ResultTaskResultHandler implementation provided In ResultHandler parameter. In the case of the example JsonPathResultHandler

cristianonicolai:
> > PR has been updated to solve the dependency issue.
> > Rather than in jbpm-flow, RestTaskHandler has been included in a project called jbpm-rest-handler and SW runtime classes (JSonPathResolver and JSonPathResultHandler) has been included in a project called jbpm-serverless-worflow-runtime.
> > This way, SW developers can update jbpm-serverless-workflow-runtime when the SE specification evolves and jbpm-flow does not depend on vert.x
> 
> @cristianonicolai based on this comment, wouldnt it make more sense to just have this whole thing as an addon then? This seems to be a little bit in-between. WDYT?

Yep, that makes sense  to me. But how about the SW specific part, are you ok to have this in the sw runtimes module that @fjtirado  proposed here? Problem was that is introduces dependecy to json path that would not be needed for bpmn projects.

fjtirado:
> > Would it be possible to get tests that call api from https://petstore.swagger.io/ or something real?
> 
> @tsurdilo @fjtirado perhaps use something like https://quarkus.io/guides/rest-client#using-a-mock-http-server-for-testsor https://www.testcontainers.org/modules/mockserver/ for integration tests so we don't depend on an external system?

Rest calls are tested in the example https://github.com/kiegroup/kogito-examples/pull/369. The example implements a dummy rest server.

fjtirado:
> se to me. But how about the SW specific part, are you ok to have this in the sw runtimes module that @fjtirado proposed here? Problem was that is introduces dependecy to json path that would not be needed for bpmn projects.

Also in one of his comments, Tihomir was asking about the possibility of chaging those classes (JsonPathResolver and JsonPathResultHander). If they are located in a separate, SW specific project, ha can freely change them without affectig bpmn code

fjtirado:
jenkins retest this

evacchi (approved):
looks good, just left 1 minor comment

evacchi:
there are a few issues with the build @fjtirado

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/WorkItemNodeVisitor.java
```diff
@@ -110,30 +117,42 @@ protected void addWorkItemParameters(Work work, BlockStmt body, String variableN
...
             }
-            body.addStatement(getFactoryMethod(variableName, METHOD_WORK_PARAMETER, new StringLiteralExpr(entry.getKey()), getParameterExpr(paramType, entry.getValue().toString())));
+            body.addStatement(getFactoryMethod(variableName, METHOD_WORK_PARAMETER, new StringLiteralExpr(entry.getKey()), getParameterExpr(paramType, entry.getValue())));
         }
     }
+    
+  
 
-    private Expression getParameterExpr(String type, String value) {
+    private Expression getParameterExpr(String type, Object value) {
```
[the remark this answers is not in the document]
fjtirado: Checked, returning null literal expr in that case

fjtirado:
> there are a few issues with the build @fjtirado

they seem unrelated, something corcerning knative-event addon

ruromero:
jenkins retest this

evacchi:
with the help of @ricardozanini there are a few issues with the pom

```
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[WARNING] 'parent.relativePath' of POM org.kie.kogito:jbpm-rest-handler:1.0.0-SNAPSHOT (/home/ricferna/dev/github.com/kiegroup/kogito-runtimes/jbpm/jbpm-rest-handler/pom.xml) points at org.kie.kogito:jbpm instead of org.kie.kogito:addons, please verify your project structure @ line 5, column 11
[FATAL] Non-resolvable parent POM for org.kie.kogito:jbpm-rest-handler:1.0.0-SNAPSHOT: Could not find artifact org.kie.kogito:addons:pom:1.0.0-SNAPSHOT and 'parent.relativePath' points at wrong local POM @ line 5, column 11
 @
```

this indicating `addons` as a parent but it is not in the `addons` directory so either you move it the right place, or don't make it an addon :)

```
 <parent>
    <artifactId>addons</artifactId>
    <version>1.0.0-SNAPSHOT</version>
  </parent>
  <artifactId>jbpm-rest-handler</artifactId>
  <name>RestHandler</name>
Edoardo Vacchi: he added that project so probably it's broken
```

fjtirado:
> with the help of @ricardozanini there are a few issues with the pom
> 
> ```
> [ERROR] [ERROR] Some problems were encountered while processing the POMs:
> [WARNING] 'parent.relativePath' of POM org.kie.kogito:jbpm-rest-handler:1.0.0-SNAPSHOT (/home/ricferna/dev/github.com/kiegroup/kogito-runtimes/jbpm/jbpm-rest-handler/pom.xml) points at org.kie.kogito:jbpm instead of org.kie.kogito:addons, please verify your project structure @ line 5, column 11
> [FATAL] Non-resolvable parent POM for org.kie.kogito:jbpm-rest-handler:1.0.0-SNAPSHOT: Could not find artifact org.kie.kogito:addons:pom:1.0.0-SNAPSHOT and 'parent.relativePath' points at wrong local POM @ line 5, column 11
>  @
> ```
> 
> this indicating `addons` as a parent but it is not in the `addons` directory so either you move it the right place, or don't make it an addon :)
> 
> ```
>  <parent>
>     <groupId>org.kie.kogito</groupId>
>     <artifactId>addons</artifactId>
>     <version>1.0.0-SNAPSHOT</version>
>   </parent>
>   <artifactId>jbpm-rest-handler</artifactId>
>   <name>RestHandler</name>
> ```

Thanks guys, I made a mistake while copy pasting. I initially tried to add RestTasKHandler as add-on, but later I change my ming and I think it should be part of bpmn core, but not part of flow, so can be added to classpath on demand. 
Pom is fixed now

evacchi:
good let's merge as soon as it is green

MarianMacik:
My review will land in a few minutes as well :)

tsurdilo:
> @tsurdilo Yep, that makes sense to me. But how about the SW specific part, are you ok to have this in the sw runtimes module that @fjtirado proposed here? Problem was that is introduces dependecy to json path that would not be needed for bpmn projects.

@cristianonicolai I do not think that these "specific" SW parts are something that goes beyond just having it run for a demo, as we havent had the chance to actually implement the "real" solution for data management of SW in Kogito. I feel a little rushed here as this implementation is based on whats there **now** and assumes its right but it may be irrelevant in the near future. 

Also this is a workitem and to bring this back one more time it should be as such a reusable component. I think talking about
a workitem and "sw specific parts" in the same sentence just shows a wrong approach tobegin with. 
I don't think having even a "serverless-workflow-runtime" or whatever package or module  makes sense at all. 

So to answer your question I would really like to see this as an addon workitem that is completely independent and does not include any "specific sw features" and then we will figure out how to incorporate its use when needed.

tsurdilo:
> good let's merge as soon as it is green

WDYM lets merge...You are the only one that approved this, and it does not fit the SW/Kogito strategy at all. I don't think that forcing this in will benefit anyone here as the right approach would be to
1) MAKE THIS A STANDALONE + REUSABLE workitem..idk how many times I need to write this hehe :) 
2) DO NOT ADD ANY SW SPECIFIC PARTS..just a workitem that can take in a JSON object body and returns a JSON response..again idk how many times I need to say this as well :)

fjtirado:
> > good let's merge as soon as it is green
> 
> WDYM lets merge...You are the only one that approved this, and it does not fit the SW/Kogito strategy at all. I don't think that forcing this in will benefit anyone here as the right approach would be to
> 
> 1. MAKE THIS A STANDALONE + REUSABLE workitem..idk how many times I need to write this hehe :)
> 2. DO NOT ADD ANY SW SPECIFIC PARTS..just a workitem that can take in a JSON object body and returns a JSON response..again idk how many times I need to say this as well :)

1) RestWorkItemHandler proposed in this PR  is fully reusable and not specific to SW
2) SW specific part is needed for the demo. You can throw it to the bin later if you want.

tsurdilo:
> > > good let's merge as soon as it is green
> > 
> > 
> > WDYM lets merge...You are the only one that approved this, and it does not fit the SW/Kogito strategy at all. I don't think that forcing this in will benefit anyone here as the right approach would be to
> > 
> > 1. MAKE THIS A STANDALONE + REUSABLE workitem..idk how many times I need to write this hehe :)
> > 2. DO NOT ADD ANY SW SPECIFIC PARTS..just a workitem that can take in a JSON object body and returns a JSON response..again idk how many times I need to say this as well :)
> 
> 1. RestWorkItemHandler proposed in this PR  is fully reusable and not specific to SW
> 2. SW specific part is needed for the demo. You can throw it to the bin later if you want.

1) You can say this as many times as you wish and just force your solution on everyone...but it is not reusable .
2) Then just do the demo in some separate repo and throw it away yourself :)

fjtirado:
> > > > good let's merge as soon as it is green
> > > 
> > > 
> > > WDYM lets merge...You are the only one that approved this, and it does not fit the SW/Kogito strategy at all. I don't think that forcing this in will benefit anyone here as the right approach would be to
> > > 
> > > 1. MAKE THIS A STANDALONE + REUSABLE workitem..idk how many times I need to write this hehe :)
> > > 2. DO NOT ADD ANY SW SPECIFIC PARTS..just a workitem that can take in a JSON object body and returns a JSON response..again idk how many times I need to say this as well :)
> > 
> > 
> > 
> > 1. RestWorkItemHandler proposed in this PR  is fully reusable and not specific to SW
> > 2. SW specific part is needed for the demo. You can throw it to the bin later if you want.
> 
> 1. You can say this as many times as you wish and just force your solution on everyone...but it is not reusable .
> 2. Then just do the demo in some separate repo and throw it away yourself :)

why is not resusable? Can you justify that assertion?

evacchi:
> WDYM lets merge...

sorry I got confused by a different PR and thought there were more approvals

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/RestTaskDescriptor.java
```diff
@@ -0,0 +1,41 @@
...
+import com.github.javaparser.ast.body.ConstructorDeclaration;
+
+import static com.github.javaparser.StaticJavaParser.parse;
+
+public class RestTaskDescriptor {
+
+    private RestTaskDescriptor() {}
+
+    public static String getClassName(ProcessMetaData processMetadata) {
+        return processMetadata.getProcessId() + "RestTaskHandler";
```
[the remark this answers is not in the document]
fjtirado: you are right, I rename it

jbpm/jbpm-flow-builder/src/main/resources/class-templates/RestTaskTemplate.java
```diff
@@ -0,0 +1,16 @@
...
+import org.jbpm.process.instance.impl.rest.RestTaskHandler;
+
+public class xxxTaskHandler extends RestTaskHandler {
+
+    public xxxRestTaskHandler() {
+        this(Vertx.vertx());
+    }
+
+    public xxxRestTaskHandler(Vertx vertx) {
+        super(WebClient.create(vertx));
```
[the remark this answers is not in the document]
fjtirado: yep

jbpm/jbpm-flow/pom.xml
```diff
@@ -54,13 +54,13 @@
...
       <artifactId>slf4j-api</artifactId>
     </dependency>
-
     <dependency>
       <groupId>javax.xml.bind</groupId>
       <artifactId>jaxb-api</artifactId>
       <scope>provided</scope>
     </dependency>
 
+
```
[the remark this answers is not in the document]
fjtirado: jbpm-flow was edited several times during this PR, thats why there is an uneeded change, Ill revert

jbpm/jbpm-rest-handler/pom.xml
```diff
@@ -0,0 +1,31 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.kie.kogito</groupId>
+    <artifactId>addons</artifactId>
+    <version>1.0.0-SNAPSHOT</version>
+  </parent>
```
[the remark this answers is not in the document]
fjtirado: yes, I think is fine, if the rest agree (It is not my intenttion to impose that on anyone)

jbpm/jbpm-rest-handler/src/main/java/org/jbpm/process/instance/impl/rest/RestTaskHandler.java
```diff
@@ -0,0 +1,155 @@
...
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.http.HttpMethod;
+import io.vertx.ext.web.client.HttpRequest;
+import io.vertx.ext.web.client.HttpResponse;
+import io.vertx.ext.web.client.WebClient;
+import org.kie.api.runtime.process.WorkItem;
+import org.kie.api.runtime.process.WorkItemHandler;
+import org.kie.api.runtime.process.WorkItemManager;
+
+public class RestTaskHandler implements WorkItemHandler {
```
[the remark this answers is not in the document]
fjtirado: yes

tsurdilo:
> > > > > good let's merge as soon as it is green
> > > > 
> > > > 
> > > > WDYM lets merge...You are the only one that approved this, and it does not fit the SW/Kogito strategy at all. I don't think that forcing this in will benefit anyone here as the right approach would be to
> > > > 
> > > > 1. MAKE THIS A STANDALONE + REUSABLE workitem..idk how many times I need to write this hehe :)
> > > > 2. DO NOT ADD ANY SW SPECIFIC PARTS..just a workitem that can take in a JSON object body and returns a JSON response..again idk how many times I need to say this as well :)
> > > 
> > > 
> > > 
> > > 1. RestWorkItemHandler proposed in this PR  is fully reusable and not specific to SW
> > > 2. SW specific part is needed for the demo. You can throw it to the bin later if you want.
> > 
> > 
> > 
> > 1. You can say this as many times as you wish and just force your solution on everyone...but it is not reusable .
> > 2. Then just do the demo in some separate repo and throw it away yourself :)
> 
> why is not resusable? Can you justify that assertion?

Well, the workitem itself is intended for demo purposes. It does not reflect anything that we actually can consider a production usable component. 
The sw specific parts of the pr are also just for demo purposes only as they do not reflect how we might actually end up implementing the data strategy around SW states and workflows.
The PR forces SW code to use it. The decision aroudn that hasnt been made yet and we might as well if we are looking into such simple stuff (just a simple rest call) do this in a script task expression to have the same outcome.
So I am arguing here because I think it's pretty clear now that this is for demo/pr purposes only, but the pr forces itself to be used within the SW code. 
The workitem itself maybe "reusable" in the sense it can be used but if you look at its scope its just again for demo :) 
Idk what else to say about this ....

evacchi:
> Well, the workitem itself is intended for demo purposes. It does not reflect anything that we actually can consider a production usable component.

@tsurdilo merging on master does not mean anything is "production-ready". If anything this is to move forward with the task. Of course I won't merge with my only approval. I understand that you do not approve of the design, and there aren't any other approving reviews so far, so let's wait for others to chime in.

But please, *everyone*, let's keep the tone friendly; there's no point in arguing over this :)

ricardozanini:
Let me try to clarify:

1. We need a feature to call REST services for BPMN and SW. I designed that doc to have an alternative based on what we have. You changed the implementation from Camel to Vertx because of technical/roadmap reasons. I'm ok with that, I don't care. I just need the feature.

2. Now I understand that this PR will become a real thing? Because was a PoC at first to try to implement the REST calls via JAX-RS/Vertx/Whatever. If so, I agree that we need a more stable code base and do not have any specifics to SW whatsoever.

3. For the Demo I'm working on, I've already written a wrapper manually and my function look like this:

```
 {
      "name": "AddLabels",
      "type": "service",
      "metadata": {
        "interface": "org.kogito.examples.sw.github.workflow.GitHubService",
        "operation": "addLabels"
      }
    },
```

Whereas whould be nice to have something like:

```
 {
      "name": "AddLabels",
      "type": "rest",
      "resource": "http://openshift/myservice/github/label"
      "metadata": {
        "method": "POST",
      }
    },
```

So people would understand that we have an alternative for calling services in a declarative way.

That's what I was looking for.

jbpm/jbpm-rest-handler/src/test/java/org/jbpm/process/instance/impl/rest/TestRestTaskHandler.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class TestRestTaskHandler {
```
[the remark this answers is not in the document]
fjtirado: ok, moving test to the end

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessGenerator.java
```diff
@@ -388,38 +388,19 @@ public ClassOrInterfaceDeclaration classDeclaration() {
...
+                    annotator
+                        .withInjection(
+                                       handlerClazz
+                                           .getConstructors()
+                                           .stream()
+                                           .filter(c -> !c.getParameters().isEmpty())
+                                           .findFirst()
+                                           .orElseThrow(
+                                                        () -> new IllegalStateException(
+                                                            "Cannot a find a non empty constructor to annotate in handler class " +
```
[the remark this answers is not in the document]
fjtirado: good catch! thanks

tsurdilo:
> > Well, the workitem itself is intended for demo purposes. It does not reflect anything that we actually can consider a production usable component.
> 
> @tsurdilo merging on master does not mean anything is "production-ready". If anything this is to move forward with the task. Of course I won't merge with my only approval. I understand that you do not approve of the design, and there aren't any other approving reviews so far, so let's wait for others to chime in.
> 
> But please, _everyone_, let's keep the tone friendly; there's no point in arguing over this :)

@evacchi And that is fine if you guys want to merge this, my problem is 
https://github.com/kiegroup/kogito-runtimes/pull/778/files#diff-552bf6d008d02f358c614dd70191c999R445

This **forces** sw implementation to use it. Remove that part and then you guys put anything you want in :)

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/RestTaskDescriptor.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public static CompilationUnit generateHandlerClassForService(String className) {
+        CompilationUnit compilationUnit =
+                parse(RestTaskDescriptor.class.getResourceAsStream("/class-templates/RestTaskTemplate.java"));
+        compilationUnit.setPackageDeclaration("org.kie.kogito.handlers");
+        compilationUnit.findFirst(ClassOrInterfaceDeclaration.class).ifPresent(c -> c.setName(className));
+        compilationUnit.findAll(ConstructorDeclaration.class).forEach(c -> c.setName(className));
+        return compilationUnit;
```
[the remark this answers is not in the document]
fjtirado: Yes, and also to reuse the same vertx instance (which is injected)

fjtirado:
How can we generate the workitemparameters without modifying current SW codebase?

fjtirado:
> Let me try to clarify:
> 
> 1. We need a feature to call REST services for BPMN and SW. I designed that doc to have an alternative based on what we have. You changed the implementation from Camel to Vertx because of technical/roadmap reasons. I'm ok with that, I don't care. I just need the feature.
> 2. Now I understand that this PR will become a real thing? Because was a PoC at first to try to implement the REST calls via JAX-RS/Vertx/Whatever. If so, I agree that we need a more stable code base and do not have any specifics to SW whatsoever.
> 3. For the Demo I'm working on, I've already written a wrapper manually and my function looks like this:
> 
> ```
>  {
>       "name": "AddLabels",
>       "type": "service",
>       "metadata": {
>         "interface": "org.kogito.examples.sw.github.workflow.GitHubService",
>         "operation": "addLabels"
>       }
>     },
> ```
> 
> Whereas would be nice to have something like:
> 
> ```
>  {
>       "name": "AddLabels",
>       "type": "rest",
>       "resource": "http://openshift/myservice/github/label"
>       "metadata": {
>         "method": "POST",
>       }
>     },
> ```
> 
> So people would understand that we have an alternative for calling services in a declarative way.
> 
> That's what I was looking for.

And this what is impelemented on this PR, in order to do that SW parser needs to be modified to generate the proper workitem.

ricardozanini (approved):

kogito-build-parent/pom.xml
```diff
@@ -350,6 +351,12 @@
         <artifactId>json-path-assert</artifactId>
         <version>${version.jayway.jsonpath}</version>
       </dependency>
+      
+      <dependency>
+        <groupId>io.vertx</groupId>
+        <artifactId>vertx-web-client</artifactId>
+        <version>${version.vertx.web.client}</version>
+    </dependency>
```
cristianonicolai, line 359: please check format

kogito-bom/pom.xml
```diff
@@ -910,6 +910,18 @@
         <version>${project.version}</version>
         <classifier>sources</classifier>
       </dependency>
+      
+       <dependency>
+        <groupId>org.kie.kogito</groupId>
+        <artifactId>jbpm-rest-handler</artifactId>
```
cristianonicolai, line 916: kogito-rest-workitem

jbpm/jbpm-workitems/pom.xml
```diff
@@ -0,0 +1,16 @@
...
+    <artifactId>jbpm</artifactId>
+    <version>1.0.0-SNAPSHOT</version>
+  </parent>
+  <artifactId>jbpm-workitems</artifactId>
+  <packaging>pom</packaging>
+  <name>Kogito :: jbpm :: Workitems</name>
+
+  <description>Kogito work items handlers</description>
+  <modules>
+    <module>jbpm-rest-handler</module>
```
cristianonicolai, line 14: kogito-rest-workitem

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/WorkItemNodeVisitor.java
```diff
@@ -85,8 +87,13 @@ public void visitNode(String factoryField, T node, BlockStmt body, VariableScope
...
             metadata.getGeneratedHandlers().put(mangledName, generatedHandler);
             workName = mangledName;
+        } else if (workName.equals("Rest Task")) {
+            workName = RestTaskDescriptor.getClassName(metadata);
+            metadata
+                .getGeneratedHandlers()
+                .computeIfAbsent(workName,RestTaskDescriptor::generateHandlerClassForService);
         }
-
+
```
cristianonicolai, line 96:
```suggestion

```

jbpm/jbpm-rest-handler/.gitignore
```diff
@@ -0,0 +1,14 @@
+/target
```
cristianonicolai, line 1: probably not needed to add this file, should be covered by the one at the root level.

jbpm/jbpm-serverless-workflow-runtime/src/test/java/RestTaskHandlerTest.java
```diff
@@ -0,0 +1,94 @@
...
+import org.mockito.stubbing.Answer;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class RestTaskHandlerTest {
+    
+    @Test
+    void testGetRestTaskHandler() {
+        WebClient webClient = Mockito.mock(WebClient.class);
```
cristianonicolai, line 46: in most tests, we end up using import static to most Mockito API calls,

jbpm/jbpm-serverless-workflow-runtime/src/test/java/RestTaskHandlerTest.java
```diff
@@ -0,0 +1,94 @@
...
+        Mockito.doAnswer((Answer<Void>) invocation -> {
+            Handler<AsyncResult<HttpResponse<Buffer>>> handler =
+                    (Handler<AsyncResult<HttpResponse<Buffer>>>) invocation.getArgument(0);
+            handler.handle(event);
+            return null;
+        }).when(request).send(Mockito.any(Handler.class));
+        Mockito.when(response.bodyAsJsonObject()).thenReturn(JsonObject.mapFrom(Collections.singletonMap("num", 1)));
+
+        Map<String, Object> parameters =
+                new HashMap<>();
```
cristianonicolai, line 64:
```suggestion
        Map<String, Object> parameters = new HashMap<>();
```

jbpm/jbpm-serverless-workflow-runtime/src/test/java/RestTaskHandlerTest.java
```diff
@@ -0,0 +1,94 @@
...
+        parameters.put("id", new JsonPathResolver("$.id"));
+        parameters.put("name", new JsonPathResolver("$.name"));
+        parameters.put(RestWorkItemHandler.ENDPOINT, "http://localhost:8080/results/{id}/names/{name}");
+        parameters.put(RestWorkItemHandler.METHOD, "GET");
+        parameters.put(RestWorkItemHandler.RESULT_HANDLER, new JSonPathResultHandler());
+        parameters.put(RestWorkItemHandler.PARAMETER, mapper.createObjectNode().put("id", 26).put("name", "pepe"));
+
+        WorkItem workItem = Mockito
+            .mock(
+                  WorkItem.class);
```
cristianonicolai, line 74:
```suggestion
        WorkItem workItem = mock(WorkItem.class);
```

jbpm/jbpm-serverless-workflow/src/main/java/org/jbpm/serverless/workflow/parser/ServerlessWorkflowParser.java
```diff
@@ -403,55 +418,41 @@ protected void handleActions(List<Function> workflowFunctions, List<Action> acti
...
+                            current = factory.humanTaskNode(idCounter.getAndIncrement(), action.getFunctionRef().getRefName(), actionFunction, process, embeddedSubProcess);
+                            break;
+                        case RULE_TYPE: 
+                            current = factory.ruleSetNode(idCounter.getAndIncrement(), action.getFunctionRef().getRefName(), actionFunction, embeddedSubProcess);
+                            break;
+                        case INTEGRATION_TYPE:
+                            current = factory.camelRouteServiceNode(idCounter.getAndIncrement(), action.getFunctionRef().getRefName(), actionFunction, embeddedSubProcess);
+                            break;
+                        case REST_TYPE: 
+                            current = factory.restServiceNode (idCounter.getAndIncrement(), action.getFunctionRef(), actionFunction,embeddedSubProcess);
```
cristianonicolai, line 446:
```suggestion
                            current = factory.restServiceNode(idCounter.getAndIncrement(), action.getFunctionRef(), actionFunction, embeddedSubProcess);
```

jbpm/jbpm-serverless-workflow/src/main/java/org/jbpm/serverless/workflow/parser/ServerlessWorkflowParser.java
```diff
@@ -403,55 +418,41 @@ protected void handleActions(List<Function> workflowFunctions, List<Action> acti
...
                     }
                 } else {
-                    LOGGER.warn("invalid function type. supported types are 'script', 'sysout', 'service', 'decision', 'ruleunit'");
-                    LOGGER.warn("defaulting to script type");
-                    String script = ServerlessWorkflowUtils.scriptFunctionScript("");
-                    current = factory.scriptNode(idCounter.getAndIncrement(), action.getFunctionRef().getRefName(), script, embeddedSubProcess);
-
-                    factory.connect(start.getId(), current.getId(), start.getId() + "_" + current.getId(), embeddedSubProcess);
-                    start = current;
+                   current = unsupportedNode(action,embeddedSubProcess);
```
cristianonicolai, line 452:
```suggestion
                   current = unsupportedNode(action, embeddedSubProcess);
```

jbpm/jbpm-serverless-workflow/src/main/java/org/jbpm/serverless/workflow/parser/ServerlessWorkflowParser.java
```diff
@@ -461,5 +462,11 @@ protected void handleActions(List<Function> workflowFunctions, List<Action> acti
             }
         }
     }
+    
+    private Node unsupportedNode (Action action, CompositeContextNode embeddedSubProcess) {
```
cristianonicolai, line 466:
```suggestion
    private Node unsupportedNode(Action action, CompositeContextNode embeddedSubProcess) {
```

jbpm/jbpm-serverless-workflow/src/main/java/org/jbpm/serverless/workflow/parser/core/JsonPathExprSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+import com.github.javaparser.ast.expr.Expression;
+import com.github.javaparser.ast.expr.ObjectCreationExpr;
+import com.github.javaparser.ast.expr.StringLiteralExpr;
+import org.jbpm.serverless.workflow.JsonPathResolver;
+
+public class JsonPathExprSupplier implements Supplier<Expression> {
+    
+    private String jsonPathExpr;
+    
+    public JsonPathExprSupplier (String jsonPathExpr) {
```
cristianonicolai, line 28:
```suggestion
    public JsonPathExprSupplier(String jsonPathExpr) {
```

jbpm/jbpm-rest-handler/pom.xml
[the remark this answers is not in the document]
cristianonicolai: @MarianMacik @fjtirado I think its ok to have the new module, but would prefer to name it kogito-workitems and kogito-rest-workitem, as this is kogito specific.

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ServiceTaskDescriptor.java
```diff
@@ -170,9 +173,25 @@ public ClassOrInterfaceDeclaration classDeclaration() {
...
         ClassOrInterfaceType serviceType = new ClassOrInterfaceType(null, interfaceName);
 
+        final String serviceName = "service";
         FieldDeclaration serviceField = new FieldDeclaration()
-                .addVariable(new VariableDeclarator(serviceType, "service"));
+                .addVariable(new VariableDeclarator(serviceType, serviceName));
         cls.addMember(serviceField);
+        cls
+            .addConstructor(Keyword.PUBLIC)
+            .setBody(new BlockStmt().addStatement(new MethodCallExpr("this",new ObjectCreationExpr().setType(serviceType))));
```
cristianonicolai, line 182:
```suggestion
            .setBody(new BlockStmt().addStatement(new MethodCallExpr("this", new ObjectCreationExpr().setType(serviceType))));
```

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/WorkItemNodeVisitor.java
```diff
@@ -85,8 +87,13 @@ public void visitNode(String factoryField, T node, BlockStmt body, VariableScope
             CompilationUnit generatedHandler = d.generateHandlerClassForService();
             metadata.getGeneratedHandlers().put(mangledName, generatedHandler);
             workName = mangledName;
+        } else if (workName.equals("Rest Task")) {
+            workName = RestTaskDescriptor.getClassName(metadata);
+            metadata
+                .getGeneratedHandlers()
+                .computeIfAbsent(workName,RestTaskDescriptor::generateHandlerClassForService);
```
cristianonicolai, line 94:
```suggestion
                .computeIfAbsent(workName, RestTaskDescriptor::generateHandlerClassForService);
```

cristianonicolai:
> > @tsurdilo Yep, that makes sense to me. But how about the SW specific part, are you ok to have this in the sw runtimes module that @fjtirado proposed here? Problem was that is introduces dependecy to json path that would not be needed for bpmn projects.
> 
> @cristianonicolai I do not think that these "specific" SW parts are something that goes beyond just having it run for a demo, as we havent had the chance to actually implement the "real" solution for data management of SW in Kogito. I feel a little rushed here as this implementation is based on whats there **now** and assumes its right but it may be irrelevant in the near future.
> 
> Also this is a workitem and to bring this back one more time it should be as such a reusable component. I think talking about
> a workitem and "sw specific parts" in the same sentence just shows a wrong approach tobegin with.
> I don't think having even a "serverless-workflow-runtime" or whatever package or module makes sense at all.
> 
> So to answer your question I would really like to see this as an addon workitem that is completely independent and does not include any "specific sw features" and then we will figure out how to incorporate its use when needed.

@tsurdilo agree that this solution is not final, but rather something for us to start showing some usecases and capabilitties, which will need evolving for sure. As for the SW part, my understanding was that its only the JSON path thing wasnt neded for processes, but I see that this can actually be reused for both, meaning, as you said, this should be just a regular work item. We can move the json path part to core as we initially thought, later on, but we sould be able to at least get something that we can start demonstraing that we can easily integrate with REST endpoints.

jbpm/jbpm-workitems/jbpm-rest-handler/pom.xml
```diff
@@ -0,0 +1,32 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+   <parent>
+    <groupId>org.kie.kogito</groupId>
+    <artifactId>jbpm-workitems</artifactId>
```
MarianMacik, line 7: As @cristianonicolai suggested, the best option would be `kogito-workitems` and the submodule either `kogito-workitems-rest` or `kogito-rest-workitem`.

fjtirado:
> > > @tsurdilo Yep, that makes sense to me. But how about the SW specific part, are you ok to have this in the sw runtimes module that @fjtirado proposed here? Problem was that is introduces dependecy to json path that would not be needed for bpmn projects.
> > 
> > 
> > @cristianonicolai I do not think that these "specific" SW parts are something that goes beyond just having it run for a demo, as we havent had the chance to actually implement the "real" solution for data management of SW in Kogito. I feel a little rushed here as this implementation is based on whats there **now** and assumes its right but it may be irrelevant in the near future.
> > Also this is a workitem and to bring this back one more time it should be as such a reusable component. I think talking about
> > a workitem and "sw specific parts" in the same sentence just shows a wrong approach tobegin with.
> > I don't think having even a "serverless-workflow-runtime" or whatever package or module makes sense at all.
> > So to answer your question I would really like to see this as an addon workitem that is completely independent and does not include any "specific sw features" and then we will figure out how to incorporate its use when needed.
> 
> @tsurdilo agree that this solution is not final, but rather something for us to start showing some usecases and capabilitties, which will need evolving for sure. As for the SW part, my understanding was that its only the JSON path thing wasnt neded for processes, but I see that this can actually be reused for both, meaning, as you said, this should be just a regular work item. We can move the json path part to core as we initially thought, later on, but we sould be able to at least get something that we can start demonstraing that we can easily integrate with REST endpoints.

JsonPath has been moved to the same project than RestTaskHandler

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/WorkItemNodeVisitor.java
```diff
@@ -85,6 +87,11 @@ public void visitNode(String factoryField, T node, BlockStmt body, VariableScope
             CompilationUnit generatedHandler = d.generateHandlerClassForService();
             metadata.getGeneratedHandlers().put(mangledName, generatedHandler);
             workName = mangledName;
+        } else if (workName.equals("Rest Task")) {
```
ricardozanini, line 90: I think it's better to use constants instead.

fjtirado:
jenkins retest this

cristianonicolai (approved):

PR #794, opened 2020-09-30
[KOGITO-3495] Error handling on RestworkItemhandler

fjtirado:
If rest invocation failed, workitem will be aborted.

### DIFF
--- a/kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
+++ b/kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
@@ -34,6 +34,8 @@ import io.vertx.ext.web.client.WebClient;
 import org.kie.api.runtime.process.WorkItem;
 import org.kie.api.runtime.process.WorkItemHandler;
 import org.kie.api.runtime.process.WorkItemManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class RestWorkItemHandler implements WorkItemHandler {
 
@@ -47,6 +49,8 @@ public class RestWorkItemHandler implements WorkItemHandler {
     public static final String PASSWORD = "password";
     public static final String HOST = "host";
     public static final String PORT = "port";
+
+    private static final Logger logger = LoggerFactory.getLogger(RestWorkItemHandler.class);
 
     // package scoped to allow unit test
     static class RestUnaryOperator implements UnaryOperator<Object> {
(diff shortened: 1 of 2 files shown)


fjtirado:
jenkins retest this

fjtirado:
jenkins retest this

fjtirado:
jenkins retest this

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
```diff
@@ -87,28 +88,24 @@ public void executeWorkItem(WorkItem workItem, WorkItemManager manager) {
...
         // execute request
-         Handler<AsyncResult<HttpResponse<Buffer>>> handler = event -> 
-                manager
-                .completeWorkItem(
-                                  workItem.getId(),
-                                  Collections
-                                      .singletonMap(RESULT, resultHandler.apply(inputModel, event.result().bodyAsJsonObject())));
+        Handler<AsyncResult<HttpResponse<Buffer>>> handler = event -> {
+            if (event.cause() != null) {
+                throw new WorkItemExecutionError("500", event.cause());
```
MarianMacik, line 99: Probably a better message would be welcome.

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
```diff
@@ -87,28 +88,24 @@ public void executeWorkItem(WorkItem workItem, WorkItemManager manager) {
...
         }
         // execute request
-         Handler<AsyncResult<HttpResponse<Buffer>>> handler = event -> 
-                manager
-                .completeWorkItem(
-                                  workItem.getId(),
-                                  Collections
-                                      .singletonMap(RESULT, resultHandler.apply(inputModel, event.result().bodyAsJsonObject())));
+        Handler<AsyncResult<HttpResponse<Buffer>>> handler = event -> {
+            if (event.cause() != null) {
```
MarianMacik, line 98: Probably more straightforward would be

```
if (event.failed()) {
    throw new WorkItemExecutionError("500", event.cause());
}
```

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
MarianMacik: As discussed, the code is expected.

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
fjtirado: thats better indeed

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
fjtirado: This is required by WorkItemExecution. Ill use the error returned by the rest service

fjtirado:
jenkins retest this

fjtirado:
jenkins retest this

fjtirado:
Including @evacchi since he were working with futures and this might interest him
Since RestWorkItemHandler is implemented asynchronously, throwing an exception when the rest invocation fails will only makes the vertx thread where the hander is invoked "angry", so I think the best we can do is to abort the work item asyncronously, in the same way we are completing it asyncrhonously when the call succeed. 
In any case, the process will continue and, in the example case (two consecutive rest invocations, where the second call depends on data retrieved by first one), since the first call failed to provide required data, the second call will fail. 
Please note the failure is not reported in the post call that initiate the process. User will have to query the process ide for resolving the status

cristianonicolai (approved):

MarianMacik (approved):

PR #802, opened 2020-10-02
KOGITO-3452: uniform error handling of build/encode/decode in CloudEventUtils

kostola:
- [JIRA task](https://issues.redhat.com/browse/KOGITO-3452)

### DIFF
--- a/addons/tracing/tracing-decision-common/src/main/java/org/kie/kogito/tracing/decision/DecisionTracingCollector.java
+++ b/addons/tracing/tracing-decision-common/src/main/java/org/kie/kogito/tracing/decision/DecisionTracingCollector.java
@@ -20,6 +20,7 @@ import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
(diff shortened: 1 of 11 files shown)


ruromero (approved):

danielezonca (approved):
LGTM

addons/tracing/tracing-decision-common/src/main/java/org/kie/kogito/tracing/decision/DecisionTracingCollector.java
```diff
@@ -82,16 +83,23 @@ public void addEvent(EvaluateEvent event) {
...
-            payloadConsumer.accept(payload);
-            LOG.debug("Generated aggregated event for evaluation {} (length {})", executionId, payload.length());
+            Optional<String> optPayload = aggregate(dmnModel, executionId, cacheMap.get(executionId));
+
+            if (optPayload.isPresent()) {
+                String payload = optPayload.get();
+                payloadConsumer.accept(payload);
+                LOG.debug("Generated aggregated event for evaluation {} (length {})", executionId, payload.length());
+            } else {
+                LOG.error("Failed aggregating data for evaluation {}", executionId);
```
jiripetrlik, line 93: Is it ok to only log an error without throwing any exception or inform user also some other way?

jiripetrlik (commented):
Hello @kostola I have one question about code and one concern. It seems that the test coverage is below our threshold of 80 %. Would it be possible to increase it? For example according to Sonar cloud we do not test any negative scenarios like wrong input etc. Would it be possible to add some test for it to verify the proper behaviour in these cases.

addons/tracing/tracing-decision-common/src/main/java/org/kie/kogito/tracing/decision/DecisionTracingCollector.java
kostola: This PR is an integration of [KOGITO-3297](https://github.com/kiegroup/kogito-runtimes/pull/750).
Yes I think this approach is correct because here in case of error we're simply not sending a trace event that the user should not be aware of. We discussed this approach with `Optional` [here](https://github.com/kiegroup/kogito-runtimes/pull/750#discussion_r490089208).

jiripetrlik (approved):
@kostola 
Thank you for your reply. It looks good to me.

danielezonca:
jenkins retest this please

danielezonca:
jenkins retest this please

PR #785, opened 2020-09-25
[DROOLS-5367] Removed kogito-specific PredictionRuleMapper

gitgabrio:
@danielezonca @mariofusco @jiripetrlik

See https://issues.redhat.com/browse/DROOLS-5367

This PR remove the code that has been ported inside drools.

It depends on https://github.com/kiegroup/drools/pull/3117

PR #818, opened 2020-10-10
[KOGITO-3505] refactor kogito according to descoping of mvel and protobuf in drools

mariofusco:


mariofusco:
Incorporated by https://github.com/kiegroup/kogito-runtimes/pull/785

PR #824, opened 2020-10-13
[KOGITO-2849] Support for Compensations

ruromero:
Fix https://issues.redhat.com/browse/KOGITO-2849

Cases covered:

- Intermediate compensation
- End Compensation
- Specifying the node to compensate
- Compensate all completed nodes

### DIFF
--- a/jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
+++ b/jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
@@ -53,6 +53,7 @@ import org.jbpm.process.instance.impl.Action;
 import org.jbpm.process.instance.impl.actions.CancelNodeInstanceAction;
 import org.jbpm.process.instance.impl.actions.ProcessInstanceCompensationAction;
 import org.jbpm.process.instance.impl.actions.SignalProcessInstanceAction;
+import org.jbpm.ruleflow.core.Metadata;
 import org.jbpm.ruleflow.core.RuleFlowProcess;
 import org.jbpm.ruleflow.core.validation.RuleFlowProcessValidator;
 import org.jbpm.workflow.core.Connection;
@@ -822,25 +823,27 @@ public class ProcessHandler extends BaseAbstractHandler implements Handler {
                                             }
                                         } else if( type.equals("Compensation") ) { 
                                             // 1. Find the parent sub-process to this event sub-process
-                                            NodeContainer parentSubProcess;  
+                                            NodeContainer parentSubProcess = null;
                                             NodeContainer subProcess = eventSubProcessNode.getParentContainer();
                                             Object isForCompensationObj = eventSubProcessNode.getMetaData("isForCompensation");
                                             if( isForCompensationObj == null ) { 
                                                 eventSubProcessNode.setMetaData("isForCompensation", true );
-                                                logger.warn( "Overriding empty or false value of \"isForCompensation\" attribute on Event Sub-Process [" 
-                                                        + eventSubProcessNode.getMetaData("UniqueId") + "] and setting it to true.");
-                                            } 
-                                            if( subProcess instanceof RuleFlowProcess ) { 
-                                                // If jBPM deletes the process (instance) as soon as the process completes.. 
+                                                logger.warn( "Overriding empty value of \"isForCompensation\" attribute on Event Sub-Process [{}] and setting it to true.",
+                                                        eventSubProcessNode.getMetaData("UniqueId"));
+                                            }
+                                            String compensationHandlerId = "";
+                                            if( subProcess instanceof RuleFlowProcess) {
+                                                // If jBPM deletes the process (instance) as soon as the process completes..
                                                 // ..how do you expect to signal compensation on the completed process (instance)?!?
                                                 throw new IllegalArgumentException("Compensation Event Sub-Processes at the process level are not supported.");
                                             }
-                                            parentSubProcess = ((Node) subProcess).getParentContainer();
-
+                                            if(subProcess instanceof  Node) {
+                                                parentSubProcess = ((Node) subProcess).getParentContainer();
+                                                compensationHandlerId = (String) ((CompositeNode) subProcess).getMetaData(Metadata.UNIQUE_ID);
+                                            }
                                             // 2. The event filter (never fires, purely for dumping purposes) has already been added
 
                                             // 3. Add compensation scope
-                                            String compensationHandlerId = (String) ((CompositeNode) subProcess).getMetaData("UniqueId");
                                             addCompensationScope(process, eventSubProcessNode, parentSubProcess, compensationHandlerId);
                                         }
                                     }
@@ -973,14 +976,14 @@ public class ProcessHandler extends BaseAbstractHandler implements Handler {
         if( throwEventNode.getMetaData("compensation-activityRef") != null ) { 
             String activityRef = (String) throwEventNode.getMetaData().remove("compensation-activityRef");
 
-            NodeContainer nodeParent = (NodeContainer) throwEventNode.getParentContainer();
+            NodeContainer nodeParent = throwEventNode.getParentContainer();
             if( nodeParent instanceof EventSubProcessNode ) { 
                 boolean compensationEventSubProcess = false;
                 List<Trigger> startTriggers = ((EventSubProcessNode) nodeParent).findStartNode().getTriggers();
                 CESP_CHECK: for( Trigger trigger : startTriggers ) { 
                     if( trigger instanceof EventTrigger ) { 
                         for( EventFilter filter : ((EventTrigger) trigger).getEventFilters() ) { 
-                            if( ((EventTypeFilter) filter).getType().equals("Compensation") ) { 
+                            if( ((EventTypeFilter) filter).getType().equals(Metadata.EVENT_TYPE_COMPENSATION) ) {
                                 compensationEventSubProcess = true;
                                 break CESP_CHECK;
                             }
@@ -989,7 +992,7 @@ public class ProcessHandler extends BaseAbstractHandler implements Handler {
                 }
                 if( compensationEventSubProcess ) { 
                     // BPMN2 spec, p. 252, p. 248: intermediate and end compensation event visibility scope
-                    nodeParent = (NodeContainer) ((NodeImpl) nodeParent).getParentContainer();
+                    nodeParent = ((NodeImpl) nodeParent).getParentContainer();
                 }
             }
             String parentId;
@@ -1000,7 +1003,7 @@ public class ProcessHandler extends BaseAbstractHandler implements Handler {
             }
 
             String compensationEvent; 
-            if( activityRef.length() == 0 ) { 
+            if(activityRef.isEmpty()) {
                 // general/implicit compensation
                 compensationEvent = CompensationScope.IMPLICIT_COMPENSATION_PREFIX + parentId;
             } else { 
@@ -1014,7 +1017,7 @@ public class ProcessHandler extends BaseAbstractHandler implements Handler {
             if( throwEventNode instanceof ActionNode ) { 
                 ((ActionNode) throwEventNode).setAction(compensationAction);
             } else if( throwEventNode instanceof EndNode ) { 
-                List<DroolsAction> actions = new ArrayList<DroolsAction>();
+                List<DroolsAction> actions = new ArrayList<>();
                 actions.add(compensationAction);
                 ((EndNode) throwEventNode).setActions(EndNode.EVENT_NODE_ENTER, actions);
             } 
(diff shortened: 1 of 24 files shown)


ruromero:
[compensations](https://trello.com/c/4VBIbcca/1-compensations)

ruromero:
[compensations](https://trello.com/c/4VBIbcca/1-compensations)

kogito-codegen/src/test/resources/logback-test.xml
```diff
@@ -17,7 +17,7 @@
 
   <!--  <logger name="org.kie.kogito.codegen" level="debug"/>-->
 
-  <root level="info">
+  <root level="debug">
```
tiagodolphine, line 20: should it be changed to debug?

kogito-codegen/src/test/java/org/kie/kogito/process/impl/CompensationTest.java
```diff
@@ -0,0 +1,94 @@
...
+import org.kie.kogito.codegen.AbstractCodegenTest;
+import org.kie.kogito.process.Process;
+import org.kie.kogito.process.ProcessInstance;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.kie.kogito.process.impl.ProcessTestUtils.assertState;
+
+class CompensationTest extends AbstractCodegenTest {
```
tiagodolphine, line 33: why not adding this test under https://github.com/kiegroup/kogito-runtimes/tree/3e7f93289ccce36edd457dd535b87fe263b59bdb/kogito-codegen/src/test/java/org/kie/kogito/codegen/tests ?

tiagodolphine (commented):
LGTM, it seems the tests in code-gen are covering most of the use cases, maybe it is worth adding some integration-tests.

kogito-codegen/src/test/resources/logback-test.xml
cristianonicolai: +1 to revert

cristianonicolai (commented):
@ruromero overall looks good, as @tiagodolphine mentioned, would be nice to have an integration test added for both quarkus and spring.

kogito-codegen/src/test/resources/logback-test.xml
ruromero: oh, yes, thanks for noticing

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ProcessVisitor.java
```diff
@@ -255,14 +261,25 @@ private void visitInterfaces(org.kie.api.definition.process.Node[] nodes, BlockS
     }
 
     private void visitConnection(Connection connection, BlockStmt body) {
-        // if the connection is a hidden one (compensations), don't dump
-        Object hidden = ((ConnectionImpl) connection).getMetaData(HIDDEN);
-        if (hidden != null && ((Boolean) hidden)) {
-            return;
+        String method = METHOD_CONNECTION;
+        Object association = ((ConnectionImpl) connection).getMetaData(ASSOCIATION);
```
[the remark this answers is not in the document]
ruromero: Actually it's the same, an association is a hidden connection so both statements are always true but to me it makes more sense to use association than hidden connection.

ruromero:
@cristianonicolai @tiagodolphine what are the integration tests going to provide in addition? There aren't any new endpoints added or data manipulation, it's just the engine and code generation which is already being tested. I don't mind adding them but I don't see any benefit

jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
```diff
@@ -822,25 +823,31 @@ private void postProcessNodes(RuleFlowProcess process, NodeContainer container)
...
-                                                throw new IllegalArgumentException("Compensation Event Sub-Processes at the process level are not supported.");
                                             }
-                                            parentSubProcess = ((Node) subProcess).getParentContainer();
-
+                                            String compensationHandlerId = "";
+                                            if( subProcess instanceof RuleFlowProcess) {
+                                                if(!process.isDynamic()) {
+                                                    // If jBPM deletes the process (instance) as soon as the process completes..
+                                                    // ..how do you expect to signal compensation on the completed process (instance)?!?
+                                                    throw new IllegalArgumentException("Compensation Event Sub-Processes at the process level are not supported.");
```
MarianMacik, line 839: AFAIK according to specs only completed activities (nodes) or subprocesses can be compensated. When we have a dynamic process (case), then it is still not finished, so why do we allow compensation at the process level? Just because we want to have this flexibility and because it is technically possible? Then I think also compensate event subprocess in "normal" process at process-level is possible - we can just run it when the compensate event is thrown, the compensate event subprocess is started and it can for example throw various compensate events for more nodes on the process level. So I think this really comes down to if we want to run compensate event subprocess even though the enclosing process is not yet finished... What do you think?

jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
```diff
@@ -973,14 +980,14 @@ protected void handleIntermediateOrEndThrowCompensationEvent(ExtendedNodeImpl th
         if( throwEventNode.getMetaData("compensation-activityRef") != null ) { 
             String activityRef = (String) throwEventNode.getMetaData().remove("compensation-activityRef");
```
MarianMacik, line 981: I know it's unrelated, just noticed this here. This is removed just so we don't process the same node twice?

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/StartNodeVisitor.java
```diff
@@ -62,19 +62,26 @@ public void visitNode(String factoryField, StartNode node, BlockStmt body, Varia
...
         } else if (node.getTriggers() != null && !node.getTriggers().isEmpty()) {
-            Map<String, Object> nodeMetaData = node.getMetaData();
-            metadata.addTrigger(new TriggerMetaData((String) nodeMetaData.get(TRIGGER_REF),
-                    (String) nodeMetaData.get(TRIGGER_TYPE),
-                    (String) nodeMetaData.get(MESSAGE_TYPE),
-                    (String) nodeMetaData.get(TRIGGER_MAPPING),
-                    String.valueOf(node.getId())).validate());
-
-            handleSignal(node, nodeMetaData, body, variableScope, metadata);
+            TriggerMetaData triggerMetaData = buildTriggerMetadata(node);
```
MarianMacik, line 65: Can it happen that triggers are not empty but `nodeMetaData.get(TRIGGER_TYPE)` returns null?

MarianMacik (commented):
Looks good to me. I just added there questions about areas which are not that well-known to me.

jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
```diff
@@ -822,25 +823,31 @@ private void postProcessNodes(RuleFlowProcess process, NodeContainer container)
...
                                         } else if( type.equals("Compensation") ) { 
                                             // 1. Find the parent sub-process to this event sub-process
-                                            NodeContainer parentSubProcess;  
+                                            NodeContainer parentSubProcess = null;
                                             NodeContainer subProcess = eventSubProcessNode.getParentContainer();
                                             Object isForCompensationObj = eventSubProcessNode.getMetaData("isForCompensation");
                                             if( isForCompensationObj == null ) { 
                                                 eventSubProcessNode.setMetaData("isForCompensation", true );
                                                 logger.warn( "Overriding empty or false value of \"isForCompensation\" attribute on Event Sub-Process [" 
                                                         + eventSubProcessNode.getMetaData("UniqueId") + "] and setting it to true.");
```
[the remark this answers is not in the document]
ruromero: Yes it defaults to true when the type is "Compensation" and "isForCompensation" is missing.

jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
ruromero: If I understood the code correctly, this is added to the node's metadata when processing the compensateEventDefinition during the xml parsing in order to be able to set up the compensation event later on.

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/StartNodeVisitor.java
ruromero: If so, the buildTriggerMetadata function will return null and the trigger will not be added.

jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
ruromero: A dynamic process is not _finished_ when one end node is reached. A dynamic process is only completed when the terminate end node is done.
Regarding whether subprocesses could be compensated within a non-dynamic process. I prefer @cristianonicolai to answer that.

MarianMacik: My point was that if we even can compensate unfinished process instance as a whole (no matter if it is dynamic or nor). As based on the spec I _think_ only completed actions can be compensated and this part compensates a dynamic process which is not completed yet, right?

jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
MarianMacik: OK, so then it is not needed anymore and removed basically?

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/StartNodeVisitor.java
MarianMacik: The question was more about if the check on the line 77 is even needed as we know triggers on the node are not empty :)

jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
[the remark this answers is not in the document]
MarianMacik: So in that case shouldn't we change the message and remove `or false` from it?

jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
ruromero: It is needed for code generation as an intermediate step

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/StartNodeVisitor.java
ruromero: Maybe not. I can remove the check.

jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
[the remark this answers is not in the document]
ruromero: Yes, good point

jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
tiagodolphine: looking at the code, it seems this specific use case, is related to Compensation Event Sub-Process (where we can execute a sub-process to compensate a parent sub-process), not other kinds of sub-processes in general. And at least looking at the spec it seems the Compensation Event Sub-Proces should be inside a parent sub-process not on the `root` let's say. So, I'm not sure if we should check the process `isDynamic`, I mean I don't think if it makes sense for this specific case. WDYT?

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/StartNodeVisitor.java
```diff
@@ -62,19 +62,23 @@ public void visitNode(String factoryField, StartNode node, BlockStmt body, Varia
...
+            if (triggerMetaData != null) {
```
MarianMacik, line 66: Hmm, now the triggerMetaData is always not null, so we can get rid of this check :)

MarianMacik (commented):
Now we can delete a check as a consequence.

jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
MarianMacik: Yep, that was my point I tried to make :+1:

ruromero:
Changed and rebased against master

MarianMacik (approved):
Thanks, @ruromero!

tiagodolphine (approved):

PR #858, opened 2020-10-28
[KOGITO-2929] remove ProjectRuntime for non-legacy use

mariofusco:


api/kogito-internal/src/main/java/org/kie/internal/conf/AlphaRangeIndexThresholdOption.java
```diff
@@ -0,0 +1,95 @@
...
+ */
+
+package org.kie.internal.conf;
+
+import org.kie.api.conf.SingleValueKieBaseOption;
+
+/**
+ * A class for the alpha node range index threshold configuration.
+ */
+public class AlphaRangeIndexThresholdOption implements SingleValueKieBaseOption {
```
evacchi, line 24: is this a leftover?

evacchi (approved):

api/kogito-internal/src/main/java/org/kie/internal/conf/AlphaRangeIndexThresholdOption.java
mariofusco: No, this is a class that has been added by @tkobayas in kie-internal and as usual I had to forward port it here.

PR #860, opened 2020-10-29
KOGITO-3738 Use JDK8-compatible readBytes impl in DecisionModelResourcesProvider

evacchi:
https://issues.redhat.com/browse/KOGITO-3738

My IDE freaks out every time the code is generated in `integration-test` because the root project is set to 8 compatibility, and all children inherits from it. For now I am just inserting a stock "readAllBytes()" impl,

1) but we can move most of this stock logic to an abstract class and inherit from there
2) alternative: we can set the pom.xml of the submodule to 11 -- don't know if this works especially with upstream Quarkus, so I am doing (1)

-----

### DIFF
--- a/kogito-codegen/src/main/resources/class-templates/DecisionModelResourcesProviderTemplate.java
+++ b/kogito-codegen/src/main/resources/class-templates/DecisionModelResourcesProviderTemplate.java
@@ -9,7 +9,7 @@ public class DecisionModelResourcesProvider implements org.kie.internal.decision
         }
 
         try {
-            byte[] bytes = stream.readAllBytes();
+            byte[] bytes = readAllBytes(stream);
             java.io.ByteArrayInputStream byteArrayInputStream = new java.io.ByteArrayInputStream(bytes);
             return new java.io.InputStreamReader(byteArrayInputStream);
         } catch (java.io.IOException e) {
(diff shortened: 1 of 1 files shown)


kogito-codegen/src/main/resources/class-templates/DecisionModelResourcesProviderTemplate.java
```diff
@@ -9,7 +9,7 @@
         }
 
         try {
-            byte[] bytes = stream.readAllBytes();
+            byte[] bytes = readAllBytes(stream);
```
tarilabs, line 12:
```suggestion
            byte[] bytes = org.drools.core.util.IoUtils.readBytesFromInputStream(stream);
```

tarilabs (changes requested):
The abstract solution is good, I agree with overall.
There was even a recent Kogito ML comment about it, this would solve that too.

The technical implementation I do not concur. Please see proposal below.

kogito-codegen/src/main/resources/class-templates/DecisionModelResourcesProviderTemplate.java
evacchi: https://media.giphy.com/media/3o8doT9BL7dgtolp7O/giphy.gif

evacchi: yes I saw that thread on ML, so that makes sense anyway

tarilabs (approved):
after ec04738 👍

ricardozanini:
Oh please merge this! :)

Thanks @evacchi

radtriste (approved):

PR #862, opened 2020-10-30
[KOGITO-3692] Add lra-addon

ruromero:

Fix https://issues.redhat.com/browse/KOGITO-3692

Depends on Narayana LRA Client 5.10.7.Final which is still under development.


kogito-codegen/src/main/resources/class-templates/RestResourceTemplate.java
```diff
@@ -56,14 +56,14 @@
...
-
             if (startFromNode != null) {
                 pi.startFrom(startFromNode);
             } else {
                 pi.start();
             }
-            UriBuilder uriBuilder = uriInfo.getAbsolutePathBuilder().path(pi.id());
-            return Response.created(uriBuilder.build())
+            URI uri = uriInfo.getAbsolutePathBuilder().path(pi.id()).build();
+            addMetadata(pi, httpHeaders, uri);
```
tiagodolphine, line 65: maybe decoupling the logic from the endpoint,  that is generic for any process... something like using Interceptors/Filters on the LRA specific template... anyway just thinking...

pom.xml
```diff
@@ -130,6 +130,7 @@
     <module>integration-tests</module>
     <module>jenkins-tests</module>
     <module>kogito-workitems</module>
+      <module>lra-springboot-integration-tests</module>
```
tiagodolphine, line 133: is the module in the right pom?

addons/lra/lra-common/src/main/java/org/kie/kogito/lra/listeners/LRAEventListener.java
```diff
@@ -0,0 +1,107 @@
...
+    public void afterProcessCompleted(ProcessCompletedEvent event) {
+        super.beforeProcessCompleted(event);
+        ProcessInstance instance = event.getProcessInstance();
+        if (!(instance instanceof RuleFlowProcessInstance)) {
+            LOGGER.warn("Unable to close LRA. Process instance is not a RuleFlowProcessInstance");
+            return;
+        }
+        RuleFlowProcessInstance wfInstance = (RuleFlowProcessInstance) instance;
+        LRAContext context = (LRAContext) wfInstance.getMetaData().get(LRA_CONTEXT);
+        if (context == null || context.getRecoverUri() != null) {
```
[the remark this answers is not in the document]
ruromero: Yes, the url is the LRA URL and the recoveryURI is the one that is returned by the coordinator each time a participant joins it.

addons/lra/lra-common/src/main/java/org/kie/kogito/lra/listeners/LRAEventListener.java
```diff
@@ -0,0 +1,107 @@
...
+            return;
+        }
+        RuleFlowProcessInstance wfInstance = (RuleFlowProcessInstance) instance;
+        LRAContext context = (LRAContext) wfInstance.getMetaData().get(LRA_CONTEXT);
+        if (context == null) {
+            context = new LRAContext();
+        }
+        //TODO: Implement all cases for org.eclipse.microprofile.lra.annotation.ws.rs.LRA#Type
+        if (context.getUri() == null) {
+            context.setUri(lraClient.startLRA(null, event.getProcessInstance().getProcessId(), 0L, ChronoUnit.SECONDS));
```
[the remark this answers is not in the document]
ruromero: Currently errors are propagated and will return a 500 error. I think it is a critical part of the process when participating in an LRA. What do you have in mind?

addons/lra/lra-quarkus-addon/src/main/java/org/kie/kogito/lra/client/LRAClientProducer.java
```diff
@@ -0,0 +1,48 @@
...
+    @ConfigProperty(name = NarayanaLRAClient.LRA_COORDINATOR_HOST_KEY, defaultValue = "localhost")
+    String host;
+
+    @ConfigProperty(name = NarayanaLRAClient.LRA_COORDINATOR_PORT_KEY, defaultValue = "8080")
+    Integer port;
+
+    @Produces
+    @Named(KogitoLRA.BEAN_NAME)
+    @ApplicationScoped
+    NarayanaLRAClient getClient() throws URISyntaxException {
```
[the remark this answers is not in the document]
ruromero: What's the difference?

pom.xml
ruromero: thanks intelliJ... haha

kogito-codegen/src/main/resources/class-templates/RestResourceTemplate.java
ruromero: I thought about it but in this case I need something in between. In the filter I don't have the process instance and in the interceptor I don't have the http headers. Am I wrong?

PR #880, opened 2020-11-10
KOGITO-3819 Ensure same Process instance is returned in Processes.fin…

cristianonicolai:
…dProcessById

https://issues.redhat.com/browse/KOGITO-3819

### DIFF
--- a/kogito-codegen/src/main/resources/class-templates/ProcessContainerTemplate.java
+++ b/kogito-codegen/src/main/resources/class-templates/ProcessContainerTemplate.java
@@ -3,14 +3,16 @@ package $Package$;
 public class Processes implements org.kie.kogito.process.Processes {
 
     private final Application application;
+    private java.util.Map<String, org.kie.kogito.process.Process<? extends org.kie.kogito.Model>> mappedProcesses = new java.util.concurrent.ConcurrentHashMap<>();
 
     public Processes(Application application) {
         this.application = application;
     }
 
     public org.kie.kogito.process.Process<? extends org.kie.kogito.Model> processById(String processId) {
-        if ("$ProcessId".equals(processId))
-            return new $ProcessClassName$(application).configure();
+        if ("$ProcessId".equals(processId)) {
+            return mappedProcesses.computeIfAbsent("$ProcessId", k -> new $ProcessClassName$(application).configure());
+        }
         return null;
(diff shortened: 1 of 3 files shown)


cristianonicolai:
jenkins please retest this

cristianonicolai:
@MarianMacik @tiagodolphine any feedback?

kogito-codegen/src/main/resources/class-templates/ProcessContainerTemplate.java
```diff
@@ -3,14 +3,16 @@
...
 
     public Processes(Application application) {
         this.application = application;
     }
 
     public org.kie.kogito.process.Process<? extends org.kie.kogito.Model> processById(String processId) {
-        if ("$ProcessId".equals(processId))
-            return new $ProcessClassName$(application).configure();
+        if ("$ProcessId".equals(processId)) {
+            return mappedProcesses.computeIfAbsent("$ProcessId", k -> new $ProcessClassName$(application).configure());
```
tiagodolphine, line 14: just a question, keeping all  Processes in-memory and never removing them, could bring issues? I mean because it has references to all processInstances for example, right?

cristianonicolai: I think is fine for Process to be available in memory, this is just the amount of process inside the project, so should be minimal number. Regarding process instances, if you use inmemory they will remain in memory anyway ( inside a map ) while they are running. To avoid that, or if that is somehow a limitation, people should switch to one of the db options.
The issue here is that we were creating more Processes ( object instances of it ) than needed, missing the correct references to the in memory map of ProcessInstances.

MarianMacik: +1 for keeping processes (process definitions) in memory. It is roughly the same as with KIE Server where all deployed processes were held in memory inside deployments.

MarianMacik (approved):

kogito-codegen/src/main/resources/class-templates/ProcessContainerTemplate.java
tiagodolphine: +1

tiagodolphine (approved):

PR #908, opened 2020-11-23
[KOGITO-3840] Forward transport headers

ruromero:
Fix https://issues.redhat.com/browse/KOGITO-3840

Adds the concept of `TransportContext` to be handled individually by each transport (HTTP, Kafka, etc.). The transport context will include key value attributes that will be used by the TaskHandlers that also interact with transports.
The purpose is to provide the possibility to propagate such properties to external services. This is very useful for distributed tracing and transactions such as LRA.

An example is an HTTP Request that contains an LRA_HTTP_CONTEXT_HEADER HttpHeader that has to be propagated to the external services invoked by the RestTaskHandler.

The way to decide which headers to propagate will be through defining TransportFilter classes that will specify which parameters are allowed to be filtered.

### DIFF
--- a/jbpm/process-serialization-protobuf/src/main/resources/org/kie/kogito/serialization/process/protobuf/kogito_types.proto
+++ b/jbpm/process-serialization-protobuf/src/main/resources/org/kie/kogito/serialization/process/protobuf/kogito_types.proto
@@ -28,6 +28,7 @@ message WorkflowContext {
     repeated NodeInstance node_instance = 2;
     repeated NodeInstanceGroup exclusive_group = 3;
     repeated IterationLevel iterationLevels = 4;
+    repeated Variable transport_context = 5;
 }
 
 message SwimlaneContext {
--- a/jbpm/process-serialization-protobuf/src/main/resources/org/kie/kogito/serialization/process/protobuf/kogito_types.proto
+++ b/jbpm/process-serialization-protobuf/src/main/resources/org/kie/kogito/serialization/process/protobuf/kogito_types.proto
@@ -28,7 +28,6 @@ message WorkflowContext {
     repeated NodeInstance node_instance = 2;
     repeated NodeInstanceGroup exclusive_group = 3;
     repeated IterationLevel iterationLevels = 4;
-    repeated Variable transport_context = 5;
 }
 
 message SwimlaneContext {
--- a/kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
+++ b/kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
@@ -28,6 +28,7 @@ import java.util.stream.Collectors;
 import org.kie.kogito.internal.process.runtime.KogitoWorkItem;
 import org.kie.kogito.internal.process.runtime.KogitoWorkItemHandler;
 import org.kie.kogito.internal.process.runtime.KogitoWorkItemManager;
+import org.kie.kogito.transport.TransportConfig;
 
 import io.vertx.core.http.HttpMethod;
 import io.vertx.mutiny.core.buffer.Buffer;
@@ -93,6 +94,8 @@ public class RestWorkItemHandler implements KogitoWorkItemHandler {
         if (user != null && !user.trim().isEmpty() && password != null && !password.trim().isEmpty()) {
             request.basicAuthentication(user, password);
         }
+        addHeadersFromContext(workItem, request);
+
         HttpResponse<Buffer> response;
         if (method == HttpMethod.POST || method == HttpMethod.PUT) {
             // if parameters is empty at this stage, assume post content is the whole input model
@@ -106,6 +109,17 @@ public class RestWorkItemHandler implements KogitoWorkItemHandler {
         }
         manager.completeWorkItem(workItem.getStringId(), Collections.singletonMap(RESULT, resultHandler.apply(inputModel,
                 response.bodyAsJsonObject())));
+    }
+
+    @SuppressWarnings("unchecked")
+    private void addHeadersFromContext(KogitoWorkItem workItem, HttpRequest<Buffer> request) {
+        if (workItem.getProcessInstance() == null) {
+            return;
+        }
+        Map<String, String> transportContext = (Map<String, String>) workItem.getProcessInstance().getMetaData().get(TransportConfig.TRANSPORT_CONTEXT);
+        if (transportContext != null) {
+            transportContext.forEach(request::putHeader);
+        }
     }
 
     @Override
--- a/api/kogito-api/pom.xml
+++ b/api/kogito-api/pom.xml
@@ -89,5 +89,10 @@
       <artifactId>jackson-databind</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 </project>
--- a/api/kogito-api/src/main/java/org/kie/kogito/conf/StaticConfigBean.java
+++ b/api/kogito-api/src/main/java/org/kie/kogito/conf/StaticConfigBean.java
@@ -18,12 +18,14 @@ package org.kie.kogito.conf;
 import java.util.Optional;
 
 import org.kie.kogito.KogitoGAV;
+import org.kie.kogito.transport.TransportConfig;
 
 public class StaticConfigBean implements ConfigBean {
 
     private String serviceUrl;
     private boolean useCloudEvents = true;
     private KogitoGAV gav;
+    private TransportConfig transportConfig = new TransportConfig();
 
     public StaticConfigBean() {
     }
@@ -46,6 +48,10 @@ public class StaticConfigBean implements ConfigBean {
         this.gav = gav;
     }
 
+    protected void setTransportConfig(TransportConfig transportConfig) {
+        this.transportConfig = transportConfig;
+    }
+
     @Override
     public boolean useCloudEvents() {
         return useCloudEvents;
@@ -59,5 +65,10 @@ public class StaticConfigBean implements ConfigBean {
     @Override
     public Optional<KogitoGAV> getGav() {
         return Optional.ofNullable(gav);
+    }
+
+    @Override
+    public TransportConfig transportConfig() {
+        return transportConfig;
     }
 }
--- a/api/kogito-api/src/main/java/org/kie/kogito/internal/process/runtime/KogitoProcessInstance.java
+++ b/api/kogito-api/src/main/java/org/kie/kogito/internal/process/runtime/KogitoProcessInstance.java
@@ -86,4 +86,17 @@ public interface KogitoProcessInstance extends ProcessInstance, KogitoEventListe
      * @return the process instance description
      */
     String getDescription();
+
+    /**
+     * @return the ProcessInstance metadata
+     */
+    Map<String, Object> getMetaData();
+
+    /**
+     * Sets a ProcessInstance metadata entry
+     * 
+     * @param name Name of the Metadata key
+     * @param data Data identified by the provided Metadata Key
+     */
+    void setMetaData(String name, Object data);
 }
--- a/api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
+++ b/api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
@@ -99,6 +99,10 @@ public interface ProcessInstance<T> {
      */
     T updateVariables(T updates);
 
+    Object getContextAttr(String name);
+
+    void setContextAttr(String name, Object value);
+
     /**
      * Returns current status of this process instance
      *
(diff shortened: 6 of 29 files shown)


ruromero:
@cristianonicolai @MarianMacik let me know if you agree with the solution proposed. I have added handling for HTTP requests and the RestTaskHandler but not yet to the CloudEvents. They might need additional changes in the interface.
Don't mind yet the commit without the JIRA, I haven't yet investigated it.

Thanks

kogito-codegen/src/main/resources/class-templates/spring/SpringRestResourceTemplate.java
```diff
@@ -144,5 +146,13 @@
                       .orElseGet(() -> ResponseEntity.notFound().build());
     }
 
+    private void addHttpHeadersToContext(ProcessInstance<$Type$> instance, HttpHeaders httpHeaders) {
+        Map<String, String> transportContext = new HashMap<>();
+        httpHeaders.entrySet()
+                .stream()
+                .filter(e -> processConfig.transportConfig().accepts(e.getKey()))
+                .forEach(e -> transportContext.put(e.getKey(), String.join(",", e.getValue())));
```
tiagodolphine, line 154: maybe extracting the logic to something like, encode/decode headers, to avoid the logic to be split/duplicated in different places.
And maybe the whole method addHttpHeadersToContext could be moved to a different class that could be used by different endpoints/templates.

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
```diff
@@ -96,6 +98,12 @@ public void executeWorkItem(WorkItem workItem, WorkItemManager manager) {
         if (user != null && !user.trim().isEmpty() && password != null && !password.trim().isEmpty()) {
             request.basicAuthentication(user, password);
         }
+        if(workItem.getProcessInstance() instanceof org.jbpm.process.instance.ProcessInstance) {
+            Map<String, String> transportContext = (Map<String, String>) ((ProcessInstance) workItem.getProcessInstance()).getMetaData().get(TransportConfig.TRANSPORT_CONTEXT);
```
tiagodolphine, line 102: why not using something like the `getContextAttr` method instead of getting the value from the metaData directly... or creating a new method that returns all attributes..

tiagodolphine (commented):
Nice @ruromero LGTM, I just left some comments.

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
ruromero: Because this `getContextAttr` I added it to the kogito's `ProcessInstance` as an abstraction to the metadata. In this case it is another type of `ProcessInstance` whose implementation already has the `getMetadata` and creating another method for the same purpose will be misleading. WDYT?

kogito-codegen/src/main/resources/class-templates/spring/SpringRestResourceTemplate.java
ruromero: I think I have reduced duplication with the current change.

drools/drools-core/src/main/java/org/drools/core/event/KogitoProcessEventSupport.java
```diff
@@ -48,11 +48,11 @@ public KogitoProcessEventSupport() {
 
     public void fireBeforeProcessStarted(final ProcessInstance instance, KieRuntime kruntime ) {
         final ProcessStartedEvent event = new ProcessStartedEventImpl(instance, kruntime);
-        unitOfWorkManager.currentUnitOfWork().intercept(WorkUnit.create(event, e -> {
+//        unitOfWorkManager.currentUnitOfWork().intercept(WorkUnit.create(event, e -> {
```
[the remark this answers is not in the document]
ruromero: yes, of course, that is not related to the PR

ruromero:
jenkins retest this

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
tiagodolphine: yeah, but in this case, the getContextAttr /set could be removed, right?

ruromero:
@danielezonca the purpose of this is to allow adding certain transport paremeters/headers/attributes (depending on the transport) to the process instance context so that they can be forwarded to any external service involved in the process (e.g. a service orchestration). I added it to the `ProcessConfig` because it's behaviour is similar to the processEventListeners but I don't mind moving it to a different place if you consider it more appropriate.

The startNode is directly used when receiving the request so I don't think it should be part of this context unless we decide to make the `start` method aware of this context and use such header.

It doesn't apply to response headers either (unless we want to).

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
ruromero: No because this method has been added to `org.kie.kogito.process.ProcessInstance` and the get/setMetadata exists in `org.kie.api.runtime.process.ProcessInstance`
I prefer the get/setContextAttr to avoid confusions with the Process's MetaData but if you agree it's better to be consistent with names I'll rename the methods.

ruromero:
@danielezonca I have moved it to the `ConfigBean` is that more aligned with what you stated in your comment?

ruromero:
@cristianonicolai @tiagodolphine @MarianMacik 
Please review when you have time, thanks

danielezonca:
@ruromero 
Thanks for your change, yes I think that `ConfigBean` is a better fit. Then we can discuss with another ticket/PR about extend or not this approach to handle all the headers :+1:

ruromero:
Ping @cristianonicolai @tiagodolphine @MarianMacik please review when possible.
Thanks

kogito-codegen/src/main/resources/class-templates/RestResourceTemplate.java
```diff
@@ -109,24 +113,29 @@
...
         return process.instances()
-                      .findById(id, ProcessInstanceReadMode.READ_ONLY)
-                      .map(pi -> pi.workItems(Policies.of(user, groups)))
-                      .orElseThrow(() -> new NotFoundException());
+                .findById(id, ProcessInstanceReadMode.READ_ONLY)
+                .map(pi -> pi.workItems(Policies.of(user, groups)))
+                .orElseThrow(() -> new NotFoundException());
+    }
+
+    private void addTransportHeaders(ProcessInstance<$Type$> instance, HttpHeaders httpHeaders) {
```
tiagodolphine, line 136: maybe not the scope of this PR, but we could have a static class (not a template) where the templated classes could extend to inherit common logic/methods. This would simplify things and avoid duplicating, methods in templates.

tiagodolphine (commented):
LGTM, just left some minor comments.

kogito-codegen/src/main/java/org/kie/kogito/codegen/process/ProcessInstanceGenerator.java
```diff
@@ -59,11 +57,10 @@ public static String qualifiedName(String packageName, String typeName) {
 
     public ProcessInstanceGenerator(String packageName, String typeName, ModelMetaData model) {
         this.packageName = packageName;
-        this.typeName = typeName;
         this.model = model;
         this.canonicalName = packageName + "." + typeName;
         this.targetTypeName = typeName + "ProcessInstance";
-        this.targetCanonicalName = packageName + "." + targetTypeName;
+        String targetCanonicalName = packageName + "." + targetTypeName;
```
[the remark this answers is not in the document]
ruromero: it is not used outside of the constructor

kogito-codegen/src/main/resources/class-templates/RestResourceTemplate.java
ruromero: As discussed, I don't see much to refactor as each transport and even framework handles headers in a different way. If at some point a parent class is defined for Rest templates we can try to move part of the code.

cristianonicolai (approved):

cristianonicolai:
@elguardian this changes the marshaller, can you just double-check its ok with your changes

tiagodolphine (approved):

tiagodolphine:
@ruromero there is a conflict to be solved.

ruromero:
> @ruromero there is a conflict to be solved.

already??? :sob: 
Thanks for the review

jbpm/process-serialization-protobuf/src/main/java/org/jbpm/marshalling/impl/AbstractProcessInstanceMarshaller.java
```diff
@@ -357,6 +344,39 @@ public int compare(NodeInstance o1, NodeInstance o2) {
...
+            stream.writeInt(keys.size());
+            for (String key : keys) {
+                stream.writeUTF(key);
+                stream.writeObject(variables.get(key));
+            }
+        }
+    }
+
+    private void marshalTransportContext(ObjectOutputStream stream, CompositeContextNodeInstance compositeNodeInstance) throws IOException {
+        if (compositeNodeInstance.getMetaData() == null && !compositeNodeInstance.getMetaData().containsKey(TransportConfig.TRANSPORT_CONTEXT)) {
```
MarianMacik, line 364: Isn't this NPE?
compositeNodeInstance.getMetaData() == null
AND
call to compositeNodeInstance.getMetaData().containsKey()

MarianMacik (commented):
Just one remark.

jbpm/process-serialization-protobuf/src/main/java/org/jbpm/marshalling/impl/AbstractProcessInstanceMarshaller.java
ruromero: thanks @MarianMacik !

MarianMacik: @ruromero Shouldn't be there rather OR? So if it is either null or doesn't contain key, we will write just 0.

ruromero: You're right, I should have paid more attention. :sweat_smile:

MarianMacik (approved):

jbpm/process-serialization-protobuf/src/main/java/org/jbpm/marshalling/impl/AbstractProcessInstanceMarshaller.java
tiagodolphine: Marian and his eagle eyes :P

ruromero: :superhero: diffman!

elguardian:
> @elguardian this changes the marshaller, can you just double-check its ok with your changes

not really sure why transport stuff should go in process state. seems weird to me, but don't really have the context.

cristianonicolai:
jenkins please retest this

ruromero:
jenkins retest this

cristianonicolai:
jenkins please retest this

cristianonicolai:
jenkins please retest this

cristianonicolai:
jenkins please retest this

ruromero:
@cristianonicolai look at the latest commit where I move some things around to fix the persistence after the latest rebase. Maybe this is somehow related to the previous errors.

jbpm/process-serialization-protobuf/src/main/resources/org/kie/kogito/serialization/process/protobuf/kogito_types.proto
```diff
@@ -28,6 +28,7 @@ message WorkflowContext {
     repeated NodeInstance node_instance = 2;
     repeated NodeInstanceGroup exclusive_group = 3;
     repeated IterationLevel iterationLevels = 4;
+    repeated Variable transport_context = 5;
```
elguardian, line 31: Hi @ruromero  this does not belong to the workflow context. In a process instance a workflow context could belong to a subprocess ,compiste contexts (like MI) and so forth. There is no point to set this in here. Please move this to a higher level (process instance) -> kogito_process_instance.proto

ruromero:
jenkins retest this

jbpm/process-serialization-protobuf/src/main/resources/org/kie/kogito/serialization/process/protobuf/kogito_types.proto
ruromero: done

ruromero:
jenkins retest this

ruromero:
jenkins retest this please

cristianonicolai:
jenkins please retest this

PR #910, opened 2020-11-24
[KOGITO-3911] [1.0.x] Fix to ack received kafka messages on QuarkusCloudEventPublisher

tiagodolphine:

### DIFF
--- a/addons/cloudevents/cloudevents-quarkus-addon/src/main/java/org/kie/kogito/addon/cloudevents/quarkus/QuarkusCloudEventPublisher.java
+++ b/addons/cloudevents/cloudevents-quarkus-addon/src/main/java/org/kie/kogito/addon/cloudevents/quarkus/QuarkusCloudEventPublisher.java
@@ -24,6 +24,7 @@ import javax.inject.Named;
 import io.quarkus.runtime.Startup;
 import io.smallrye.mutiny.Multi;
 import org.eclipse.microprofile.reactive.messaging.Channel;
+import org.eclipse.microprofile.reactive.messaging.Message;
 import org.kie.kogito.event.KogitoEventStreams;
 import org.reactivestreams.Publisher;
 
@@ -35,13 +36,15 @@ import org.reactivestreams.Publisher;
 @ApplicationScoped
 public class QuarkusCloudEventPublisher {
     @Channel(KogitoEventStreams.INCOMING)
-    Multi<String> events;
+    Multi<Message<String>> events;
 
     @Produces
     @ApplicationScoped
     @Named(KogitoEventStreams.PUBLISHER)
     public Multi<String> makeMulti() {
-        return events.broadcast().toAllSubscribers();
-    }
-
+        return events
+                .invoke(Message::ack)
+                .map(Message::getPayload)
+                .broadcast().toAllSubscribers();
+    }    
 }


evacchi (approved):

evacchi:
honestly looks trivial enough that it makes sense to merge and restart the build WDYT @cristianonicolai ?

tiagodolphine:
> honestly looks trivial enough that it makes sense to merge and restart the build WDYT @cristianonicolai ?

Yeah, I'd say the same +1

cristianonicolai (approved):
@tiagodolphine @evacchi agree, +1 to merge and restart.

addons/cloudevents/cloudevents-quarkus-addon/src/main/java/org/kie/kogito/addon/cloudevents/quarkus/QuarkusCloudEventPublisher.java
```diff
@@ -35,13 +36,15 @@
...
     @Produces
     @ApplicationScoped
     @Named(KogitoEventStreams.PUBLISHER)
     public Multi<String> makeMulti() {
-        return events.broadcast().toAllSubscribers();
-    }
-
-}
+        return events
+                .invoke(Message::ack)
```
radtriste, line 46: so we ack the message but what happens if the message cannot be read by the subscriber ? isn't there already a ack process on subscriber side ?

tiagodolphine: Yeah, this is a possible issue @radtriste, this fix is basically to ack the message when it is received before it is handled by the subscriber, without the fix no acks were being done, there is no way to ack the message on the subscriber side, or after it, with the current, because we are passing the payload (as a String), and we need the `Message` to do the ack. We could have an abstraction for the massage and have an impl for Quarkus and Spring, anyway, IMO this would be an improvement that could be done in a new Jira because this requires bigger refactoring. 
@evacchi @cristianonicolai ^

radtriste: that was my concern. we had a similar issue on data-index when a protobuf is being updated. you may receive messages with new format before it is updated on data-index for example. and thus, the message raised an exception and the it was not ack to have a retry later.
Would that change have such an impact here ?
cc @sutaakar also

tiagodolphine: nop, this is an unrelated change, basically, this is how the messages are received in the runtime, like if in a process there is a Start Message Event or an Intermediate Message Event, that consumes a message from a topic to interact with a process in the runtime. The part the publishes the events that are consumed by the data-index is decoupled from this.

radtriste: ok for quick fix here then for me but on master we should have a better correction IMO :)

radtriste (approved):

PR #915, opened 2020-11-26
KOGITO-3849: kogito-events-api module + pluggable CloudEvent consumers

kostola:
* [Related Jira task](https://issues.redhat.com/browse/KOGITO-3849)
* [Related PR in kogito-apps](https://github.com/kiegroup/kogito-apps/pull/568)

## Goal

Give the developer of a Kogito app the capability to write custom consumers of events processed by the cloudevents addon with minimum effort. While doing so, try to isolate all the event-related APIs in a dedicated module.

### Target

* Have a clearly identifiable place where to put all generic event-related code.
* Have a cleaner dependency graph.

### Implementation

* I moved almost all the generic classes and interfaces from `kogito-api` and `kogito-services` to a new module called `kogito-events-api`, with minimal dependencies (`jackson-annotations` and `reactive-streams`).
* I only left `org.kie.kogito.event.EventManager` inside `kogito-api` because otherwise it would generate a circular dependency between that and the new module: in fact, it depends from `org.kie.kogito.Addons` and is a dependency of `org.kie.kogito.uow.UnitOfWorkManager`.
* I left several classes/interfaces in `org.kie.kogito.services.event` inside `kogito-services` because they're strictly related to processes.
* All the events addons now depends from `kogito-events-api` (it's even the only dependency of `cloudevents-common-addon`).
* `kogito-events-api` is now an [_optional_ dependency](https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html) of both `kogito-api` and `kogito-services`: this allows to build without issues but forces the developer who needs events to explicitly declare it as dependency or include one of the events addons. 

### Implementation

I added a `CloudEventReceiver` class, which in my view is the dual of `CloudEventEmitter`, that hides the details about the wiring with the publisher.

The developer will simply need to declare a new bean that receives an injected `CloudEventReceiver` instance, which can be used to subscribe to incoming events.

### Custom consumer example for Quarkus

[Here](https://github.com/kostola/kogito-examples/blob/KOGITO-3849/custom-event-consumer-quarkus/src/main/java/org/kie/kogito/examples/CustomCloudEventConsumer.java) is an example of a custom event consumer written for Quarkus.

### Custom consumer example for Spring Boot

[Here](https://github.com/kostola/kogito-examples/blob/KOGITO-3849/custom-event-consumer-springboot/src/main/java/org/kie/kogito/examples/CustomCloudEventConsumer.java) is an example of a custom event consumer written for Spring Boot.

### Known issues

* Only `String` payloads are supported at the moment. But that's how it works for the rest of the implementation.


kostola:
jenkins retest this please

kostola:
jenkins retest this please

ricardozanini (approved):
Nice move!

r00ta (approved):
Cool :+1:

api/kogito-api/pom.xml
```diff
@@ -35,6 +35,12 @@
       <artifactId>jackson-annotations</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>org.kie.kogito</groupId>
+      <artifactId>kogito-events-api</artifactId>
+      <optional>true</optional>
```
danielezonca, line 41: Please add a comment to mention why this dependency is optional and that is a transitive dependency in the addon so user is expected to only provide the addon

api/kogito-services/pom.xml
```diff
@@ -24,6 +24,11 @@
       <groupId>org.kie.kogito</groupId>
       <artifactId>kogito-api</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.kie.kogito</groupId>
+      <artifactId>kogito-events-api</artifactId>
+      <optional>true</optional>
```
danielezonca, line 30: Same as above

jbpm/jbpm-flow/pom.xml
```diff
@@ -25,6 +25,10 @@
       <groupId>org.kie.kogito</groupId>
       <artifactId>kogito-api</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.kie.kogito</groupId>
+      <artifactId>kogito-events-api</artifactId>
```
danielezonca, line 30: I think it should be nice in another PR if this dependency can become optional also in `jbpm-flow` module. I expect we will need to do some changes in the code (i.e. create a `jbpm-events` module or similar).
Wdyt? Do you think it makes sense to create a similar ticket/task?

danielezonca (approved):
Well done :+1: just some minor comments

cristianonicolai (approved):

jbpm/jbpm-flow/pom.xml
cristianonicolai: +1

api/kogito-api/pom.xml
kostola: Done :+1:

api/kogito-services/pom.xml
kostola: Done :+1:

kostola:
jenkins retest this please

MarianMacik (approved):

MarianMacik:
@kostola just one question about the optional `kogito-events-api` dependency. So it is needed for compilation, but in case an addon is not on classpath, the code generation doesn't use the abstract classes and their imports, so we don't need it in that case?

jbpm/jbpm-flow/pom.xml
kostola: Created [a dedicated Jira task](https://issues.redhat.com/browse/KOGITO-4058)

PR #917, opened 2020-11-27
[KOGITO-3759] PoC of outbox pattern for resolving the Kogito consistency issue with Debezium

xieshenzh:
PoC of outbox pattern for resolving the Kogito consistency issue (i.e. persistence and events consistency) with Debezium:
https://debezium.io/blog/2019/02/19/reliable-microservices-data-exchange-with-the-outbox-pattern/

Please refer to [README](https://github.com/kiegroup/kogito-runtimes/blob/5fad81783d861cc9c0655154890ecd1300a549b6/addons/events/kogito-events-reactive-messaging-addon/debezium/README.md) for the steps of running the PoC.

Jira: https://issues.redhat.com/browse/KOGITO-3759

### DIFF
--- a/addons/events/kogito-events-reactive-messaging-addon/src/main/java/org/kie/kogito/events/rm/document/AbstractDataEventDocument.java
+++ b/addons/events/kogito-events-reactive-messaging-addon/src/main/java/org/kie/kogito/events/rm/document/AbstractDataEventDocument.java
@@ -1,0 +1,122 @@
+package org.kie.kogito.events.rm.document;
+
+public class AbstractDataEventDocument {
+
+    String id;
+    String specVersion;
+    String source;
+    String type;
+    String time;
+    String subject;
+    String dataContentType;
+    String dataSchema;
+    String kogitoProcessinstanceId;
+    String kogitoRootProcessinstanceId;
+    String kogitoProcessId;
+    String kogitoRootProcessId;
+    String kogitoAddons;
+
+    public String getId() {
+        return id;
+    }
+
+    public void setId(String id) {
+        this.id = id;
+    }
+
+    public String getSpecVersion() {
+        return specVersion;
+    }
+
+    public void setSpecVersion(String specVersion) {
+        this.specVersion = specVersion;
+    }
+
+    public String getSource() {
+        return source;
+    }
+
+    public void setSource(String source) {
+        this.source = source;
+    }
+
+    public String getType() {
+        return type;
+    }
+
+    public void setType(String type) {
+        this.type = type;
+    }
+
+    public String getTime() {
+        return time;
+    }
+
+    public void setTime(String time) {
+        this.time = time;
+    }
+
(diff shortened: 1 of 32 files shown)


addons/events/kogito-events-reactive-messaging-addon/src/main/java/org/kie/kogito/events/rm/document/AbstractDataEventDocument.java
```diff
@@ -0,0 +1,122 @@
...
+    String type;
+    String time;
+    String subject;
+    String dataContentType;
+    String dataSchema;
+    String kogitoProcessinstanceId;
+    String kogitoRootProcessinstanceId;
+    String kogitoProcessId;
+    String kogitoRootProcessId;
+    String kogitoAddons;
```
MarianMacik, line 17: Do these fields need to be package-private?

api/kogito-api/src/main/java/org/kie/kogito/event/EventManager.java
```diff
@@ -36,7 +36,7 @@
      * processed only when there are any publishers available. 
      * @param batch batch to be published
      */
-    void publish(EventBatch batch);
+    void publish(EventBatch batch, Object ... options);
```
MarianMacik, line 39: I know this is just a PoC, but in the final code we could use some kind of abstraction here instead of `Object`.

MarianMacik (approved):
Looks good to me from the design point of view.

cristianonicolai:
@xieshenzh I think we have two main areas here:
- transaction management for Mongo client
- Event persistence
What do you think if we split these into separated PRs? I think would be good for us to first see how to abstract a way for tm to be plugged in, independently from the persistence type, which is quite coupled right now.
Next, look into how can we allow events to be persisted instead of sending out, I think in there we might want to create a new addon, and try to reuse the event classes that we already have, but anyway maybe having different PRs could help us start moving this code into master.

xieshenzh:
> @xieshenzh I think we have two main areas here:
> 
> * transaction management for Mongo client
> * Event persistence
>   What do you think if we split these into separated PRs? I think would be good for us to first see how to abstract a way for tm to be plugged in, independently from the persistence type, which is quite coupled right now.
>   Next, look into how can we allow events to be persisted instead of sending out, I think in there we might want to create a new addon, and try to reuse the event classes that we already have, but anyway maybe having different PRs could help us start moving this code into master.

@cristianonicolai it makes sense. I'll work on that.

MarianMacik:
One thing that came to my mind is that we maybe could decouple persistence configuration separately for runtime data and events. So runtime data could be for example still persisted in Infinispan but events could be persisted and emitted via MongoDB. That would help as this would make just eventing dependent on Debezium, so in case Debezium doesn't support a particular storage, e.g. Infinispan, it could still be used for eventing with MongoDB for example.

PR #922, opened 2020-12-07
[KOGITO-3974] Make DecisionModels and PredictionModels injectable

danielezonca:
https://issues.redhat.com/browse/KOGITO-3974

Details:
- `DecisionModels` generation has now platform specific templates
- `PredictionModels` generation has now platform specific templates
- Created `AbstractDecisionModels` to share common code
- Created `AbstractPredictionModels` to share common code
- Updated `Application` generation to use injection (where available) also for `DecisionModels` and `PredictionModels`
- ~~Now all `Applicatation` templates correctly delegate to `StaticApplication`~~ postponed to [KOGITO-3975](https://issues.redhat.com/browse/KOGITO-3975)


danielezonca:
jenkins retest this please

drools/kogito-dmn/src/main/java/org/kie/kogito/dmn/AbstractDecisionModels.java
```diff
@@ -0,0 +1,76 @@
...
+package org.kie.kogito.dmn;
+
+import org.kie.dmn.api.core.DMNRuntime;
+import org.kie.kogito.Application;
+import org.kie.kogito.ExecutionIdSupplier;
+import org.kie.kogito.decision.DecisionModels;
+
+public abstract class AbstractDecisionModels implements DecisionModels {
+
+    private final static boolean IS_NATIVE_IMAGE = System.getProperty("org.graalvm.nativeimage.imagecode") != null;
```
tarilabs, line 24:
```suggestion
    private final static boolean IS_NATIVE_IMAGE = org.kie.dmn.feel.util.ClassLoaderUtil.CAN_PLATFORM_CLASSLOAD;
```

integration-tests/integration-tests-quarkus-decisions/src/main/java/org/kie/kogito/integrationtests/InjectDecisionModels.java
```diff
@@ -0,0 +1,34 @@
...
+
+import javax.inject.Inject;
+
+@Startup
+public class InjectDecisionModels {
+
+    @Inject
+    public InjectDecisionModels(DecisionModels decisionModels, Application application) {
+        if(decisionModels != application.decisionModels()) {
+            throw new IllegalStateException("DecisionModels should be injectable and same instance application.decisionModels()");
```
tarilabs, line 31: isn't this "pattern" which I see repeated for processes and rules, a sympthom we don't have a clear API design at the Kogito platform level?

kogito-codegen/src/main/java/org/kie/kogito/codegen/ApplicationContainerGenerator.java
```diff
@@ -73,8 +74,10 @@ public CompilationUnit getCompilationUnitOrThrow() {
                         templatePath(),
                         "Compilation unit doesn't contain a class or interface declaration!"));
 
-        for (String section : sections) {
-            replaceSectionPlaceHolder(cls, section);
+        if (annotator == null) {
+            for (String section : sections) {
+                replaceSectionPlaceHolder(cls, section);
+            }
```
tarilabs, line 80: I'm not sure I get this change, can you explain please?

kogito-quarkus-extension/deployment/src/main/java/org/kie/kogito/quarkus/deployment/KogitoAssetsProcessor.java
```diff
@@ -192,7 +193,7 @@ public void generateModel() throws IOException {
                 new ApplicationGenerator(
                         appPackageName,
                         new File(appPaths.getFirstProjectPath().toFile(), "target"))
-                        .withDependencyInjection(new CDIDependencyInjectionAnnotator())
+                        .withDependencyInjection(dependencyInjectionAnnotator)
```
tarilabs, line 196: since this change, and others below, shouldn't that reflect also on ALL the `...CompilationProvider` classes??

tarilabs (commented):
thank you, this is good stuff, I have some questions if you can kindly help me understand better below please

drools/kogito-pmml/src/main/java/org/kie/kogito/pmml/AbstractPredictionModels.java
```diff
@@ -0,0 +1,41 @@
...
+    protected Function<String, KieRuntimeFactory> getKieRuntimeFactoryFunction() {
+        return kieRuntimeFactoryFunction;
+    }
+
+    protected void setKieRuntimeFactoryFunction(Function<String, KieRuntimeFactory> kieRuntimeFactoryFunction) {
+        this.kieRuntimeFactoryFunction = kieRuntimeFactoryFunction;
+    }
+
+    private org.kie.pmml.api.runtime.PMMLRuntime getPMMLRuntime(java.lang.String modelName) {
+        return getKieRuntimeFactoryFunction().apply(modelName).get(org.kie.pmml.api.runtime.PMMLRuntime.class);
```
gitgabrio, line 39: @danielezonca 
Maybe just a leftover, but I guess better to use directly "kieRuntimeFactoryFunction" instead of the getter method.
Beside that, as a side note, there is not null check/NPE protection. 
Would it be possible to put the "kieRuntimeFactoryFunction" in the constructor ?
I see the setter is called in the template
`
public PredictionModels() {
        super();
        setKieRuntimeFactoryFunction(sKieRuntimeFactoryFunction);
    } 
`
so - unless for other reasons - it would just be a matter of moving "sKieRuntimeFactoryFunction" inside the "super" invocation.

kogito-codegen/src/main/java/org/kie/kogito/codegen/prediction/PredictionModelsGenerator.java
```diff
@@ -30,60 +26,87 @@
...
-    public PredictionModelsGenerator(String applicationCanonicalName, List<PMMLResource> resources) {
-        super("PredictionModels", "predictionModels", PredictionModels.class);
+    public PredictionModelsGenerator(String packageName, String applicationCanonicalName, List<PMMLResource> resources) {
+        super(SECTION_CLASS_NAME, "predictionModels", PredictionModels.class);
         this.applicationCanonicalName = applicationCanonicalName;
         this.resources = resources;
+
+        this.templatedGenerator = new TemplatedGenerator(
+                packageName,
+                SECTION_CLASS_NAME,
```
gitgabrio, line 55: @danielezonca 
If I'm not wrong, elsewhere in the code we have a strict naming convention.
As far as I see all this templates have the same schema: {context}{base model template} where "context" is "Cdi" or "Spring" and "base model template" is (for example) "PredictionModelsTemplate".
Would not be easier to just pass the latter (e.g. "PredictionModelsTemplate") to the "TemplatedGenerator", and have it infer the Cdi/Spring ones ? This would also enforce consistence between models, that is a good thing (IMHO).

Beside that, the only differences I see in the "models" templates, between different contexts, are the annotations on the class declaration (Cdi vs Spring ones). If/where this is true, I think it would be much better to remove all that useless duplications, have a "plain" template, and use JavaParser to dynamically add required annotation (Cdi/Spring) when needed. 
All those templates are pretty cumbersome; whenever a modification is needed, it must be done twice; and, after all, they not really needed (again, IMVHO)

gitgabrio (changes requested):
@danielezonca 
Some minor comments, but I would really love to get rid of all those templates, at least the "duplicated" ones

integration-tests/integration-tests-quarkus-decisions/src/main/java/org/kie/kogito/integrationtests/InjectDecisionModels.java
danielezonca: You are probably right: we generate classes using templates it is easy to remove an annotation and breaks without an integration tests that test it. I don't know if we can enforce this at API/design level to be honest :thinking: 
Any idea/suggestion?

kogito-codegen/src/main/java/org/kie/kogito/codegen/ApplicationContainerGenerator.java
danielezonca: If you look at `*ApplicationTemplate`s files there was a "replacement" pattern like
`this.process = null /* $Processes$ */;` -> `this.process = new Processes(this);`
Now that all the models are injectable this is no more needed except for the scenario with `annotator == null` (no Quarkus/Spring) so I decided to make this explicit and skip the replacement in this case

kogito-quarkus-extension/deployment/src/main/java/org/kie/kogito/quarkus/deployment/KogitoAssetsProcessor.java
danielezonca: Sorry but I don't understand what do you mean. I just created a local variable instead of instantiate multiple times  `CDIDependencyInjectionAnnotator`.
Can you please clarify your comment?

drools/kogito-dmn/src/main/java/org/kie/kogito/dmn/AbstractDecisionModels.java
danielezonca: Done

drools/kogito-pmml/src/main/java/org/kie/kogito/pmml/AbstractPredictionModels.java
danielezonca: Done, I reviewed all the templates and now ~all the code is inside the abstract :+1:

kogito-codegen/src/main/java/org/kie/kogito/codegen/prediction/PredictionModelsGenerator.java
danielezonca: Done, I reviewed all the templates and now ~all the code is inside the abstract :+1:

danielezonca:
jenkins retest this please

kogito-quarkus-extension/deployment/src/main/java/org/kie/kogito/quarkus/deployment/KogitoAssetsProcessor.java
tarilabs: I believe you did it with 574cf923f0cce1da08ea0675c4a8ab3c9c0d8cea

tarilabs: do other classes of `...CompilationProvider` need this too?

integration-tests/integration-tests-quarkus-decisions/src/main/java/org/kie/kogito/integrationtests/InjectDecisionModels.java
tarilabs: No I don't think this is solvable within the scope of this PR.
I'm saying this sounded to me a symptom the design of the Kogito API has a problem.

kogito-codegen/src/main/java/org/kie/kogito/codegen/ApplicationContainerGenerator.java
tarilabs: thanks for the feedback; worthy of adding it as a `// <code comment>`?

drools/kogito-pmml/src/main/java/org/kie/kogito/pmml/AbstractPredictionModels.java
```diff
@@ -14,28 +14,27 @@
...
-
 public abstract class AbstractPredictionModels implements PredictionModels {
 
-    private java.util.function.Function<String, org.kie.api.runtime.KieRuntimeFactory> kieRuntimeFactoryFunction;
+    public static java.util.function.Function<String, org.kie.api.runtime.KieRuntimeFactory> kieRuntimeFactoryFunction;
 
-    public org.kie.kogito.prediction.PredictionModel getPredictionModel(java.lang.String modelName) {
-        return new org.kie.kogito.pmml.PmmlPredictionModel(getPMMLRuntime(modelName), modelName);
-    }
+    protected static void init(String ... pmmlFiles) {
```
gitgabrio, line 23: @danielezonca 
Nice refactoring. Out of curiosity: is it needed for this method to be "static" ? I see it invoked inside a constructor

kogito-codegen/src/main/java/org/kie/kogito/codegen/ApplicationContainerGenerator.java
danielezonca: Done :+1:

gitgabrio (approved):
Thanks for the refactoring.
LGTM

kogito-quarkus-extension/deployment/src/main/java/org/kie/kogito/quarkus/deployment/KogitoAssetsProcessor.java
danielezonca: Thanks a lot for the comment, I reviewed the usages and I realized that all `Generator`s have a `setDependencyInjection` that is automatically provided by `ApplicationGenerator` during the initialization so updated the code to use that one

drools/kogito-pmml/src/main/java/org/kie/kogito/pmml/AbstractPredictionModels.java
danielezonca: It is not invoked inside the constructor but with a `static` block (i.e. [link](https://github.com/kiegroup/kogito-runtimes/blob/3769ea2644548e7e397cb39ee229c316bd759a2c/kogito-codegen/src/main/resources/class-templates/CdiPredictionModelsTemplate.java#L5))

gitgabrio: Thx, misread :+1:

evacchi (approved):
Approved, but I would move the rename of `withGenerator()` to a separate JIRA :)

drools/kogito-pmml/src/main/java/org/kie/kogito/pmml/AbstractPredictionModels.java
```diff
@@ -14,20 +14,27 @@
...
-    public static java.util.function.Function<String, org.kie.api.runtime.KieRuntimeFactory> kieRuntimeFactoryFunction;
+    private static final AtomicReference<Function<String, KieRuntimeFactory>> functionReference = new AtomicReference<>();
+    public static final Function<String, KieRuntimeFactory> kieRuntimeFactoryFunction = s -> functionReference.get().apply(s);
 
     protected static void init(String ... pmmlFiles) {
-        final java.util.Map<org.kie.api.KieBase, org.kie.api.runtime.KieRuntimeFactory> kieRuntimeFactories = org.kie.kogito.pmml.PMMLKogito.createKieRuntimeFactories(pmmlFiles);
-        kieRuntimeFactoryFunction = s -> kieRuntimeFactories.keySet().stream()
-                .filter(kieBase -> org.kie.pmml.evaluator.core.utils.KnowledgeBaseUtils.getModel(kieBase, s).isPresent())
+        final java.util.Map<KieBase, KieRuntimeFactory> kieRuntimeFactories = PMMLKogito.createKieRuntimeFactories(pmmlFiles);
+        final Function<String, KieRuntimeFactory> function = s -> kieRuntimeFactories.keySet().stream()
```
gitgabrio, line 32: @danielezonca 
Fine if it works.
Anyway, all that still look as a painful workaround related to https://issues.redhat.com/browse/DROOLS-5724.
If I'm not wrong it all depends on the fact that DMN requires PMML for validation at building time (inside its Kogito-specific builder) even if NULL (as currently is), while model validation should be managed in previous step and should be model specific.

gitgabrio (approved):

drools/kogito-pmml/src/main/java/org/kie/kogito/pmml/AbstractPredictionModels.java
tarilabs: > If I'm not wrong it all depends on the fact that DMN requires PMML for validation at building time

This assessment is incorrect.
This was already explained in the JIRA history (ref https://issues.redhat.com/browse/DROOLS-5724?focusedCommentId=15374787&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15374787 ) and followed up with the Kogito leads responsible for the previous Vs current directives.
Yes DMN internals will be refactored as part of DROOLS-5724 anyway.

danielezonca: I added a comment to explain the reason of the `AtomicReference`. I tested it both jvm and native mode so we have the possibility to approach other refactoring incrementally without block this PR

integration-tests/integration-tests-quarkus-decisions/src/main/java/org/kie/kogito/integrationtests/InjectDecisionModels.java
MarianMacik, line 31: `DecisionModels should be injectable and same instance application.decisionModels()` - probably `same instance as application.decisionModels()`?

integration-tests/integration-tests-quarkus-predictions/src/main/java/org/kie/kogito/integrationtests/InjectPredictionModels.java
```diff
@@ -0,0 +1,34 @@
...
+
+import javax.inject.Inject;
+
+@Startup
+public class InjectPredictionModels {
+
+    @Inject
+    public InjectPredictionModels(PredictionModels predictionModels, Application application) {
+        if(predictionModels != application.predictionModels()) {
+            throw new IllegalStateException("PredictionModels should be injectable and same instance application.predictionModels()");
```
MarianMacik, line 31: Same here.

integration-tests/integration-tests-quarkus-processes/src/main/java/org/kie/kogito/integrationtests/InjectProcesses.java
```diff
@@ -0,0 +1,34 @@
...
+
+import javax.inject.Inject;
+
+@Startup
+public class InjectProcesses {
+
+    @Inject
+    public InjectProcesses(Processes processes, Application application) {
+        if(processes != application.processes()) {
+            throw new IllegalStateException("Processes should be injectable and same instance application.processes()");
```
MarianMacik, line 31: And here.

kogito-codegen/src/main/resources/class-templates/CdiApplicationTemplate.java
```diff
@@ -1,30 +1,16 @@
 package $Package$;
 
-import org.kie.kogito.Config;
-import org.kie.kogito.StaticApplication;
-import org.kie.kogito.process.Processes;
-import org.kie.kogito.rules.RuleUnits;
-
 @javax.inject.Singleton
-public class Application extends StaticApplication {
+public class Application extends org.kie.kogito.StaticApplication {
```
MarianMacik, line 4: I know that generated code probably shouldn't be readable that much, but wouldn't it be better to leave the imports here to make it cleaner? If the class is moved in future to a different package, we will have to change the templates anyway.

drools/kogito-pmml/src/main/java/org/kie/kogito/pmml/AbstractPredictionModels.java
```diff
@@ -0,0 +1,51 @@
...
+public abstract class AbstractPredictionModels implements PredictionModels {
+
+    // This AtomicReference is needed to have a deterministic/stable value for kieRuntimeFactoryFunction.
+    // DecisionModels statically refers kieRuntimeFactoryFunction but it can be loaded before PredictionModels (it depends
+    // on dependency injection container initialisation) so it is not possible to make an assumption on the order.
+    // It can be probably removed (and make kieRuntimeFactoryFunction private) after resolving https://issues.redhat.com/browse/DROOLS-5724
+    private static final AtomicReference<Function<String, KieRuntimeFactory>> functionReference = new AtomicReference<>();
+    public static final Function<String, KieRuntimeFactory> kieRuntimeFactoryFunction = s -> functionReference.get().apply(s);
+
+    protected static void init(String ... pmmlFiles) {
```
MarianMacik, line 34: So the static block is used because of the CDI as we have to have a no-arg constructor? (Otherwise we would use just parameters of the constructor...)

MarianMacik (commented):
Just have one question + I found the exception message confusing after first read :)

drools/kogito-pmml/src/main/java/org/kie/kogito/pmml/AbstractPredictionModels.java
danielezonca: No here the problem is that both DMN and PMML has some `static` code to be executed to load files. This is a trick for native compilation (and avoid parsing at runtime). So instead of duplicate this static code in all the templates, I moved it to the abstract class and created this init static method that is invoked in a `static` block codegenerated with the arguments.

integration-tests/integration-tests-quarkus-decisions/src/main/java/org/kie/kogito/integrationtests/InjectDecisionModels.java
danielezonca: Done

integration-tests/integration-tests-quarkus-processes/src/main/java/org/kie/kogito/integrationtests/InjectProcesses.java
danielezonca: Done

integration-tests/integration-tests-quarkus-predictions/src/main/java/org/kie/kogito/integrationtests/InjectPredictionModels.java
danielezonca: Done

kogito-codegen/src/main/resources/class-templates/CdiApplicationTemplate.java
danielezonca: In general I prefer to be on the safe side with templates: when we add code to the template it could produce name clashing while with FQCN this should not happen. Btw no strong opinion, if you prefer I can revert

MarianMacik: Makes sense, thanks!

MarianMacik (approved):
All sorted, thanks!

PR #925, opened 2020-12-08
KOGITO-3659 Process codegen won't generate duplicate classes

ricardozanini:
Signed-off-by: Ricardo Zanini <zanini@redhat.com>

See: https://issues.redhat.com/browse/KOGITO-3659

On Serverless Workflows, sometimes we want to reuse a given set of functions on subprocesses. In such scenarios, the codegen was duplicating the `WorkItemHandlers` responsible for handling these functions.

An example of such a case can be viewed in: <PR to be sent>

Instead of using a `List`, we use a `Set` to not have duplicates in the final generated code. This can be re-think in the future. For example: would make sense to generate all the handlers beforehand on a specific package instead? Today each process take care of their handlers, we don't have "global" scope for such thing. Wouldn't that make sense? For a handler to be reused on subprocesses/flows?

### DIFF
--- a/kogito-codegen/src/main/java/org/kie/kogito/codegen/GeneratedFile.java
+++ b/kogito-codegen/src/main/java/org/kie/kogito/codegen/GeneratedFile.java
@@ -16,6 +16,7 @@
 package org.kie.kogito.codegen;
 
 import java.nio.charset.StandardCharsets;
+import java.util.Objects;
 
 public class GeneratedFile {
 
@@ -86,5 +87,22 @@ public class GeneratedFile {
                 "type=" + type +
                 ", relativePath='" + relativePath + '\'' +
                 '}';
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        GeneratedFile that = (GeneratedFile) o;
+        return Objects.equals(relativePath, that.relativePath) && type == that.type;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(relativePath, type);
     }
 }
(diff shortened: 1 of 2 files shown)


ricardozanini:
Jenkins test this please

tsurdilo (approved):

cristianonicolai (approved):

evacchi:
as much as it was silently failing before, this is still silently failing now; 

- with List<T> if the same path is found twice it will be overwritten on disk. 
- if I remember correctly the behavior of Set<T>, this is now "ignoring" an incoming file when it has the same path of an existing one, effectively writing to disk the "first" one instead of the "last" one.

In both cases, the result is that only the contents of one of the GeneratedFiles that you have added will be written to disk.

Proposal: wouldn't it be better to error out when a duplicate is found? This way we can fix the generation code instead of silently failing.

ricardozanini:
> Proposal: wouldn't it be better to error out when a duplicate is found? This way we can fix the generation code instead of silently failing.

Unfortunately in this particular scenario we can't unless we have a major refactoring in Process codegen. Each subflow generates its handler, not knowing if it's a duplicated handler or not. My first idea was to not generate the handlers individually within the generation tree (for each individual) but as a whole. I don't think we are in a moment to do this refactor now. We need to rethink the codegen as a whole first, which is something that we already do. I'll open a follow-up JIRA to investigate this generation.

What we can do now, is to add to the hash computation the bytes of the file as well. If it's the same, we will ignore, if not, it will fail in the compilation, so we won't silence ignore this scenario. wdyt @evacchi ?

ricardozanini:
@evacchi now we will fail if the file contents are different on compile time.

evacchi (approved):

ricardozanini:
Jenkins test this please

PR #926, opened 2020-12-09
KOGITO-3959 Upgrade Xstream version to 1.4.14

cristianonicolai:


radtriste (approved):

evacchi (approved):
do we have to do this on v7 too?

cristianonicolai:
@evacchi I would say so, @mareknovotny could you take a look there?

mareknovotny (approved):

mareknovotny:
> do we have to do this on v7 too?

no, it is already upgraded there

PR #930, opened 2020-12-10
[KOGITO-3969] Reviewed ApplicationGenerator.logger usage + removed dead code

danielezonca:
https://issues.redhat.com/browse/KOGITO-3969

Details:
- `ApplicationGenerator.logger` now is a private field and each class has its own logger (to prevent misleading logging messages from a different package)
- Removed `ApplicationGenerator.log` method (a debug method that logs generated content) and moved the logic to `GeneratedFile` constructor: now if debug is enabled, all generated classes are logged with also type and relativePath
- Removed unused private fields in `ConfigGenerator`
- Removed unused `hasRuleUnits` field from `ApplicationGeneator` and updated all the existing usages (all in tests)


kogito-codegen/src/main/java/org/kie/kogito/codegen/prediction/PredictionCodegen.java
```diff
@@ -234,6 +233,6 @@ private void addModels(final List<KiePMMLModel> kiepmmlModels, final PMMLResourc
 
 
     private void storeFile(GeneratedFile.Type type, String path, String source) {
-        generatedFiles.add(new GeneratedFile(type, path, log(source).getBytes(StandardCharsets.UTF_8)));
+        generatedFiles.add(new GeneratedFile(type, path, source));
```
evacchi, line 236: I know that you did it to avoid repetition but I don't love that we are now logging within the constructor of this class, because now it will probably log stuff that we may not really care about (eg resources)

OTOH: I never loved this logging; this stuff is dumped to disk sooner or later, so  it's much easier to read it from a file, rather than displayed on the console log.

So I am ok with the change, although it feels a bit strange (to me).

evacchi (approved):

kogito-codegen/src/main/java/org/kie/kogito/codegen/prediction/PredictionCodegen.java
danielezonca: Yes I agree that it produce a lot of noise in the logs but my goal was to centralize it for now.
I moved the logging to `generate()` method of `ApplicationGenerator` so that it is not a side effect inside `GeneratedFile` constructor and it will be easier to change in the future (i.e. dump to disk instead of log).
Wdyt?

evacchi: I like it much better now!

kogito-codegen/src/main/java/org/kie/kogito/codegen/ApplicationGenerator.java
```diff
@@ -41,7 +40,7 @@
 
 public class ApplicationGenerator {
 
-    public static final Logger logger = LoggerFactory.getLogger(ApplicationGenerator.class);
+    private static final Logger logger = LoggerFactory.getLogger(ApplicationGenerator.class);
```
radtriste, line 43: could we put the logger as `LOGGER` ?

kogito-codegen/src/main/java/org/kie/kogito/codegen/decision/DecisionCodegen.java
```diff
@@ -222,25 +220,25 @@ private void generateStronglyTypedInput(DMNModel model) {
             DMNTypeSafeTypeGenerator generator = new DMNTypeSafeTypeGenerator(model, index, factory).withJacksonAnnotation();
             boolean useMPAnnotations = isMPAnnotationsPresent();
             if (useMPAnnotations) {
-                logger.debug("useMPAnnotations");
+                LOG.debug("useMPAnnotations");
```
radtriste, line 223: could we rename as `LOGGER` ?

kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/DeclaredTypeCodegen.java
```diff
@@ -46,12 +46,15 @@
...
 import org.kie.kogito.codegen.rules.config.RuleConfigGenerator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import static java.util.stream.Collectors.toList;
-import static org.kie.kogito.codegen.ApplicationGenerator.logger;
 
 public class DeclaredTypeCodegen extends AbstractGenerator {
 
+    private static final Logger logger = LoggerFactory.getLogger(DeclaredTypeCodegen.class);
```
radtriste, line 56: `LOGGER`

kogito-codegen/src/main/java/org/kie/kogito/codegen/rules/IncrementalRuleCodegen.java
```diff
@@ -80,16 +80,18 @@
...
 import static java.util.stream.Collectors.toList;
 
 import static com.github.javaparser.StaticJavaParser.parse;
 import static org.drools.compiler.kie.builder.impl.KieBuilderImpl.setDefaultsforEmptyKieModule;
-import static org.kie.kogito.codegen.ApplicationGenerator.log;
-import static org.kie.kogito.codegen.ApplicationGenerator.logger;
 
 public class IncrementalRuleCodegen extends AbstractGenerator {
 
+    private static final Logger logger = LoggerFactory.getLogger(IncrementalRuleCodegen.class);
```
radtriste, line 93: `LOGGER`

radtriste (changes requested):
just small convention

radtriste (approved):
lgtm

PR #938, opened 2020-12-15
[KOGITO-4014] Runtime Process to wait for a process to finish

fjtirado:
Adding waitForEnd method
Tested with https://github.com/kiegroup/kogito-examples/pull/489

### DIFF
--- a/kogito-codegen/src/main/resources/class-templates/RestResourceTemplate.java
+++ b/kogito-codegen/src/main/resources/class-templates/RestResourceTemplate.java
@@ -2,6 +2,7 @@ package com.myspace.demo;
 
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 import javax.ws.rs.Consumes;
@@ -51,16 +52,20 @@ public class $Type$Resource {
     public Response createResource_$name$(@Context HttpHeaders httpHeaders,
                                               @Context UriInfo uriInfo,
                                               @QueryParam("businessKey") String businessKey,
+                                              @QueryParam("forceSync") @DefaultValue("false") boolean forceSync,
+                                              @QueryParam("forceTimeout") @DefaultValue("2000") long timeout,
                                               $Type$Input resource) {
         return UnitOfWorkExecutor.executeInUnitOfWork(application.unitOfWorkManager(), () -> {
             $Type$Input inputModel = resource != null ? resource : new $Type$Input();
             ProcessInstance<$Type$> pi = process.createInstance(businessKey, inputModel.toModel());
             String startFromNode = httpHeaders.getHeaderString("X-KOGITO-StartFromNode");
-
             if (startFromNode != null) {
                 pi.startFrom(startFromNode);
             } else {
                 pi.start();
+            }
+            if (forceSync) {
+                pi.waitForEnd(timeout,TimeUnit.MILLISECONDS);
             }
             UriBuilder uriBuilder = uriInfo.getAbsolutePathBuilder().path(pi.id());
             return Response.created(uriBuilder.build())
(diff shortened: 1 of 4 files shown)


kogito-codegen/src/main/resources/class-templates/RestResourceTemplate.java
```diff
@@ -51,17 +52,21 @@
     public Response createResource_$name$(@Context HttpHeaders httpHeaders,
                                               @Context UriInfo uriInfo,
                                               @QueryParam("businessKey") String businessKey,
+                                              @QueryParam("forceSync") @DefaultValue("false") boolean forceSync,
```
fjtirado, line 55: maybe it makes sense to add this as header parameter? In such case, which name?

ricardozanini: `Kogito-[X]` would be my guess, maybe `Kogito-Process-ForceSync` and `Kogito-Process-WaitTimeout`?

One thing to consider thou is the HTTP timeout VS Process timeout. If the former is less than later, that could mess with the use case. Needs to really document this, or we can just rely on HTTP Timeout and once the request thread goes, the process thread is terminated as well.

fjtirado:
jenkins retest this

evacchi (commented):
the PR looks good, only note this means we are effectively making the process API blocking -- or at least, we are acknowledging so, which wasn't being done before. Is this ok for everyone?

fjtirado:
> the PR looks good, only note this means we are effectively making the process API blocking -- or at least, we are acknowledging so, which wasn't being done before. Is this ok for everyone?

I see this "forceSync" alternative only suitable for "demo" scenarios where we want to show the final result of the process just doing a POST to execute it. We see that need now because REST WIH is async (or better said, reactive), but I guess there will be more async WIH in the future if reactive paradigm remains strong in the computer science world.

tsurdilo (changes requested):
this is so backwards.....

tsurdilo:
@fjtirado @ricardozanini changes in this pr are not needed and backwards from workflow/process perspective. please provide a sync rest workitem handler that ppl can use instead.

ricardozanini:
> please provide a sync rest workitem handler that ppl can use instead.

Hi @tsurdilo! Yes, I think that would make sense. The problem we are having right now with the WIH async implementation is that we can't get the processed data easily.

So if I have a workflow that composes other services together and produces an output, my alternatives are:

1. Produce an event at the end of the process to emit the result to some topic/broker/queue. This approach would require a messaging infrastructure. Right now either a Kafka or Knative broker.
2. Have the workflow to be persisted, so I `POST` to the endpoint, and poll the service with a `GET <business key>` from time to time to get the processed data.
3. Persist the data into a third-party service, having an end operation state to call such service, having my data in there I could fetch that later.

All of these approaches require a lot of setup and infrastructure. Not even only simple "demos" like the one I'm working on right now, but I can see real use cases that a user would like to do simple workflows facing these restrictions giving up.

If the WIH is implemented as sync instead, can I have this scenario supported? Like sending a `POST` to the generated REST service and receive my data back without having to rely on events or persistence?

fjtirado:
> @fjtirado @ricardozanini changes in this pr are not needed and backwards from workflow/process perspective. please provide a sync rest workitem handler that ppl can use instead.

I need to understand why waiting for the process to finish to get the result is backwards, and making the wih sync in an inherently async flow is not.

cristianonicolai:
@fjtirado a service tasks ( work item handler ) can be set as sync ( default ) or async at the BPMN level ( Is Async ), there is some context in https://docs.jboss.org/jbpm/release/7.47.0.Final/jbpm-docs/html_single/#_asynchronous_execution. In saying that we probably don't have an infrastructure in place to reliably execute async WIH as in V7 yet and based on the thread here, I would say that it makes sense to look in this property to tweak the execution. So in the case of SW, it will be sync by default, solving the issue reported by @ricardozanini. In general BPM/process execution, users will then be able to select/choose the approach they want either sync or async. So as @tsurdilo mentioned, making the WIH able to executed synchronously seems like the best way to tackle this issue.

fjtirado:
Ok, got it now, I will change the WIH to be sync and the user will be able to decide if the service task is executed sync or async using that property. 
Thanks @tsurdilo, good catch!!

PR #939, opened 2020-12-15
[KOGITO-4015] Register WorkItemHandler wont work with persistence enable

fjtirado:
Adding property to force handler to be sync


ricardozanini:
This is just a workaround for a demo, don't merge.

ricardozanini:
@fjtirado I believe that we need to properly fix the bug instead of making this workaround. Thanks for helping me.

PR #942, opened 2020-12-16
[KOGITO-4014] Kogito Runtime Process to wait for a process to finish

fjtirado:
Making WIH sync
Please read https://github.com/kiegroup/kogito-runtimes/pull/938 to find out why are we changing WIH to sync

### DIFF
--- a/kogito-build-parent/pom.xml
+++ b/kogito-build-parent/pom.xml
@@ -75,8 +75,8 @@
     <version.taglist.plugin>2.4</version.taglist.plugin>
     <version.versions.plugin>2.1</version.versions.plugin>
     <version.jayway.jsonpath>2.4.0</version.jayway.jsonpath>
-    <version.vertx.web.client>3.9.4</version.vertx.web.client>
-
+    <version.io.smallrye.reactive.mutiny-vertx-web-client>1.2.1</version.io.smallrye.reactive.mutiny-vertx-web-client>
+    
     <!-- this version property is used in plugins but also in dependencies too -->
     <version.io.quarkus>1.10.3.Final</version.io.quarkus>
     <quarkus.logging.manager>org.jboss.logmanager.LogManager</quarkus.logging.manager>
@@ -385,13 +385,11 @@
         <artifactId>json-path-assert</artifactId>
         <version>${version.jayway.jsonpath}</version>
       </dependency>
-
       <dependency>
-        <groupId>io.vertx</groupId>
-        <artifactId>vertx-web-client</artifactId>
-        <version>${version.vertx.web.client}</version>
+        <groupId>io.smallrye.reactive</groupId>
+        <artifactId>smallrye-mutiny-vertx-web-client</artifactId>
+        <version>${version.io.smallrye.reactive.mutiny-vertx-web-client}</version>
       </dependency>
-
       <!-- http communication -->
       <dependency>
         <groupId>com.squareup.okhttp3</groupId>
--- a/kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
+++ b/kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
@@ -24,18 +24,14 @@ import java.util.Set;
 import java.util.function.UnaryOperator;
 import java.util.stream.Collectors;
 
-import io.vertx.core.AsyncResult;
-import io.vertx.core.Handler;
-import io.vertx.core.buffer.Buffer;
 import io.vertx.core.http.HttpMethod;
-import io.vertx.ext.web.client.HttpRequest;
-import io.vertx.ext.web.client.HttpResponse;
-import io.vertx.ext.web.client.WebClient;
+import io.vertx.mutiny.core.buffer.Buffer;
+import io.vertx.mutiny.ext.web.client.HttpRequest;
+import io.vertx.mutiny.ext.web.client.HttpResponse;
+import io.vertx.mutiny.ext.web.client.WebClient;
 import org.kie.api.runtime.process.WorkItem;
 import org.kie.api.runtime.process.WorkItemHandler;
 import org.kie.api.runtime.process.WorkItemManager;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public class RestWorkItemHandler implements WorkItemHandler {
 
@@ -49,8 +45,6 @@ public class RestWorkItemHandler implements WorkItemHandler {
     public static final String PASSWORD = "password";
     public static final String HOST = "host";
     public static final String PORT = "port";
-
-    private static final Logger logger = LoggerFactory.getLogger(RestWorkItemHandler.class);
 
     // package scoped to allow unit test
     static class RestUnaryOperator implements UnaryOperator<Object> {
@@ -96,28 +90,18 @@ public class RestWorkItemHandler implements WorkItemHandler {
         if (user != null && !user.trim().isEmpty() && password != null && !password.trim().isEmpty()) {
             request.basicAuthentication(user, password);
         }
-        // execute request
-        Handler<AsyncResult<HttpResponse<Buffer>>> handler = event -> {
-            if (event.failed()) {
-                logger.error("Rest invocation failed", event.cause());
-                manager.abortWorkItem(workItem.getId());
-            } else if (event.result().statusCode() >= 300) {
-                logger.error("Rest invocation returns invalid code {}", event.result().statusCode());
-                manager.abortWorkItem(workItem.getId());
-            } else {
-                manager.completeWorkItem(workItem.getId(), Collections.singletonMap(RESULT, resultHandler.apply(
-                        inputModel, event.result().bodyAsJsonObject())));
-            }
-        };
+        HttpResponse<Buffer> response;
         if (method == HttpMethod.POST || method == HttpMethod.PUT) {
             // if parameters is empty at this stage, assume post content is the whole input model
             // if not, build a map from parameters remaining
             Object body = parameters.isEmpty() ? inputModel : parameters.entrySet().stream().collect(Collectors.toMap(
                     Entry::getKey, e -> resolver.apply(e.getValue())));
-            request.sendJson(body, handler);
+            response = request.sendJsonAndAwait(body);
         } else {
-            request.send(handler);
+            response = request.sendAndAwait();
         }
+        manager.completeWorkItem(workItem.getId(), Collections.singletonMap(RESULT, resultHandler.apply(inputModel,
+                response.bodyAsJsonObject())));
     }
 
     @Override
--- a/kogito-workitems/kogito-rest-workitem/src/test/java/org/kogito/workitem/rest/RestTaskHandlerTest.java
+++ b/kogito-workitems/kogito-rest-workitem/src/test/java/org/kogito/workitem/rest/RestTaskHandlerTest.java
@@ -20,28 +20,23 @@ import java.util.Map;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
-import io.vertx.core.AsyncResult;
-import io.vertx.core.Handler;
-import io.vertx.core.buffer.Buffer;
 import io.vertx.core.http.HttpMethod;
 import io.vertx.core.json.JsonObject;
-import io.vertx.ext.web.client.HttpRequest;
-import io.vertx.ext.web.client.HttpResponse;
-import io.vertx.ext.web.client.WebClient;
+import io.vertx.mutiny.core.buffer.Buffer;
+import io.vertx.mutiny.ext.web.client.HttpRequest;
+import io.vertx.mutiny.ext.web.client.HttpResponse;
+import io.vertx.mutiny.ext.web.client.WebClient;
 import org.junit.jupiter.api.Test;
 import org.kie.api.runtime.process.WorkItem;
 import org.kie.api.runtime.process.WorkItemManager;
 import org.kogito.workitem.rest.jsonpath.functions.JSonPathResultHandler;
 import org.kogito.workitem.rest.jsonpath.functions.JsonPathResolver;
 import org.mockito.ArgumentCaptor;
-import org.mockito.stubbing.Answer;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyString;
-import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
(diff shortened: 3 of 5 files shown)


kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
```diff
@@ -71,7 +65,7 @@ public Object apply(Object value) {
     private WebClient client;
 
     public RestWorkItemHandler(WebClient client) {
-        this.client = client;
+          this.client = client;
```
ricardozanini, line 68: I think there's an extra space here :sweat_smile:

kogito-workitems/kogito-rest-workitem/src/test/java/org/kogito/workitem/rest/RestTaskHandlerTest.java
```diff
@@ -154,39 +142,4 @@ public void testGetRestTaskHandler() {
         assertTrue(result instanceof ObjectNode);
         assertEquals(1, ((ObjectNode) result).get("num").asInt());
     }
-
-    @SuppressWarnings("unchecked")
-    @Test
-    public void testGetRestTaskHandlerFail() {
```
ricardozanini, line 160: Isn't necessary to have a "fail" test case anymore?

kogito-build-parent/pom.xml
```diff
@@ -76,7 +76,8 @@
     <version.versions.plugin>2.1</version.versions.plugin>
     <version.jayway.jsonpath>2.4.0</version.jayway.jsonpath>
     <version.vertx.web.client>3.9.4</version.vertx.web.client>
```
ricardozanini, line 78: Do we still need this one?

ricardozanini (commented):
LGTM, only minor comments

kogito-workitems/kogito-rest-workitem/src/test/java/org/kogito/workitem/rest/RestTaskHandlerTest.java
fjtirado: Not really, the previous one was to ensure that when executing asyncrhonous in case of failure the WIH performs abortworkitem call

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
fjtirado: done

kogito-build-parent/pom.xml
fjtirado: Im not sure, so I prefer to leave since it is only a managed dependency, if no one use it, it wont be included anymore

ricardozanini: I just asked because it's getting difficult to manage all these dependencies when we rollout Quarkus version. @evacchi is there an easy way to know if a version is being used by other modules? I would run `mvn clean package` from the root path to find that out. :stuck_out_tongue:

ricardozanini (approved):

kogito-build-parent/pom.xml
fjtirado: managed dependency removed (crossing fingers)

fjtirado:
jenkins retest this (github was down, I repeat git hub was down AHHHHH!!!!)

fjtirado:
jenkins test this

ricardozanini:
@fjtirado see: https://kie.zulipchat.com/#narrow/stream/236603-kogito-ci/topic/Failed.20to.20connect.20to.20GitHub

ricardozanini:
Right now the example is failing to compile the generated code with:

```log
Caused by: javax.enterprise.inject.spi.DeploymentException: 
javax.enterprise.inject.UnsatisfiedResolutionException: Unsatisfied dependency for type io.vertx.mutiny.core.Vertx and qualifiers [@Default]
	- java member: org.kie.kogito.handlers.fahrenheit_to_celsiusRestWorkItemHandler#<init>()
	- declared on CLASS bean [types=[org.kogito.workitem.rest.RestWorkItemHandler, org.kie.kogito.handlers.fahrenheit_to_celsiusRestWorkItemHandler, java.lang.Object, org.kie.api.runtime.process.WorkItemHandler], qualifiers=[@Default, @Any], target=org.kie.kogito.handlers.fahrenheit_to_celsiusRestWorkItemHandler]
Caused by: javax.enterprise.inject.UnsatisfiedResolutionException: 
Unsatisfied dependency for type io.vertx.mutiny.core.Vertx and qualifiers [@Default]
	- java member: org.kie.kogito.handlers.fahrenheit_to_celsiusRestWorkItemHandler#<init>()
	- declared on CLASS bean [types=[org.kogito.workitem.rest.RestWorkItemHandler, org.kie.kogito.handlers.fahrenheit_to_celsiusRestWorkItemHandler, java.lang.Object, org.kie.api.runtime.process.WorkItemHandler], qualifiers=[@Default, @Any], target=org.kie.kogito.handlers.fahrenheit_to_celsiusRestWorkItemHandler]
```

I'm investigating:

```shell
# adding mutiny-vertx-core manually

mvn clean dependency:tree | grep vertx
[INFO] |  |  +- io.quarkus.http:quarkus-http-vertx-backend:jar:3.0.16.Final:compile
[INFO] |  \- io.smallrye.reactive:smallrye-mutiny-vertx-web-client:jar:1.2.1:compile
[INFO] |     +- io.vertx:vertx-web-client:jar:3.9.4:compile
[INFO] |     \- io.vertx:vertx-web-common:jar:3.9.4:compile
[INFO] |  +- io.quarkus:quarkus-vertx-http:jar:1.10.3.Final:compile
[INFO] |  |  +- io.quarkus:quarkus-vertx-core:jar:1.10.3.Final:compile
[INFO] |  |  \- io.vertx:vertx-web:jar:3.9.4:compile
[INFO] |  |     +- io.vertx:vertx-auth-common:jar:3.9.4:compile
[INFO] |  |     \- io.vertx:vertx-bridge-common:jar:3.9.4:compile
[INFO] +- io.smallrye.reactive:smallrye-mutiny-vertx-core:jar:1.2.2:compile
[INFO] |  +- io.smallrye.reactive:smallrye-mutiny-vertx-runtime:jar:1.2.2:compile
[INFO] |  +- io.vertx:vertx-core:jar:3.9.4:compile
[INFO] |  \- io.smallrye.reactive:vertx-mutiny-generator:jar:1.2.2:compile
[INFO] |     \- io.vertx:vertx-codegen:jar:3.9.4:compile
```

fjtirado:
jenkins test this

ricardozanini:
@fjtirado I had to manually add `quarkus-vertx` to my project for the CDI to inject the `Vertx` bean to the generated code:

```java
package org.kie.kogito.handlers;

import io.vertx.mutiny.ext.web.client.WebClient;
import io.vertx.mutiny.core.Vertx;
import org.kogito.workitem.rest.RestWorkItemHandler;

@javax.enterprise.context.ApplicationScoped()
public class fahrenheit_to_celsiusRestWorkItemHandler extends RestWorkItemHandler {

    public fahrenheit_to_celsiusRestWorkItemHandler() {
        this(Vertx.vertx());
    }

    @javax.inject.Inject()
    public fahrenheit_to_celsiusRestWorkItemHandler(Vertx vertx) {
        super(WebClient.create(vertx));
    }
}
```

Is this expected? I think we could take another approach to not make users to add this dependency manually to their project. Or at least we could make `quarkus-vertx` be dependant of `kogito-quarkus`. I haven't tested with SB yet, but I believe we will have the same problem.

What about just using Apache HTTPClient that is already included in our context maybe? cc @danielezonca @cristianonicolai

ricardozanini:
@fjtirado @cristianonicolai @danielezonca @tsurdilo 
It's working great with the example I put together, but having to add two deps in my project, from a UX perspective might not be that cool:

```xml
<dependency>
      <artifactId>kogito-rest-workitem</artifactId>
    </dependency>
    <dependency>
      <groupId>io.quarkus</groupId>
      <artifactId>quarkus-vertx</artifactId>
    </dependency>
```

We could either:

1. Have a `workitem` for each runtime that brings the right deps or;
2. Add Vertx to the `kofito-quarkus|sb` as a direct dependency or;
3. Use another HTTP Client that won't depend on CDI or;
4. Do nothing, users will add the deps to their pom manually. :D

But I agree that this can be handled in another JIRA. Just let me know which approach you found better.

fjtirado:
@ricardozanini yes, this is expected. It was also done for serverless-worflow-function-quarkus

ricardozanini:
Hi @fjtirado! Thanks for your reply, but about this:

>  So I think this is way beyond this PR, which purpose was to revert (I first implement it a sync, then asyns as per comment, now sync again) to sync mode.

I'd like to highlight this quote from my last comment:

> But I agree that this can be handled in another JIRA. Just let me know which approach you found better.

This is just a concern from UX point of view. Don't worry, the PR is good and fulfills its purpose, thanks.

ricardozanini:
@tsurdilo @cristianonicolai can you guys please take a look?

cristianonicolai (approved):

kogito-build-parent/pom.xml
```diff
@@ -385,13 +385,11 @@
...
         <version>${version.jayway.jsonpath}</version>
       </dependency>
-
       <dependency>
-        <groupId>io.vertx</groupId>
-        <artifactId>vertx-web-client</artifactId>
-        <version>${version.vertx.web.client}</version>
+        <groupId>io.smallrye.reactive</groupId>
+        <artifactId>smallrye-mutiny-vertx-web-client</artifactId>
+        <version>${version.mutiny-vertx-web-client}</version>
```
mareknovotny, line 391: i saw this as increased a trend, please @ricardozanini use our BSIG and jboss syntax for version properties
version.groupId

ricardozanini: @fjtirado can you please change the version property as requested?

fjtirado: @mareknovotny this implies that all artifacts for the same group will have the same version dependency, Im not sure thats expected in this particular case. 
since I was unsure, I used artifactId, trying locally right with already defined version to see if it works

fjtirado: @mareknovotny @ricardozanini 
As I suspected, that change wont work
Failed to execute goal on project kogito-rest-workitem: Could not resolve dependencies for project org.kie.kogito:kogito-rest-workitem:jar:2.0.0-SNAPSHOT: Could not find artifact io.smallrye.reactive:smallrye-mutiny-vertx-web-client:jar:2.5.0 in jboss-public-repository-group (https://repository.jboss.org/nexus/content/groups/public/) -> [Help 1]

mareknovotny: yes, exactly @ricardozanini.
I wasn't asking to change existing properties as that is a bigger task right now and it should be done in close future by one person to be more effective. My comment was primarily to new property addition as I noticed it and raising the awareness of bad pattern ;)

tsurdilo:
I don't see a problem with providing 2 handlers for rest, one sync the other async. Maybe we can do that in future.

tsurdilo (approved):

kogito-build-parent/pom.xml
fjtirado: @mareknovotny 
I was aware of the pattern, but in this particular case this pattern cannot be used because there are different version for different artifact below the same group. 
Now Ricardo is asking, in such case, what pattern we should use, I think version.<artifactId> is enough, Ricardo propose version.<groupid>.<artifactId>, which is too long in my opinion.
Which one should we use, as a general approach for such a legit case?

mareknovotny: for the reference at  https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/pom.xml#L51-L55

> CONVENTIONS:
>       - A version property must be specified in the format "version.{groupId}", optionally with a suffix to make it unique.
>       - Version properties must be sorted alphabetically (other form of sorting were found to be unclear and ambiguous).

I created  https://issues.redhat.com/browse/KOGITO-4064

ricardozanini: @mareknovotny so what do you propose? `version.io.smallrye.reactive.mutiny-vertx` maybe? https://search.maven.org/search?q=mutiny-vertx

All of them seem to follow the same release train.

ricardozanini:
> I don't see a problem with providing 2 handlers for rest, one sync the other async. Maybe we can do that in future.

Agreed. We can read the `IsAsync` property from ServiceTask, right? I'll open a JIRA to follow this up.

kogito-build-parent/pom.xml
fjtirado: @mareknovotny 
Ok, then, in my  opinion. the suffix to make it unique should be <artifact_id> not a random one. 
Are we agree on that?

fjtirado:
> Agreed. We can read the `IsAsync` property from ServiceTask, right? I'll open a JIRA to follow this up.

But if we do that, in my opinion, this should be generic and not depending on WIH, which should be alway sync (to allow aync handling on service task that invoked the sync WIH depending on that property)

ricardozanini:
> > Agreed. We can read the `IsAsync` property from ServiceTask, right? I'll open a JIRA to follow this up.
> 
> But if we do that, in my opinion, this should be generic (meaning common code) and not depending on WIH implementation, which should be always sync (to allow async handling on service task that invoked the sync WIH depending on that property)

We can bring this topic to the next meeting, next year. I'll hold this JIRA until there.

kogito-build-parent/pom.xml
ricardozanini: It's not random, it's the prefix used by the project. So if one would add let's say, `io.smallrye.reactive:smallrye-mutiny-vertx-amqp-client`, they could use the same property...

fjtirado:
> > > Agreed. We can read the `IsAsync` property from ServiceTask, right? I'll open a JIRA to follow this up.
> > 
> > 
> > But if we do that, in my opinion, this should be generic (meaning common code) and not depending on WIH implementation, which should be always sync (to allow async handling on service task that invoked the sync WIH depending on that property)
> 
> We can bring this topic to the next meeting, next year. I'll hold this JIRA until there.

@ricardozanini you should merge this JIRA, because this will be a matter of a different one. Right now, you need sync WIH. So please merge that one and open a new one to discuss a common approach for handling WIH asynchornously based on "IsAsync" property

kogito-build-parent/pom.xml
fjtirado: @Ricardo I think we are saying the same thing. If group id needs to be there and is not enough, the suffix should be artifcat id (Im just pointing out to Marek that the convention he linked is lacking a good definition of the suffix to be used in that case)
I updated the PR to use version.groupId.artifactId

ricardozanini:
> > > > Agreed. We can read the `IsAsync` property from ServiceTask, right? I'll open a JIRA to follow this up.
> > > 
> > > 
> > > But if we do that, in my opinion, this should be generic (meaning common code) and not depending on WIH implementation, which should be always sync (to allow async handling on service task that invoked the sync WIH depending on that property)
> > 
> > 
> > We can bring this topic to the next meeting, next year. I'll hold this JIRA until there.
> 
> @ricardozanini I think you can merge this PR now, because the async question will be a matter of a broader discussion. Right now, you need sync WIH. So I would merge that one and open another JIRA to discuss a common approach for handling any WIH (not just REST) asynchronously based on "IsAsync" property

I did not say that I wasn't going to merge this PR. I'm waiting for @mareknovotny to approve the property version change.

kogito-build-parent/pom.xml
ricardozanini: @mareknovotny ^^

tsurdilo:
@ricardozanini I would not tie the workflow control flow logic with the WIH impl, but rather give users a choice to pick which one they would like to use. You could have one rest handler and provide a parameter (fire and forget vs wait for response), that would be fine as well. 
IMHO I don't really understand all the discussions and "issues" with this rather simple rest workitem. We have close to 100 workitem handlers here https://github.com/kiegroup/jbpm-work-items and never had such issues before :)

fjtirado:
> @ricardozanini I would not tie the workflow control flow logic with the WIH impl, but rather give users a choice to pick which one they would like to use. You could have one rest handler and provide a parameter (fire and forget vs wait for response), that would be fine as well.
> IMHO I don't really understand all the discussions and "issues" with this rather simple rest workitem. We have close to 100 workitem handlers here https://github.com/kiegroup/jbpm-work-items and never had such issues before :)

Yes, if finally async depends on WIH implementation itself, the vertx library that is used now (not the previous one, that only support async handling) also easily allows to take a fire and forget approach (after, as you indicating, checking the proper work item param), but I think that the "isAsync" generic property is worth to explore path, because it might be useful for other WIH.

tsurdilo:
@fjtirado tying control flow logic and wih code is not really a good idea no matter the wih. Maybe we could look at something similar to jBPM executor for provide users with abilities for powerful async executions.

fjtirado:
> @fjtirado tying control flow logic and wih code is not really a good idea no matter the wih. Maybe we could look at something similar to jBPM executor for provide users with abilities for powerful async executions.

@tsurdilo Agreed, thats why I prefer to handle async outside WIH (the reason I implemented WIH async was that I feel, after receiving comments on that regard, it will perform better and have not real impact on flow)

ricardozanini:
Since we are all good, I'm merging. @radtriste needs to start branching and this feature is a must for 1.1.x. 
@fjtirado @tsurdilo we talk more about the other approaches to have also `async` in the next sync meeting (no pun intended). Thank you all for your contribution.

PR #946, opened 2020-12-17
KOGITO-3936: event-driven evaluation of decision models

kostola:
* [Related Jira task](https://issues.redhat.com/browse/KOGITO-3936)
* [Related PR in kogito-apps](https://github.com/kiegroup/kogito-apps/pull/571)
* [Related PR in kogito-examples](https://github.com/kiegroup/kogito-examples/pull/497)

## Goal

Enable evaluation of decision models in an event-driven fashion, so that it can be used as part of an event processing pipeline.

## Implementation

I created a new addon called _event-driven-decisions_ that comes in two flavour: Quarkus and Spring Boot, like the tracing addon.

The developer, to be able to trigger decisions via events, only needs to include the correct version as dependency of his Kogito app and configure the input/output topics. The wiring is done by the Kogito codegen and framework specific CDI.

The execution is triggered upon receiving an event containing the initial context from a specific Kafka input topic. The result is then sent to a Kafka output topic (which may be the same). Both input and output events are formatted as CloudEvents.

As side effect, I created a new _cloudevents-utils_ to share some pre existing CloudEvent-related logic between these new modules and others.


kostola:
jenkins retest this please

kostola:
jenkins retest this please

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/DecisionResponseEvent.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @JsonInclude(NON_EMPTY)
+    private String executionId;
+
+    @JsonProperty(access = READ_ONLY) // temporary fix of deserialization issues in tests
+    @JsonInclude(NON_NULL)
+    private DMNResult result;
+
+    private DecisionResponseEvent() {
+    }
```
r00ta, line 42: I might be wrong, but if you remove `@JsonProperty(access = READ_ONLY)` and create a public default constructor the issue should go away?

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/DecisionRequestEvent.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.Map;
+
+public class DecisionRequestEvent {
+
+    private String modelName;
+    private String modelNamespace;
+    private String decisionServiceName;
+    private Map<String, Object> inputContext;
+
```
r00ta, line 27: During the review I've realized that there is an implicit contract: if `decisionServiceName` is specified, only that decision service is evaluated, otherwise all the services are evaluated. 

I'm wondering if it's better to make it explicit, so that it's more clear to read, consume, and extend in the future. wdyt?

addons/events/event-driven-decisions/event-driven-decisions-common/src/test/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionControllerTest.java
```diff
@@ -0,0 +1,335 @@
...
+            "{\"modelNamespace\": \"aNamespace\"}",
+            "{\"inputContext\": {}}",
+            "{\"modelName\": \"aName\",\"modelNamespace\": \"aNamespace\"}",
+            "{\"modelName\": \"aName\",\"inputContext\": {}}",
+            "{\"modelNamespace\": \"aNamespace\",\"inputContext\": {}}"
+    ).collect(Collectors.toList());
+
+    private static final String DATA_EVENT_NOT_FOUND = "{\"modelName\": \"aName\",\"modelNamespace\": \"aNamespace\",\"inputContext\": {}}";
+
+    private static final String DATA_EVENT_OK_EVALUATE_ALL = "" +
```
r00ta, line 95: I think it would be better to put the files under the `resource` folder and then read them here in the tests.

addons/events/event-driven-decisions/event-driven-decisions-quarkus-addon/pom.xml
```diff
@@ -0,0 +1,38 @@
...
+    <dependencies>
+      <dependency>
+        <groupId>io.quarkus</groupId>
+        <artifactId>quarkus-bom</artifactId>
+        <version>${version.io.quarkus}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
```
r00ta, line 25: Do we need this?

r00ta (commented):
Hi @kostola , cool feature! I have some comments here and there, I did not finish the review 'cause I have a major question around the API and I wanted to know wdyt about it before moving forward

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
```diff
@@ -0,0 +1,177 @@
...
+import org.kie.kogito.dmn.rest.DMNJSONUtils;
+import org.kie.kogito.event.CloudEventEmitter;
+import org.kie.kogito.event.CloudEventReceiver;
+
+import static org.kie.kogito.eventdriven.decision.DecisionRequestType.EVALUATE_ALL;
+import static org.kie.kogito.eventdriven.decision.DecisionRequestType.EVALUATE_DECISION_SERVICE;
+
+public class EventDrivenDecisionController {
+
+    private static final String VALID_REQUEST_EVENT_TYPE = DecisionRequestEvent.class.getName();
```
danielezonca, line 43: What about?
```suggestion
    private static final String VALID_REQUEST_EVENT_TYPE = DecisionRequestEvent.class.getSimpleName();
```
I think it is enough without adding the package

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
```diff
@@ -0,0 +1,177 @@
...
+        setup();
+    }
+
+    protected void setup() {
+        eventReceiver.subscribe(this::handleEvent);
+    }
+
+    void handleEvent(String event) {
+        CloudEventUtils.decode(event)
+                .filter(e -> VALID_REQUEST_EVENT_TYPE.equals(e.getType()))
```
danielezonca, line 75: @evacchi @ricardozanini 
Here we are using event type to decide if accept of skip the message. Do we agree in general?

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
```diff
@@ -0,0 +1,177 @@
...
+        }
+        if (ctx.requestType == EVALUATE_DECISION_SERVICE) {
+            return CloudEventUtils.buildDecisionSource(config.getServiceUrl(), ctx.request.getModelName(), ctx.request.getDecisionServiceName());
+        }
+        return CloudEventUtils.buildDecisionSource(config.getServiceUrl());
+    }
+
+    private String buildResponseCloudEventSubject(EvaluationContext ctx) {
+        try {
+            return MAPPER.writeValueAsString(
```
danielezonca, line 154: To be honest I have mixed feeling to use a json to encode the subject of the message. As far as I remember this is used to correlate input event to output event. Am I right?

@evacchi @ricardozanini 
Wdyt? Do you like to have a json string as subject?

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
```diff
@@ -0,0 +1,177 @@
...
+                    new DecisionResponseCloudEventSubject(
+                            ctx.requestCloudEvent.getId(),
+                            ctx.requestCloudEvent.getSource().toString()
+                    )
+            );
+        } catch (JsonProcessingException e) {
+            return null;
```
danielezonca, line 161: What about at least log the error?

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
```diff
@@ -0,0 +1,177 @@
...
+    private DecisionModels decisionModels;
+    private ConfigBean config;
+    private CloudEventEmitter eventEmitter;
+    private CloudEventReceiver eventReceiver;
+
+    protected EventDrivenDecisionController() {
+    }
+
+    protected EventDrivenDecisionController(Application application, ConfigBean config, CloudEventEmitter eventEmitter, CloudEventReceiver eventReceiver) {
+        this.decisionModels = application.get(DecisionModels.class);
```
danielezonca, line 55: What about directly inject `DecisionModels` instead of `Application`?

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
```diff
@@ -0,0 +1,177 @@
...
+        this.config = config;
+        this.eventEmitter = eventEmitter;
+        this.eventReceiver = eventReceiver;
+    }
+
+    protected void setup(Application application, ConfigBean config, CloudEventEmitter eventEmitter, CloudEventReceiver eventReceiver) {
+        this.decisionModels = application.get(DecisionModels.class);
```
danielezonca, line 62: Same as above

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
```diff
@@ -0,0 +1,177 @@
...
+
+    private DecisionResponseEvent buildDecisionResponseEventFromResult(DecisionRequestEvent event, DMNResult result) {
+        String executionId = DecisionExecutionIdUtils.get(result.getContext());
+        return new DecisionResponseEvent(
+                executionId,
+                new org.kie.kogito.dmn.rest.DMNResult(event.getModelNamespace(), event.getModelName(), result)
+        );
+    }
+
+    private Optional<CloudEvent> buildResponseCloudEvent(EvaluationContext ctx) {
```
danielezonca, line 136: Why `Optional<CloudEvent>`? After the evaluation we should be sure to emit the event

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
```diff
@@ -0,0 +1,177 @@
...
+                .filter(e -> VALID_REQUEST_EVENT_TYPE.equals(e.getType()))
+                .ifPresent(this::handleRequest);
+    }
+
+    private void handleRequest(CloudEvent event) {
+        CloudEventUtils.decodeData(event, DecisionRequestEvent.class)
+                .map(data -> new EvaluationContext(event, data))
+                .map(this::processRequest)
+                .flatMap(this::buildResponseCloudEvent)
+                .flatMap(CloudEventUtils::encode)
```
danielezonca, line 84: We should not lose messages especially after processing it. What about a log at least?

Btw it should be nice to `ack` the request only if we are able to process it (probably this change requires a different PR). Wdyt?

addons/events/event-driven-decisions/event-driven-decisions-quarkus-addon/src/main/java/org/kie/kogito/eventdriven/decision/QuarkusEventDrivenDecisionController.java
```diff
@@ -0,0 +1,47 @@
...
+import org.kie.kogito.Application;
+import org.kie.kogito.conf.ConfigBean;
+import org.kie.kogito.event.CloudEventEmitter;
+import org.kie.kogito.event.CloudEventReceiver;
+
+@Startup
+public class QuarkusEventDrivenDecisionController extends EventDrivenDecisionController {
+
+    @Inject
+    Application application;
```
danielezonca, line 32: You can inject `DecisionModels` directly

addons/events/event-driven-decisions/event-driven-decisions-quarkus-addon/src/main/java/org/kie/kogito/eventdriven/decision/QuarkusEventDrivenDecisionController.java
```diff
@@ -0,0 +1,47 @@
...
+import javax.annotation.PostConstruct;
+import javax.inject.Inject;
+
+import io.quarkus.runtime.Startup;
+import org.kie.kogito.Application;
+import org.kie.kogito.conf.ConfigBean;
+import org.kie.kogito.event.CloudEventEmitter;
+import org.kie.kogito.event.CloudEventReceiver;
+
+@Startup
```
danielezonca, line 28: As far as I remember we got some problem in the past with `@Startup` bean because of uncompleted kafka initialization and we had to observe `StartupEvent` to solve ([see here](https://github.com/kiegroup/kogito-runtimes/blob/master/addons/tracing/tracing-decision-quarkus-addon/src/main/java/org/kie/kogito/tracing/decision/QuarkusStartupModelPublisher.java#L38))
@r00ta 
Can you please double check if I'm right?

addons/events/event-driven-decisions/event-driven-decisions-springboot-addon/src/main/java/org/kie/kogito/eventdriven/decision/SpringBootEventDrivenDecisionController.java
```diff
@@ -0,0 +1,40 @@
...
+import org.kie.kogito.event.CloudEventEmitter;
+import org.kie.kogito.event.CloudEventReceiver;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+public class SpringBootEventDrivenDecisionController extends EventDrivenDecisionController {
+
+    @Autowired
+    public SpringBootEventDrivenDecisionController(Application application, ConfigBean config, CloudEventEmitter eventEmitter, CloudEventReceiver eventReceiver) {
```
danielezonca, line 32: You can inject `DecisionModels` directly

danielezonca:
@tarilabs 
FYI, this PR is to let kogito receive a cloud event and evaluate a decision.
If you want to have a look the post important class is  `EventDrivenDecisionController` class https://github.com/kiegroup/kogito-runtimes/pull/946/files#diff-33f564fb5874fb7b0283ee8c5b146dd9e4b08f9028f9e97245f03841c3e22790

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
ricardozanini: That's the model we've decided to follow. But I think won't hurt adding an or condition to check for the `event Source` as well.

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
```diff
@@ -0,0 +1,177 @@
...
+                    : EVALUATE_DECISION_SERVICE;
+        }
+        return DecisionRequestType.INVALID;
+    }
+
+    private Optional<DecisionModel> getDecisionModel(DecisionRequestEvent event) {
+        try {
+            return Optional.ofNullable(decisionModels.getDecisionModel(event.getModelNamespace(), event.getModelName()));
+        } catch (IllegalStateException e) {
+            return Optional.empty();
```
jiripetrlik, line 117: Can we log something here?

kogito-test-utils/src/main/java/org/kie/kogito/decision/DecisionTestUtils.java
```diff
@@ -62,55 +53,53 @@ public static DMNModel createDMNModel() {
...
     public static Map<String, Object> getEvaluateAllContextForError() {
-        return new HashMap<String, Object>() {{
-            put("Violation", getViolation("speed", 115, 100));
-        }};
+        HashMap<String, Object> map = new HashMap<>();
+        map.put("Violation", getViolation("speed", 115, 100));
+        return map;
     }
 
     public static Map<String, Object> getEvaluateDecisionServiceContext() {
```
jiripetrlik, line 75: Is it intentional that this method has the same implementation as "getEvaluateAllContextForError"?

jiripetrlik (commented):
I have few minor commets/suggestions.

addons/cloudevents/cloudevents-utils/src/main/java/org/kie/kogito/cloudevents/CloudEventUtils.java
```diff
@@ -70,6 +82,19 @@
         }
     }
 
+    public static <T> Optional<T> decodeData(CloudEvent event, Class<T> dataClass) {
+        return decodeData(event, dataClass, Mapper.mapper());
+    }
+
+    public static <T> Optional<T> decodeData(CloudEvent event, Class<T> dataClass, ObjectMapper mapper) {
```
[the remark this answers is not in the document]
kostola: Good catch. No it was not used, probably a leftover of previous implementations. I will remove it :+1:

addons/cloudevents/cloudevents-utils/src/main/java/org/kie/kogito/cloudevents/CloudEventUtils.java
```diff
@@ -94,6 +119,35 @@
...
+    public static URI buildDecisionSource(String serviceUrl) {
+        return buildDecisionSource(serviceUrl, null, null);
+    }
+
+    public static URI buildDecisionSource(String serviceUrl, String decisionModelName) {
+        return buildDecisionSource(serviceUrl, decisionModelName, null);
+    }
+
+    public static URI buildDecisionSource(String serviceUrl, String decisionModelName, String decisionServiceName) {
+        String modelChunk = Optional.ofNullable(decisionModelName)
```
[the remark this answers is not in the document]
kostola: No they weren't. I added the tests.

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/DecisionResponseEvent.java
kostola: Nope...apparently it doesn't work

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
```diff
@@ -0,0 +1,177 @@
...
+                : getDecisionModel(request)
+                        .map(model -> evaluateRequest(request, type, model))
+                        .map(result -> buildDecisionResponseEventFromResult(request, result))
+                        .orElseGet(() -> new DecisionResponseEvent(DecisionResponseStatus.NOT_FOUND, "Model not found"));
+
+        return ctx;
+    }
+
+    private DecisionRequestType getRequestType(DecisionRequestEvent event) {
+        if (event != null && event.getModelName() != null && event.getModelNamespace() != null && event.getInputContext() != null) {
```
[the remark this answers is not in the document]
kostola: Good idea. Done :+1:

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
kostola: I used `getName` with the whole package to be consistent with `TraceEvent`s which do the same.
I prefer being consistent over saving the length of the package in this specific event but if you prefer otherwise it's fine for me to change it.

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
kostola: @danielezonca I did it this way because, as Ricardo said, it's the current way it works.

@ricardozanini I assumed that any source can send events to trigger the executions as long as they're formatted correctly. If we want to filter by source we must discuss a logic to distinguish valid sources from invalid ones. Are there already some conventions about this?

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
kostola: Good idea. Done :+1:

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
kostola: Good idea. Done :+1:

addons/events/event-driven-decisions/event-driven-decisions-quarkus-addon/src/main/java/org/kie/kogito/eventdriven/decision/QuarkusEventDrivenDecisionController.java
kostola: Good idea. Done :+1:

addons/events/event-driven-decisions/event-driven-decisions-springboot-addon/src/main/java/org/kie/kogito/eventdriven/decision/SpringBootEventDrivenDecisionController.java
kostola: Good idea. Done :+1:

addons/cloudevents/cloudevents-utils/src/main/java/org/kie/kogito/cloudevents/CloudEventUtils.java
```diff
@@ -102,8 +156,13 @@
         public static ObjectMapper mapper() {
             return OBJECT_MAPPER;
         }
+
+        private Mapper() {
+            throw new IllegalStateException("Utility class");
```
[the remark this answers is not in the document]
kostola: Done :+1:

addons/cloudevents/cloudevents-utils/src/main/java/org/kie/kogito/cloudevents/CloudEventUtils.java
```diff
@@ -102,8 +156,13 @@
...
+        }
     }
 
     private CloudEventUtils() {
+        throw new IllegalStateException("Utility class");
```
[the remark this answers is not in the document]
kostola: Done :+1:

kogito-test-utils/src/main/java/org/kie/kogito/decision/DecisionTestUtils.java
kostola: Yes because they're used in two different test cases and I'd like to let anyone who reads this code to clearly understand which input triggers the specific behavior. What do you think?

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/DecisionRequestEvent.java
kostola: Good point. What's your idea for making it explicit?

addons/events/event-driven-decisions/event-driven-decisions-common/src/test/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionControllerTest.java
kostola: This is the good old question :smile:
I'd like to leave as it is at the moment because in another PR I got asked exactly the opposite :sweat_smile:

kogito-test-utils/pom.xml
```diff
@@ -36,6 +36,11 @@
   </dependencyManagement>
 
   <dependencies>
+    <dependency>
+      <groupId>org.kie.kogito</groupId>
+      <artifactId>kogito-dmn</artifactId>
```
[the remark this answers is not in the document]
kostola: No. The standard scope here is required because some classes from `kogito-dmn` are used in the main code of this module, which is then intended to be included with `test` scope in other modules.

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
ruromero: IMO the subject should be the one received in the request event (if any). The subject can be understood as a correlation key and is usually generated by the producer.
Subjects are attributes commonly used to create subscriptions or correlate multiple events.

addons/events/event-driven-decisions/event-driven-decisions-quarkus-addon/pom.xml
kostola: Honestly I always found it in all the other Quarkus addons I looked at, so I blindly assumed it was needed.
@danielezonca is this necessary or should we remove this?

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/DecisionRequestType.java
```diff
@@ -0,0 +1,23 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.kie.kogito.eventdriven.decision;
+
+public enum DecisionRequestType {
+    INVALID,
```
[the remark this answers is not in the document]
kostola: I refactored this method so that:
1. the `DecisionRequestType` enum is now a private inner class of `EventDrivenDecisionController`, since it's only used inside there (named `RequestType`).
2. the `INVALID` value is now replaced by a local boolean variable, since, as you correctly pointed out, semantically it is not a real type of request.

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
kostola: >What about a log at least

The `CloudEventUtils::encode` method already logs an error if it can't encode the data. Should we double log in case the logic in that class is changed?

>Btw it should be nice to ack the request only if we are able to process it

Yes I totally agree. Imho we should open a new PR to harden the error handling.

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
kostola: At this point the response CloudEvent should be built always except for unexpected errors. In this case, with the same logic described in the first part of [this comment](https://github.com/kiegroup/kogito-runtimes/pull/946#discussion_r556443892), the `CloudEventUtils::build` method would log the message. Shall we add a dedicated log here?

addons/events/event-driven-decisions/event-driven-decisions-quarkus-addon/pom.xml
danielezonca: Well, this module has no additional/explicit dependency on `quarkus` artifacts so I think it is better to remove this bom import. It is probably already inheriting it from one of the dependencies

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
ricardozanini: [Sources](https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#source-1) are `URI` attribute, so I think we can start from there. But FWIW, I believe it's better to just open a follow up JIRA to filter by sources and have some discussions around it and how we are going to implement.

From the Knative eventing perspective, we can always use [filters](https://knative.dev/docs/eventing/triggers/#trigger-filtering) to get these events into our platform, so we can cover some use cases until there.

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/DecisionResponseEvent.java
```diff
@@ -0,0 +1,70 @@
...
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.kie.kogito.dmn.rest.DMNResult;
+
+import static com.fasterxml.jackson.annotation.JsonInclude.Include.NON_EMPTY;
+import static com.fasterxml.jackson.annotation.JsonInclude.Include.NON_NULL;
+import static com.fasterxml.jackson.annotation.JsonProperty.Access.READ_ONLY;
+
+public class DecisionResponseEvent {
```
ruromero, line 27: Wouldn't it be better for management, tracing and monitoring purposes that the DMNResult is the object returned and any other attributes are added as CloudEvent metadata?
In case of error, the error message should be the event content itself as I assume the result will be empty.

ruromero:
After using the solution I would like to show a request/response for the TrafficViolation example with some comments and concerns to try to make it more usable:

**Request event:**
```
{
   "specversion":"1.0",
   "id":"21627e26-31eb-43e7-8343-92a696fd96b1",
   "source":"",
   "type":"org.kie.kogito.eventdriven.decision.DecisionRequestEvent",
   "data":{
      "modelName":"TrafficViolation",
      "modelNamespace":"https://github.com/kiegroup/drools/kie-dmn/_A4BCA8B8-CF08-433F-93B2-A2598F19ECFF",
      "decisionServiceName": "Fine",
      "inputContext":{
         "Driver":{
            "Points":2
         },
         "Violation":{
            "Type":"speed",
            "Actual Speed":120,
            "Speed Limit":100
         }
      }
   }
}
```
In this request I would like the data to be only the decision input data (i.e. the inputContext content) and any other attribute be sent as extension attributes, similarly to what is being done for processes.
The type should be the simple name instead of the full class name for simplicity.
Please, consider using extension attributes for model, namespace and decisionName:
```
kogitodmnmodelname: "TrafficViolation"
kogitodmnmodelns: "https://github.com/kiegroup/drools/kie-dmn/_A4BCA8B8-CF08-433F-93B2-A2598F19ECFF"
kogitodmndecision: "Fine"
```

**Response event:**
```
{
   "specversion":"1.0",
   "id":"09a0e9fa-da06-4262-95ce-da76f1c7e2f0",
   "source":"TrafficViolation",
   "type":"org.kie.kogito.eventdriven.decision.DecisionResponseEvent",
   "subject":"{\"id\":\"21627e26-31eb-43e7-8343-92a696fd96b1\",\"source\":\"\"}",
   "data":{
      "status":"OK",
      "result":{
         "namespace":"https://github.com/kiegroup/drools/kie-dmn/_A4BCA8B8-CF08-433F-93B2-A2598F19ECFF",
         "modelName":"TrafficViolation",
         "dmnContext":{
            "Violation":{
               "Type":"speed",
               "Speed Limit":100,
               "Actual Speed":120,
               "Code":null,
               "Date":null
            },
            "Driver":{
               "Points":2,
               "State":null,
               "City":null,
               "Age":null,
               "Name":null
            },
            "Fine":{
               "Points":3,
               "Amount":500
            },
            "Should the driver be suspended?":"No"
         },
         "messages":[
            
         ],
         "decisionResults":[
            {
               "decisionId":"_4055D956-1C47-479C-B3F4-BAEB61F1C929",
               "decisionName":"Fine",
               "result":{
                  "Points":3,
                  "Amount":500
               },
               "messages":[
                  
               ],
               "evaluationStatus":"SUCCEEDED"
            },
            {
               "decisionId":"_8A408366-D8E9-4626-ABF3-5F69AA01F880",
               "decisionName":"Should the driver be suspended?",
               "result":"No",
               "messages":[
                  
               ],
               "evaluationStatus":"SUCCEEDED"
            }
         ]
      }
   }
}
```

IMO the response event is too verbose and depending on the use case can seem to be unnecessarily heavy. I suggest the following changes:

* Type: Use the simple class name. Consider using a different type for failures, if not use an extension attribute
* Subject: (Already in one of my comments). Keep subject from request (if any)
* Data: The data content should be the decision result or the error message if it failed.
* Decision results: I might lack some understanding here. If multiple decisions are processed for one request, the consumer should only worry about the "final" decision. The rest is only for understanding the result, i.e. debugging or explainability.
* Status seems to always be OK. I'd remove it.
* Context: I don't think it's necessary, it is useless information for processing a decision result. Same as intermediate decision results.
* Messages: When the decision fails, contains the error message and the decision results are empty. I think this should be also the data content and have the CloudEvent type to distinguish whether this is a success or a failure.
* Intermediate decision results and context should be sent to the data index for tracing but not be part of the result type.

Maybe these are breaking changes or maybe some are not justified and I'm lacking background or context. Feel free to split it in multiple PR if needed.

danielezonca:
Thanks @ruromero for the analysis and the comment :+1: 

> * Type: Use the simple class name. Consider using a different type for failures, if not use an extension attribute

I agree, I think we should avoid FQCN because it makes the event less reusable. I think we should verify and apply the same logic to all the event we produce. @kostola what about already implement the change here and then in another ticket align other event usages?

> * Subject: (Already in one of my comments). Keep subject from request (if any)

I agree to preserve the subject. At the same time we need a way to link the output with the input so what about two extension attributes like `input-event-id` and `input-event-source`. Wdyt?

> * Data: The data content should be the decision result or the error message if it failed.

Well in DMN spec this is not usually possible because there is a status for each decision and it is completely fine to have some failed decisions, some skipped and some succeeded

> * Status seems to always be OK. I'd remove it.

Status OK is probably an aggregation of the decision statuses: if all of them are OK -> global status OK. @kostola please correct me if I'm wrong

> * Context: I don't think it's necessary, it is useless information for processing a decision result. Same as intermediate decision results.
> * Messages: When the decision fails, contains the error message and the decision results are empty. I think this should be also the data content and have the CloudEvent type to distinguish whether this is a success or a failure.
> * Intermediate decision results and context should be sent to the data index for tracing but not be part of the result type.

I think it makes sense to follow the same approach we already have with REST: two different event types (like we have two different endpoints) to obtain only the data (DMNContext with input + output) or the whole DMNResult with messages and similar.
In this way output is a map and user can get the decision out of it. We can also add an optional extension attribute to exclude input from result :thinking: 
@kostola @tarilabs Wdyt?

kostola:
@ruromero @danielezonca some considerations

>Type: Use the simple class name. Consider using a different type for failures, if not use an extension attribute

>what about already implement the change here and then in another ticket align other event usages?

Perfectly fine for me :+1: 

>Subject: (Already in one of my comments). Keep subject from request (if any)

Fine for me as well :+1: 

>Status seems to always be OK. I'd remove it. (@ruromero)

>Status OK is probably an aggregation of the decision statuses (@danielezonca)

>I think it makes sense to follow the same approach we already have with REST: two different event types

@danielezonca what do you mean here exactly? Send two different events (one with only the output context and one with the full result) for every evaluation?

kostola:
Sorry I clicked the "Close with comment" button (which is *very* close to the "Comment" button, nice UI Github :unamused:)  by mistake...

tarilabs:
Given anyway a CE needs to be unwrapped, and given the "domain oriented" approach have shown limitations even in some use cases of REST, wouldn't make sense to just have the complete answer back? (The one with DMNResult) 
I would exercise this opportunity to revise one last time its serialisation "naming conventions" as previously required by Lead, since it will be even more difficult to change it later

danielezonca:
> @danielezonca what do you mean here exactly? Send two different events (one with only the output context and one with the full result) for every evaluation?

I mean having two different input event types and based on that produce the output with only data or with the whole dmnresult. Something like `DecisionRequestEvent` that produce `DecisionResponseEvent` with only DMNContext as result and `DMNRequestEvent` that produce `DMNResponseEvent` with the whole DMNResult (so also messages etc).
As alternative we can use an extension attribute like `dmnresult` `true`/`false` instead of having two different event types

In addition we can add an extension attribute like `filter-output` (feel free to use a better name :) ) to let the user ask to remove the input from DMNContext

In summary having the flexibility to have:
- full DMNResult or only DMNContext
- input data in the output or not

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/DecisionRequestEvent.java
r00ta: Something like
```java
public class DecisionRequestEvent {

    private String modelName;
    private String modelNamespace;
    private boolean evaluateAll;
    private String decisionServiceName;
    private Map<String, Object> inputContext;
```
?

addons/events/event-driven-decisions/event-driven-decisions-quarkus-addon/src/main/java/org/kie/kogito/eventdriven/decision/QuarkusEventDrivenDecisionController.java
r00ta: I had a look and it should be fine. Btw the issue I opened to Quarkus has been fixed in SmallRye Reactive Messaging 2.6.0  https://github.com/quarkusio/quarkus/issues/12820 (I didn't check if we have the fix in the version we are using, but we should not have any issue here)

kostola:
@danielezonca @ruromero I refactored the structure of input/output events following your suggestions. Here are the main points:

1. Input event type is now `DecisionRequest`.
2. Two different output event types are now used: `DecisionResponse` if the request event successfully triggers an evaluation of the model, `DecisionResponseError` otherwise.
3. Only the DMN context is returned in the `data` field of `DecisionResponse` events, as it happens when triggering the evaluation via the default REST endpoint.
4. There are two errors at the moment: `BAD_REQUEST` and `MODEL_NOT_FOUND`.
5. Some extension attributes are used to specify the behavior: `kogitodmnmodelname` + `kogitodmnmodelnamespace` (required) to identify the DMN model to be evaluated and the optional `kogitodmnevaldecision` to evaluate a specific decision service (which is more explicit imho as suggested by @r00ta).
6. `kogitoevaluationid` extension attribute is added in the output event if the evaluation result contains an execution ID.
7. The subject of the output event is the same as the corresponding input and it's now responsibility of the caller how to use it (e.g. custom correlation ID).

### Example: evaluate all (with execution ID)

#### Input

```json
{
  "specversion": "1.0",
  "id": "a89b61a2-5644-487a-8a86-144855c5dce8",
  "source": "SomeEventSource",
  "type": "DecisionRequest",
  "subject": "TheSubject",
  "kogitodmnmodelname": "Traffic Violation",
  "kogitodmnmodelnamespace": "https://github.com/kiegroup/drools/kie-dmn/_A4BCA8B8-CF08-433F-93B2-A2598F19ECFF",
  "data": {
    "Driver": {
      "Age": 25,
      "Points": 13
    },
    "Violation": {
      "Type": "speed",
      "Actual Speed": 115,
      "Speed Limit": 100
    }
  }
}
```

#### Output

```json
{
  "specversion": "1.0",
  "id": "f8bf840b-9f6b-4d26-bc90-38c1fffc148e",
  "source": "Traffic+Violation",
  "type": "DecisionResponse",
  "subject": "TheSubject",
  "kogitoexecutionid": "11ecbb6f-fb25-4597-88c8-ac7976efe078",
  "kogitodmnmodelnamespace": "https://github.com/kiegroup/drools/kie-dmn/_A4BCA8B8-CF08-433F-93B2-A2598F19ECFF",
  "kogitodmnmodelname": "Traffic Violation",
  "data": {
    "Violation": {
      "Type": "speed",
      "Speed Limit": 100,
      "Actual Speed": 115,
    },
    "calculateTotalPoints": "function calculateTotalPoints( driver, fine )",
    "Driver": {
      "Points": 13,
      "Age": 25
    },
    "Fine": {
      "Points": 3,
      "Amount": 500
    },
    "Should the driver be suspended?": "No"
  }
}
```

### Example: evaluate decision service (with execution ID)

#### Input

```json
{
  "specversion": "1.0",
  "id": "a89b61a2-5644-487a-8a86-144855c5dce8",
  "source": "SomeEventSource",
  "type": "DecisionRequest",
  "subject": "TheSubject",
  "kogitodmnmodelname": "Traffic Violation",
  "kogitodmnmodelnamespace": "https://github.com/kiegroup/drools/kie-dmn/_A4BCA8B8-CF08-433F-93B2-A2598F19ECFF",
  "kogitodmnevaldecision": "FineService",
  "data": {
    "Violation": {
      "Type": "speed",
      "Actual Speed": 115,
      "Speed Limit": 100
    }
  }
}
```

#### Output

```json
{
  "specversion": "1.0",
  "id": "90fd1bd1-2bf8-4277-a757-85c32cfe8075",
  "source": "Traffic+Violation/FineService",
  "type": "DecisionResponse",
  "subject": "TheSubject",
  "kogitoexecutionid": "11ecbb6f-fb25-4597-88c8-ac7976efe078",
  "kogitodmnmodelnamespace": "https://github.com/kiegroup/drools/kie-dmn/_A4BCA8B8-CF08-433F-93B2-A2598F19ECFF",
  "kogitodmnmodelname": "Traffic Violation",
  "kogitodmnevaldecision": "FineService",
  "data": {
    "Violation": {
      "Type": "speed",
      "Speed Limit": 100,
      "Actual Speed": 115
    },
    "Fine": {
      "Points": 3,
      "Amount": 500
    }
  }
}
```

### Example: MODEL_NOT_FOUND error

#### Input

```json
{
  "specversion": "1.0",
  "id": "a89b61a2-5644-487a-8a86-144855c5dce8",
  "source": "SomeEventSource",
  "type": "DecisionRequest",
  "subject": "TheSubject",
  "kogitodmnmodelname": "aName",
  "kogitodmnmodelnamespace": "aNamespace",
  "data": {}
}
```

#### Output

```json
{
  "specversion": "1.0",
  "id": "115d3b99-28f5-4051-b86b-038c06f10f08",
  "source": "aName",
  "type": "DecisionResponseError",
  "subject": "TheSubject",
  "kogitodmnmodelnamespace": "aNamespace",
  "kogitodmnmodelname": "aName",
  "data": "MODEL_NOT_FOUND"
}
```

### Example: BAD_REQUEST error

Bad request is caused because the `data` field of the input event is `null`.

#### Input

```json
{
  "specversion": "1.0",
  "id": "a89b61a2-5644-487a-8a86-144855c5dce8",
  "source": "SomeEventSource",
  "type": "DecisionRequest",
  "subject": "TheSubject",
  "kogitodmnmodelname": "Traffic Violation",
  "kogitodmnmodelnamespace": "https://github.com/kiegroup/drools/kie-dmn/_A4BCA8B8-CF08-433F-93B2-A2598F19ECFF",
  "data": null
}
```

#### Output

```json
{
  "specversion": "1.0",
  "id": "39cb6883-2f9d-4f3a-bce4-878511c5f9fb",
  "source": "Traffic+Violation",
  "type": "DecisionResponseError",
  "subject": "TheSubject",
  "kogitodmnmodelnamespace": "https://github.com/kiegroup/drools/kie-dmn/_A4BCA8B8-CF08-433F-93B2-A2598F19ECFF",
  "kogitodmnmodelname": "Traffic Violation",
  "data": "BAD_REQUEST"
}
```

addons/events/event-driven-decisions/event-driven-decisions-quarkus-addon/pom.xml
kostola: Removed :+1:

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/DecisionRequestEvent.java
kostola: Resolving this conversation because the new refactoring completely changed this part.

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
kostola: New event types use plain string instead of class full names.

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
kostola: The new version behaves exactly as @ruromero described about event subjects :+1:

addons/events/event-driven-decisions/event-driven-decisions-quarkus-addon/src/main/java/org/kie/kogito/eventdriven/decision/QuarkusEventDrivenDecisionController.java
kostola: Yes Kogito is now using Quarkus 1.11.0.Final, which contains SmallRye Reactive Messaging 2.7.0 :+1:

addons/events/event-driven-decisions/event-driven-decisions-common/src/main/java/org/kie/kogito/eventdriven/decision/DecisionResponseEvent.java
kostola: The new version behaves like you described :+1:

addons/events/event-driven-decisions/event-driven-decisions-common/src/test/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionControllerTest.java
```diff
@@ -0,0 +1,440 @@
...
+                        }
+                    } else {
+                        fail("Can't decode emitted CloudEvent data of: " + emittedCloudEventJson);
+                    }
+                } else {
+                    fail("No Kogito extension in emitted CloudEvent: " + emittedCloudEventJson);
+                }
+            } else {
+                fail("Can't decode emitted CloudEvent");
+            }
```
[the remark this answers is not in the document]
kostola: I did it only because you used the emoticon in the comment :stuck_out_tongue_closed_eyes:

kostola:
jenkins retest this please

jiripetrlik (approved):
@kostola 
Thank you for your changes. It looks good to me now.

kogito-test-utils/src/main/java/org/kie/kogito/decision/DecisionTestUtils.java
kostola: Resolving this after receiving [your approval](https://github.com/kiegroup/kogito-runtimes/pull/946#pullrequestreview-573225653) :+1:

danielezonca:
jenkins retest this please

danielezonca (approved):
Well done :+1:

r00ta (commented):
Just a small comment/question from my side! Thanks :+1:

addons/cloudevents/cloudevents-utils/src/main/java/org/kie/kogito/cloudevents/extension/KogitoExtension.java
```diff
@@ -0,0 +1,190 @@
...
+        readStringExtension(extensions, KOGITO_DMN_MODEL_NAME, this::setDmnModelName);
+        readStringExtension(extensions, KOGITO_DMN_MODEL_NAMESPACE, this::setDmnModelNamespace);
+        readStringExtension(extensions, KOGITO_DMN_EVALUATE_DECISION, this::setDmnEvaluateDecision);
+        readBooleanExtension(extensions, KOGITO_DMN_FULL_RESULT, this::setDmnFullResult);
+        readBooleanExtension(extensions, KOGITO_DMN_FILTERED_CTX, this::setDmnFilteredCtx);
+    }
+
+    private static void readStringExtension(CloudEventExtensions extensions, String key, Consumer<String> consumer) {
+        Optional.ofNullable(extensions.getExtension(key))
+                // there seems to be a bug in the cloudevents sdk so that, when a extension attributes is null,
```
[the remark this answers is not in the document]
kostola: Yes I will do this today :+1:

PR #955, opened 2020-12-23
[KOGITO-4076] Monitoring addon is broken with SpringBoot

danielezonca:
https://issues.redhat.com/browse/KOGITO-4076

Details:
- Fixed `DecisionRestResourceGenerator` Spring template for monitoring codegen
- Made `DecisionCodegenTest` parametric to test all `KogitoBuildContext`s
- Updated monitoring-springboot-addon to only use Spring Web

PRs:
- https://github.com/kiegroup/kogito-runtimes/pull/955
- https://github.com/kiegroup/kogito-examples/pull/493


tarilabs (approved):
if it works for @r00ta works for me too,
but after PTO I would appreciate some cascade about all these refactorings, please.

r00ta (approved):
Thank you very much for this!

jiripetrlik (commented):
Looks good, but it seems we have low code coverage. Can you please for example write some test to cover QuarkusMetricsInterceptor. According to Sonar cloud the filter method is not tested.

danielezonca:
@jiripetrlik 
Coverage improved :)
Let me know if it is fine for you :+1:

jiripetrlik (approved):
@danielezonca 
Thank you for additional tests!

kogito-codegen/src/main/java/org/kie/kogito/codegen/context/JavaKogitoBuildContext.java
```diff
@@ -40,5 +40,10 @@ protected JavaKogitoBuildContextBuilder() {
         public JavaKogitoBuildContext build() {
             return new JavaKogitoBuildContext(packageName, classAvailabilityResolver, targetDirectory, addonsConfig, applicationProperties);
         }
+
+        @Override
+        public String toString() {
+            return "Java";
```
MarianMacik, line 46: This has been probably done due to jUnit parameterized tests output so it is easier to spot which parameter is currently being tested. I haven't seen the exact output, but having just plain `Java` in the output of the Java program might cause some confusion/overlooking. Wouldn't it be better to define it for example `JavaKogitoBuildContextBuilder` or `KogitoBuildContextBuilder - Java` etc?

MarianMacik (approved):
Looks good, just one suggestion.

kogito-codegen/src/main/java/org/kie/kogito/codegen/context/JavaKogitoBuildContext.java
danielezonca: A similar change is done (with a proper context name) in this [other PR](https://github.com/kiegroup/kogito-runtimes/pull/968/files#diff-dae50530c92279e8b434b829e9fac007f77af1d5661446ae5f5dcaaa8a37bedf)

MarianMacik: Probably there it is fine, but if here the purpose is for jUnit parametrization, I would add there also at least info that it is a builder and not the context maybe? As otherwise both the context and its builder will have just `Java`. But up to you, it was just a point to discuss.

danielezonca:
jenkins retest this please

danielezonca:
jenkins retest this please

danielezonca:
jenkins retest this please

PR #969, opened 2021-01-11
[DROOLS-5930] allow an accumulate function to state if it supports re…

mariofusco:
…verse for the specific data to be reversed


mariofusco:
This addition is a trivial forward port of a change in v7 kie public API. Merging.

PR #987, opened 2021-01-19
[KOGITO-3759] Implement plugable transaction management for MongoDB persistence

xieshenzh:
Implement plugable transaction management for MongoDB to support outbox pattern.

Jira: https://issues.redhat.com/browse/KOGITO-3759

### DIFF
--- a/addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/MongoDBProcessInstances.java
+++ b/addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/MongoDBProcessInstances.java
@@ -23,6 +23,7 @@ import java.util.function.Supplier;
 
 import org.bson.Document;
 import org.kie.kogito.Model;
+import org.kie.kogito.mongodb.transaction.MongoDBTransactionManager;
 import org.kie.kogito.process.MutableProcessInstances;
 import org.kie.kogito.process.ProcessInstance;
 import org.kie.kogito.process.ProcessInstanceDuplicatedException;
@@ -33,6 +34,8 @@ import org.kie.kogito.serialization.process.ProcessInstanceMarshallerService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.mongodb.client.ClientSession;
+import com.mongodb.client.FindIterable;
 import com.mongodb.client.MongoClient;
 import com.mongodb.client.MongoCollection;
 import com.mongodb.client.MongoCursor;
@@ -49,11 +52,16 @@ public class MongoDBProcessInstances<T extends Model> implements MutableProcessI
     private org.kie.kogito.process.Process<?> process;
     private ProcessInstanceMarshallerService marshaller;
     private final MongoCollection<Document> collection;
+    private MongoDBTransactionManager transactionManager;
 
-    public MongoDBProcessInstances(MongoClient mongoClient, org.kie.kogito.process.Process<?> process, String dbName) {
+    public MongoDBProcessInstances(MongoClient mongoClient, org.kie.kogito.process.Process<?> process, String dbName, MongoDBTransactionManager transactionManager) {
         this.process = process;
-        collection = getCollection(mongoClient, process.id(), dbName);
-        marshaller = ProcessInstanceMarshallerService.newBuilder().withDefaultObjectMarshallerStrategies().withContextEntries(singletonMap(MarshallerContextName.MARSHALLER_FORMAT, "json")).build();
+        this.collection = getCollection(mongoClient, process.id(), dbName);
+        this.marshaller = ProcessInstanceMarshallerService.newBuilder()
+                .withDefaultObjectMarshallerStrategies()
+                .withContextEntries(singletonMap(MarshallerContextName.MARSHALLER_FORMAT, "json"))
+                .build();
+        this.transactionManager = transactionManager;
     }
 
     @Override
@@ -68,8 +76,11 @@ public class MongoDBProcessInstances<T extends Model> implements MutableProcessI
 
     @Override
     public Collection<ProcessInstance<T>> values(ProcessInstanceReadMode mode) {
+        FindIterable<Document> docs = Optional.ofNullable(transactionManager.getClientSession())
+                .map(collection::find)
+                .orElseGet(collection::find);
         List<ProcessInstance<T>> list = new ArrayList<>();
-        try (MongoCursor<Document> cursor = collection.find().iterator()) {
+        try (MongoCursor<Document> cursor = docs.iterator()) {
             while (cursor.hasNext()) {
                 list.add(unmarshall(cursor.next(), mode));
             }
@@ -93,14 +104,26 @@ public class MongoDBProcessInstances<T extends Model> implements MutableProcessI
     }
 
     protected void updateStorage(String id, ProcessInstance<T> instance, boolean checkDuplicates) {
-        if (isActive(instance)) {
-            Document doc = Document.parse(new String(marshaller.marshallProcessInstance(instance)));
-            if (checkDuplicates) {
-                if (exists(id)) {
-                    throw new ProcessInstanceDuplicatedException(id);
+        if (!isActive(instance)) {
+            reloadProcessInstance(instance, id);
+            return;
+        }
+
+        ClientSession clientSession = transactionManager.getClientSession();
+        Document doc = Document.parse(new String(marshaller.marshallProcessInstance(instance)));
+        if (checkDuplicates) {
+            if (exists(id)) {
+                throw new ProcessInstanceDuplicatedException(id);
+            } else {
+                if (clientSession != null) {
+                    collection.insertOne(clientSession, doc);
                 } else {
                     collection.insertOne(doc);
                 }
+            }
+        } else {
+            if (clientSession != null) {
+                collection.replaceOne(clientSession, Filters.eq(PROCESS_INSTANCE_ID, id), doc);
             } else {
                 collection.replaceOne(Filters.eq(PROCESS_INSTANCE_ID, id), doc);
             }
@@ -109,7 +132,9 @@ public class MongoDBProcessInstances<T extends Model> implements MutableProcessI
     }
 
     private Document find(String id) {
-        return collection.find(Filters.eq(PROCESS_INSTANCE_ID, id)).first();
+        return Optional.ofNullable(transactionManager.getClientSession())
+                .map(r -> collection.find(r, Filters.eq(PROCESS_INSTANCE_ID, id)).first())
+                .orElseGet(() -> collection.find(Filters.eq(PROCESS_INSTANCE_ID, id)).first());
     }
 
     @Override
@@ -119,7 +144,12 @@ public class MongoDBProcessInstances<T extends Model> implements MutableProcessI
 
     @Override
     public void remove(String id) {
-        collection.deleteOne(Filters.eq(PROCESS_INSTANCE_ID, id));
+        ClientSession clientSession = transactionManager.getClientSession();
+        if (clientSession != null) {
+            collection.deleteOne(clientSession, Filters.eq(PROCESS_INSTANCE_ID, id));
+        } else {
+            collection.deleteOne(Filters.eq(PROCESS_INSTANCE_ID, id));
+        }
     }
 
     private void reloadProcessInstance(ProcessInstance<T> instance, String id) {
@@ -136,6 +166,8 @@ public class MongoDBProcessInstances<T extends Model> implements MutableProcessI
 
     @Override
     public Integer size() {
-        return (int) collection.countDocuments();
+        return Optional.ofNullable(transactionManager.getClientSession())
+                .map(r -> (int) collection.countDocuments(r))
+                .orElseGet(() -> (int) collection.countDocuments());
     }
 }
--- a/addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/transaction/MongoDBTransactionManager.java
+++ b/addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/transaction/MongoDBTransactionManager.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.mongodb.transaction;
+
+import org.kie.kogito.uow.events.UnitOfWorkAbortEvent;
+import org.kie.kogito.uow.events.UnitOfWorkEndEvent;
+import org.kie.kogito.uow.events.UnitOfWorkEventListener;
+import org.kie.kogito.uow.events.UnitOfWorkStartEvent;
+
+import com.mongodb.ReadConcern;
+import com.mongodb.ReadPreference;
+import com.mongodb.TransactionOptions;
+import com.mongodb.WriteConcern;
+import com.mongodb.client.ClientSession;
+import com.mongodb.client.MongoClient;
+
+public abstract class MongoDBTransactionManager implements UnitOfWorkEventListener {
+
+    private MongoClient mongoClient;
+
+    private ThreadLocal<ClientSession> clientSessionLocal = new ThreadLocal<>();
+
+    public MongoDBTransactionManager(MongoClient mongoClient) {
+        this.mongoClient = mongoClient;
+    }
+
+    @Override
+    public void onBeforeStartEvent(UnitOfWorkStartEvent event) {
+        if (!enabled()) {
+            return;
+        }
+
+        ClientSession clientSession = mongoClient.startSession();
+        this.clientSessionLocal.set(clientSession);
+        TransactionOptions txnOptions = TransactionOptions.builder()
+                .readPreference(ReadPreference.primary())
+                .readConcern(ReadConcern.MAJORITY)
+                .writeConcern(WriteConcern.MAJORITY)
+                .build();
+        clientSession.startTransaction(txnOptions);
+    }
+
+    @Override
+    public void onAfterEndEvent(UnitOfWorkEndEvent event) {
+        if (!enabled()) {
+            return;
+        }
+
+        try (ClientSession clientSession = this.getClientSession()) {
+            clientSession.commitTransaction();
+        } finally {
+            clientSessionLocal.remove();
+        }
+    }
+
+    @Override
+    public void onAfterAbortEvent(UnitOfWorkAbortEvent event) {
+        if (!enabled()) {
+            return;
+        }
+
+        try (ClientSession clientSession = this.getClientSession()) {
+            clientSession.abortTransaction();
+        } finally {
+            clientSessionLocal.remove();
+        }
+    }
+
+    public ClientSession getClientSession() {
+        return clientSessionLocal.get();
+    }
+
+    public abstract boolean enabled();
+}
--- a/addons/persistence/mongodb-persistence-addon/src/test/java/org/kie/kogito/mongodb/transaction/MongoDBTransactionManagerIT.java
+++ b/addons/persistence/mongodb-persistence-addon/src/test/java/org/kie/kogito/mongodb/transaction/MongoDBTransactionManagerIT.java
@@ -1,0 +1,254 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.mongodb.transaction;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import org.bson.Document;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.kie.kogito.testcontainers.KogitoMongoDBContainer;
+import org.kie.kogito.uow.events.UnitOfWorkAbortEvent;
+import org.kie.kogito.uow.events.UnitOfWorkEndEvent;
+import org.kie.kogito.uow.events.UnitOfWorkStartEvent;
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+import com.mongodb.client.MongoClient;
+import com.mongodb.client.MongoClients;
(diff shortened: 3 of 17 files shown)


api/kogito-api/src/main/java/org/kie/kogito/transaction/TransactionHandler.java
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.transaction;
+
+import java.io.Closeable;
+import java.util.Collection;
+
+public interface TransactionHandler extends Closeable {
```
tiagodolphine, line 21: I think it is missing the abortTransaction, which should be called in case of any failure.

api/kogito-api/src/main/java/org/kie/kogito/transaction/TransactionHandler.java
```diff
@@ -0,0 +1,38 @@
...
+
+    void startTransaction();
+
+    void commitTransaction();
+
+    <R, T> T find(R repository, String id);
```
tiagodolphine, line 27: I'm not sure the find, delete, create, update, values, size methods should be in this interface, they are related to the MutableProcessInstances/ProcessInstances that represent the DAO let's say...

tiagodolphine (commented):
@xieshenzh great work, I have some questions and suggestions, it seems the changes introduced by the PR is much more complex than it could be, from my understanding the `UnitOfWork` represents a Transaction and `UnitOfWorkExecutor` represents the TransactionManager, so in theory, everything that should be executed in a transaction is already set in the code, it is just necessary to add the abstraction you created like the TransactionHandler (with start, commit, abort) into the UnitOfWorkExecutor, an call the proper methods from there...  something like this... I don't see the need to change the whole codebase where we have the ProcessInstances because transaction should not coupled with them, we could reuse the UnitOfWork stuff that we already have, WDYT?

@danielezonca @cristianonicolai @evacchi ^

api/kogito-api/src/main/java/org/kie/kogito/transaction/TransactionHandler.java
xieshenzh: Thanks @tiagodolphine . It makes sense. I'll redo the work.

danielezonca:
> @xieshenzh great work, I have some questions and suggestions, it seems the changes introduced by the PR is much more complex than it could be, from my understanding the `UnitOfWork` represents a Transaction and `UnitOfWorkExecutor` represents the TransactionManager, so in theory, everything that should be executed in a transaction is already set in the code, it is just necessary to add the abstraction you created like the TransactionHandler (with start, commit, abort) into the UnitOfWorkExecutor, an call the proper methods from there... something like this... I don't see the need to change the whole codebase where we have the ProcessInstances because transaction should not coupled with them, we could reuse the UnitOfWork stuff that we already have, WDYT?
> 
> @danielezonca @cristianonicolai @evacchi ^

I still have to review but I agree we should be able to wrap the whole transaction logic inside the unitOfWork 👍

xieshenzh:
@tiagodolphine @danielezonca , I have changed the implementation. Please take a look. Thanks.

xieshenzh:
Jenkins retest please

xieshenzh:
Jenkins retest this

addons/persistence/mongodb-persistence-addon/src/test/java/org/kie/kogito/mongodb/MongoDBProcessInstancesTest.java
```diff
@@ -15,24 +15,33 @@
 
 package org.kie.kogito.mongodb;
 
-import java.util.Collections;
-import java.util.Optional;
-
-import com.mongodb.client.MongoClient;
+import com.mongodb.client.*;
```
cristianonicolai, line 18: please avoid star imports

addons/persistence/mongodb-persistence-transaction-addon/pom.xml
```diff
@@ -0,0 +1,94 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
cristianonicolai, line 1: @xieshenzh could we simply have a property in the current mongo addon to enable/disable transaction? instead of a new module?

cristianonicolai (commented):
@xieshenzh @tiagodolphine  another idea, maybe instead of changing the unit of work directly, what if it simply throws events, lets say ( uow started, completed, aborted ) ? That way the addon would be self contained to handled it and start/commit/rollback as needed, wdyt?

xieshenzh:
> @xieshenzh @tiagodolphine another idea, maybe instead of changing the unit of work directly, what if it simply throws events, lets say ( uow started, completed, aborted ) ? That way the addon would be self contained to handled it and start/commit/rollback as needed, wdyt?

@cristianonicolai a `ClientSession` must be created before executing the unit of work, because it must be passed to all the MongoDB operations in order for the transaction to work. And only from `UnitOfWork`, we can know when to create or close the `ClientSession`. 

Do you mean we should modify the `UnitOfWork` in a different way, and let it send events to transaction manager to start/commit/abort transaction, instead of injecting a transaction manager to the `UnitOfWork` directly? 

Please advise. Thanks.

xieshenzh:
Jenkins retest this

cristianonicolai:
> > @xieshenzh @tiagodolphine another idea, maybe instead of changing the unit of work directly, what if it simply throws events, lets say ( uow started, completed, aborted ) ? That way the addon would be self contained to handled it and start/commit/rollback as needed, wdyt?
> 
> @cristianonicolai a `ClientSession` must be created before executing the unit of work, because it must be passed to all the MongoDB operations in order for the transaction to work. And only from `UnitOfWork`, we can know when to create or close the `ClientSession`.
> 
> Do you mean we should modify the `UnitOfWork` in a different way, and let it send events to transaction manager to start/commit/abort transaction, instead of injecting a transaction manager to the `UnitOfWork` directly?
> 
> Please advise. Thanks.

@xieshenzh yep, so if the unitofwork sends out events such as before start. after start, then I think you can handle the creation of client session in the addon itself. 
@tiagodolphine wdyt about this approach, so sending out events from unit of work instead of directly modifying it, I think this can allow us to plug different mechanisms moving forward, a little more flexible.

xieshenzh:
> @xieshenzh yep, so if the unitofwork sends out events such as before start. after start, then I think you can handle the creation of client session in the addon itself.

@cristianonicolai The event mechanism for springboot is very different from quarkus. So we need to create two sets of event publishers and consumers and inject them to `UnitOfWork`. It would be very complicated to create them with code-generation and it is hard to maintain. So I need to create different modules for quarkus event and springboot event. Or should I use some third party event solution instead, and inject that to `UnitOfWork`?

cristianonicolai:
> > @xieshenzh yep, so if the unitofwork sends out events such as before start. after start, then I think you can handle the creation of client session in the addon itself.
> 
> @cristianonicolai The event mechanism for springboot is very different from quarkus. So we need to create two sets of event publishers and consumers and inject them to `UnitOfWork`. It would be very complicated to create them with code-generation and it is hard to maintain. So I need to create different modules for quarkus event and springboot event. Or should I use some third party event solution instead, and inject that to `UnitOfWork`?

@xieshenzh agree eventing is probably not the best solution, is not uniform with the way we handle this elsewhere in the platform. But I still think it's worth disconnecting these concepts. Maybe what could be done is a UnitOfWork listener interface that we could inject based on runtime instead? Similar to process instance event listeners. @tiagodolphine wdyt?

xieshenzh:
> @xieshenzh agree eventing is probably not the best solution, is not uniform with the way we handle this elsewhere in the platform. But I still think it's worth disconnecting these concepts. Maybe what could be done is a UnitOfWork listener interface that we could inject based on runtime instead? Similar to process instance event listeners. @tiagodolphine wdyt?

@cristianonicolai I have updated this PR to make UnitOfWork send events. Hope it will be helpful with your investigation on eventing.

xieshenzh:
Jenkins retest this

xieshenzh:
Jenkins retest this

xieshenzh:
Jenkins retest this

xieshenzh:
@cristianonicolai I have rebased the PR to use UnitOfWork event listeners.

cristianonicolai (approved):

cristianonicolai:
@danielezonca @tiagodolphine @MarianMacik if you guys could take a quick look again, thanks.

cristianonicolai:
@xieshenzh needs rebase

addons/persistence/mongodb-persistence-addon/src/test/java/org/kie/kogito/mongodb/transaction/MongoDBTransactionManagerIT.java
```diff
@@ -0,0 +1,253 @@
...
+                assertEquals(2, size2);
+            } finally {
+                latch2.countDown();
+                assertTrue(latch1.await(1, TimeUnit.MINUTES));
+            }
+            transactionManager.onAfterEndEvent(new UnitOfWorkEndEvent(null));
+            return null;
+        });
+
+        execution1.get();
```
cristianonicolai, line 144: @xieshenzh could you review so that all .get methods have an explicit timeout so they don't wait indefinitely?

cristianonicolai:
@xieshenzh needs rebase.

@MarianMacik can you take a look?

kogito-test-utils/src/main/java/org/kie/kogito/testcontainers/KogitoMongoDBContainer.java
```diff
@@ -28,10 +29,12 @@
 
     public static final String NAME = "mongodb";
     public static final int MONGODB_INTERNAL_PORT = 27018;
+    public static final String IMAGE_TAG = "4.4";
 
     private static final Logger LOGGER = LoggerFactory.getLogger(KogitoMongoDBContainer.class);
 
     public KogitoMongoDBContainer() {
+        super(DockerImageName.parse("mongo").withTag(IMAGE_TAG));
```
MarianMacik, line 37: Shouldn't we externalize this as with other containers?

MarianMacik (approved):
Looks good to me.

xieshenzh:
Jenkins retest this

xieshenzh:
Jenkins retest this

cristianonicolai:
jenkins please retest this

cristianonicolai:
jenkins please retest this

cristianonicolai:
jenkins please retest this

cristianonicolai:
@xieshenzh needs rebase

kogito-test-utils/src/test/java/org/kie/kogito/testcontainers/quarkus/MongoDBQuarkusTestResourceTest.java
```diff
@@ -27,6 +29,13 @@
 
     private MongoDBQuarkusTestResource resource;
 
+    private static final String IMAGE = "mongo:4.2";
+
+    @BeforeEach
+    public void setup() {
+        System.setProperty(KogitoMongoDBContainer.MONGODB_PROPERTY, IMAGE);
```
cristianonicolai, line 36: @xieshenzh can you just ensure that any system property set for a test is properly clean up after it?

kogito-test-utils/src/test/java/org/kie/kogito/testcontainers/springboot/MongoDBSpringBootTestResourceTest.java
```diff
@@ -35,6 +36,13 @@
 
     private MongoDBSpringBootTestResource resource;
 
+    private static final String IMAGE = "mongo:4.2";
+
+    @BeforeEach
+    public void setup() {
+        System.setProperty(KogitoMongoDBContainer.MONGODB_PROPERTY, IMAGE);
```
cristianonicolai, line 43: same here

kogito-test-utils/src/test/java/org/kie/kogito/testcontainers/KogitoMongoDBContainerTest.java
```diff
@@ -26,13 +26,16 @@
 class KogitoMongoDBContainerTest {
 
     private static final int MAPPED_PORT = 27020;
+    private static final String IMAGE = "mongo:4.2";
 
     private KogitoMongoDBContainer container;
 
     @BeforeEach
     public void setup() {
+        System.setProperty(KogitoMongoDBContainer.MONGODB_PROPERTY, IMAGE);
```
cristianonicolai, line 35: here also

cristianonicolai:
@xieshenzh sorry another conflict now :/

cristianonicolai:
jenkins please retest this

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/MongoDBProcessInstances.java
```diff
@@ -93,14 +106,26 @@ public void update(String id, ProcessInstance<T> instance) {
...
-            Document doc = Document.parse(new String(marshaller.marshallProcessInstance(instance)));
-            if (checkDuplicates) {
-                if (exists(id)) {
-                    throw new ProcessInstanceDuplicatedException(id);
+        if (!isActive(instance)) {
+            reloadProcessInstance(instance, id);
+            return;
+        }
+
+        ClientSession clientSession = Optional.ofNullable(transactionManager).map(TransactionManager::getResource).orElse(null);
```
tiagodolphine, line 114: why not keeping `Optional<ClientSession>` this would avoid the IF checks when null later on the method.

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/MongoDBProcessInstances.java
```diff
@@ -49,11 +53,16 @@
     private org.kie.kogito.process.Process<?> process;
     private ProcessInstanceMarshallerService marshaller;
     private final MongoCollection<Document> collection;
+    private MongoDBTransactionManager transactionManager;
```
tiagodolphine, line 56: it seems if you keep the transacptionManager as an Optional it would avoid the `Optional.ofNullable(transactionManager)` in many places, it could be just in the constructor.

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/transaction/MongoDBTransactionManager.java
```diff
@@ -0,0 +1,98 @@
...
+import com.mongodb.TransactionOptions;
+import com.mongodb.WriteConcern;
+import com.mongodb.client.ClientSession;
+import com.mongodb.client.MongoClient;
+
+public abstract class MongoDBTransactionManager implements TransactionManager<ClientSession>, UnitOfWorkEventListener {
+
+    private MongoClient mongoClient;
+
+    private ThreadLocal<ClientSession> clientSessionLocal = new ThreadLocal<>();
```
tiagodolphine, line 36: Isn't risk do rely on `ThreadLocal` context? I mean mainly because Quarkus/Vertx works in an async/eventloop approach with just a few threads. Probably it would be an alternative to use context propagation https://quarkus.io/guides/context-propagation, or even a `@RequestScoped` bean that would be injected here in the `MongoDBTransactionManager`, something like that.

tiagodolphine (commented):
LGTM, my main concern is about the ThreadLocal usage, take a look at my comment.
Thanks.

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/MongoDBProcessInstances.java
xieshenzh: It is not optional. I have removed the nullable check.

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/MongoDBProcessInstances.java
xieshenzh: I removed Optional here. Because the code is compiled with java 8, so the IF checks cannot be avoided with Optional.

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/transaction/MongoDBTransactionManager.java
xieshenzh: The code the for both Quarkus and Springboot, and async approach is not used here. 
For Quarkus, MongoDB transaction is supported with Panache and not with reactive approach: https://github.com/loicmathieu/quarkus/commit/212bfa803847f5d8eb316e442e4f7591aae725c9#diff-5e9957008764952bfb939eb4dc65acb9a37d4774ad16cb26cbd075bda2ef779a

kogito-test-utils/src/test/java/org/kie/kogito/testcontainers/springboot/MongoDBSpringBootTestResourceTest.java
```diff
@@ -35,6 +37,18 @@
 
     private MongoDBSpringBootTestResource resource;
 
+    private static final String IMAGE = "mongo:4.2";
+
```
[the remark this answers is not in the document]
tiagodolphine: we already have a method to get the image+version from a system property that is set in the pom, maybe we could use the same https://github.com/kiegroup/kogito-runtimes/blob/1de75a2e01beec3a53a7f31db255c0277543947f/kogito-test-utils/src/main/java/org/kie/kogito/testcontainers/KogitoGenericContainer.java#L40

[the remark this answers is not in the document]
danielezonca: I don't think we can use it because `KogitoGenericContainer` is an abstract class so it is fine only when we define the container from scratch. Here (like with `Kafka`), we extend another class ([`MongoDBContainer`](https://github.com/kiegroup/kogito-runtimes/blob/master/kogito-test-utils/src/main/java/org/kie/kogito/testcontainers/KogitoMongoDBContainer.java#L27)) so we cannot reuse this logic.
We should try to move common logic to a delegate/utility class (or even `interface` with default methods) to make it more reusable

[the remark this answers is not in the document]
xieshenzh: This property is already added to `kogito-build-parent/pom.xml`: https://github.com/kiegroup/kogito-runtimes/pull/987/files#diff-0887b6a8a10f8b2df645c943a46afbd388f7bce7ca8298df14ed60733ed138ae
Or do you mean something else?

kogito-test-utils/src/main/java/org/kie/kogito/testcontainers/KogitoMongoDBContainer.java
```diff
@@ -47,4 +53,8 @@ public int getMappedPort() {
     public String getResourceName() {
         return NAME;
     }
+
+    private static String mongoImage() {
+        return System.getProperty(MONGODB_PROPERTY, DEFAULT_IMAGE);
```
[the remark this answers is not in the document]
xieshenzh: There is a `DEFAULT_IMAGE`. So it will not be null here.

addons/persistence/mongodb-persistence-addon/src/main/java/org/kie/kogito/mongodb/transaction/MongoDBTransactionManager.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    @Override
+    public void onAfterAbortEvent(UnitOfWorkAbortEvent event) {
+        if (!enabled()) {
+            return;
+        }
+
+        ClientSession clientSession = this.getResource();
+        if (clientSession != null) {
```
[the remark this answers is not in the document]
xieshenzh: I'll remove the null checks.

kogito-test-utils/src/test/java/org/kie/kogito/testcontainers/springboot/MongoDBSpringBootTestResourceTest.java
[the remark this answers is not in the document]
danielezonca: I mean to have a `version.org.mongo-springboot` with version `4.2` so that here you can just load property instead of have 4.2 in the code. In this way it is clear also in build-parent that we have two different versions (and easier to upgrade)

tiagodolphine (approved):

kogito-test-utils/src/test/java/org/kie/kogito/testcontainers/springboot/MongoDBSpringBootTestResourceTest.java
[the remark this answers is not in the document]
cristianonicolai: +1

api/kogito-api/src/main/java/org/kie/kogito/persistence/transaction/TransactionManager.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.persistence.transaction;
+
+public interface TransactionManager<T> {
```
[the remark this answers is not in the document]
cristianonicolai: we do intend to have similar implementation around `postgresql` but that really should relly on the `UnitOfWorkEventListener`, not a real dependency on the Kogito API. @xieshenzh any particular reason to have this in the API module?

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/persistence/PersistenceGenerator.java
```diff
@@ -453,12 +460,14 @@ private CompilationUnit generateProtostreamBaseMarshaller(TemplatedGenerator gen
 
             persistenceProviderClazz.addMember(dbNameField);
             persistenceProviderClazz.addMember(dbNameMethod);
+            generatedTMFile = mongodbBasedTransaction(persistenceProviderClazz);
             generatedClientFile = generatePersistenceProviderClazz(persistenceProviderClazz,
```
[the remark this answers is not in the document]
cristianonicolai: yeah, in fact should we check this at all? isn't DI mandatory anyway?

api/kogito-api/src/main/java/org/kie/kogito/persistence/transaction/TransactionManager.java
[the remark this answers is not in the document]
xieshenzh: `UnitOfWorkEventListener` has nothing to do with transaction management.
I thought there should be an interface for kogito transaction management in general. 
Is this interface needed? If it is needed, where should it be?

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/persistence/PersistenceGenerator.java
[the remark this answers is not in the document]
danielezonca: Well we still have the `no DI` codepath even if it is only used by codegen integration test (using `AbstractCodegenIT` class). So for now it is mainly our decision what to have in codepath without DI.
I think we should be able to have the "engines" and the wiring of `Application` at least so persistence could be part of it.
Btw it is definitely out of scope for this PR

kogito-test-utils/src/test/java/org/kie/kogito/testcontainers/springboot/MongoDBSpringBootTestResourceTest.java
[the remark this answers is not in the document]
xieshenzh: I have changed to set properties with surefire plugin, and changed the tests to verify the image name explicitly. 
The value of the version here is just for testing purpose, so it doesn't really matter. 
The value of `version.org.mongo.springboot` for kogito is 4.1.0. But there is no mongo image released for this particular version. And the `MongoDBContainer` won't start if the image doesn't exist. So we cannot use this property for mongo image tag. Property `container.image.mongodb` is used instead for the tests.

xieshenzh:
Jenkins retest this

api/kogito-api/src/main/java/org/kie/kogito/persistence/transaction/TransactionManager.java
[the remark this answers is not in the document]
cristianonicolai: Thats the point, I think we dont need this extra interface, only implement `UnitOfWorkEventListener`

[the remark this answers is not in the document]
xieshenzh: deleted

cristianonicolai:
Jenkins rerun apps tests

PR #991, opened 2021-01-20
KOGITO-4213 - Throw exception when failing to parse cloud event content

cristianonicolai:


evacchi (commented):
IIRC this was intended because currently we broadcast events to all consumers

danielezonca:
Can you please make sure that message `ack()` handling is correct after this change (aka no ack and message available to other consumers)

evacchi:
> What do you mean by intended? So we don't throw any exception because otherwise it fails the producer and an event won't be broadcast to every consumer?

all consumers are (currently) subscribed to the same event bus; so they may receive some messages that are not intended for them -- in that case they should just ignore it, even (especially?) if they fail to parse it. They are just not the intended receiver.

At least that's what I remember; if we throw here, then all subscribers that are not expecting that message type will throw and fail, when they see someone else's message -- which is probably _not_ what we want :)

the best we can do here is raise a warning -- then an improved solution would send messages only to the "right" sender.

cristianonicolai:
@evacchi sorry for the delay, catching up here.
So regarding your points, what brought up this improvement is a failure to parse the cloud event message, see https://issues.redhat.com/browse/KOGITO-4249 and https://gist.github.com/jbride/9b3c4c53b051c1f9c4268a7cb6f03662
In this case, it's before any logic to decide if the message should be ignored by the subscriber or not. Without knowing the event structure there is little any consumer will be able to do.
Another point is that the object mapper is shared for all consumer, so is not like one will be able to unmarshall and others don't, they will all fail or succeed. So in this case for the exception, we are handling here ( JsonProcessingException ) I think we should throw an error and simply stop processing but I agree that for other scenarios a warn would be more appropriate.

cristianonicolai:
@evacchi @MarianMacik does my comment makes sense? should we go ahead with this change?

MarianMacik:
@cristianonicolai Yes, makes more sense now. The mapper is shared so in case there is an error when mapping, we should stop everything as others will also fail.

MarianMacik (approved):

PR #1005, opened 2021-01-25
[KOGITO-4263] unfork source code of drools-core-dynamic

mariofusco:


kogito-build-parent/pom.xml
```diff
@@ -205,6 +205,11 @@
         <artifactId>drools-core-reflective</artifactId>
         <version>${version.org.kie7}</version>
       </dependency>
+      <dependency>
+        <groupId>org.drools</groupId>
+        <artifactId>drools-core-dynamic</artifactId>
+        <version>${version.org.kie7}</version>
```
danielezonca, line 211: This is probably a more general comment, I don't think we should add to dependency management all the modules whose we have a wrapped version in kogito so that we don't risk to include a wrong artifact by mistake.
In this way the wrapped module has to declare explicitly the version (using kie7 property) but then nobody else will find it in the depMan.
Wdyt?
If you agree we can cleanup this think in a different PR.

mariofusco: I agree, but I thought there was an enforcer not allowing you to do that, am I wrong?

danielezonca: Well at the moment I don't think there is an enforcer configured and btw even in that case it is enough to create a `dependencyManagement` section in the module where you want to use the dependency to solve it

danielezonca (approved):

PR #1006, opened 2021-01-25
[KOGITO-4019] Add onSignal and onMessage to ProcessEventListener

fjtirado:
- onSignal and onMessage methods has been added to ProcessEventListener
- Code generator has been changed to deal with signals both for IntermediateThrowEvent and EndEvent. It delegates on existing SignalProcessInstanceAction, conveniently modified to call onSignal method. 
- Code generator has been changed to add onMessage listener call on action lambda for messages
- I refrained myself and do not change bpmn2 project substantially, since removing Signal and Message handler from xml handlers affects a number of legacy test. Eventually we will need to refactor that part, but I prefer to wait the result of the on parallel v7 unfork exercise


tiagodolphine (commented):
@fjtirado +1 I'm just wondering if you can add some tests for these new Events, maybe on the `ProcessEventListenerTest` wdyt?

fjtirado:
> @fjtirado +1 I'm just wondering if you can add some tests for these new Events, maybe on the `ProcessEventListenerTest` wdyt?

Sure, I forgot to mention on the description that this was still on draft because I was working on unit test ;)

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/actions/SignalProcessInstanceAction.java
```diff
@@ -78,24 +80,31 @@ public SignalProcessInstanceAction(String signalName, Function<ProcessContext, O
...
+    @Override
     public void execute(ProcessContext context) throws Exception {
         String variableName = VariableUtil.resolveVariable(this.variableName, context.getNodeInstance());
         Object variable = variableName == null ? eventDataSupplier.apply(context) : context.getVariable(variableName);
 
         if (transformation != null) {
             variable = new org.jbpm.process.core.event.EventTransformerImpl(transformation).transformEvent(context.getProcessInstance().getVariables());
         }
+        
+        String signalType = VariableUtil.resolveVariable(signalName, context.getNodeInstance());
```
MarianMacik, line 92: I would maybe take this opportunity to make this part clear once and for all as the `signalName` variable doesn't contain the name of the signal (as it may contain also something like `#{signalVarInProcess}`, rather it is a value of the "signal" variable from the definition - but this is somehow clear. What is not that much clear is that `variable` is in fact the payload of the signal. Likewise, `variableName` is the name of the payload variable. Maybe we can make this clearer? Same goes for the Message action.

Also, we don't want to have an option to use `#{varFromProcess}` replacements with messages? I am missing that part in the Message action.

MarianMacik (commented):
So far so good, just some comments to discuss/improve things.

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/actions/SignalProcessInstanceAction.java
fjtirado: I agree on the clarification of the variable names, Ill change it
Regarding message replacement, I guess that for messages we want to have just predefined strings. This will be handy for Kafka topic resolution base on message name (https://issues.redhat.com/browse/KOGITO-4267, if message name is variable, that will be really messy), so my option will be to leave it.

kogito-codegen/src/test/java/org/kie/kogito/codegen/tests/MessageIntermediateEventTest.java
```diff
@@ -50,7 +60,8 @@ public void testMessageThrowEventProcess() throws Exception {
         
         assertThat(processInstance.status()).isEqualTo(ProcessInstance.STATE_COMPLETED);
         Model result = (Model)processInstance.variables();
-        assertThat(result.toMap()).hasSize(1).containsKeys("customerId");        
+        assertThat(result.toMap()).hasSize(1).containsKeys("customerId");
+        verify(listener,times(1)).onMessage(any(MessageEvent.class));
```
fjtirado, line 64: test is failing because onMessage is not being invoked.
After some debugging, I find that code generate replace HandleMessageAction.execute method by this

actionNode3.action(kcontext -> {
            producer_3.produce(kcontext.getProcessInstance(), (java.lang.String) kcontext.getVariable("customerId"));
        });

which invokes this

public class MessageIntermediateEventMessageProducer_3 {

    Object emitter;

    EventMarshaller marshaller = new DefaultEventMarshaller();

    public void configure() {
    }

    public void produce(ProcessInstance pi, java.lang.String eventData) {
    }

    private String marshall(ProcessInstance pi, java.lang.String eventData) {
        return marshaller.marshall(eventData, e -> new MessageIntermediateEventMessageDataEvent_3("", e, pi.getId(), pi.getParentProcessInstanceId(), pi.getRootProcessInstanceId(), pi.getProcessId(), pi.getRootProcessId(), String.valueOf(pi.getState()), pi.getReferenceId() == null || pi.getReferenceId().trim().isEmpty() ? null : pi.getReferenceId()), useCloudEvents);
    }
}
which is basically doing nothing. 
We probably need to discuss this.

kogito-codegen/src/test/java/org/kie/kogito/codegen/tests/SignalEventTest.java
```diff
@@ -15,35 +15,58 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+
 public class SignalEventTest extends AbstractCodegenTest {
     
+    
+    @Test
+    public void testIntermediateThrowSignal() throws Exception {
```
fjtirado, line 51: This is also failing, ActionNodeVisitor is not capable of generating code to deal with the signal action
https://github.com/kiegroup/kogito-runtimes/blob/master/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ActionNodeVisitor.java#L66-L68

fjtirado:
I think we have a general problem with DroolsActions created by this method
https://github.com/kiegroup/kogito-runtimes/blob/master/jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java#L1076-L1080
This actions has empty string consequence and hence are basically ignored by ActionNodeVisitor (which means that throw signal and message events are not working) 
We need to discuss a solution. I vote for generating lambda that invoke execute method of all classes under package org.jbpm.process.instance.impl.actions. We can generate a special Action implementation JavaAction that is handled differently by ActionNodeVisitor. 
My concern is the impact this might have for code that is already working (as far as I know Kafka add on for messages)

tiagodolphine:
@fjtirado from my understanding the Actions were replaced by lambda expressions during the codegen. For instance, the messaging part you mentioned is done here https://github.com/kiegroup/kogito-runtimes/blob/820408b268bc116b380f88d2dc9c4c1e2417d9af/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/TriggerMetaData.java#L126 where is basically call the `producer` generated by the codegen (quarkus/spring). So in summary the `HandleMessageAction` is not used, and other actions the same (from what I understand)... you can check for example **https://github.com/kiegroup/kogito-runtimes/blob/820408b268bc116b380f88d2dc9c4c1e2417d9af/jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/EndNodeVisitor.java#L72** where the code generated for end node (message, signal or compensation) is generated using lambdas based on information on the TriggerMetada or in some cases the in attributes from the Action object, but the Action itself is not used to `execute`.

fjtirado:
@tiagodolphine 
Yes, thats my understanding too. But this has some consequences in behaviour that needs to be confirmed.
HandleMessageAction (as its v7 counterpart) is invoking a WorkItemHandler named "Send Task". This will stop working when migrating from v7 to Kogito. 
But messages are not my main concern. My main concern are signals, as indicated in one of the test, Kogito is even unable to complete the code generation process when a signalDefinition is included in an IntermediateThrowEvent.

tiagodolphine:
> @tiagodolphine
> Yes, thats my understanding too. But this has some consequences in behaviour that needs to be confirmed.
> HandleMessageAction (as its v7 counterpart) is invoking a WorkItemHandler named "Send Task". This will stop working when migrating from v7 to Kogito.
> But messages are not my main concern. My main concern are signals, as indicated [here](https://github.com/kiegroup/kogito-runtimes/pull/1006#discussion_r565223826), Kogito is even unable to complete the code generation process when a signalDefinition is included in an IntermediateThrowEvent.

Yeah, the  WorkItemHandler "Send Task", is not used anymore for messages, since this is executed by the `Producer` classes generated at build time... for signals, I'm not sure all use cases were implemented at least the ones I know use lambda expressions to send the signal, @ruromero  since you worked with this part, do you have anything to add?
Anyway, for signals, I don't see issues to use the Action approach if it makes things more clear, since we don't depend on classes from the codegen....

ruromero:
@tiagodolphine @fjtirado 
I added lambda actions for few cases, specially for the actions used as conditions and consequences for Milestones and maybe also Ad-Hoc Subprocesses that I can remember.
So not "any action" has been replaced by a lambda there are many cases and I wasn't confident to cover them all appropriately.

fjtirado:
There are several issues:
1) A BPMN which defines a throw intermediate event that include a signal breaks code generation. This is crearly broken compared with V7.
2) A backward compatibility one. In V7, when a message is inside a throw event (intermediate or end), a work item named "send task" is created and executed. One of the use cases was to register a workitemhandler for "send task" that publish jms events. I guess we can break backward compatibility here, but needs to be clearly stated somewhere
3) If we do not have work item handler to handle actions any more, we need something to replace that. The processor currently generated by generator (no addon enabled) is empty  
```
public class MessageIntermediateEventMessageProducer_3 {

    Object emitter;

    EventMarshaller marshaller = new DefaultEventMarshaller();

    public void configure() {
    }

    public void produce(ProcessInstance pi, java.lang.String eventData) {
    }

    private String marshall(ProcessInstance pi, java.lang.String eventData) {
        return marshaller.marshall(eventData, e -> new MessageIntermediateEventMessageDataEvent_3("", e, pi.getId(), pi.getParentProcessInstanceId(), pi.getRootProcessInstanceId(), pi.getProcessId(), pi.getRootProcessId(), String.valueOf(pi.getState()), pi.getReferenceId() == null || pi.getReferenceId().trim().isEmpty() ? null : pi.getReferenceId()), useCloudEvents);
    }
}
```
So, Im wondering which is the idea here. Manullay implementr the producer through scaffolding? Assume that if not addon is enable, messages are ignored?

krisv:
Shouldn't we do something like fail during validation / build of the service?  Or at least write out warn/error statements at runtime?  Because to me it seems that defining message events but then not having an implementation seems to be a bug that needs to be corrected.  I think we should try to catch that at build time, or at least log during execution, as I don't think not doing anything is ever going to be acceptable?

fjtirado:
@krisv 
Thanks for the feedback. 
Since workitem mechanism is not longer used, in the spirit of what you are recommending (logging something) I suggest adding to the generated producer (fixing also the scenarios in which generation is broken), the invocation to process event listener (onMessage/onSignal), so message/signal actually do "something" when no addon is enabled.

tiagodolphine:
@fjtirado 
1. Agree it seems to be a bug, or it was not even implemented and tested, so we can fix it.
2. As Kris mentioned I agree, we don't need a work item for Send Task, the approach we have in kogito is basically to call the generated Producer that should perform the action to send the message (kafka, jms, whatever) that is under configuration at build time, so the generated producer handle this, and currently, we don't use addon for this, maybe this could be a good approach, the addon we have for messaging is just to publish runtime related events that are consumed by data-index.
3. The generated producer is empty because this is the fallback when no context is set (quarkus/spring),  this is in the codegen, see https://github.com/kiegroup/kogito-runtimes/blob/6ce20bcc68c234fa8b28bfbd59bb105cfcced831/kogito-codegen/src/main/resources/class-templates/MessageProducerJavaTemplate.java#L22-L21 so the test should consider this IMO, @danielezonca ^

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/TriggerMetaData.java
```diff
@@ -122,30 +135,68 @@ public TriggerMetaData validate() {
     public String toString() {
         return "TriggerMetaData [name=" + name + ", type=" + type + ", dataType=" + dataType + ", modelRef=" + modelRef + "]";
     }
+    
+    public static ObjectCreationExpr buildAction(String signalName, String variable, String scope) {
```
tiagodolphine, line 139: nice, so for signal the idea is to replace the old lambda with 'SignalProcessInstanceAction', I think that's fine because we don't depend on any codegen class and basically it solve the issue we had with 'scope' that wasn't being taken into account, just sending the signal to the process instance scope.

tiagodolphine (approved):
@fjtirado looks good to me, so now for Signals, we will use the Action and for Messages we still use the lambda with the producer + the event call you inserted.

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/TriggerMetaData.java
fjtirado: yes, it is nice when other developer understood your purpose ;)

fjtirado:
@kiegroup/gatekeepers ready to go

drools/drools-core/src/main/java/org/drools/core/event/KogitoProcessEventSupport.java
```diff
@@ -197,7 +211,26 @@ public void fireAfterWorkItemTransition(final ProcessInstance instance, WorkItem
                 }
             }));
     }
+    
+    public void fireOnSignal (final ProcessInstance instance, NodeInstance nodeInstance, String signalName, String signalObject, KieRuntime kruntime) {
```
cristianonicolai, line 215:
```suggestion
    public void fireOnSignal(final ProcessInstance instance, NodeInstance nodeInstance, String signalName, String signalObject, KieRuntime kruntime) {
```

drools/drools-core/src/main/java/org/drools/core/event/KogitoProcessEventSupport.java
```diff
@@ -197,7 +211,26 @@ public void fireAfterWorkItemTransition(final ProcessInstance instance, WorkItem
...
+        final SignalEvent event = new KogitoSignalEvent(instance, kruntime, nodeInstance, signalName, signalObject);
+        unitOfWorkManager.currentUnitOfWork().intercept(WorkUnit.create(event, e -> {
+            if ( hasListeners() ) {
+                notifyAllListeners( event, ( l, e1 ) -> l.onSignal( e1 ) );
+            }
+        }));
+    }
+    
+    public void fireOnMessage (final ProcessInstance instance, NodeInstance nodeInstance, String messageName, String messageObject, KieRuntime kruntime) {
```
cristianonicolai, line 224:
```suggestion
    public void fireOnMessage(final ProcessInstance instance, NodeInstance nodeInstance, String messageName, String messageObject, KieRuntime kruntime) {
```

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ActionNodeVisitor.java
```diff
@@ -62,7 +67,12 @@ public void visitNode(String factoryField, ActionNode node, BlockStmt body, Vari
         } else if (node.getMetaData(TRIGGER_REF) != null) { // if there is trigger defined on end event create TriggerMetaData for it
             LambdaExpr lambda = TriggerMetaData.buildLambdaExpr(node, metadata);
             body.addStatement(getFactoryMethod(getNodeId(node), METHOD_ACTION, lambda));
-        } else {
+        } 
+        else if (node.getMetaData(REF) != null && EVENT_TYPE_SIGNAL.equals(node.getMetaData(EVENT_TYPE))) {
```
cristianonicolai, line 71:
```suggestion
        } else if (node.getMetaData(REF) != null && EVENT_TYPE_SIGNAL.equals(node.getMetaData(EVENT_TYPE))) {
```

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ActionNodeVisitor.java
```diff
@@ -62,7 +67,12 @@ public void visitNode(String factoryField, ActionNode node, BlockStmt body, Vari
...
+            body.addStatement(getFactoryMethod(getNodeId(node), METHOD_ACTION, TriggerMetaData.buildAction((String)node.getMetaData(REF),
+                (String)node.getMetaData(VARIABLE), (String) node.getMetaData(CUSTOM_SCOPE))));
+        }
+        else {
```
cristianonicolai, line 75:
```suggestion
        } else {
```

cristianonicolai (approved):
some minor formatting

fjtirado:
unrelated but jenkins please test this

MarianMacik (approved):

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/TriggerMetaData.java
```diff
@@ -122,30 +134,66 @@ public TriggerMetaData validate() {
...
+        return new ObjectCreationExpr(null,
+                parseClassOrInterfaceType(SignalProcessInstanceAction.class.getCanonicalName()),
+                new NodeList<>(new StringLiteralExpr(signalName), variable != null ? new StringLiteralExpr(variable.replace("\"", "\\\"")) : new CastExpr(
+                        parseClassOrInterfaceType(String.class.getCanonicalName()), new NullLiteralExpr()),
+                        scope != null ? new StringLiteralExpr(scope) : new CastExpr(
+                                parseClassOrInterfaceType(String.class.getCanonicalName()), new NullLiteralExpr())));
```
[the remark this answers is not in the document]
fjtirado: yes, to avoid ambiguous constructor compilation issue

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/TriggerMetaData.java
```diff
@@ -122,30 +134,66 @@ public TriggerMetaData validate() {
...
+                piName), new MethodCallExpr(new NameExpr("kcontext"),
+                        "getProcessInstance"), Operator.ASSIGN);
+        MethodCallExpr listenerMethodCall = new MethodCallExpr(
+                new MethodCallExpr(process, "getProcessEventSupport"), "fireOnMessage")
+                        .addArgument(pi)
+                        .addArgument(new MethodCallExpr(kExpr, "getNodeInstance"))
+                        .addArgument(runtime)
+                        .addArgument(new StringLiteralExpr(messageName)).addArgument(object);
+        MethodCallExpr producerMethodCall = new MethodCallExpr(new NameExpr("producer_" + node.getId()), "produce")
+                .addArgument(pi).addArgument(object);
```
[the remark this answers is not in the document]
fjtirado: yes, thats the idea, so in case not addon is provided, users using kogito as embedded engine can guess when the message has been thrown

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/TriggerMetaData.java
```diff
@@ -122,30 +134,66 @@ public TriggerMetaData validate() {
...
+        actionBody.addStatement(objectExpr);
+        actionBody.addStatement(runtimeExpr);
+        actionBody.addStatement(processInstanceAssignment);
+        actionBody.addStatement(processExpr);
+        actionBody.addStatement(listenerMethodCall);
```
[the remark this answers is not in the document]
fjtirado: sure, adding it

fjtirado:
jenkins retest this (unrelated again)

MarianMacik (approved):

fjtirado:
@kiegroup/gatekeepers ready to merge

PR #1014, opened 2021-01-28
 [KOGITO-3915] Introducing OpenAPI client codegen for SW 

ricardozanini:
See: https://issues.redhat.com/browse/KOGITO-3915

In this PR, we introduce a new codegen, the `kogito-codegen-openapi` module. Its role is to generate code based on the OpenApi specification file from a given Kogito resource file. For now, only `*.sw.json|yaml|yml` files are supported. 

For example:

```json
  "functions": [
    {
      "name": "countryInfoFunction",
      "operation": "restcountries.json#getByName1"
    }]
```

In this `function` definition, we can see an operation pointing to `restcountries.json`, which is an OpenApi spec file defined in the user's classpath:

```
src
├── main
│   └── resources
│       ├── restcountries.json

```

We support `classpath:`, `file:` and `http|s` resources. The `#getName1` portion of the operation identifies the REST endpoint we want to call:

```json
"/v2/name/{name}" : {
      "get" : {
        "operationId" : "getByName_1",
        "parameters" : [ {
          "name" : "name",
          "in" : "path",
          "required" : true,
          "schema" : {
            "type" : "string"
          }
        }, {
          "name" : "fullText",
          "in" : "query",
          "schema" : {
            "type" : "boolean"
          }
        }, {
          "name" : "fields",
          "in" : "query",
          "schema" : {
            "type" : "string"
          }
        } ],
```

All the parameters are also processed by our engine, resulting in a method like this:

```java
public JsonNode getByName1(JsonNode var1, JsonNode var2, JsonNode var3) throws ApiException 
```

This class is generated by the OpenApi generator tool and post-processed by our engine.

The function call in the SW file looks like this:

```json
{
      "name": "GetCountryInfo",
      "type": "operation",
      "start": true,
      "actionMode": "sequential",
      "actions": [
        {
          "name": "getCountryInfo",
          "functionRef": {
            "refName": "countryInfoFunction",
            "parameters": {
              "name": "$.name"
            }
          }
        }
      ],
      "transition": {
        "nextState": "ClassifyByPopulation"
      }
    },
```

If a parameter is required, the generator will fail with a specific exception for the user to fix.

Every parameter and return object is a JsonNode. This is because the SW engine treats everything as JSON. This can be changed in the future to adapt to other scenarios, like a jBPM Service Task, each in/out parameter can be a POJO, which is also generated by the tool. 

A spec file contains the information about the server, which we can call, for example:

```json
{
  "openapi" : "3.0.1",
  "info" : {
    "title" : "REST Countries",
    "description" : "REST Countries",
    "termsOfService" : "https://restcountries.eu/",
    "license" : {
      "name" : "Mozilla Public License MPL 2.0",
      "url" : "https://www.mozilla.org/en-US/MPL/2.0/"
    },
    "version" : "2.0.5"
  },
  "servers" : [ {
    "url" : "https://restcountries.eu/rest/"
  } ],
}
```

If not provided, the tool defaults to `http://localhost`. A user can change this property in runtime via a Kogito property in their `application.properties` file, for example:

```properties
org.kogito.openapi.client.multiplication.base_path=http://localhost:8282
```

`multiplication` is the name of the spec file defined in the Kogito resource file.

If the spec defines security objects, one can also configure username, password, and API Keys via properties. This scenario can work now, but it would take much more time to cover everything, so I left this use case to another JIRA.

## Key Takeaways

1. Since the code generated by this module can be a requirement for other resources (such as processes), it MUST generate its files first.
2. This generator is a wrapper for the original [OpenApi Generator Tool](https://github.com/OpenAPITools/openapi-generator/). I used this tool to avoid reinventing the wheel. Their community is huge, active, and licensed under Apache v2. I don't see any problems with using it nor productizing it in the future.
3. The generated code is acting like a "Service" class to make REST calls. We also filter the code generated by the tool to adjust to our requirements and work on SpringBoot, Quarkus, and Native scenarios. You can see the examples here: https://github.com/kiegroup/kogito-examples/pull/591
4. All the files generated by the original tool are deleted after processed. We only keep the files that a user would require for their application. The generated files are then embedded in the final application inside the `org.kie.kogito.app.openapi` package. An SW file can have multiple OpenApi spec definitions.
5. The Service Task Java implementation is now being manually crafted by the Java Parser library, which is not ideal. I'll remove refactor this module in a later JIRA to use templates instead.
6. The `jbpm-serverless-workflow` module will be refactored to be a "translator" from an SW definition to a jBPM process only. Any generated code will be migrated to `kogito-modules-process` instead (and we will leverage the template files).
7. The `kogito-workitems` module will also be refactored. Ideally, it should be added on-demand via an extension, not manually added in the user's pom file.
8. The OpenApi generator tool generates REST client code based on resteasy client. It was the middle-ground solution I've found to generate code for SB, Quarkus, and Native. It's the same code, same post-processing. Although they support SB HTTP Client, the easiest approach was to generate the same code to avoid maintenance. We can change this in the future, of course. 


kogito-codegen/src/main/java/org/kie/kogito/codegen/openapi/OpenAPIClientCodegen.java
```diff
@@ -0,0 +1,96 @@
...
+                        .flatMap(workflow -> workflow.getFunctions().getFunctionDefs().stream())
+                        .map(OpenAPISpecResource::fromSWFunctionDefinition)
+                        .collect(toList());
+        return new OpenAPIClientCodegen(context, "openapispecs", openAPISpecResources);
+    }
+
+    private static boolean acceptOnlyWithOpenAPIOperation(final Workflow workflow) {
+        return workflow.getFunctions() != null &&
+                workflow.getFunctions().getFunctionDefs() != null &&
+                !workflow.getFunctions().getFunctionDefs().isEmpty() &&
```
ricardozanini, line 75: remove this

ricardozanini:
Jenkins retest this please

ricardozanini:
> Jenkins retest this please

kogito-codegen-modules/kogito-codegen-openapi/src/main/java/org/kie/kogito/codegen/openapi/client/OpenApiClientCodegen.java
```diff
@@ -0,0 +1,168 @@
...
+    public Optional<ApplicationSection> section() {
+        return Optional.empty();
+    }
+
+    @Override
+    public Collection<GeneratedFile> generate() {
+        final List<GeneratedFile> generatedFiles = new ArrayList<>();
+        this.openApiSpecDescriptors.forEach(descriptor -> {
+            // save the descriptor file in a temporary location
+            final String resolvedPath =
```
danielezonca, line 106: Is it really necessary to save the file? I would like to keep the generator as an in memory only process (to be thread safe and idempotent).
If there are no other option, what about add a finally block to remove it?
Moreover, can you please make it explicitly in method name if it saves a file (instead of just `resolve`)?

kogito-quarkus-extension/deployment/pom.xml
```diff
@@ -74,6 +74,10 @@
             <groupId>org.kie.kogito</groupId>
             <artifactId>kogito-codegen-predictions</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.kie.kogito</groupId>
+            <artifactId>kogito-codegen-openapi</artifactId>
```
danielezonca, line 79: Make sure to do the same for `kogito-maven-plugin`

kogito-codegen-modules/kogito-codegen-openapi/src/main/resources/custom-templates/resteasy/README
```diff
@@ -0,0 +1,5 @@
+These are the custom templates used by the OpenAPI Generator Tool to create the REST client code.
```
danielezonca, line 1: What about add `md` extension to the file?

kogito-codegen-modules/kogito-codegen-openapi/src/main/java/org/kie/kogito/codegen/openapi/client/OpenApiClientCodegen.java
```diff
@@ -0,0 +1,168 @@
...
+import static java.util.stream.Collectors.toList;
+
+/**
+ * Entry point for the OpenAPIClient generator code.
+ **/
+public class OpenApiClientCodegen extends AbstractGenerator {
+
+    private static final String BASE_PACKAGE = "openapi.client";
+    private static final String GEN_BASE_PATH = "openapi";
+    private static final int PRIORITY = 0;
```
danielezonca, line 53: Please don't use priority 0, any number between 1 and 9 should be fine. As alternative you can even update all other priorities to shift by 10 and make this priority 10.

kogito-codegen-modules/kogito-codegen-openapi/src/main/java/org/kie/kogito/codegen/openapi/client/OpenApiUtils.java
```diff
@@ -0,0 +1,38 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+public final class OpenApiUtils {
+
+    private OpenApiUtils() {
+    }
+
+    public static String getEndUserTargetDir(final ClassLoader classLoader) {
+        return new File(classLoader.getResource("").getFile()).getParentFile().getPath();
```
danielezonca, line 28: Please use `context.getAppPaths().getFirstProjectPath()`. It should be equivalent and it handled in a centralised way.

kogito-codegen-modules/kogito-codegen-openapi/src/main/java/org/kie/kogito/codegen/openapi/client/OpenApiUtils.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    public static void requireValidSpecURI(final OpenApiSpecDescriptor resource) {
+        requireNonNull(resource);
+        requireNonNull(resource.getURI());
```
danielezonca, line 33: Please add a message too, otherwise it will just throw a NPE

kogito-codegen-modules/kogito-codegen-openapi/src/main/java/org/kie/kogito/codegen/openapi/client/ServerlessWorkflowCodegenUtils.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.Set;
+
+import io.serverlessworkflow.api.Workflow;
+import io.serverlessworkflow.api.functions.FunctionDefinition;
+import io.serverlessworkflow.api.workflow.Functions;
+import org.jbpm.serverless.workflow.parser.util.ServerlessWorkflowUtils;
+
+final class ServerlessWorkflowCodegenUtils {
+
+    public static final Set<String> SUPPORTED_SW_EXTENSIONS = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(".sw.json", ".sw.yaml", ".sw.yml")));
```
danielezonca, line 33: We should centralise this somewhere else with a future ticket, same for other generator files. Wdyt?

kogito-codegen-modules/kogito-codegen-api/src/main/java/org/kie/kogito/codegen/api/context/KogitoBuildContext.java
```diff
@@ -73,6 +74,28 @@ default boolean isValidationSupported() {
...
+    /**
+     * <strong>Note: This method is on experimental phase. Can disappear in future releases.</strong>
+     * <p>
+     * Attributes shared among generators and client code.
+     * Any generator can write or read from this context.
+     * </p>
+     * @see ContextAttributesConstants for a list of possible attributes
+     * @return An unmodifiable map with all attributes. If you need to write to the context, use {@link #addContextAttribute(String, Object)}
+     */
+    Map<String, Object> getContextAttributes();
```
danielezonca, line 86: @evacchi 
FYI
I agreed with @ricardozanini to create this map to make generators share info for now and then migrate to symbol table as soon as available.

kogito-codegen-modules/kogito-codegen-api/src/main/java/org/kie/kogito/codegen/api/context/impl/AbstractKogitoBuildContext.java
```diff
@@ -118,6 +123,28 @@ public AppPaths getAppPaths() {
...
+    @Override
+    public <T> T getContextAttribute(String key, Class<T> asClass) {
+        final Object output = this.contextAttributes.get(key);
+        if (output == null) {
+            return null;
+        }
+        if (asClass.isAssignableFrom(output.getClass())) {
+            return asClass.cast(output);
+        }
+        throw new AssertionError("Impossible to cast key value " + output + " as " + asClass.getName());
```
danielezonca, line 140:
```suggestion
        throw new AssertionError("Impossible to cast '" + key + "' key value as " + asClass.getName() + ", found " + output.getClass().getCanonicalName());
```

danielezonca (commented):
Nice work, well done 👍 
I did initial review so added some comments :)

kogito-codegen-modules/kogito-codegen-integration-tests/src/test/java/org/kie/kogito/codegen/openapi/client/OpenApiClientCodegenTest.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.Collections;
+
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+import org.kie.kogito.codegen.AbstractCodegenTest;
+import org.kie.kogito.codegen.api.GeneratedFile;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class OpenApiClientCodegenTest extends AbstractCodegenTest {
```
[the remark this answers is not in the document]
ricardozanini: Oh, I'll expand these unit tests today to do the actual compilation. Otherwise, I'll move them. It's there since you did the modules refactoring.

kogito-codegen-modules/kogito-codegen-openapi/src/main/java/org/kie/kogito/codegen/openapi/client/OpenApiUtils.java
ricardozanini: I didn't know about this method :sweat:

kogito-codegen-modules/kogito-codegen-openapi/src/main/java/org/kie/kogito/codegen/openapi/client/ServerlessWorkflowCodegenUtils.java
ricardozanini: I will open a JIRA to centralize all the core modules' extensions. It's on my plate. wdyt?

kogito-codegen-modules/kogito-codegen-openapi/src/main/java/org/kie/kogito/codegen/openapi/client/OpenApiClientCodegen.java
ricardozanini: Everyone is using `Interger.MAX`, IIRC. :/

kogito-codegen-modules/kogito-codegen-openapi/src/main/java/org/kie/kogito/codegen/openapi/client/OpenApiClientCodegen.java
ricardozanini: I'll take a look. The interface is `String resolve(spec)`. For some resolvers (e.g., classpath or file), it won't "save" the file but only translate it. For HTTP(S), we need to download it. My first plan was to save on OS temp location instead, but we lack control. The file is saved on `target` dir, cleaned after any `mvn clean` command. Plus, a user might want to inspect the downloaded file for debugging reasons.

Not sure if we should delete it. The process is idempotent.

ricardozanini:
> The Linux check has **failed**. Please check [the logs](https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/KIE/job/kogito/job/pullrequest/job/kogito-runtimes/144/console).

LOL

```
java.io.IOException: No space left on device
	at java.base/sun.nio.ch.FileDispatcherImpl.write0(Native Method)
	at java.base/sun.nio.ch.FileDispatcherImpl.write(FileDispatcherImpl.java:62)
	at java.base/sun.nio.ch.IOUtil.writeFromNativeBuffer(IOUtil.java:113)
	at java.base/sun.nio.ch.IOUtil.write(IOUtil.java:79)
	at java.base/sun.nio.ch.FileChannelImpl.write(FileChannelImpl.java:280)
```

kogito-quarkus-extension/deployment/src/test/java/org/kie/kogito/quarkus/serverless/workflow/OpenApiClientTest.java
```diff
@@ -0,0 +1,53 @@
...
+import org.jboss.shrinkwrap.api.asset.EmptyAsset;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import static io.restassured.RestAssured.given;
+
+@Disabled("Need to add resteasy http client to the test classpath in order to compile the generated code")
+public class OpenApiClientTest {
```
danielezonca, line 30: I think it is better to move this test to [integration-tests-quarkus-processes](https://github.com/kiegroup/kogito-runtimes/tree/master/integration-tests/integration-tests-quarkus-processes) module (you can create an integration test for SprintBoot too if you need).

ricardozanini: Oh, this slipped. I just added it for debugging purposes.

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/WorkItemNodeVisitor.java
```diff
@@ -80,17 +81,14 @@ public void visitNode(String factoryField, T node, BlockStmt body, VariableScope
...
-                .computeIfAbsent(workName, RestTaskDescriptor::generateHandlerClassForService);
+        if (TaskDescriptorBuilder.isBuilderSupported(workName)) {
+            final TaskDescriptor taskDescriptor = new TaskDescriptorBuilder(workName)
+                    .withProcessMetadata(metadata)
+                    .withWorkItemNode(node)
+                    .withClassloader(contextClassLoader)
+                    .build();
+            workName = taskDescriptor.getName();
+            metadata.getGeneratedHandlers().put(workName, taskDescriptor.generateHandlerClassForService());
         }
```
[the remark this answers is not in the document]
ricardozanini: Oh! Brain fart, I should've done it.

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/descriptors/TaskDescriptorBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+    private ClassLoader classLoader;
+    private ProcessMetaData metadata;
+
+    public TaskDescriptorBuilder(final String descriptorName) {
+        this.descriptorName = descriptorName;
+    }
+
+    public static boolean isBuilderSupported(final String descriptorName) {
+        return SUPPORTED_DESCRIPTORS.contains(descriptorName);
+    }
```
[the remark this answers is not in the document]
ricardozanini: I'll refactor it again. :D

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/WorkItemNodeVisitor.java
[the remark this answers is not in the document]
ricardozanini: nvm, I kept the same algorithm. If it's not supported, it's not supposed to create a service task, we are fine.

ricardozanini:
Not sure why these files have 0% coverage:

```
kogito-workitems/kogito-openapi-workitem/src/main/java/org/kogito/workitem/openapi/JsonNodeParameterResolver.java
kogito-workitems/kogito-openapi-workitem/src/main/java/org/kogito/workitem/openapi/JsonNodeParser.java
kogito-workitems/kogito-openapi-workitem/src/main/java/org/kogito/workitem/openapi/JsonNodeResultHandler.java
```

Since this is  a new module, should I do something about it? Does someone know what I have to do?

tsurdilo (approved):

tsurdilo:
@ricardozanini great job!! 💯

cristianonicolai (approved):
nice work @ricardozanini !

danielezonca (approved):
Thanks for the PR, approving 👍 
What about including as README of `open-api-codegen` module the description of the PR?

kogito-build-parent/pom.xml
```diff
@@ -175,7 +177,7 @@
     <version.org.infinispan.protostream>4.3.4.Final</version.org.infinispan.protostream>
     <version.org.infinispan.starter>2.3.3.Final</version.org.infinispan.starter>
     <version.org.jboss.spec.javax.ws.rs.jboss-jaxrs-api_2.1_spec>1.0.1.Final</version.org.jboss.spec.javax.ws.rs.jboss-jaxrs-api_2.1_spec>
-    <version.org.jboss.resteasy>4.6.0.Final</version.org.jboss.resteasy>
+    <version.org.jboss.resteasy>4.5.9.Final</version.org.jboss.resteasy>
```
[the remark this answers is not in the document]
ricardozanini: https://github.com/quarkusio/quarkus/blob/1.11/bom/application/pom.xml#L21

ricardozanini:
For some reason, my description update with the follow-up JIRAS was lost. But here they are https://issues.redhat.com/browse/KOGITO-3915.

ricardozanini:
@danielezonca I've removed the `package` modifiers and added the README to describe the new codegen module. Thanks!

evacchi (approved):

evacchi:
Please notice that we are downgrading master to Quarkus 1.11 so if there are Classpath related issues I'd invite to verify if those still hold rebasing on top of https://github.com/kiegroup/kogito-runtimes/pull/1115. This has a high priority as it should go in 1.4.x /cc @radtriste @spolti

ricardozanini:
@evacchi I think we can wait, yes. Also, I've removed the `resteasy` dependencies using `exclusions` on `jbpm-serverless-workflow` module since it's used only by the end-user project. That dependencies were messing with the DMN integration tests.

ricardozanini:
@danielezonca, the CI is failing because of this: 
https://github.com/kiegroup/kogito-runtimes/blob/master/addons/cloudevents/cloudevents-spring-boot-addon-it/pom.xml#L37-L38

It looks like the local repo doesn't have the updated Kogito BOM with the new openapi module, thus the failure on the maven plugin dependency. I checked locally and Maven is downloading the BOM file from the jboss repository instead of the local `.m2`. I don't see a reason to configure this option in the invoker since it's using the same .m2 repository from the machine where the build is running. :thinking: 

I just pushed a new commit without this configuration to verify if it works. The error vanished locally. If you're ok with this change, I can remove this configuration for good.

danielezonca:
> @danielezonca, the CI is failing because of this:
> https://github.com/kiegroup/kogito-runtimes/blob/master/addons/cloudevents/cloudevents-spring-boot-addon-it/pom.xml#L37-L38
> 
> It looks like the local repo doesn't have the updated Kogito BOM with the new openapi module, thus the failure on the maven plugin dependency. I checked locally and Maven is downloading the BOM file from the jboss repository instead of the local `.m2`. I don't see a reason to configure this option in the invoker since it's using the same .m2 repository from the machine where the build is running. 🤔
> 
> I just pushed a new commit without this configuration to verify if it works. The error vanished locally. If you're ok with this change, I can remove this configuration for good.

I think it is related to the same issue reported here
https://github.com/kiegroup/kogito-runtimes/pull/1051
Try to apply the change in your branch and see if CI becomes green.
We don't agree on that PR (see comments) but at least we can check if it is the same problem

ricardozanini:
> I think it is related to the same issue reported here
> #1051
> 
> @cristianonicolai
> Can you please have a look to the changes done by Ricardo with this commit [377066d](https://github.com/kiegroup/kogito-runtimes/commit/377066d6d0cb715c60fe405183f98bf9ab0b5ae4) ?

It's working now. I'm willing to make my changes permanent and rebase with Quarkus downgrade. @cristianonicolai @danielezonca wdyt?

ricardozanini:
Jenkins retest please

ricardozanini:
@danielezonca @cristianonicolai once you guys wake up, mind taking a look? I think we should be fine to merge.

ricardozanini:
Jenkins retest this please

radtriste:
@tsurdilo @ricardozanini what is the targeted version for this PR ?

addons/cloudevents/cloudevents-spring-boot-addon-it/pom.xml
```diff
@@ -34,8 +34,6 @@
             <configuration>
                 <streamLogs>true</streamLogs>
                 <cloneProjectsTo>${project.build.directory}/it</cloneProjectsTo>
-                <settingsFile>src/it/settings.xml</settingsFile>
-                <localRepositoryPath>${project.build.directory}/local-repo</localRepositoryPath>
```
evacchi, line 38: remember you'll have to restore these before we can merge

evacchi (changes requested):

evacchi:
I had left a comment about restoring the invoker configuration, but I see that was intended now? @danielezonca

ricardozanini:
> I had left a comment about restoring the invoker configuration, but I see that was intended now? @danielezonca
> 
> > Timeout has been exceeded
> > Finished: ABORTED
> 
> AFAIK the reason for those lines is time outs :/

It's timing out in the last step, testing the persistence with ISPN. I believe it's an infra problem?

ricardozanini:
Jenkins retest this please

ricardozanini:
@cristianonicolai @danielezonca feel free to merge.

PR #1015, opened 2021-01-28
KOGITO-4119: add StringMatchesUUIDPattern to kogito-test-utils

kostola:


kostola:
This is not needed anymore

PR #1031, opened 2021-02-03
KOGITO-4341 Infinispan upgrade to version 12

cristianonicolai:
Assembly:
- https://github.com/kiegroup/kogito-runtimes/pull/1031
- https://github.com/kiegroup/kogito-apps/pull/815
- https://github.com/kiegroup/kogito-examples/pull/692

### DIFF
--- a/integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-kafka-it/pom.xml
+++ b/integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-kafka-it/pom.xml
@@ -18,22 +18,24 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
-    <parent>
-        <groupId>org.springframework.boot</groupId>
-        <artifactId>spring-boot-starter-parent</artifactId>
-        <version>@version.org.springframework.boot@</version>
-    </parent>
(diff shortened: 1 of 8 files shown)


integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-kafka-it/pom.xml
```diff
@@ -30,6 +30,7 @@
 
     <properties>
         <java.version>@maven.compiler.release@</java.version>
+        <version.infinispan.starter>@version.org.infinispan.starter@</version.infinispan.starter>
```
evacchi, line 33: would it make sense to create `integration-tests-springboot-infinispan-it` instead?

cristianonicolai: I would say so, as a follow up of the work @tiagodolphine started on https://github.com/kiegroup/kogito-runtimes/pull/1226

nmirasch:
@cristianonicolai I tried this upgrade with 'process-usertasks-quarkus-with-console' and it works really fine..  I could create a new instances, data-index service indexed them and the console was able to execute the query. Even infinispan throws this message: 

[org.infinispan.encoding.impl.StorageConfigurationManager] ISPN000599: Configuration for cache 'hiring_store' does not define the encoding for keys or values. If you use operations that require data conversion or queries, you should configure the cache with a specific MediaType for keys or values.

PR #1037, opened 2021-02-04
KOGITO-4353 Add missing Jackson dependencies

gsmet:
This should fix the native build with Quarkus master.

https://issues.redhat.com/browse/KOGITO-4353

### DIFF
--- a/kogito-build-parent/pom.xml
+++ b/kogito-build-parent/pom.xml
@@ -388,6 +388,31 @@
         <version>${version.com.fasterxml.jackson}</version>
       </dependency>
       <dependency>
+        <groupId>com.fasterxml.jackson.datatype</groupId>
+        <artifactId>jackson-datatype-jdk8</artifactId>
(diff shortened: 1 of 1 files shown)


evacchi:
ok to test

evacchi:
@radtriste

gsmet:
So to give more details about this: in the Quarkus IT, when using the master branch of Quarkus, you have a mix of Jackson artifacts in 2.11 (versions coming from Kogito) and some in 2.12 (the ones not defined in Kogito build-parent and thus it's Quarkus defining the version).

Given we want to support both Quarkus 1.11 (which uses Jackson 2.11) and Quarkus 1.12 (which uses Jackson 2.12), for now, we have to force the versions of all the Jackson artifacts used in the tests to have consistent Jackson versions.

evacchi (approved):

radtriste (approved):

PR #1042, opened 2021-02-08
[KOGITO-4373] Remove kogito-internal and post kie-api unfork cleanup

danielezonca:
https://issues.redhat.com/browse/KOGITO-4373

## Summary
After `kie-api` unfork there are some additional cleanup that can be performed to isolate usages of old code. The goal of this PR is to remove `kogito-internal` module, remove `kie7` dependencies from `kogito-bom` (so that users will not find them in the dependencyManagement and then update part of the codebase to use the `Kogito*` specific classes (`KogitoProcessInstance` etc)

## Details
- Removed `kogito-internal` module because it is empty now
- Removed kie7 dependencies from kogito-bom, in this way user will not find them in the dependencyManagement and we don't want to use them directly (now where are used version is explicit)
- Removed all `kie-api` and `kie-internal` exclusions
- `KogitoEventListener` now is no more extending `EventListener` so it is decoupled from 7
- `KogitoWorkItemHandler` now is no more extending `WorkItemHandler` so it is decoupled from 7
- `KogitoWorkItemManager` now is no more extending `WorkItemManager` so it is decoupled from 7
- Renamed `org.kie.kogito.process.workitems.KogitoWorkItem` as `org.kie.kogito.process.workitems.InternalKogitoWorkItem` to avoid name clash
- Renamed `org.kie.kogito.process.workitems.KogitoWorkItemManager` as `org.kie.kogito.process.workitems.InternalKogitoWorkItemManager` to avoid name clash
- Updated most of the usages of `org.kie.api.runtime.process.ProcessInstance` to `org.kie.kogito.internal.process.runtime.KogitoProcessInstance` (not updated where I was not sure)
- Same for `WorkItemManager` -> `KogitoWorkItemManager`, `KogitoWorkItem` etc for all the new `Kogito*` classes

## PR lists
- https://github.com/kiegroup/kogito-runtimes/pull/1042
- https://github.com/kiegroup/kogito-apps/pull/662
- https://github.com/kiegroup/kogito-examples/pull/589


danielezonca:
jenkins retest this please

jbpm/jbpm-flow-builder/src/test/java/org/jbpm/integrationtests/ProcessStartTest.java
```diff
@@ -88,23 +90,31 @@ public void testStartConstraintTrigger() throws Exception {
...
-        
+
+		/*
+		 * This test cannot be migrated to use KogitoProcessRuntime because during ProcessRuntime initialization
+		 * multiple listeners are registered and this break the semantic of this test:
+		 * see ProcessRuntimeImpl#initProcessActivationListener() -> startProcessWithParamsAndTrigger (same for LightProcessRuntime)
+		 * This listener triggers a start process when the insert is performed (but no rules are fired) that is an expected behavior
+		 * for BPMN2 processes (see org.jbpm.bpmn2.StartEventTest#testConditionalStart() ) while for DRF processes produces an additional
+		 * unexpected (broken) execution
+		 */
```
danielezonca, line 106: @cristianonicolai @krisv 
I found a different behaviour of DRF processes compared to BPMN2 (a bug?) and added this comment because it is not possible to migrate to `KogitoProcessRuntime` because of that.
Personally I don't know if we want to consider it dead/legacy code or if we want to fix it.
Let me know if I should raise a Jira or not

danielezonca:
jenkins retest this please

jbpm/jbpm-flow-builder/src/test/java/org/jbpm/integrationtests/ProcessStartTest.java
cristianonicolai: @danielezonca sounds like legacy to me, @krisv wdyt?

cristianonicolai (approved):
@danielezonca looks great, thanks!

danielezonca:
jenkins retest this please

mariofusco (approved):

danielezonca:
jenkins retest this please

danielezonca:
jenkins retest this please

danielezonca:
jenkins retest this please

tiagodolphine (approved):

danielezonca:
jenkins retest this please

danielezonca:
jenkins retest this please

danielezonca:
Manually killed, try again...

jenkins retest this please

danielezonca:
jenkins retest this please

jbpm/process-workitems/src/main/java/org/kie/kogito/process/workitems/InternalKogitoWorkItemManager.java
```diff
@@ -19,25 +19,32 @@
 import java.util.Set;
 
 import org.drools.core.process.instance.WorkItem;
+import org.kie.api.runtime.process.WorkItemHandler;
+import org.kie.kogito.internal.process.runtime.KogitoWorkItemHandler;
 
-public interface KogitoWorkItemManager extends org.drools.core.process.instance.WorkItemManager, org.kie.kogito.internal.process.runtime.KogitoWorkItemManager {
+public interface InternalKogitoWorkItemManager extends org.drools.core.process.instance.WorkItemManager, org.kie.kogito.internal.process.runtime.KogitoWorkItemManager {
```
MarianMacik, line 25: Maybe I would do it the other way, that `InternalKogitoWorkItemManager` resides in the package which has `internal` in its name, but up to you.

MarianMacik (approved):

api/kogito-api/pom.xml
```diff
@@ -55,12 +55,7 @@
     <dependency>
       <groupId>org.kie</groupId>
       <artifactId>kie-dmn-api</artifactId>
-      <exclusions>
-        <exclusion>
-          <groupId>org.kie</groupId>
-          <artifactId>kie-internal</artifactId>
-        </exclusion>
-      </exclusions>
+      <version>${version.org.kie7}</version>
```
MarianMacik, line 58: Why is this no longer handled by build parent?

jbpm/process-workitems/src/main/java/org/kie/kogito/process/workitems/InternalKogitoWorkItemManager.java
danielezonca: Yes I agree it is misleading but I don't think we can improve it for now: `KogitoWorkItemManager` is in an `internal` package of the `public` API (`kogito-api`) because it is not yet a stable API for user while `InternalKogitoWorkItemManager` is an jbpm specific interface.
I expect in the future to move `KogitoWorkItemManager` out of `internal` package (or replace it with a new public stable API) and keep `InternalKogitoWorkItemManager` as impl detail of process core

api/kogito-api/pom.xml
danielezonca: I did this change on purpose because we have modules with the same artifactId of kie7 but different groupid (i.e. `org.drools:drools-core` and `org.kie.kogito:drools-core`) both available as managed dependencies so it is quite error prone to include the wrong one (I found at least two wrong usages).
Now kie7 dependencies are not included anymore in the dependency management so you need to explicitly declare one of them if and when you really need it and use Kogito "alternative" otherwise.
I'm open to revisit this with another ticket if needed but in any case I would prefer we create a specific `kie7-bom` to explicitly include where necessary instead of use the same shared `kogito-build-parent`.
/cc @evacchi @mareknovotny

MarianMacik: OK, makes sense.

PR #1051, opened 2021-02-09
[KOGITO-4404] Fix spring boot testing environment

fjtirado:
Removing mirror section from settings.xml so dependencies are checked
from external repo if not present

### DIFF
--- a/integration-tests/integration-tests-springboot/src/it/settings.xml
+++ b/integration-tests/integration-tests-springboot/src/it/settings.xml
@@ -34,11 +34,4 @@
             </pluginRepositories>
         </profile>
     </profiles>
-    <mirrors>
-        <mirror>
-            <id>local.mirror</id>
-            <mirrorOf>*</mirrorOf>
-            <url>@localRepositoryUrl@</url>
-        </mirror>
-    </mirrors>
 </settings>


cristianonicolai:
as mentioned in the JIRA this will cause Maven to download artefacts again and would slow down the build.

fjtirado:
> as mentioned in the JIRA this will cause Maven to download artefacts again and would slow down the build.

 I'm aware of that, but with current mirror settings.xml, depending on the content of you local maven repo, the test of spring boot integration might fail (Since I spend 4 hours to find out the reason, I guess that might be an issue for future contributors)

danielezonca:
> as mentioned in the JIRA this will cause Maven to download artefacts again and would slow down the build.

@cristianonicolai 
I agree that this PR introduce a CI performance regression but at the moment the problem is that if you start with a clean environment (empty m2) and you compile it fails because of spring boot started and then many other jars (for example `maven-jar` version 3.2 instead of 3.1).
Any idea about how to improve it? What about document in contribution file how to "init" the environment? Personally I don't like it as solution but at least it could help to understand what is going on.
Wdyt?

jstastny-cz:
Confirming that this affect users building just the integration-tests package. Could not find the spring-boot-starter-parent. This is blocking us in running these tests with productized binaries. Would be good to resolve somehow. Please include me in the discussion if there's any.

fjtirado:
In my opinion, I think we should just merge this PR (which will fix the issue building the environment) and assume the performance cost (which is not really a big deal, in my opinion)

jstastny-cz:
@fjtirado is there a reason then to actually keep the settings.xml? Maven ivoker should be able to set the local repository just by setting the configuration property

localRepositoryPath | File | 1.2 | The path to the local repository into which the project artifacts should be installed for the integration tests.
-- | -- | -- | --

fjtirado:
> @fjtirado is there a reason then to actually keep the settings.xml? Maven ivoker should be able to set the local repository just by setting the configuration property
> localRepositoryPath 	File 	1.2 	The path to the local repository into which the project artifacts should be installed for the integration tests.

I guess yes, it can be removed. I was just doing the least minimum change to make my environment work (this is the first time I interact with maven invoker)

r00ta:
Jenkins retest this please

Hi there, we just added a maven plugin to check the codestyle. I retrigger CI because of that

cristianonicolai:
jenkins please retest this

jstastny-cz:
@fjtirado wouldn't the solution be instead of changing the settings.xml to actually make sure the dependecnies are in the local repository and can be downloaded?
I've made the integration-tests-springboot work by adding following dependencies into its pom.xml:
```
    <dependency>
      <artifactId>process-management-springboot-addon</artifactId>
      <scope>test</scope>
    </dependency>
    <dependency>
      <artifactId>task-management-springboot-addon</artifactId>
      <scope>test</scope>
    </dependency>
    <dependency>
      <artifactId>kogito-test-utils</artifactId>
      <scope>test</scope>
    </dependency>
    <dependency>
      <artifactId>kogito-cloudevents-spring-boot-addon</artifactId>
      <scope>test</scope>
    </dependency>
    <dependency>
      <artifactId>infinispan-persistence-addon</artifactId>
      <scope>test</scope>
    </dependency>
```
This makes sure that these are downloaded into local repository before invoker tests are started (and use that settings.xml that takes dependencies from there only).
This way you tell Maven reactor that the order of build should be all deps first and integration-tests after - so it does also allow for running from top-level as `mvn clean verify -pl :integration-tests-springboot -am`.
Could you please check if that solves the issue you were tackling?

fjtirado:
@jstastny-cz
Tried, but similar error (missing dependency depends on what is present in my local repo)
 Failed to execute goal on project integration-tests-springboot-it: Could not resolve dependencies for project org.kie.kogito:integration-tests-springboot-it:jar:2.0.0-SNAPSHOT: The following artifacts could not be resolved: org.hamcrest:hamcrest-core:jar:2.2, org.hamcrest:hamcrest-library:jar:2.2, com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:jar:2.11.0: Could not find artifact org.hamcrest:hamcrest-core:jar:2.2 in local.mirror (file:///home/ftirados/.m2/repository) -> [Help 1]

PR #1052, opened 2021-02-10
bump to kie 7.50.0.t20210209

mariofusco:


triceo:
OptaPlanner build will fail and it is expected - https://github.com/kiegroup/optaplanner/pull/1153 fixes it.

danielezonca (approved):

mariofusco:
All failing tests seems to be in optaplanner. @triceo please give a look and confirm if it is ok to merge this PR.

triceo:
@mariofusco Indeed these failures are expected and the referenced PR fixes them. I will merge it after you merge this.

drools/drools-core-static/src/main/java/org/drools/statics/StaticKieAssemblers.java
```diff
@@ -40,6 +40,10 @@ public StaticKieAssemblers() {
         //         instance("org.kie.dmn.core.assembler.DMNAssemblerService"));
     }
 
+    @Override
+    public void addResourceBeforeRules( Object o, Resource resource, ResourceType resourceType, ResourceConfiguration resourceConfiguration ) throws Exception {
```
evacchi, line 44: I think this should work 

```suggestion
    public void addResourceBeforeRules(Object knowledgeBuilder, Resource resource, ResourceType type, ResourceConfiguration configuration) throws Exception {
        KieAssemblerService assembler = assemblers.get(type);
        if (assembler != null) {
            assembler.addResourceBeforeRules(knowledgeBuilder,
                                             resource,
                                             type,
                                             configuration);
        } else {
            throw new RuntimeException("Unknown resource type: " + type);
        }
    
```

PR #1054, opened 2021-02-10
[KOGITO-4411] null pointer exception when process is started

fjtirado:
Setting variable context from marshaller even if there are no properties

### DIFF
--- a/integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-it/src/test/java/org/kie/kogito/integrationtests/springboot/BasicRestTest.java
+++ b/integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-it/src/test/java/org/kie/kogito/integrationtests/springboot/BasicRestTest.java
@@ -17,6 +17,7 @@
 package org.kie.kogito.integrationtests.springboot;
 
 import java.util.HashMap;
+
 import java.util.Map;
 import java.util.UUID;
 
@@ -33,6 +34,8 @@ import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.Matchers.emptyOrNullString;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
+
+import org.acme.travels.Traveller;
 
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT, classes = KogitoSpringbootApplication.class)
 @ContextConfiguration(initializers = InfinispanSpringBootTestResource.Conditional.class)
@@ -64,6 +67,31 @@ class BasicRestTest extends BaseRestTest {
                 .statusCode(200)
                 .body("id", equalTo(id))
                 .body("var1", equalTo("Kermit"));
+    }
+    
+    @Test
+    void testWithInaccurateModel () {
+        
+        Traveller traveller = new Traveller ("Javierito","Dimequienes","pepe@pepe.com","Spanish", null);
+        String processId = given()
+                .contentType(ContentType.JSON)
+                .body(traveller)
+                .when()
+                .post("/approvals")
+                .then()
+                .statusCode(201)
+                .extract()
+                .path("id");
+        
+        given()
+                .contentType(ContentType.JSON)
+                .queryParam("user", "admin")
+                .queryParam("group", "managers")
+                .pathParam("processId", processId)
+                .when()
+                .get("/approvals/{processId}/tasks")
+                .then()
+                .statusCode(200);
     }
 
     @Test
--- a/jbpm/process-serialization-protobuf/src/main/java/org/jbpm/marshalling/impl/AbstractProtobufProcessInstanceMarshaller.java
+++ b/jbpm/process-serialization-protobuf/src/main/java/org/jbpm/marshalling/impl/AbstractProtobufProcessInstanceMarshaller.java
@@ -802,11 +802,11 @@ public abstract class AbstractProtobufProcessInstanceMarshaller
             }
         }
 
+        Context variableScope = ((org.jbpm.process.core.Process) process)
+                .getDefaultContext( VariableScope.VARIABLE_SCOPE );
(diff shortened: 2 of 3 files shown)


jbpm/process-serialization-protobuf/src/main/java/org/jbpm/marshalling/impl/AbstractProtobufProcessInstanceMarshaller.java
```diff
@@ -802,11 +802,11 @@ public ProcessInstance readProcessInstance(MarshallerReaderContext context) thro
...
 
+        Context variableScope = ((org.jbpm.process.core.Process) process)
+                .getDefaultContext( VariableScope.VARIABLE_SCOPE );
+        VariableScopeInstance variableScopeInstance = (VariableScopeInstance) processInstance
+                .getContextInstance( variableScope );
         if ( _instance.getVariableCount() > 0 ) {
-            Context variableScope = ((org.jbpm.process.core.Process) process)
-                    .getDefaultContext( VariableScope.VARIABLE_SCOPE );
-            VariableScopeInstance variableScopeInstance = (VariableScopeInstance) processInstance
-                    .getContextInstance( variableScope );
```
danielezonca, line 809: Can you add a test?
Btw as far as I can see in this way you are preventing the NPE but `variableScope` can be `null` right?

fjtirado: Sure, test added (it was raising null pointer exception before, now it goes through) 
yes, I prefer, in order to avoid  refactoring all the context part, just set variable context when deserializing in any case, not only when there are variables found.

danielezonca (approved):
Thanks for the PR approving 👍

tiagodolphine (commented):
LGTM, and @fjtirado any idea why this was this issue happening only in Spring and not in Quarkus?

fjtirado:
> LGTM, and @fjtirado any idea why this was this issue happening only in Spring and not in Quarkus?

The only explanation that occurs to me is that when using Quarkus, processInstance is not deserialized, but I did not investigate why.

danielezonca:
> > LGTM, and @fjtirado any idea why this was this issue happening only in Spring and not in Quarkus?
> 
> The only explanation that occurs to me is that when using Quarkus, processInstance is not deserialized, but I did not investigate why.

It might be related to protogenerator different impls 🤔

integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-it/src/test/java/org/kie/kogito/integrationtests/springboot/BasicRestTest.java
```diff
@@ -65,6 +68,31 @@ void testGeneratedId() {
                 .body("id", equalTo(id))
                 .body("var1", equalTo("Kermit"));
     }
+    
+    @Test
+    void testWithEmptyModel () {
```
cristianonicolai, line 73: what do you mean by empty model? a task that doesn't have any input params? Just confused because you seen to be sending the traveller object as model anyway.

integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-it/src/test/java/org/kie/kogito/integrationtests/springboot/BasicRestTest.java
```diff
@@ -34,6 +35,8 @@
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
 
+import org.acme.travels.Traveller;
+
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT, classes = KogitoSpringbootApplication.class)
```
cristianonicolai, line 40: @fjtirado is this only reproducible in Spring, should a test be also added for Quarkus?

cristianonicolai: Seems that's the case, but anyway should we add in there to be on the safe side?

integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-it/src/test/java/org/kie/kogito/integrationtests/springboot/BasicRestTest.java
fjtirado: The other option was wrong, whatever the case is confusing. the point is to manage to have an empty model when serializing and this can be achieved by passing an object in the wrong position of the hierarchy

integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-it/src/test/java/org/kie/kogito/integrationtests/springboot/BasicRestTest.java
fjtirado: Yes, I will add. 
I did not initially to not have a comment "why this is needed?" ;)

danielezonca (approved):

tiagodolphine (approved):

fjtirado:
> > > LGTM, and @fjtirado any idea why this was this issue happening only in Spring and not in Quarkus?
> > 
> > 
> > The only explanation that occurs to me is that when using Quarkus, processInstance is not deserialized, but I did not investigate why.
> 
> It might be related to protogenerator different impls thinking

I was doing some test regarding this. When debugging an example  in quarkus with persistence add enable,  I got the same null pointer exception than with spring boot. So I guess when quarkus integration tests were run in Jenkins I guess  persistence profile was not being used for some unknown reason. If I try to run quarkus integration test locally with -Ppersistence the test broke during code generation phase, so I suspect this profile is disable in jenkins.

fjtirado:
@kiegroup/gatekeepers please merge this one

MarianMacik (approved):

PR #1039, opened 2021-02-05
KOGITO-4357 - Add formatter plugin + reformat codebase

r00ta:
JIRA: https://issues.redhat.com/browse/KOGITO-4357

Hi @danielezonca @mariofusco @cristianonicolai @ricardozanini @triceo @evacchi , this is a follow up of the proposal made here https://github.com/kiegroup/kogito-runtimes/pull/1022. 

I configured the plugins `formatter-maven-plugin` and `impsort-maven-plugin` so that: 
1) the CI checks if the codebase is properly formatted, the imports are sorted and there are no unused imports.
2) The code is automatically formatted every time it is compiled 
3) You can check if your codebase is formatted appending `-Dvalidate-formatting` to your mvn command
4) You can avoid formatting checks appending `-Dformatter.skip=true` to your mvn command 

The ide configuration is the same of `Optaplanner`: see https://github.com/kiegroup/optaplanner/blob/master/build/optaplanner-ide-config/src/main/resources/eclipse-format.xml . 
We can discuss if we want to change some rules. I personally think the Optaplanner one is very reasonable. 

The contributors have to read the README at `kogito-ide-config/README.md` to setup eclipse or intellij properly. 

## Related PRs: 
kogito-apps: https://github.com/kiegroup/kogito-apps/pull/625
kogito-examples: https://github.com/kiegroup/kogito-examples/pull/551

Many thanks for submitting your Pull Request :heart:! 

Please make sure that your PR meets the following requirements:

- [x] You have read the [contributors guide](CONTRIBUTING.md)
- [x] Your code is properly formatted according to [this configuration](https://github.com/kiegroup/droolsjbpm-build-bootstrap/tree/master/ide-configuration)
- [x] Pull Request title is properly formatted: `KOGITO-XYZ Subject`
- [x] Pull Request title contains the target branch if not targeting master: `[0.9.x] KOGITO-XYZ Subject`
- [x] Pull Request contains link to the JIRA issue
- [x] Pull Request contains link to any dependent or related Pull Request
- [x] Pull Request contains description of the issue
- [x] Pull Request does not include fixes for issues other than the main ticket

PR #1075, opened 2021-02-16
KOGITO-4376 Enable Copyright validation during maven build and fix in…

cristianonicolai:
…valid files

Assembly:
- https://github.com/kiegroup/kogito-runtimes/pull/1075
- https://github.com/kiegroup/kogito-examples/pull/566

### DIFF
--- a/kogito-build-parent/pom.xml
+++ b/kogito-build-parent/pom.xml
@@ -23,6 +23,7 @@
       - properties must be sorted alphabetically (other form of sorting were found to be unclear and ambiguous).
   -->
   <properties>
+    <checkstyle.failOnViolation>true</checkstyle.failOnViolation>
     <checkstyle.header.template><![CDATA[
 ^\/\*$\n^
 \* Copyright \d\d\d\d Red Hat, Inc\. and\/or its affiliates\.$\n^
@@ -42,6 +43,7 @@
 ]]>
     </checkstyle.header.template>
     <checkstyle.header.extensions>java</checkstyle.header.extensions>
+    <checkstyle.logViolationsToConsole>true</checkstyle.logViolationsToConsole>
 
     <!-- container images for testing -->
     <container.image.infinispan>infinispan/server:${version.org.infinispan}</container.image.infinispan>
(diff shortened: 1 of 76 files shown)


MarianMacik (approved):
Looks good to me. BTW why don't we use the license plugin also for check? Is it that slow when it compares the year against git?

cristianonicolai:
@MarianMacik the license git plugin showed to be problematic and potentially slow ( we don't need to revalidate the year during builds ), there are caveats to how the year is retrieved, see https://github.com/mycila/license-maven-plugin/tree/master/license-maven-plugin-git#noteworthy and also concerns regarding the breaking PRs during end of the year.

evacchi (approved):

kogito-build-parent/pom.xml
```diff
@@ -23,6 +23,7 @@
       - Version properties must be sorted alphabetically (other form of sorting were found to be unclear and ambiguous).
   -->
   <properties>
+    <checkstyle.failOnViolation>true</checkstyle.failOnViolation>
```
danielezonca, line 26: I'm fine to enable this check by default but I think we can disable it in daily builds to avoid that snapshots are not generated because of a malformed header.
Wdyt?
In general this should never happen so it should be safe to disable

cristianonicolai: In theory, this should never happen, maybe now when we get the first version in as some PRs might not be updated but still green. I think I would still prefer to have the build consistent between different runs, `mvn clean install` to get it all done in the same way as locally. We can always review this if becomes problematic.

danielezonca: Yes in general my concern is when we merge a PR that is green but it has not been compiled since a while: if automatic merge works we have no warning from GitHub but then the final output could be not valid from style perspective.
Now, considering that this PR only cover license check I agree that it should not be an issue but for example the same approach applied for formatting plugin ( Jacopo's PR https://github.com/kiegroup/kogito-runtimes/pull/1039 ) could break more frequently.
Btw fine for me to proceed and then we can revisit if this generate any problem to master :)

danielezonca (approved):

cristianonicolai:
jenkins please retest this

PR #1099, opened 2021-02-25
KOGITO-4545 Disable header validation for generated sources and fix t…

cristianonicolai:
…emplates to create valid header


nmirasch (approved):

danielezonca (commented):
@cristianonicolai 
Thanks for the PR, can you please remove license from templates? I don't think they are necessary at all

cristianonicolai:
@danielezonca agree but that will require some work. If fine with you, I would like to just fix the examples and look into that in a follow-up PR/JIRA.

danielezonca:
> @danielezonca agree but that will require some work. If fine with you, I would like to just fix the examples and look into that in a follow-up PR/JIRA.

Sure, fine for me 👍

danielezonca (approved):

MarianMacik (approved):
FYI it worked for me (both the check and formatting) even without package declarations.

PR #1109, opened 2021-03-03
KOGITO-4439 Allow plugging UnitOfWorkEvent listeners

cristianonicolai:


api/kogito-api/src/main/java/org/kie/kogito/uow/events/UnitOfWorkEventListener.java
```diff
@@ -0,0 +1,32 @@
...
+    default void onStartEvent(UnitOfWorkStartEvent event) {
+        //NO-OP
+    }
+
+    default void onEndEvent(UnitOfWorkEndEvent event) {
+        //NO-OP
+    }
+
+    default void onAbortEvent(UnitOfWorkAbortEvent event) {
+        //NO-OP
```
MarianMacik, line 30: Maybe a JavaDoc here could clarify that onStartEvent is emitted before the actual start of the `UnitOfWork` and the other two events are emitted after the end or abort has happened.

MarianMacik (approved):
Looks good to me, maybe a JavaDoc can be added to the listener interface.

api/kogito-api/src/main/java/org/kie/kogito/uow/events/UnitOfWorkEventListener.java
tiagodolphine: maybe `onBeforeStartEvent`, but what if we change the order like the event is called after each method execution in `ManagedUnitOfWork`. Anyway, this is not consistent, because the start event is called before and the others are called after....

tiagodolphine (approved):
@cristianonicolai LGTM, just a minor comment related to the order the events are published, to have a consistent approach like after the operation execution or before, anyway and besides, we could add the javadoc on them.

api/kogito-api/src/main/java/org/kie/kogito/uow/events/UnitOfWorkEventListener.java
tiagodolphine: +1 for a javadoc for each method.

cristianonicolai: I renamed the methods to make it explicit that is fired before and after.

cristianonicolai:
jenkins please retest this

cristianonicolai:
jenkins please retest this

cristianonicolai:
jenkins please retest this

PR #1112, opened 2021-03-03
KOGITO-4569: Additional tests

jomarko:
Additional tests of executing models containing:
- Relation Expression
- Collect Hit Policy Decision Table Expressions
- Multiple DRDs
- DMN model including another DMN model

For more details see: https://issues.redhat.com/browse/KOGITO-4569

- [ ] Your code is properly formatted according to [this configuration](https://github.com/kiegroup/kogito-runtimes/tree/master/kogito-ide-config)

### DIFF
--- a/integration-tests/integration-tests-quarkus-decisions/pom.xml
+++ b/integration-tests/integration-tests-quarkus-decisions/pom.xml
@@ -93,6 +93,12 @@
       </exclusions>
     </dependency>
 
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <scope>test</scope>
+    </dependency>
+
     <!-- this is used implicitly by quarkus tests so let's make Maven aware of it -->
     <dependency>
       <groupId>org.kie.kogito</groupId>
--- a/integration-tests/integration-tests-quarkus-decisions/src/main/java/org/kie/kogito/integrationtests/quarkus/FaceMaskTestData.java
+++ b/integration-tests/integration-tests-quarkus-decisions/src/main/java/org/kie/kogito/integrationtests/quarkus/FaceMaskTestData.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.kogito.integrationtests.quarkus;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class FaceMaskTestData {
+    @JsonProperty(value = "Human")
+    private Human human;
+
+    @JsonProperty(value = "Location")
+    private String location;
+
+    public Human getHuman() {
+        return human;
+    }
+
+    public void setHuman(Human human) {
+        this.human = human;
+    }
+
+    public String getLocation() {
+        return location;
+    }
+
+    public void setLocation(String location) {
+        this.location = location;
+    }
+}
--- a/integration-tests/integration-tests-quarkus-decisions/src/main/java/org/kie/kogito/integrationtests/quarkus/HospitalStatusTestData.java
+++ b/integration-tests/integration-tests-quarkus-decisions/src/main/java/org/kie/kogito/integrationtests/quarkus/HospitalStatusTestData.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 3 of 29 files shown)


jstastny-cz (approved):
LGTM, thank you.

jstastny-cz:
Just one note to reported code-coverage, we might exclude the `integration-tests` folder from the assessment.

jomarko:
- https://github.com/kiegroup/kogito-runtimes/pull/1112#pullrequestreview-607427691 @kiegroup/gatekeepers  please merge
- https://github.com/kiegroup/kogito-runtimes/pull/1112#issuecomment-793928151 @rsynek could you please point me to some conf. where I should exclude some folder/module from sonar chekcs?

rsynek:
> @jomarko I think this should help placing into integration-tests/pom.xml
> 
> ```
>   <profiles>
>     <profile>
>       <id>sonarcloud-analysis</id>
>       <properties>
>         <sonar.exclusions>**/*</sonar.exclusions>
>       </properties>
>     </profile>
>   </profiles>
> ```

Yes, this should work. Just one nitpick comment: not sure that wrapping the property in a profile is worth the maintenance effort (think keeping the profile ID in sync across modules).

danielezonca:
Hi @jomarko 
Thanks for the PR, I agree that a DMN model that includes another DMN model is a valid additional scenario to cover because Quarkus/Spring are new runtimes so we need to make sure classpath handling/resource wiring to be correct.

It is not clear to me the reason of the other tests (DRD, Hit policies, etc): Kogito just include `dmn-core` from `drools.git` repo so we don't need to check the behaviour of the engine itself of a single DMN execution.
Wdyt?
/cc @tarilabs 

Side note, please duplicate Quarkus tests for Native execution (see [this example](https://github.com/kiegroup/kogito-runtimes/blob/master/kogito-quarkus-parent/kogito-quarkus-decisions-extension/kogito-quarkus-decisions-integration-test/src/test/java/org/kie/kogito/quarkus/dmn/NativeDMNIT.java)). You need to manually run native compilation because we only have a job on master for now

jstastny-cz:
@danielezonca Unless there are specific tests for the mentioned areas and also more complex models somewhere in Kogito code-base that we're not aware of, this seems like a coverage gap.
Principally we need to test also more complex cases in Kogito, even though we have the tests somewhere else. Integration tests should also guard against version mismatch and similar, so we thought this would be a good place to put these.

tarilabs:
I fully agree with @danielezonca .
DRD is not something to be tested in `kogito-runtimes` pertains only to the editor/tooling, and we have already plenty of test on the engine codebase repo.
HitPolicies, etc. has also engine semantic, and that is fully tested since years now on the engine codebase.
As Daniele pointed out, having a DMN-Import test is good since this is the only real gap of how Kogito codegen uses a DMN Engine feature of the builder (so we have import test on engine codebase, and it is to make sure it works here for kogito too, since that is something like an SPI).

jstastny-cz:
@tarilabs @danielezonca I do interpret your comments as "put your tests elsewhere", in that case, could you advise on correct location for these (in your view redundant) test cases? Any guidance much appretiated.

danielezonca:
> @tarilabs @danielezonca I do interpret your comments as "put your tests elsewhere", in that case, could you advise on correct location for these (in your view redundant) test cases? Any guidance much appretiated.

Well my comment is more that these tests (all expect the DMN import one) already exist inside `dmn-core` so we should be already ok without additional coverage.

From a hierarchy perspective we have:
- Kogito endpoint (kogito-runtime) that just marshall/unmarshall data
- Kogito API (kogito-runtime) (`DecisionModel`) that just delegate to dmn impl (`DMNRuntime` and `DMNModel`)
- DMNRuntime + DMNModel (dmn-core) that actually execute the code and that is already well covered inside `drools.git`

While the way to DMN files are linked together in Kogito is a bit different so I think that test improves the coverage, I don't see how the other checks/examples can influence dmn-core behaviour

jomarko:
@tarilabs @danielezonca hello. I can simplify this PR to contain just tests of included models feature. However I am not sure I understand non agreement with adding tests into our **integration-tests** module.

**Cons** from my point of view
- increased build/test time: I think this is something we can not avoid with any PR merging into kiegroup. For sure there needs to be a balance. That is why I added just 5 DMN models and will not add more.
- increased risk of random fail, as more non trivial tests we have, the chance for fail is higher. However added tests follow structure of existing quarkus and springboot tests. If the original tests are stable, I do not expect added tests will make a difference.

**Pros** from my point of view
- we increase ratio of tests, where we "eat our own food". I produced DMN models using current VS Code tooling, while ~half of existing integration-tests onsume third party generated DMN models. I see this is useful as confirmation kogito-runtimes-xxx is fine running models produced by vscode-extension-yyy
- we decrease risk of regression for kogito users, if DMN standard will change behavior of 'Relation' (e.g. minimal amount of columns will be changed), 'DRDs' (e.g. it will be mandatory belonging to DRD's for each node), 'Hit Policy Collecting Values' (e.g. some collect strategy will be deprecated). I am not saying any of these listed changes is probable. However as drools repository and kogito-runtimes repository are two separate builds, from my point of view adding few tests into kogito-runtimes should help to reveal an inconsistency sooner

I understand everyone in this conversation would list different **Pros** and **Cons**, however as we all collaborate on open source project, where anyone from world has an opportunity to contribute, I would like to understand if really this PR has more **Cons** than **Pros**  before I do changes to the PR.

jomarko:
@danielezonca https://github.com/kiegroup/kogito-runtimes/pull/1112#issuecomment-794033677 - native version of test added.

baldimir:
"Kogito API (kogito-runtime) (DecisionModel) that just delegate to dmn impl (DMNRuntime and DMNModel)" - are we sure this is tested well? Also from regression point of view?

danielezonca (approved):
As agreed with a call, I'm fine to proceed 👍

PR #1117, opened 2021-03-05
KOGITO-4594 kogito-runtimes has dependency for narayana

evacchi:
Exclude useless transitive dependency on narayana-bom

### DIFF
--- a/kogito-bom/pom.xml
+++ b/kogito-bom/pom.xml
@@ -22,6 +22,12 @@
         <groupId>org.kie.kogito</groupId>
         <artifactId>kogito-api</artifactId>
         <version>${project.version}</version>
+        <exclusions>
+          <exclusion>
+            <groupId>org.kie</groupId>
+            <artifactId>narayana-integration-bom</artifactId>
+          </exclusion>
+        </exclusions>
       </dependency>
       <dependency>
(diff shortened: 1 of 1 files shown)


danielezonca (commented):
What about create a `kie7-bom` where we declare all 7 modules with exclusions (if needed) instead of this?
The current PR is probably solving the issue because `kie-api` is a transitive dependency if many modules and it is enough to reach any kie7 API to have again to add a similar exclusion

jomarko:
reading the discussion on jira, I think we should close his PR, am I right?

MarianMacik:
@danielezonca As I said in the JIRA, even if you create such a bom, it won't help as the pulled dependencies are BOMs (Narayana BOM, Thorntail BOM etc.) which Maven needs to download in order to properly resolve kie-parent.

mareknovotny:
First the idea of BOM is just for dependency management not for including dependency into build chain.
Second we know that drools or kie-api will pull the  v7 dependencies and kie-parent behind it. So better to limit only real dependencies, not the BOM itself.

Is it really problem to let maven resolve BOMs or actual dependencies?

evacchi:
Closing this as it doesn't really solve the issue (well, if any)

PR #1125, opened 2021-03-10
[KOGITO-4593] Improve testing strategy and coverage for Processes

elguardian:
https://issues.redhat.com/browse/KOGITO-4593


danielezonca:
Hi @elguardian 
I noticed this PR and I would like to understand if the plan is to use it to replace process unit tests.

As far as I can see this PR generalised some of the code of `AbstractCodegenTest` with a Junit extension but if the goal is just to test process engine, have you consider to use the approach implemented in [ActivityGenerationModelTest](https://github.com/kiegroup/kogito-runtimes/blob/master/jbpm/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/canonical/ActivityGenerationModelTest.java#L74) unit test?
In this way you avoid the overhead to generate/compile REST endpoints and many other classes and other unnecessary step like extracting `Application` impl

On the other hand if this is intended to be a generic Kogito "framework" to test different resources we should probably generalise it because for example [`KogitoUnitTestContext`](https://github.com/kiegroup/kogito-runtimes/pull/1125/files#diff-9990929638a9e6193b9c21e4afb10d22758b5378a22bea5bb86758d5f20343aa) is tied to processes at the moment.
I personally prefer integration tests to test Kogito because there is a lot of code (REST/Messages/Persistence) that is probably hardly testable with a similar approach :)

Said differently, is this intended to test at "engine" level (bpmn/sw) or at "wiring" level (endpoint)?

/cc @tiagodolphine @evacchi FYI

elguardian:
Hi Daniel, please read the jira and sync with others before this sort of comments. Jira states to intent to migrate test coverage and try to bring the fixing in the component of jbpm-flow (which is not here)

1) this is no replacing any integrations test. This is for engine testing + façade (not ws)
2) this is tight to process because it is the intent of the testing (not from endpoints as beforemention)
3) the activity generation model test has several problem. The biggest one is using inheritance. 
    This causes sustaining problem, code duplications and large etc. you just need to see test coverage, or even integrations testing). Anyone who has performed testing at engine level knows. sometimes you need async + persistence... sometimes you just need persistence... sometimes you need an environment flagged, etc... inheritances is bad for that and should be avoided.
4) testing engine with integration test is not the right approach and it is overkilling. that will cause slowness in the testing. current implementation support parallelization.
5) this is going to be a slow PR.


for instance if yout test the engine alone you would not notice that façade has a limitation over the name (cannot contain "-")
a variable in the process cannot be set as "id" it is reserved in the model
there is a data conversion problem caused by compilation façada as well. 
Some code refactos has changed what is mandatory and what is not regarding for instance script actions (now requires the script)
Also ruleflow groups have been deprecated but the code is still in the engine...
and a long etc...

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstances.java
```diff
@@ -30,6 +30,19 @@
         return values(ProcessInstanceReadMode.READ_ONLY);
     }
 
+    default Optional<ProcessInstance<T>> findByBusinessKey(String businessKey) {
+        return findByBusinessKey(businessKey, ProcessInstanceReadMode.READ_ONLY);
+    }
+
+    default Optional<ProcessInstance<T>> findByBusinessKey(String businessKey, ProcessInstanceReadMode mode) {
+        for(ProcessInstance<T> instance : values(mode)) {
```
cristianonicolai, line 38: @elguardian we should instead implement a method to `findByBusinessKey` if you think that is necessary, simply because the `values` method should be avoided as it loads all process instances.

elguardian: this is WIP. business key is used for instance in PAM for searching process. it works like subrogate key. 
In any case this is frozen till is reached an agreement about all this.

elguardian:
closing this one. As planning to try something more complete.

PR #1122, opened 2021-03-09
KOGITO-4475 Upgrade Spring Boot to 2.3.4

evacchi:
Many thanks for submitting your Pull Request :heart:! 

Please make sure that your PR meets the following requirements:

- [ ] You have read the [contributors guide](CONTRIBUTING.md)
- [ ] Your code is properly formatted according to [this configuration](https://github.com/kiegroup/kogito-runtimes/tree/master/kogito-ide-config)
- [ ] Pull Request title is properly formatted: `KOGITO-XYZ Subject`
- [ ] Pull Request title contains the target branch if not targeting master: `[0.9.x] KOGITO-XYZ Subject`
- [ ] Pull Request contains link to the JIRA issue
- [ ] Pull Request contains link to any dependent or related Pull Request
- [ ] Pull Request contains description of the issue
- [ ] Pull Request does not include fixes for issues other than the main ticket

PR #1128, opened 2021-03-10
Add no-snapshot property 

r00ta:
This PR aims to disable the download of SNAPSHOTS artifacts in CI environment. In addition to that, the setup of the integration tests for springboot is changed: there is no need anymore to download the artifacts upfront and fail if the requested versions are not in the local m2

- [x] Your code is properly formatted according to [this configuration](https://github.com/kiegroup/kogito-runtimes/tree/master/kogito-ide-config)

<details>
<summary>
</summary>

* <b>Pull Request</b>  
 
* <b>Quarkus LTS checks</b>  

</details>


r00ta:
Hi @radtriste @evacchi @cristianonicolai , wdyt about this one? We would remove the setup that force the artifacts to be fetched upfront and disable the download of snapshots. 
By consequence, the bump to the new version of springboot _should be_ painless from now on

evacchi:
sounds great to me. I suck at Jenkins, so I'll accept if @radtriste says that's fine

evacchi (approved):

radtriste:
This is not Jenkins specific I think.

So if I get it right, in PR check, we would no more get snapshots from https://repository.jboss.org/nexus/content/groups/public/ ?

What is the final purpose of it ? I am missing some context on integration-tests I guess ...

Also invoking @mareknovotny here as he may have more knowledge for that

evacchi:
If I recall correctly, integration-tests with invoker plugin does not use the local .m2, so, unless configured differently, if fetches all snapshots _again_ causing the build to time out. 

Current configuration is to use the local .m2 _as if it were a remote repository_. This means that, however, if some new artefacts have not been fetched from the previous modules of the Maven reactor, the nested maven build will fail (artefact missing).

With this change, snapshots are no longer fetched from nexus, so we can always hit a remote repository, and disable the special invoker config. 

@danielezonca may have more details as well.

radtriste:
but in that case we would always download from remote or ?
wouldn't that create some other timeouts ?

evacchi:
that's what I would think, but from the logs timestamps look reasonable to me 🤷

radtriste (approved):
I am ok with the code change itself. Let's see if it works on Edoardo's PR.

Would that create problems in the nightly if we don't use the `no-snapshot` ?

danielezonca:
In general the idea of this change is to disable snapshot fetching during PR builds: all the snapshots should be generated during the build or just fail.
This is based on the assumption that jenkins executor's m2 is empty at the beginning or at least it doesn't contain previous snapshots. @mareknovotny @mbiarnes Is this assumption correct?

danielezonca:
> I am ok with the code change itself. Let's see if it works on Edoardo's PR.
> 
> Would that create problems in the nightly if we don't use the `no-snapshot` ?

Master is supposed to be correct so we can leave the default setup or add the same property, I have no strong opinion

mareknovotny:
> In general the idea of this change is to disable snapshot fetching during PR builds: all the snapshots should be generated during the build or just fail.
> This is based on the assumption that jenkins executor's m2 is empty at the beginning or at least it doesn't contain previous snapshots. @mareknovotny @mbiarnes Is this assumption correct?

i am not sure that all snapshots are built here you need, but generally i agree.

r00ta:
Hi @mareknovotny , if not all the snapshots were not built the CI should have failed or am I missing something?

radtriste:
> In general the idea of this change is to disable snapshot fetching during PR builds: all the snapshots should be generated during the build or just fail.
> This is based on the assumption that jenkins executor's m2 is empty at the beginning or at least it doesn't contain previous snapshots. @mareknovotny @mbiarnes Is this assumption correct?

This should be. If you really want to be sure, then you can use `maven.cleanRepository()` in the `Initialize` stage => https://github.com/kiegroup/jenkins-pipeline-shared-libraries/blob/master/vars/maven.groovy#L138

evacchi:
failure on https://github.com/kiegroup/kogito-runtimes/pull/1122 is unrelated, shall we merge both? @mareknovotny @danielezonca @radtriste ?

radtriste:
shouldn't https://github.com/kiegroup/kogito-runtimes/pull/1128/commits/a079d968d3f273ed739eda51b36e9f0470f6a100 be reverted before ?

r00ta:
> shouldn't [a079d96](https://github.com/kiegroup/kogito-runtimes/commit/a079d968d3f273ed739eda51b36e9f0470f6a100) be reverted before ?

Hi @radtriste , unfortunately I have a strange behaviour here on CI. I pushed that commit to break `kogito-test-utils` module on purpose. I have also cleaned up the m2 repo on the node with the Jenkinsfile. 
Locally I have the expected behaviour that  Kogito :: Integration Tests :: SpringBoot fails using the new property `-Dno-snapshot`, but here on the CI the module is build and run with no problem at all. Do you have any tip? It's super strange but apparently the m2 repo on the node is not empty

radtriste:
@r00ta I guess this is is because we do use a Maven mirror in Jenkins, and so the `kogito-test-utils` can be retrieved from it: https://bxms-qe.rhev-ci-vms.eng.rdu2.redhat.com:8443/nexus/content/groups/kogito-runtimes-pr-artifacts-group/org/kie/kogito/kogito-test-utils/2.0.0-SNAPSHOT/ :S
Thinking about how to disable that easily

r00ta:
> @r00ta I guess this is is because we do use a Maven mirror in Jenkins, and so the `kogito-test-utils` can be retrieved from it: https://bxms-qe.rhev-ci-vms.eng.rdu2.redhat.com:8443/nexus/content/groups/kogito-runtimes-pr-artifacts-group/org/kie/kogito/kogito-test-utils/2.0.0-SNAPSHOT/ :S
> Thinking about how to disable that easily

Thank you very much, let me know if I can help somehow or if we have to find another solution

radtriste:
Once https://github.com/kiegroup/jenkins-pipeline-shared-libraries/pull/147 is merged, you should add `.withMirrorDisabledForRepoInSettings('jboss-public-repository-group')` to after that line: https://github.com/kiegroup/kogito-runtimes/blob/master/Jenkinsfile#L239

Jenkinsfile
```diff
@@ -57,10 +58,12 @@ pipeline {
                     if(getQuarkusBranch()) {
                         // Only run against given quarkus branch
                         getMavenCommand('kogito-runtimes')
+                            .withProperty('no-snapshot')
```
radtriste, line 61:
```suggestion
                            .withProperty('no-snapshot')
                            .withMirrorDisabledForRepoInSettings('jboss-public-repository-group')
```

Jenkinsfile
```diff
@@ -57,10 +58,12 @@ pipeline {
...
                             .run('clean install')
                     } else {
                         // Build & Perform the analysis
                         getMavenCommand('kogito-runtimes')
+                            .withProperty('no-snapshot')
```
radtriste, line 66:
```suggestion
                            .withProperty('no-snapshot')
                            .withMirrorDisabledForRepoInSettings('jboss-public-repository-group')
```

radtriste:
ok now I see what is the other problem ... :S

danielezonca:
> Once [kiegroup/jenkins-pipeline-shared-libraries#147](https://github.com/kiegroup/jenkins-pipeline-shared-libraries/pull/147) is merged, you should add `.withMirrorDisabledForRepoInSettings('jboss-public-repository-group')` to after that line: https://github.com/kiegroup/kogito-runtimes/blob/master/Jenkinsfile#L239

@radtriste 
If I understand correctly this change, it removes the "internal" nexus entirely from the build (both for snapshot and final dependencies).
If I'm right this might have other impacts to consider:
- the snapshot property introduced in this PR can be probably removed 🤔  
- all dependencies are not fetched any more (even final) from the local nexus slowing down the whole build

Am I right?

radtriste:
yeah you are right, we may need to have more fine-grained selection...

radtriste:
in kogito `settings.xml`, we have 
```
<repository>
          <id>jboss-public-repository-group</id>
          <name>JBoss Public Repository Group</name>
          <url>https://repository.jboss.org/nexus/content/groups/public/</url>
          <releases>
            <enabled>true</enabled>
            <updatePolicy>never</updatePolicy>
          </releases>
          <snapshots>
            <enabled>true</enabled>
            <updatePolicy>never</updatePolicy>
          </snapshots>
        </repository>
```
trying to find a way to disable snapshots if needed ... would that be enough ?

danielezonca:
> in kogito `settings.xml`, we have
> 
> ```
> <repository>
>           <id>jboss-public-repository-group</id>
>           <name>JBoss Public Repository Group</name>
>           <url>https://repository.jboss.org/nexus/content/groups/public/</url>
>           <releases>
>             <enabled>true</enabled>
>             <updatePolicy>never</updatePolicy>
>           </releases>
>           <snapshots>
>             <enabled>true</enabled>
>             <updatePolicy>never</updatePolicy>
>           </snapshots>
>         </repository>
> ```
> 
> trying to find a way to disable snapshots if needed ... would that be enough ?
> 
> and the solution to not use the settings.xml is not one because that means we would not at all use the mirror repo ...

Well, PR builds should not fetch any snapshot at all (we always build everything), so what if remove snapshot configuration/create a custom setting for that?

radtriste:
gonna create a `kogito_pr_settings` then

jstastny-cz:
Shouldn't we have a Jira for this?

I will now focus just on integration-tests and maven-invoker related changes - btw tracked as KOGITO-4404.

- When not using `localRepositoryPath`, I am afraid that the `install` goal of maven invoker will pollute the local repository, while this is not problematic in CI, where clean repo is used, it is issue in local env, where you install potentially broken artifacts (and kinda unexpected to change your local repo when running just `mvn clean verify`). You can overcome this issue by keeping `localRepositoryPath` + adding `useLocalRepository` property.
I am investigating the safest approach as part of #1161 .

- By removing dependencies from the integration-test module's pom.xml we're probably changing dependency tree of the reactor, meaning that `mvn clean verify -pl :integration-tests-springboot -am` might not have all the predecessors built correctly and will end up in downloading the artifacts from remote repository - which you typically don't want when using this kind of maven command.

I would suggest at least reducing scope of your changes to omit this part and we can resolve uniformly across the repo.

evacchi:
status? we need to merge https://github.com/kiegroup/kogito-runtimes/pull/1122 ASAP

radtriste:
> gonna create a `kogito_pr_settings` then with snapshots disabled. I let you know once that is available.

FYI, adopted a different approach and we could directly disable snapshots download from settings.xml via `.withSnapshotsDisabledInSettings()` on Maven Command (would replace the `.withMirrorDisabledForRepoInSettings(...)` line)
https://github.com/kiegroup/jenkins-pipeline-shared-libraries/pull/149

Don't know if still useful after https://github.com/kiegroup/kogito-runtimes/pull/1161, let's see ...

r00ta:
> > gonna create a `kogito_pr_settings` then with snapshots disabled. I let you know once that is available.
> 
> FYI, adopted a different approach and we could directly disable snapshots download from settings.xml via `.withSnapshotsDisabledInSettings()` on Maven Command (would replace the `.withMirrorDisabledForRepoInSettings(...)` line)
> [kiegroup/jenkins-pipeline-shared-libraries#149](https://github.com/kiegroup/jenkins-pipeline-shared-libraries/pull/149)
> 
> Don't know if still useful after #1161, let's see ...

Hi @radtriste , I think so. Closing

PR #1129, opened 2021-03-10
[KOGITO-4557] Adding attachment and comments to tasks

fjtirado:
- Added Attachment and Comment POJOs to Kogito API (although already present in V7, I think is better to define new ones, Attachment and Comment are basically the same object, but the latter accept just strings as "content" and the former an URI)
- Attachment use URI reference rather than byte[] because of memory usage concerns
- Added Attachment and Comment  to WorkItemTask  (since they are entities with an id, a map is used to hold them in memory)
- Added CRUD operations for Process Resource (added HumanTaskHelper class to reduce logic present in generated code)
- Modified marshaller and protobuf classes to store comments and attachments as part of runtime info. 
- Rename workItemId as taskId
- Removing the work done by https://issues.redhat.com/browse/KOGITO-4303. Once the user has separated API to add comments and attachments to task, PUT rather than PATCH is a best fit for changing model without changing transition


integration-tests/integration-tests-quarkus-processes/src/test/java/org/kie/kogito/integrationtests/quarkus/TaskTest.java
```diff
@@ -89,29 +102,125 @@ void testUpdateTask() {
...
+
+        final String commentText = "We have done everything we can";
         given().contentType(ContentType.JSON)
                 .when()
                 .queryParam("user", "admin")
                 .queryParam("group", "managers")
                 .pathParam("processId", processId)
                 .pathParam("taskId", taskId)
-                .body(Collections.singletonMap("approved", true))
-                .patch("/approvals/{processId}/firstLineApproval/{taskId}")
```
MarianMacik, line 99: Why was this use case removed? I think it was added as a part of task outputs update capability...

integration-tests/integration-tests-quarkus-processes/src/test/java/org/kie/kogito/integrationtests/quarkus/TaskTest.java
```diff
@@ -63,8 +66,18 @@ void testJsonSchema() {
...
+        public Date updatedAt = new Date();
+
+        public TaskInfoClient(T content) {
+            this.content = content;
+        }
+    }
+
     @Test
-    void testUpdateTask() {
+    void testCommentAndAttachment() {
```
MarianMacik, line 80: Why are we getting rid of the update task test case?

integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-it/src/test/java/org/kie/kogito/integrationtests/springboot/TaskTest.java
```diff
@@ -59,14 +62,24 @@ void testJsonSchema() {
...
+        public Date updatedAt = new Date();
+
+        public TaskInfoClient(T content) {
+            this.content = content;
+        }
+    }
+
     @Test
-    void testUpdateTask() {
+    void testCommentAndAttachment() {
```
MarianMacik, line 76: Same for Spring Boot.

api/kogito-api/src/main/java/org/kie/kogito/process/workitem/TaskMetaEntity.java
```diff
@@ -0,0 +1,62 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.kogito.process.workitem;
+
+public class TaskMetaEntity<T> extends TaskMetaInfo<T> {
+
+    private static final long serialVersionUID = 1L;
+    private Object id;
```
MarianMacik, line 21: This is UUID so it is always String, isn't it?

jbpm/process-serialization-protobuf/pom.xml
```diff
@@ -39,5 +39,32 @@
...
 
   </dependencies>
+  <profiles>
+  <profile>
+  <id>proto</id>
+  <build>
+  <plugins>
+    <plugin>
+      <groupId>org.xolstice.maven.plugins</groupId>
+      <artifactId>protobuf-maven-plugin</artifactId>
```
MarianMacik, line 49: We didn't need to regenerate protobuf files until now? Interesting...

jbpm/process-serialization-protobuf/pom.xml
```diff
@@ -39,5 +39,32 @@
...
+          <goals>
+            <goal>compile</goal>
+          </goals>
+        </execution>
+      </executions>
+    </plugin>
+  </plugins>
+</build>
+</profile>
+</profiles>
```
MarianMacik, line 68: Formatting is a little bit off.

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/RestResourceQuarkusTemplate.java
```diff
@@ -48,12 +49,17 @@
...
+import org.kie.kogito.process.workitem.Comment;
 import org.kie.kogito.process.workitem.Policies;
+import org.kie.kogito.process.workitem.TaskMetaInfo;
 import org.kie.kogito.process.impl.Sig;
 import org.kie.kogito.services.uow.UnitOfWorkExecutor;
 import org.kie.kogito.auth.IdentityProvider;
+import org.jbpm.process.instance.impl.humantask.HumanTaskHelper;
 import org.jbpm.process.instance.impl.humantask.HumanTaskTransition;
 
+
```
MarianMacik, line 62: Unnecessary :)

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/RestResourceSpringTemplate.java
```diff
@@ -53,6 +58,8 @@
 import org.springframework.web.util.UriComponents;
 import org.springframework.web.util.UriComponentsBuilder;
 
+
+
```
MarianMacik, line 62: Same here.

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/RestResourceUserTaskQuarkusTemplate.java
```diff
@@ -176,11 +162,221 @@ public Response signal(@PathParam("id") final String id, @Context UriInfo uriInf
...
                                                                        return pi.variables().toOutput();
                                                                    }))
                                                                    .orElseThrow(() -> new NotFoundException());
     }
+    
+    @POST
+    @Path("/{id}/$taskName$/{taskId}/comments")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response addComment (@PathParam("id") final String id,
```
MarianMacik, line 177: Unnecessary space.

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/RestResourceUserTaskQuarkusTemplate.java
```diff
@@ -176,11 +162,221 @@ public Response signal(@PathParam("id") final String id, @Context UriInfo uriInf
...
+                                            .build()).entity(comment).build();
+                                })
+                                .orElseThrow(() -> new NotFoundException()));
+    }
+    
+    @PUT
+    @Path("/{id}/$taskName$/{taskId}/comments/{commentId}")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response updateComment (@PathParam("id") final String id,
```
MarianMacik, line 204: Same here.

integration-tests/integration-tests-quarkus-processes/src/test/java/org/kie/kogito/integrationtests/quarkus/TaskTest.java
fjtirado: I updated description of the PR indicating why we are removing this. Basically once users has the possibility to add comment and attachments, any performed on the task model should be part of a transition phase

integration-tests/integration-tests-quarkus-processes/src/test/java/org/kie/kogito/integrationtests/quarkus/TaskTest.java
fjtirado: same reason

api/kogito-api/src/main/java/org/kie/kogito/process/workitem/TaskMetaEntity.java
fjtirado: Yes, now ID is UUID for sure, but if in future we decided a different ID strategy (which might not be a string) making Id an object does not force us to modify API to change ID generation strategy

jbpm/process-serialization-protobuf/pom.xml
fjtirado: Maybe it was dones, but nobody do it using maven plugin and a profile ;)

integration-tests/integration-tests-quarkus-processes/src/test/java/org/kie/kogito/integrationtests/quarkus/TaskTest.java
fjtirado: I will add PUT api for updating whole model (which is kind of different from the removed PATCH one)

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/RestResourceQuarkusTemplate.java
fjtirado: I liked the extra line (just kidding)

integration-tests/integration-tests-quarkus-processes/src/test/java/org/kie/kogito/integrationtests/quarkus/TaskTest.java
```diff
@@ -89,29 +103,177 @@ void testUpdateTask() {
...
                 .pathParam("taskId", taskId)
-                .body(Collections.singletonMap("approved", true))
-                .patch("/approvals/{processId}/firstLineApproval/{taskId}")
+                .body(model)
+                .put("/approvals/{processId}/firstLineApproval/{taskId}")
                 .then()
                 .statusCode(200)
-                .body("approved", is(true));
+                .extract()
+                .as(Map.class));
```
MarianMacik, line 118: Just thinking if we shouldn't have also an endpoint for obtaining the saved TaskOutput. Right now after save a client has to save it somewhere for future use, otherwise the save action doesn't make sense as there is no possibility to get back to the saved work. Wdyt? I think this could be handled as a separate JIRA, just thinking about it.

MarianMacik (approved):
Looks good, thanks for a quick resolution and a fruitful discussion :)

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/humantask/HumanTaskWorkItemImpl.java
```diff
@@ -182,4 +198,14 @@ protected void checkAssignedOwners(String user, Collection<String> roles) {
             throw new NotAuthorizedException("User " + user + " is not authorized to access task instance with id " + getStringId());
         }
     }
+
+    @Override
+    public Map<Object, Attachment> getAttachments() {
```
tiagodolphine, line 203: just a concern, all instances of attachments and comments are kept in-memory in the HumanTaskWorkItem, so this may cause memory issues, like imagine a request to fetch for N HumanTasks and each could have N attachments, so we need at least to limit the size of the byte[] for each attachment because this may cause memory issues easily. 
Anyway IMO it is dangerous to keep the attachments in-memory, we need to keep just a reference like the ID an get on demand in case the user really needs this data, but with an specific operation for this. Got my point?

tiagodolphine (commented):
@fjtirado LGTM in general, I just added a comment related to keeping the references with the content (byte[]) in-memory.

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/humantask/HumanTaskWorkItemImpl.java
tiagodolphine: Things to consider:
- Restrict the content size the byte[]
- Restrict the number of attachments allowed for each humanTask
- Keep in-memory just the Ids of each attachment, and fetch them when really needed
- Maybe creating a test to create a lot of humanTasks + put a lot of huge attachments for each and getting all humanTasks, monitoring the memory consumption in for jvm would show possible issues.

integration-tests/integration-tests-quarkus-processes/src/test/java/org/kie/kogito/integrationtests/quarkus/TaskTest.java
fjtirado: I opened this one for that a couple of weeks ago https://issues.redhat.com/browse/KOGITO-4502

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/humantask/HumanTaskWorkItemImpl.java
fjtirado: Good point, I need to think about it, problem is that attachment storage is linked to persistence add on. 
If we just keep a reference to the stored content, we will be enforcing the presence of a backstore to store them, regardless the fact the persistence add on is present or not

fjtirado: Changed byte[] to URI

api/kogito-api/src/main/java/org/kie/kogito/process/workitem/TaskMetaEntity.java
cristianonicolai: so maybe have the type for the id as generic and the extending class defines it?

api/kogito-api/src/main/java/org/kie/kogito/process/workitem/HumanTaskWorkItem.java
```diff
@@ -94,4 +95,19 @@
      * @return excluded users
      */
     Set<String> getExcludedUsers();
+
+    /**
+     * Returns task attachments
+     * 
+     * @return A map which key is the attachment id and value the attachment object
+     */
+    Map<Object, Attachment> getAttachments();
```
cristianonicolai, line 104: wouldn't be more consistent if we use a Set in here? that way the key impl is also not so important but then using getId from the actual object

fjtirado: This is a convenient way to store attachemnt and comment so update operation can be performed smoothly. 
Otherwise, you will have to search the set for the right key

fjtirado:
jenkins retest this

api/kogito-services/src/main/java/org/kie/kogito/services/event/impl/UserTaskInstanceEventBody.java
```diff
@@ -45,6 +49,9 @@
     private Map<String, Object> inputs;
     private Map<String, Object> outputs;
 
+    private Collection<Comment> comments;
```
tiagodolphine, line 52: I think this requires changes in data-index to deserialize and persist the new fields.
@cristianonicolai ^

jbpm/process-serialization-protobuf/pom.xml
```diff
@@ -37,7 +37,32 @@
         </exclusion>
       </exclusions>
     </dependency>
-
   </dependencies>
-
+  <profiles>
+    <profile>
+    <id>proto</id>
```
tiagodolphine, line 43: why was it necessary to add this profile?

tiagodolphine (commented):
LGTM the changes, just made some commetns

api/kogito-services/src/main/java/org/kie/kogito/services/event/impl/UserTaskInstanceEventBody.java
cristianonicolai: yep, created https://issues.redhat.com/browse/KOGITO-4710 to track

api/kogito-api/src/main/java/org/kie/kogito/process/workitem/HumanTaskWorkItem.java
cristianonicolai: you only need to search when you're updating right?  To be honest I find actually more natural that a Task has a list/set of Comments and Attachments, the id is part of the entity and to me is usually used for situations where you want to make it easier to avoid dups or get access to an object as you said but again, not something I see as part of the model. And yes the set enforces no duplicates so you need to have the equals properly implemented to drive that. That is just my personal preference, maybe we can get others opinion as well to balance a bit. @tiagodolphine @MarianMacik wdyt?

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/humantask/HumanTaskHelper.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    private static boolean isSearchWorkItem(KogitoNodeInstance ni, String taskId, Policy<?>... policies) {
+        return ni instanceof WorkItemNodeInstance && ((WorkItemNodeInstance) ni).getWorkItemId().equals(
+                taskId) && ((WorkItemNodeInstance) ni).getWorkItem().enforce(policies) &&
+                ((WorkItemNodeInstance) ni).getWorkItem() instanceof HumanTaskWorkItem;
+    }
+
+    public static Comment updateComment(KogitoWorkItem item, Object id, String commentInfo, String user) {
+        Comment comment = asHumanTask(item).getComments().computeIfPresent(id, (k, v) -> buildComment(k,
```
cristianonicolai, line 71: @fjtirado @tiagodolphine @MarianMacik should we constrain that a comment/attachment can only be updated by the same user that created it?

jbpm/process-serialization-protobuf/pom.xml
fjtirado: To allow "semi automatic" generation of java objects from proto file

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/humantask/HumanTaskHelper.java
fjtirado: What I implemented is that any potential owner can add a comment. I deduce that from the fact that v7 task has a user field that wont be needed if only the creator can add comments. 
In any case, I would say that we can restrict it to the actual owner (which unless the creator might change, justifying the existence of such feld) , but I prefer to be more liberal, as if a task was a JIRA, where any user interested on the completion of the task can add information to it

fjtirado:
Answering https://github.com/kiegroup/kogito-runtimes/pull/1129#discussion_r596515643, I tried yesterday to change Map to Collection just to avoid this discussion and the resulting code was really confusing. 
I tried is to expose the collection in the interface and keep the map for updating (I would like to see a non cumbersome code proposal for updating an instance living  in a set, until I saw that, I tend to think the map should be somehow kept) but it turns out that, in order to add comments and attachments during unmarshall, either the map should be make public (marshalling is in a different package) or I have to add "setAttachments" and "setComments" methods to HumanTaskWorkItem in order to be used by the marshaller. If the latter approach is fine (which is not consistent with the current getXXXList.add approach followed for setting other collection contents on marshaller) I guess is the best solution to not expose the Map in the interface (goal which I disagree in the first turn)

api/kogito-services/src/main/java/org/kie/kogito/services/event/impl/UserTaskInstanceEventBody.java
fjtirado: Thanks @cristianonicolai (I have doubts about that need, that why I asked about it in google meet)

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/humantask/HumanTaskHelper.java
cristianonicolai: I agree, we should allow any user to add comments, not only the actual owner. My mention was more regarding who can update or remove a comment. In that case, I think only the user that added can do. Unless we provide some other API where admins can do it as well.

fjtirado: ok, got it now.  lets implement that only the user that added the comment can update or delete it

api/kogito-api/src/main/java/org/kie/kogito/process/workitem/HumanTaskWorkItem.java
MarianMacik: @cristianonicolai We discussed all the possibilities with @fjtirado and agreed that there is nothing "wrong" with having comments in a map with their id as a key. I think it makes it easy to access them in case an external system is referencing them using an id. For other cases, e.g. expose Set in interface and use a Map internally for optimized access, it would either mean to add addComment() and setComments() method to the interface to avoid a cast in a marshaller, or use these methods just in the implementation but then cast the object in the marshaller, so these methods are visible.

fjtirado:
unrelated, jenkins test this

api/kogito-api/src/main/java/org/kie/kogito/process/workitem/HumanTaskWorkItem.java
cristianonicolai: agree, I think this proposal would be a better approach and hide the map details

fjtirado: Although I agree Tiago proposal is more generic (if we changed Set by Collection and also include update and delete methods) than the proposed implementation, what I find surprising of this whole discussion is the resistance to expose a Map in the interface and the fact we have not issue with exposing a Set. Both Set and Map are part of Collections API and for me they belong to the "same" level of "generality". If interface were just exposing Collection, it will be a different story to me, because we will be saying that it just holds collections, without indicating if they are used for random access, for avoiding duplication, or in the case of Comments and Attachment, to allow retrieval, update and removal by id.   
Also, the problem with that proposal is that it adds "mutable" methods to an interface that so far do not have any setter, so it is potentially "inmutable" (btw, the fact that the interface is exposed as inmutable, provoke issues in Marshalling interaction, issues that were so far solved using getColleciton().add() antipattern)
Also, there is a fundamental question that I wanted to skip as part of this PR, but since the conversation is still active, it worth consideration, is HumanTaskWorkImpl suppose to be a pojo with no logic (as it seems by its ussage from Marshalling classes) or a business object with "inteligence", as can be deduced by the presence of "enforce" method to deal with security policies?.  
So, for me the simple solution (without having to discuss the inmutability of the interface or fixing the marshaller antipattern) was to include a Map in the interface (on the already mention issue that a Map or a Set provide the same level of abstraction) 
I think using a Map we can still decouple marshalling, in fact, the map is not serialized in my proposal, I'm just serializing a collection. What Im doing when using a map is telling users of HumanWorkItem interface that they can search by id (if the interface is really inmutable) or event update or remove (if we lated decide that interface is mutable)

fjtirado:
unrelated, jenkins test this

api/kogito-api/src/main/java/org/kie/kogito/process/workitem/HumanTaskWorkItem.java
tiagodolphine: @fjtirado as I mentioned before, IMO it is just a matter of choice, and no matter the choice I'm ok, I just gave an alternative idea. +1

tiagodolphine (approved):

cristianonicolai (approved):

fjtirado:
jenkins retest this

fjtirado:
unrelated data index test failure, but jenkins retest this in any case (;

cristianonicolai:
@fjtirado changes in the message are affecting  data index
`org.kie.kogito.index.event.KogitoUserTaskCloudEvent["data"]->org.kie.kogito.index.model.UserTaskInstance["comments"])
2021-03-23 17:39:17.803  INFO 32311 --- [eam--1007246844] o.k.k.i.t.AbstractDataIndexContainer     : STDOUT: 	at com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException.from(UnrecognizedPropertyException.java:61)`
it should just ignore the unknown property, need to check whats happening

fjtirado:
jenkins retest this

cristianonicolai:
jenkins please retest this

fjtirado:
@kiegroup/gatekeepers please merge this

PR #1132, opened 2021-03-11
KOGITO-4657 Experiment with Kogito Quarkus Platform BOM + Quarkus

evacchi:
This is an experimental PR to add support to our own Quarkus-style Kogito Platform BOM. 

This will be superseded in the future, but it may be necessary ad-interim for supported releases. It may be also useful as an alternative to Quarkus archetypes (we may drop those in favor of this)

I am creating this as a PoC to start a discussion and to see if this is something feasible for support with the given timeframe.

How this works
---------------

A new `kogito-quarkus-parent-quarkus-platform-descriptor` is added with a plugin. The plug-in generates a JSON file that contains metadata for the Quarkus CLI. When the Quarkus CLI is invoked, the metadata is fetched alongside the BOM and a list of the given extensions is then displayed for users to pick.

[Read more in the Quarkus documentation](https://github.com/quarkusio/quarkus/blob/master/docs/src/main/asciidoc/platform.adoc#quarkus-platform-descriptor)

How to use. 
------------

Once all is installed locally:

```
mvn io.quarkus:quarkus-maven-plugin:1.11.6.Final:create -DplatformGroupId=org.kie.kogito -DplatformArtifactId=kogito-quarkus-parent -DplatformVersion=2.0.0-SNAPSHOT
```

follow the instructions. You can add the `kogito` extension here or later; once you are in the fresh project directory, you can now:

```
$ mvn quarkus:list-extension
```

the Kogito extension will be displayed alongside the other core extensions:

```
Current Quarkus extensions available:
Kogito                                             kogito-quarkus
Kogito - Decisions                                 kogito-quarkus-decisions
Kogito - Predictions                               kogito-quarkus-predictions
Kogito - Rules                                     kogito-quarkus-rules
Amazon Alexa                                       quarkus-amazon-alexa
Amazon DynamoDB                                    quarkus-amazon-dynamodb
Amazon IAM                                         quarkus-amazon-iam
Amazon KMS                                         quarkus-amazon-kms
AWS Lambda                                         quarkus-amazon-lambda
AWS Lambda HTTP                                    quarkus-amazon-lambda-http
...

```

Caveats
--------

- I am using kogito-quarkus-parent as a BOM. This is suboptimal for 2 reasons
     1. it is not designed as a BOM 
     2. it is named -parent (weird naming convention for a BOM)
  so we may want to create an ad-hoc BOM

- kogito-bom declares artifacts that are in `kogito-apps`. Because the plug-in fetches dependent artifacts while generating the metadata, and because this platform BOM is runtimes, they are probably being pulled from nexus at the moment. Not sure if this is a real issue, because they are being pulled just for processing. ~~Solutions may be:~~ 
   1. ~~move this platform BOM to kogito-apps (not great) or~~
   2. ~~create another BOM **specific to the extension** where `kogito-apps` artifacts are not present (this is probably what we want)~~
   3. this may be a non-issue because the plugin may be further configured to ignore some artifacts
   
- I haven't created an integration test for this; but we should probably create an invoker-style integration test with the same contents of the Quarkus archetype

- As part of the platform building process, this is pulling the dependencies from the entire Quarkus Core BOM, so this will slow down our builds (at least when we bump Quarkus). Is this OK? Shall we create a specific profile?


aloubyansky:
Note, the reason the json generating plug-in resolves all the artifacts from the bon is that it is trying to catch all the artifacts that are Quarkus extensions. You can configure the plugin to not process artifacts, e.g. from a specific groupId.
As to the BOM, it has to be convenient for the end user of your extensions. So if you can have a single bom that contains everything you customers need, that would be the best, IMO.

kogito-build-parent/pom.xml
```diff
@@ -117,7 +117,7 @@
     <version.org.jsonschema2pojo-maven-plugin>1.0.1</version.org.jsonschema2pojo-maven-plugin>
 
     <!-- this version property is used in plugins but also in dependencies too -->
-    <version.io.quarkus>1.11.5.Final</version.io.quarkus>
+    <version.io.quarkus>1.11.6.Final</version.io.quarkus>
```
evacchi, line 120: .6 contains a fix that address some issue with the BOM processing plugin

kogito-quarkus-platform/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+                <executions>
+                    <execution>
+                        <phase>${pluginPhase}</phase>
+                        <goals>
+                            <goal>generate-extensions-json</goal>
+                            <!-- <goal>generate-platform-descriptor-json</goal> -->
+                        </goals>
+                        <configuration>
+                            <!-- <ignoredGroupIds>
+                                <ignoredGroupId>software.amazon.awssdk</ignoredGroupId>
```
evacchi, line 60: here some groupIds can be ignored -- however our group ids are all `org.kie.kogito`

kogito-build-parent/pom.xml
aloubyansky: Actually it's it fixes an issue in the devtools.

kogito-quarkus-platform/pom.xml
aloubyansky: We can certainly evolve this to our/your needs. But this is what's available today.

maxandersen:
This means kogito is creating its own platform. That when quarkus releases users won't be update and community and camel extensions not available. 

I'm not a fan of this approach unless kogito will release same time as quarkus and consider use  universal platform as base. 


Any reason not going for the just BOM approach ?

kogito-quarkus-platform/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+    <modelVersion>4.0.0</modelVersion>
+         
+    <parent>
+        <groupId>org.kie.kogito</groupId>
+        <artifactId>kogito-build-parent</artifactId>
+        <version>2.0.0-SNAPSHOT</version>
+        <relativePath>../kogito-build-parent/pom.xml</relativePath>
+    </parent>
+
+    <artifactId>kogito-quarkus-parent-quarkus-platform-descriptor</artifactId>
```
mareknovotny, line 15: What about this name?
```suggestion
    <artifactId>kogito-quarkus-platform-descriptor</artifactId>
```

kogito-quarkus-parent/pom.xml
```diff
@@ -25,6 +25,18 @@
...
       </dependency>
+      <dependency>
+        <groupId>org.kie.kogito</groupId>
+        <artifactId>kogito-quarkus-parent-quarkus-platform-descriptor</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.kie.kogito</groupId>
+        <artifactId>kogito-quarkus-parent-quarkus-platform-descriptor</artifactId>
+        <classifier>${project.version}</classifier>
```
Ginxo, line 36: sorry, I probably missed something, but... is the `classifier` correct?

Ginxo (approved):
LGTM, just check my comments

kogito-quarkus-parent/pom.xml
mareknovotny: very good catch, I think it is not.

aloubyansky: Yes, it actually is.

aloubyansky:
> LGTM, just check my comments

Please, make sure you understood Max's points above. Perhaps you already have considered that. I just want to be sure we are making conscious decisions here.

kogito-quarkus-parent/pom.xml
Ginxo: @aloubyansky just curious, can you please explain? thanks

aloubyansky: Sure, I explained it here https://github.com/quarkusio/quarkus/blob/main/docs/src/main/asciidoc/platform.adoc#quarkus-platform-descriptor

mareknovotny: ok, thnx @aloubyansky!

evacchi:
I think, for the given timeframe, we'll have to keep archetypes. We may consider using this as a way to test kogito-examples (on master) before they land in Quarkus platform, though. I don't know how that would work though

evacchi:
closed as outdated

PR #1175, opened 2021-03-24
[KOGITO-4300] Implementing deadlines

fjtirado:
- Added task notification addon. It create emitters for publishing start and completed deadlines.
- Added mail notification addon. it subscribe to emitters created by task notification add on and send e-mail when event is received
- Modifications on engine to parse bpmn notification string (same format than in v7 in order to maintain editor compatibility), schedule notification using jobs service and once received "timerTriggered" event type, propagate it properly to the emitter in the addon. 
-  Handling of Rx/duration/endDate has been modified compared with v7. Rather than going beyond enddate, the first ocurrent of the timer will happen at enddate - number of repeats * duration. If just R, then the first ocurrent will happen at now() + duration as many as needed till endData is reached
- Couple of modification on JobsService client api and InMemoryService to properly remove timers on engine 

Merge with https://github.com/kiegroup/kogito-examples/pull/633


fjtirado:
jenkins retest this

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/humantask/HumanTaskWorkItemImpl.java
```diff
@@ -172,7 +172,8 @@ protected void enforceAuthorization(IdentityProvider identity) {
             String user = identity.getName();
             String currentOwner = getActualOwner();
             // if actual owner is already set always enforce same user
-            if (currentOwner != null && !currentOwner.trim().isEmpty() && !user.equals(currentOwner)) {
+            if (currentOwner != null && !currentOwner.trim().isEmpty() && !user.equals(currentOwner) &&
```
fjtirado, line 175: I was reviewing current general  security constraints for updating tasks. A user that is in the list of admins should have permission to do anything. 
Open issue is how this admin user include himself in the list (we probably need to revisit security policies soon anyway)

addons/task-notification/task-notification-quarkus-addon/pom.xml
```diff
@@ -0,0 +1,101 @@
...
+
+  <dependencies>
+    <dependency>
+      <groupId>javax.inject</groupId>
+      <artifactId>javax.inject</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-smallrye-reactive-messaging-kafka</artifactId>
```
tiagodolphine, line 34: So this comes to my point from the other comment, I think we could rely on a more generic approach for emitters, like let the user choose the implementation even with an in-memory emitter, or any other... in this way depending on the dependency in the pom... so we just rely on the API not on the impl... like `smallrye-reactive-messaging-provider`

addons/task-notification/task-notification-quarkus-addon/src/main/java/org/kie/kogito/task/notification/quarkus/NotificationEventPublisher.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    protected void publishToTopic(DataEvent<?> event, Emitter<String> emitter, String topic) {
+        if (emitter.hasRequests()) {
+            logger.debug("Emitter {} is not ready to send messages", topic);
+        }
+        logger.debug("About to publish event {} to topic {}", event, topic);
+        try {
+            String eventString = json.writeValueAsString(event);
```
tiagodolphine, line 80: so what about letting the event be serialized at the end, like https://quarkus.io/guides/kafka#serializing-via-jackson in case of kafka... because the serialization might change in this way we decouple it from our publisher... in fact we could rethink this strategy for other addons... wdyt?

addons/task-notification/task-notification-springboot-addon/src/main/java/org/kie/kogito/task/notification/spring/SpringNotificationEventPublisher.java
```diff
@@ -0,0 +1,80 @@
...
+@Component
+public class SpringNotificationEventPublisher implements EventPublisher {
+
+    private static final Logger logger = LoggerFactory.getLogger(SpringNotificationEventPublisher.class);
+
+    @Autowired
+    private ObjectMapper json;
+
+    @Autowired
+    private KafkaTemplate<String, String> eventsEmitter;
```
tiagodolphine, line 40: well the same, basically we are coupling the implementation with kafka..

api/kogito-api/src/main/java/org/kie/kogito/internal/process/event/HumanTaskDeadlineEvent.java
```diff
@@ -0,0 +1,55 @@
...
+        Start,
+        Completed
+    }
+
+    /**
+     * Returns work item which timeout expires
+     * 
+     * @return work item
+     */
+    HumanTaskWorkItem getWorkItem();
```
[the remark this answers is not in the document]
tiagodolphine: mmm this is more an internal event, not the event that goes over the wire let's say...

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/HumanTaskNodeInstance.java
```diff
@@ -51,6 +75,23 @@ protected InternalKogitoWorkItem newWorkItem() {
         return new HumanTaskWorkItemImpl();
     }
 
+    public Map<String, Map<String, Object>> getStartDeadlineTimers() {
```
tiagodolphine, line 78: same here.. the signature is not clear...  Map<String, Map<String, Object>>  I mean this could be anything... maybe a way to let this more clear.. like creating some Classes to encapsulate... don't know.. in the last case adding a documentaiton...

tiagodolphine (commented):
@fjtirado thanks, great job, I just added some comments.

addons/mail/mail-quarkus-addon/pom.xml
```diff
@@ -0,0 +1,100 @@
...
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <includes>
+            <include>org/kie/kogito/task/management/test/*</include>
+          </includes>
+          <systemPropertyVariables>
+            <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
+            <container.image.infinispan>${container.image.infinispan}</container.image.infinispan>
```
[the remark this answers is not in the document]
fjtirado: yes, thats copy paste issue ( I mean, I used an already existing POM as baseline for this specific one), Ill remove

api/kogito-api/src/main/java/org/kie/kogito/internal/process/event/HumanTaskDeadlineEvent.java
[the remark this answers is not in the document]
fjtirado: yes, this is the internal event, so there is no problem on holding the whole human task work item. The serialized event is an specific one that does not contain the whole workitem, only the needed information

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/Work.java
```diff
@@ -42,4 +46,14 @@
 
     ParameterDefinition getParameterDefinition(String name);
 
+    Set<String> getMetaParameters();
+
+    Collection<DeadlineInfo<Map<String, Object>>> getStartDeadlines();
```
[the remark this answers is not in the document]
fjtirado: Im reusing DeadLineInfo as a kind of union, to hold timer info and notification or reassigment info. Reassigment info is a well-known structure but notification can (and needs to be) any info (not just information to send an e-mail as in v7), so Map<String,Object> needs to be used here. Ill document

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/HumanTaskNodeInstance.java
fjtirado: Same rationale, a map with timer id (a string) and notification info (Map<String,Object>) which needs to be kept on memory because jobs service does not provide any mean to pass the notification information as part of the timer expiration. So Ill documment too

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/HumanTaskNodeInstance.java
```diff
@@ -125,6 +299,14 @@ public void triggerCompleted(InternalKogitoWorkItem workItem) {
         super.triggerCompleted(workItem);
     }
 
+    private <T> void cancelTimers(Map<String, T> timers) {
+        Iterator<String> iter = timers.keySet().iterator();
+        while (iter.hasNext()) {
+            getJobsService().cancelJob(iter.next());
```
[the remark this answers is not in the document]
fjtirado: yes, we need to cancel repeteable timers when the work item is completed

jbpm/process-serialization-protobuf/pom.xml
```diff
@@ -61,6 +61,14 @@
               </execution>
             </executions>
           </plugin>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-checkstyle-plugin</artifactId>
+            <version>3.1.2</version>
+            <configuration>
+              <failOnViolation>false</failOnViolation>
```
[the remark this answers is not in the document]
fjtirado: This is needed to avoid proto generation profile to fail because of checkstyle. 
Since protobuffer generator does not provide a header, it has to be manually added later, but at least with this property the build with proto profile will not fail.

addons/task-notification/task-notification-springboot-addon/src/main/java/org/kie/kogito/task/notification/spring/SpringNotificationEventPublisher.java
fjtirado: Im following here the same approach that was used for publishing events on springboot. I mean, yes, this is an addon that use kafka to publish notifications. There can be other addons that use a didffer

addons/task-notification/task-notification-quarkus-addon/src/main/java/org/kie/kogito/task/notification/quarkus/NotificationEventPublisher.java
fjtirado: Im following the same approach that is followed for data-index events https://github.com/kiegroup/kogito-runtimes/blob/master/addons/events/kogito-events-reactive-messaging-addon/src/main/java/org/kie/kogito/events/rm/ReactiveMessagingEventPublisher.java#L104-L113, why do we want to change that?

addons/mail/mail-quarkus-addon/src/main/java/org/kie/kogito/mail/QuarkusMailSender.java
```diff
@@ -0,0 +1,75 @@
...
+    public void onStartDeadline(String event) throws IOException {
+        onDeadline(event);
+    }
+
+    @Incoming("kogito-completed-deadline-consumer")
+    public void onCompletedDeadline(String event) throws IOException {
+        onDeadline(event);
+    }
+
+    private void onDeadline(String event) throws IOException {
```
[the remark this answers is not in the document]
fjtirado: Ill try, but it does not work, the Incoming annotation needs a String there (or adidng a construcotr to UserTaskDeadlineDataEvent that accepts a String). 
Anyway, the string contains the UserTaskDeadlineDataEvent, that is deserialized using the injected mapper

addons/task-notification/task-notification-quarkus-addon/pom.xml
fjtirado: There is currently an asymmetry between https://github.com/kiegroup/kogito-runtimes/blob/master/addons/events/kogito-events-reactive-messaging-addon/pom.xml and https://github.com/kiegroup/kogito-runtimes/blob/master/addons/events/kogito-events-spring-boot-addon/pom.xml. The first one do not assume anything about Kafka. The second does. But this implies that in the first one an additiional dependency needs to be added in final pom
Since I do not like assimetries, Im including the Kafka dependency on both. Ill change the name of the addons so Kafka appears on them

addons/mail/mail-quarkus-addon/src/main/java/org/kie/kogito/mail/QuarkusMailSender.java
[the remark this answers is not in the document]
fjtirado: I do not think the json usage is related with the kafka question. See my comments below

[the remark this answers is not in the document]
fjtirado: Ill change to use Emitter<DataEvent> (and mail addon to use DataEvent and add documentation to use proper deserializer)

addons/task-notification/task-notification-quarkus-addon/pom.xml
fjtirado: What I can do is use is to use Emitter<DataEvent> (and change mail addon to use DataEvent directly rather than string and add documentation to use proper deserializer)

addons/task-notification/task-notification-quarkus-addon/src/main/java/org/kie/kogito/task/notification/quarkus/NotificationEventPublisher.java
fjtirado: Ill change that to use Emitter<DataEvent<?> rather than the injected jackson mapper

addons/mail/mail-quarkus-addon/src/main/java/org/kie/kogito/mail/QuarkusMailSender.java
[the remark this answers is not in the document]
fjtirado: Changed to use the deserialize the POJO

fjtirado:
jenkins retest this

addons/mail/mail-quarkus-addon/src/main/java/org/kie/kogito/mail/QuarkusMailSender.java
```diff
@@ -0,0 +1,61 @@
...
+import io.quarkus.mailer.reactive.ReactiveMailer;
+
+@ApplicationScoped
+public class QuarkusMailSender {
+
+    private static final Logger logger = LoggerFactory.getLogger(QuarkusMailSender.class);
+    @Inject
+    private ReactiveMailer mailer;
+
+    @Incoming("kogito-start-deadline-consumer")
```
cristianonicolai, line 38: should we keep it consistent as in start/complete or started/completed?

addons/task-notification/task-notification-quarkus-addon/src/main/java/org/kie/kogito/task/notification/quarkus/NotificationEventPublisher.java
```diff
@@ -0,0 +1,78 @@
...
+    @Channel(CD_TOPIC_NAME)
+    Emitter<DataEvent<?>> completedDeadlinesEventsEmitter;
+
+    @Inject
+    @ConfigProperty(name = "kogito.events.deadlines.enabled")
+    Optional<Boolean> deadlineEvents;
+
+    @Override
+    public void publish(DataEvent<?> event) {
+        if (event.getType().equals("UserTaskDeadlineStart") && deadlineEvents.orElse(true)) {
```
cristianonicolai, line 53: @fjtirado @tiagodolphine if we think in terms of how Task and Process are structured, we would have a single topic where UserTaskDeadline events are published with different status ( start / complete ) wdyt?

api/kogito-api/src/main/java/org/kie/kogito/internal/process/event/KogitoProcessEventSupport.java
```diff
@@ -63,5 +65,15 @@ void fireAfterVariableChanged(String name, String id, Object oldValue, Object ne
 
     void fireOnMessage(KogitoProcessInstance instance, KogitoNodeInstance nodeInstance, KieRuntime kruntime, String messageName, Object messageObject);
 
+    void fireNotStartNotification(KogitoProcessInstance instance,
```
cristianonicolai, line 68:
```suggestion
    void fireNotStartedNotification(KogitoProcessInstance instance,
```

api/kogito-api/src/main/java/org/kie/kogito/internal/process/event/KogitoProcessEventSupport.java
```diff
@@ -63,5 +65,15 @@ void fireAfterVariableChanged(String name, String id, Object oldValue, Object ne
...
+            HumanTaskWorkItem workItem,
+            Map<String, Object> notification,
+            KieRuntime kruntime);
+
+    void fireNotEndNotification(KogitoProcessInstance instance,
```
cristianonicolai, line 73:
```suggestion
    void fireNotCompletedNotification(KogitoProcessInstance instance,
```

api/kogito-api/src/main/java/org/kie/kogito/jobs/DurationExpirationTime.java
```diff
@@ -72,7 +72,7 @@ public static DurationExpirationTime repeat(long delay, Long repeatInterval, Int
     }
 
     public static DurationExpirationTime repeat(long delay, Long repeatInterval, TemporalUnit unit) {
-        return new DurationExpirationTime(ZonedDateTime.now().plus(delay, unit), repeatInterval, 0);
+        return new DurationExpirationTime(ZonedDateTime.now().plus(delay, unit), repeatInterval, -1);
```
cristianonicolai, line 75: should this be 0 as in do not repeat?

api/kogito-services/src/main/java/org/kie/kogito/services/event/impl/UserTaskDeadlineEventBody.java
```diff
@@ -0,0 +1,245 @@
...
+
+import java.util.Collection;
+import java.util.Date;
+import java.util.Map;
+import java.util.Set;
+
+import org.kie.kogito.process.workitem.Attachment;
+import org.kie.kogito.process.workitem.Comment;
+
+public class UserTaskDeadlineEventBody {
```
cristianonicolai, line 26: @fjtirado could we reuse `UserTaskInstanceEventBody` instead? they look very similar

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/impl/WorkImpl.java
```diff
@@ -34,63 +39,134 @@
     private Map<String, Object> parameters = new LinkedHashMap<String, Object>();
     private Map<String, ParameterDefinition> parameterDefinitions = new LinkedHashMap<String, ParameterDefinition>();
 
+    private Collection<DeadlineInfo<Map<String, Object>>> startDeadlines;
+    private Collection<DeadlineInfo<Map<String, Object>>> endDeadlines;
+    private Collection<DeadlineInfo<Reassignment>> startReassigments;
+    private Collection<DeadlineInfo<Reassignment>> endReassigments;
+
+    private static final String NOT_STARTED = "NotStartedNotify";
```
cristianonicolai, line 47: @fjtirado maybe is worth syncing with @romartin to check if there is a way for these new attributes to be available in the editor. Also, we need to make sure these are documented.

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/humantask/HumanTaskWorkItemImpl.java
cristianonicolai: yeah, it should be a system wide setting, having per task admin is probably too fine grained.

jbpm/process-serialization-protobuf/pom.xml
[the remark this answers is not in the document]
cristianonicolai: @fjtirado maybe you could just exclude the file(s) here instead of completely disabling the validation?

jbpm/process-serialization-protobuf/pom.xml
```diff
@@ -61,6 +61,14 @@
               </execution>
             </executions>
           </plugin>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-checkstyle-plugin</artifactId>
+            <version>3.1.2</version>
```
cristianonicolai, line 67: shouldnt need to have the version as it is managed in the parent.

api/kogito-services/src/main/java/org/kie/kogito/services/event/impl/UserTaskDeadlineEventBody.java
cristianonicolai: actually, do we actually need a new/specific module for task notification or should this be just another event that is published in the reactive-messaging-addon? @fjtirado @tiagodolphine

jbpm/process-serialization-protobuf/pom.xml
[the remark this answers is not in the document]
fjtirado: @cristianonicolai I tried, but did not find an easy way (exclusion was not working). Sinee normally after changing the proto and generating the code, the marshaller must be updated (and validation is not disabled in that profile), that wont have any consequence beyond avoinding the generation to fail with a "false" positive

jbpm/process-serialization-protobuf/pom.xml
fjtirado: sure, will remove

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/impl/WorkImpl.java
fjtirado: well, actually these attribures are not "new", they are already there for v7 (which shares editor with Kogito) and the main purpose of this JIRA is to handle them in Kogito (where so fat they were ignored)

api/kogito-services/src/main/java/org/kie/kogito/services/event/impl/UserTaskDeadlineEventBody.java
fjtirado: Initial implementation I made (basically because was far less work) was to add two new emitters to existing reactive addon, but this will force all Quarkus users to provide configuration for these emitters, regardless they are using the functionality or not. This will broke most existing examples using events, so it is not really a good idea, addon is far better approach and I will say more consistent with Kogito current architecture
We need UserTaskDeadLineEventBody (again I was using UserTaskNotification initially ;)) because we do not want to publish, on the new emitter, details about the list of admin users. Besides, this gives us flexibility to refine the data we are publising.

addons/task-notification/task-notification-quarkus-addon/src/main/java/org/kie/kogito/task/notification/quarkus/NotificationEventPublisher.java
fjtirado: I think this should be user decission, thats why I include two channels, that might be configured to use the same topic. Thats adds more flexibility to user

addons/mail/mail-quarkus-addon/src/main/java/org/kie/kogito/mail/QuarkusMailSender.java
fjtirado: I will use "complete" rather than "completed". somehow I think the name should be as short as possible

api/kogito-api/src/main/java/org/kie/kogito/jobs/DurationExpirationTime.java
fjtirado: I changed to  -1 because I wanted to distinguish between not repeat (0) and repeat(-1)

addons/task-notification/task-notification-quarkus-addon/src/main/java/org/kie/kogito/task/notification/quarkus/NotificationEventPublisher.java
fjtirado: @cristianonicolai after giving this another thought, I think that most frequent scenario if we have two channels  is customer mapping these two channels to the same topic. And in such case, due to limitations of CDI approach taken by Smallrye, the event will be received twice in Quarkus env. 
So, it is probably better to define just one channel, Im changing that

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/WorkItemNodeInstance.java
```diff
@@ -200,7 +202,14 @@ protected InternalKogitoWorkItem createWorkItem(WorkItemNode workItemNode) {
         workItem = newWorkItem();
         workItem.setName(work.getName());
         workItem.setProcessInstanceId(getProcessInstance().getStringId());
-        workItem.setParameters(new HashMap<>(work.getParameters()));
+
+        Collection<String> metaParameters = work.getMetaParameters();
```
fjtirado, line 206: Here I'm preventing "magic" strings used by editor to configure notification to appear as work item parameters (which are currently returned by getTask list process REST api and included in the event)

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/humantask/DeadlineHelper.java
```diff
@@ -0,0 +1,187 @@
...
+        throw new IllegalArgumentException("Invalid formar for dead line expression " + text);
+    }
+
+    private static Collection<ScheduleInfo> getReassignmentSchedule(String timeStr) {
+        ScheduleInfo info = new ScheduleInfo();
+        if (!timeStr.startsWith("PT")) {
+            timeStr = "PT" + timeStr;
+        }
+        info.setDuration(Duration.parse(timeStr));
+        return Collections.singletonList(info);
```
MarianMacik, line 109: We don't support the same format as for notifications?

jbpm/process-serialization-protobuf/src/main/java/org/jbpm/marshalling/impl/AbstractProtobufProcessInstanceMarshaller.java
```diff
@@ -535,6 +544,28 @@ public int compare(NodeInstance o1,
...
+                builder.putContent(pair.getKey(), pair.getValue().toString());
+            }
+            consumer.accept(deadlineTimer.getKey(), builder.build());
+        }
+    }
+
+    private void writeReassignmentMap(Map<String, Reassignment> timers,
+            BiConsumer<String, JBPMMessages.ProcessInstance.NodeInstanceContent.Reassignment> consumer) {
+        for (Map.Entry<String, Reassignment> timer : timers.entrySet()) {
+            consumer.accept(timer.getKey(), JBPMMessages.ProcessInstance.NodeInstanceContent.Reassignment.newBuilder()
```
MarianMacik, line 562: Why not to import the Reassignment builder as well?

jbpm/jbpm-flow/src/test/java/org/jbpm/process/instance/impl/humantask/DeadlineHelperTest.java
```diff
@@ -0,0 +1,209 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class DeadlineHelperTest {
+
+    @Test
+    public void testRepetition() {
```
MarianMacik, line 37: FYI you can find all 4 use cases we support for deadlines ISO timers [here](https://github.com/kiegroup/jbpm/blob/951346254c8917308cba8dc21ed85978f1c9dea0/jbpm-human-task/jbpm-human-task-workitems/src/test/java/org/jbpm/services/task/wih/util/HumanTaskHandlerHelperTest.java#L1766).

MarianMacik (commented):
BTW we should also thing about support of full ISO, which means also years, months, weeks. This is called Period in Java. This was added a little bit later as a part of kiegroup/jbpm#1553.

addons/mail/mail-quarkus-addon/src/main/java/org/kie/kogito/mail/QuarkusMailSender.java
```diff
@@ -0,0 +1,60 @@
...
+    @Inject
+    private ReactiveMailer mailer;
+
+    @Incoming("kogito-deadline-consumer")
+    public void onDeadline(UserTaskDeadlineDataEvent event) {
+        MailInfo mailInfo = MailInfo.of(event.getData());
+        logger.info("Sending e-mail {}", mailInfo);
+        Mail message = new Mail();
+        if (mailInfo.to() != null) {
+            message.setTo(Arrays.asList(mailInfo.to()));
```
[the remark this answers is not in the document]
fjtirado: I can confirm mailer writes to log (that I check during testing thats why nothing else is done)

addons/mail/mail-common-addon/src/main/java/org/kie/kogito/mail/MailInfo.java
```diff
@@ -0,0 +1,83 @@
...
+
+    public static MailInfo of(UserTaskDeadlineEventBody data) {
+        Map<String, Object> info = data.getNotification();
+        return new MailInfo(Optional.ofNullable((String) info.get(TO_PROPERTY)).map(s -> s.split(",")).orElse(null),
+                (String) info.get(FROM_PROPERTY), evalTemplate((String) info.get(SUBJECT_PROPERTY), data),
+                (String) info.get(REPLY_TO_PROPERTY), evalTemplate((String) info.get(BODY_PROPERTY), data));
+    }
+
+    private static String evalTemplate(String template, UserTaskDeadlineEventBody data) {
+        return template != null ? TemplateRuntime.eval(template, data).toString() : template;
```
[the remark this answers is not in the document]
fjtirado: Thats a typo, thanks

addons/mail/mail-springboot-addon/src/main/java/org/kie/kogito/mail/SpringBootMailSender.java
```diff
@@ -0,0 +1,62 @@
...
+    private JavaMailSender emailSender;
+
+    @KafkaListener(id = "${kogito.addon.mail.group.id:mail}",
+            topics = "${kogito.events.deadline.topic:kogito-deadline-events}")
+    public void onMessage(UserTaskDeadlineDataEvent record) {
+        MailInfo mailInfo = MailInfo.of(record.getData());
+        logger.info("Sending e-mail {}", mailInfo);
+        SimpleMailMessage message = new SimpleMailMessage();
+        if (mailInfo.to() != null) {
+            message.setTo(mailInfo.to());
```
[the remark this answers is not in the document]
fjtirado: If something goes wrong, trace will be written by mailer, nothing else we can do

addons/mail/pom.xml
```diff
@@ -0,0 +1,32 @@
...
+    <dependencies>
+      <dependency>
+        <groupId>org.kie.kogito</groupId>
+        <artifactId>kogito-bom</artifactId>
+        <version>${project.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
```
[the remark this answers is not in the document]
fjtirado: yes, thats not needed, thanks

jbpm/jbpm-flow/src/test/java/org/jbpm/process/instance/impl/humantask/DeadlineHelperTest.java
fjtirado: Ill take a look, but I think Duration should be able to deal with days.

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/humantask/DeadlineHelper.java
```diff
@@ -0,0 +1,187 @@
...
+                expirationTime = DurationExpirationTime.repeat(
+                        Instant.now().until(info.getEndDate().minus(info.getDuration().multipliedBy(info
+                                .getNumRepetitions())), ChronoUnit.MILLIS),
+                        getDuration(info), numRepetitions);
+            } else {
+                numRepetitions = (int) (Instant.now().until(info.getEndDate(), ChronoUnit.MILLIS) / info.getDuration()
+                        .toMillis());
+                long duration = getDuration(info);
+                expirationTime = numRepetitions == 0 ? DurationExpirationTime.after(duration)
+                        : DurationExpirationTime.repeat(duration, duration, numRepetitions);
```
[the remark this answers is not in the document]
fjtirado: The case where the handlling is different is when the user specify a number of repetitions, a duration and end date. In V7, this was handled by substracting the period from end date, so if number of repetitions was 3, the real end date will be end date +2*duration. This does not make sense at all to me, so the reintepretation I made was, if user specify number of repetitions equals to 3, duration x and end date y, the first notification will ocurr at y - 3*x. End date will still be the specified end date.

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/humantask/DeadlineHelper.java
```diff
@@ -0,0 +1,187 @@
...
+                scheduleInfo.setDuration(Duration.parse(dateComponents[2]));
+            }
+
+        } else if (isDuration(dateComponents[0])) {
+            scheduleInfo.setDuration(Duration.parse(dateComponents[0]));
+        } else {
+            scheduleInfo.setEndDate(ZonedDateTime.parse(dateComponents[0]));
+        }
+        return scheduleInfo;
+    }
```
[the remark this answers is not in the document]
fjtirado: The case you highlighted is the exact date case, where there is no Repetition.

fjtirado:
Answering https://github.com/kiegroup/kogito-runtimes/pull/1175#discussion_r609792327,
Thats true for the InMemoryJobService but untrue for Kogito-apps job service

jbpm/jbpm-flow/src/test/java/org/jbpm/process/instance/impl/humantask/DeadlineHelperTest.java
fjtirado: I took a look. 
The two cases that I think are not covered were: 
- when user specify an interval greater than a Day
- when user specify Rx/StartDate/EndDate, which Im not sure how should be intepreted. 
Ill take a look at the docs and incorporate them

addons/task-notification/task-notification-quarkus-addon/src/main/java/org/kie/kogito/task/notification/quarkus/NotificationEventPublisher.java
```diff
@@ -0,0 +1,61 @@
...
+import javax.inject.Singleton;
+
+import org.eclipse.microprofile.reactive.messaging.Channel;
+import org.eclipse.microprofile.reactive.messaging.Emitter;
+import org.kie.kogito.event.DataEvent;
+import org.kie.kogito.event.EventPublisher;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Singleton
```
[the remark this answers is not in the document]
fjtirado: I was actually wondering the same, this is because I copy paste from Reatvice publisher that is using Singleton (and I decided to follow the same approach), I will change

api/kogito-api/src/main/java/org/kie/kogito/internal/process/event/KogitoProcessEventSupport.java
```diff
@@ -63,5 +65,15 @@ void fireAfterVariableChanged(String name, String id, Object oldValue, Object ne
 
     void fireOnMessage(KogitoProcessInstance instance, KogitoNodeInstance nodeInstance, KieRuntime kruntime, String messageName, Object messageObject);
 
+    void fireNotStartedNotification(KogitoProcessInstance instance,
```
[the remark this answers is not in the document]
fjtirado: too lager, but what about fireOnTaskNotStartedDeadline?

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/Work.java
```diff
@@ -42,4 +46,32 @@
 
     ParameterDefinition getParameterDefinition(String name);
 
+    Set<String> getMetaParameters();
+
+    /**
+     * Retrieve information about non started deadlines.
```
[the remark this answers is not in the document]
fjtirado: Changed

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/humantask/DeadlineHelper.java
```diff
@@ -0,0 +1,187 @@
...
+
+    private static <T> DeadlineInfo<T> parseDeadline(String text,
+            Function<String, T> notificationFunction,
+            Function<String, Collection<ScheduleInfo>> scheduleFunction) {
+        text = text.trim();
+        if (text.startsWith("[") && text.endsWith("]")) {
+            DeadlineInfo<T> deadline = new DeadlineInfo<>();
+            int indexOf = text.indexOf("]@[");
+            deadline.setNotification(notificationFunction.apply(text.substring(1, indexOf).trim()));
+            deadline.setScheduleInfo(scheduleFunction.apply(text.substring(indexOf + 3, text.length() - 1).trim()));
```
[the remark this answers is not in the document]
fjtirado: ExpirationDate might depend on now(). The moment in which the string is parsed and the moment is which the node is executed does not necesarily match. Please note that DeadlineInfo is information tied to a process, triggering that node is information tied to a process instance.

addons/mail/mail-quarkus-addon/src/main/java/org/kie/kogito/mail/QuarkusMailSender.java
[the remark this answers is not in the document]
fjtirado: @MarianMacik you were right, I changed to log in case of error

addons/mail/mail-springboot-addon/src/main/java/org/kie/kogito/mail/SpringBootMailSender.java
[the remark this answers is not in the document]
fjtirado: Springboot seems to be calling the mail sync (or partially sync) so we can relay on the stack trace of the catch exception

fjtirado:
I finished incorporating Marian comments, so please re-review, have a nice weekend

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/humantask/DeadlineHelper.java
```diff
@@ -0,0 +1,210 @@
...
+                // duration/enddate with 1 or more repetitions
+                expirationTime = DurationExpirationTime.repeat(
+                        Instant.now().until(info.getEndDate().minus(info.getDuration().multipliedBy(info
+                                .getNumRepetitions())), ChronoUnit.MILLIS),
+                        getDuration(info), numRepetitions);
+            } else {
+                numRepetitions = (int) (Instant.now().until(info.getEndDate(), ChronoUnit.MILLIS) / getDuration(info));
+                long duration = getDuration(info);
+                // duration/enddate with "unlimited" repetitions, since there is an end date, it is not really unlimited
+                expirationTime = numRepetitions == 0 ? DurationExpirationTime.after(duration)
```
MarianMacik, line 72: Maybe to be more precise, shouldn't we in this case reduce the duration, so it ends exactly with the endDate? Because right now in case the R/PT1H/00:00 is defined and it is currently 23:50, it will fire at 00:50, but the endDate says the last trigger should occur at 00:00.

Also in case there is more than one repetition calculated, e.g. R/PT1H/00:00 and it is 20:30, it will currently fire 3 times: 21:30, 22:30, 23:30, so maybe we could use the same approach as from the defined number of repetitions, so it will fire 22:00, 23:00 and 00:00?

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/humantask/DeadlineHelper.java
```diff
@@ -0,0 +1,210 @@
...
+            scheduleInfo.setDuration(parseDuration(dateComponents[0]));
+        } else {
+            // not repeatable exact date
+            scheduleInfo.setEndDate(ZonedDateTime.parse(dateComponents[0]));
+        }
+        return scheduleInfo;
+    }
+
+    private static Duration parseDuration(String text) {
+        return text.startsWith("PT") ? Duration.parse(text) : getDuration(Period.parse(text));
```
MarianMacik, line 166: You can have both at the same time, e.g. years with hours. See [DateTimeUtils](https://github.com/kiegroup/jbpm/blob/5bab5adda9db44a90cab8567b783cf210d0947a3/jbpm-flow/src/main/java/org/jbpm/process/core/timer/DateTimeUtils.java#L52-L67).

MarianMacik:
> Answering [#1175 (comment)](https://github.com/kiegroup/kogito-runtimes/pull/1175#discussion_r609792327),
> Thats true for the InMemoryJobService but untrue for Kogito-apps job service

So what is the meaning of repeatLimit 0 and 1 in Kogito-apps job service?
-1 is infinite repetition
0 is execute once, no repetition
1 is execute once and repeat once?

In such case `R1/...` will mean executing the interval twice I think. But according to the spec it should be executed just once.
Or maybe 0 and 1 have the same outcome here, in which case it's OK.

api/kogito-api/src/main/java/org/kie/kogito/jobs/DurationExpirationTime.java
fjtirado: Now I remember the reason why I changed to -1 (this happens when the review occurs one week after I have changed the code and I forgot to put a comment explaining the change, I have to do the same review work twice). 
-1 and 0 have the same semantic from both jobservice (unlimited number of repetitions) but I need to distinguish the case, in the engine, from a timer that has reach the 0 count (and hence should be removed from the engine internal list) and an infinite one. In order to be sure that jobs service application do not pass 0 for an infinite (it is included in the callback call), it is better to put -1 there. In fact I vaguely remember testing with 0 and the timers did not being erased from the node list, thats why I changed to -1. Can we move forward?

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/humantask/DeadlineHelper.java
fjtirado: ok

fjtirado:
> > Answering [#1175 (comment)](https://github.com/kiegroup/kogito-runtimes/pull/1175#discussion_r609792327),
> > Thats true for the InMemoryJobService but untrue for Kogito-apps job service
> 
> So what is the meaning of repeatLimit 0 and 1 in Kogito-apps job service?
> -1 is infinite repetition
> 0 is execute once, no repetition
> 1 is execute once and repeat once?
> 
> In such case `R1/...` will mean executing the interval twice I think. But according to the spec it should be executed just once.
> Or maybe 0 and 1 have the same outcome here, in which case it's OK.

0 and -1 means unlimited (but the value sent to the rest callback might be 0 both for an unlimited and a non limited timer that has been exhausted)
1 means 1 execution, 2 means 2 execution and so on

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/humantask/DeadlineHelper.java
fjtirado: ok

fjtirado:
jenkins retest this

winklerm:
Jenkins retest this

Ginxo:
Jenkins retest this

fjtirado:
jenkins retest this

cristianonicolai (approved):

MarianMacik (approved):
Looks good, thanks!

tiagodolphine (approved):

fjtirado:
@kiegroup/cloud-gatekeepers please merge together with kiegroup/kogito-examples#633

PR #1260, opened 2021-04-26
[KOGITO-4956] avoid registering anonymous rule units

mariofusco:

- [ ] Your code is properly formatted according to [this configuration](https://github.com/kiegroup/kogito-runtimes/tree/master/kogito-ide-config)

<details>
<summary>
</summary>

* <b>Pull Request</b>  
 
* <b>Quarkus LTS checks</b>  

* <b>Native checks</b>  

</details>

### DIFF
--- a/drools/kogito-ruleunits/src/main/java/org/kie/kogito/rules/units/impl/AbstractRuleUnit.java
+++ b/drools/kogito-ruleunits/src/main/java/org/kie/kogito/rules/units/impl/AbstractRuleUnit.java
@@ -41,7 +41,9 @@ public abstract class AbstractRuleUnit<T extends RuleUnitData> implements RuleUn
     @Override
     public RuleUnitInstance<T> createInstance(T data, String name) {
         RuleUnitInstance<T> instance = internalCreateInstance(data);
-        app.get(RuleUnits.class).register(name, instance);
+        if (name != null) {
+            app.get(RuleUnits.class).register(name, instance);
+        }
         return instance;
     }
 }
--- a/api/kogito-api/src/main/java/org/kie/kogito/rules/RuleUnit.java
+++ b/api/kogito-api/src/main/java/org/kie/kogito/rules/RuleUnit.java
@@ -15,14 +15,12 @@
  */
 package org.kie.kogito.rules;
 
-import java.util.UUID;
-
 public interface RuleUnit<T extends RuleUnitData> {
 
     String id();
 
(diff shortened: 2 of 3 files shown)


mariofusco:
@evacchi @danielezonca I didn't remove the `unitRegistry` map, but only populating it when a name is explicitly given instead of generating an `UUID` that anyway is not reachable as I did before.

This allowed me to also keep [this test](https://github.com/kiegroup/kogito-runtimes/blob/4aa991409dab43747e55819370cd9a934d98cbd4/kogito-codegen-modules/kogito-codegen-integration-tests/src/test/java/org/kie/kogito/codegen/RuleUnitCompilerTest.java#L172) that is the only one using that registry so far and also demonstrating how I intended to use it. If you think that this usage pattern is not useful or anyway in contrast with the direction of the other APIs we're implementing please let me know and I'll remove both the registry and the test. At the moment I'm not strongly opinionated about this.

evacchi (approved):

evacchi:
well @mariofusco that won't eliminate the "leak" unless we provide some sort of "remove/dispose" method. Something we need to be aware of in the future.

drools/kogito-ruleunits/src/main/java/org/kie/kogito/rules/units/impl/AbstractRuleUnit.java
```diff
@@ -41,7 +41,9 @@ public String id() {
     @Override
     public RuleUnitInstance<T> createInstance(T data, String name) {
         RuleUnitInstance<T> instance = internalCreateInstance(data);
-        app.get(RuleUnits.class).register(name, instance);
+        if (name != null) {
```
danielezonca, line 44: What about add this null check inside `AbstractRuleUnits.register` (maybe logging error or throwing an exception as you prefer)?

mariofusco: @danielezonca do you mean adding the null check **also** there (for defensive reasons) and throwing an exception if it is called with a null? I cannot remove the null check here, correct?

danielezonca: Yes I mean in addition to this code.
`register` method is public so it is better to be on the safe side

danielezonca (approved):

PR #1265, opened 2021-04-27
KOGITO-4963 Refactor WorkItem APIs on processes

evacchi:
🍼 This is a first baby step towards the brand-new awesome public API that will solve the energetic crisis, make you a better programmer, and feel a warm fuzzy feeling in your heart ❤️  

💡 General guiding principle: if you have a series of methods `getFoo()` `updateFoo()` `removeFoo()` etc, chances are you should want to create class `Foo` with methods `get()`, `update()`, `remove()` etc. 

So this PR moves all the work-item related methods to a `WorkItemCatalog` (*) object

- `@Deprecate` all methods `${verb}WorkItem$()`
- Delegate all deprecated methods to `this.workItems().${verb}()`

🟢  Next iteration(s) -- i.e. separate PRs: 
- update all usages (possibly one method at a time)
- remove all deprecated methods 

(*) I don't feel particularly attached to the name, I just wanted a synonym for Collection that was not Container. Not many options there.

### DIFF
--- a/api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
+++ b/api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
+++ b/api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
@@ -177,14 +177,13 @@ public interface ProcessInstance<T> {
      * Returns list of currently active work items.
      *
      * <strong>this is an experimental API.</strong>
-     *  Will be renamed to <code>workItems</code> when
-     *  {@link #workItems(Policy[])} is removed
+     * Will be renamed to <code>workItems</code> when
+     * {@link #workItems(Policy[])} is removed
      *
      * @param policies optional list of policies to be enforced
      * @return non empty list of identifiers of currently active tasks.
      */
     WorkItemCatalog getWorkItems(Policy<?>... policies);
-
 
     /**
      * Returns identifier of this process instance
--- a/api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
+++ b/api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
@@ -30,6 +30,7 @@ import org.kie.kogito.process.flexible.AdHocFragment;
 import org.kie.kogito.process.flexible.Milestone;
 import org.kie.kogito.process.workitem.Policy;
 import org.kie.kogito.process.workitem.Transition;
+import org.kie.kogito.process.workitem.WorkItemCatalog;
 
 public interface ProcessInstance<T> {
 
@@ -113,6 +114,7 @@ public interface ProcessInstance<T> {
      * @param variables optional variables
      * @param policies optional list of policies to be enforced
      */
+    @Deprecated
     void completeWorkItem(String id, Map<String, Object> variables, Policy<?>... policies);
 
     /**
@@ -123,6 +125,7 @@ public interface ProcessInstance<T> {
      * @param policies optional security information
      * @return result of the operation performed by the updater
      */
+    @Deprecated
     <R> R updateWorkItem(String id, Function<KogitoWorkItem, R> updater, Policy<?>... policies);
 
     /**
@@ -131,6 +134,7 @@ public interface ProcessInstance<T> {
      * @param id id of the work item to complete
      * @param policies optional list of policies to be enforced
      */
+    @Deprecated
     void abortWorkItem(String id, Policy<?>... policies);
 
     /**
@@ -139,6 +143,7 @@ public interface ProcessInstance<T> {
      * @param id id of the work item to complete
      * @param transition target transition including phase, identity and data
      */
+    @Deprecated
     void transitionWorkItem(String id, Transition<?> transition);
 
     /**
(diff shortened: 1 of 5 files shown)


fjtirado:
As an alternative to Catalog you can maybe use Service.

fjtirado:
I do not know if it will be better to implement this in one shot, I mean, remove the methods and change the resource templates, but I agree with the steps anyway. 
If we do that in multi PR way, I will do it in two steps, once this is merged, open PR changing template to use pi.workItems().xxx and remove the deprecated

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
```diff
@@ -164,8 +170,21 @@
...
+     * Returns list of currently active work items.
+     *
+     * <strong>this is an experimental API.</strong>
+     * Will be renamed to <code>workItems</code> when
+     * {@link #workItems(Policy[])} is removed
+     *
+     * @param policies optional list of policies to be enforced
+     * @return non empty list of identifiers of currently active tasks.
+     */
+    WorkItemCatalog getWorkItems(Policy<?>... policies);
```
fjtirado, line 186: you can remove policies from the list of arguments (it is a parameter for every method in WorkItemCatalog). 
I will also rename to workItemCatalog() (so you avoid the get in the name, collision with existing method, confusion in persons that might think that workItems() returns a list of items rather than a interface to handle workitems)

api/kogito-api/src/main/java/org/kie/kogito/process/workitem/WorkItemCatalog.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Aborts work item belonging to this process instance
+     *
+     * @param id id of the work item to complete
+     * @param policies optional list of policies to be enforced
+     */
+    void abort(String id, Policy<?>... policies);
+
+    /**
+     * Transition work item belonging to this process instance not another life cycle phase
```
fjtirado, line 54: change _not_ to _to_ (this was an already existing typo that now is more visible ;))

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcess.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
```
fjtirado, line 2: unneeded?

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcessInstance.java
```diff
@@ -121,10 +120,20 @@ public AbstractProcessInstance(AbstractProcess<T> process, T variables, ProcessR
         this.process = process;
         this.rt = (InternalProcessRuntime) rt;
         this.variables = variables;
+        this.workItemCatalog = makeWorkItemCatalog();
+
         syncProcessInstance((WorkflowProcessInstance) wpi);
         reconnect();
     }
 
+    private WorkItemCatalogImpl<T> makeWorkItemCatalog() {
```
fjtirado, line 129: Just a though. WorkItems were handled here because it needs to interact with process instance to notify updates and remove when finish. 
Although we are maintaining that logic here, the namings of the runnables are pretty specific (removeWhenFinish and updateNotifier)
I believe that to really decouple workItemCatalog from processInstance we should better provide to WorkItemCatalogImpl a runnable per WorkItemCatalog method, ideally though a builder rather than a constructor, so, something like this WorkItemCatalogImpl.builder.addCompleteHandler(this::removeOnFinish).addAbortHandler(this::removeOnFinish).addUpdateHandler(this::notifyUpdate).addTransitionhandler(this::removeOnFinish)

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/workitem/impl/WorkItemCatalogImpl.java
```diff
@@ -0,0 +1,114 @@
...
+    private final Runnable updateNotifier;
+
+    public WorkItemCatalogImpl(
+            Supplier<KogitoWorkItemManager> workItemManagerSupplier,
+            Supplier<WorkflowProcessInstance> processInstanceSupplier,
+            Runnable removeOnFinish,
+            Runnable updateNotifier) {
+        this.workItemManagerSupplier = workItemManagerSupplier;
+        this.processInstanceSupplier = processInstanceSupplier;
+        this.removeOnFinish = removeOnFinish;
```
fjtirado, line 49: Please check my comment about WorkItemCatelogImpl runnables

fjtirado (commented):
As an alternative to `Catalog` you might want to use `Service`, which you can abreviate using `svc`, so my proposal to `workItemCatalog() `become `workItemSvc()`

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
evacchi: > (it is a parameter for every method in WorkItemCatalog).

good point

> I will also rename to workItemCatalog() (so you avoid the get in the name, collision with existing method, confusion in persons that might think that workItems() returns a list of items rather than a interface to handle workitems)

explained the reason for this naming convention in another comment. But probably need further explanation. Please do take a look at the new API design doc :)

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcessInstance.java
evacchi: yes this PR probably needs a better introduction. This is the first step of a sequence of refactors. It is not intended to be good design, but to introduce the new public API ASAP, we can certainly revisit the design if the way it suits us better.

evacchi:
@fjtirado 
> If we do that in multi PR way, I will do it in two steps, once this is merged, open PR changing template to use pi.workItems().xxx and remove the deprecated

yep this is a self-contained PR, I plan to merge this and then follow up

> As an alternative to Catalog you might want to use Service

the design of the PR follows the document for the new API; take a look at that, and why the naming follows those conventions :) Service will be (probably) used for runtime execution, while APIs such as these will be turned into simple "path constructors". We can probably follow up online since you guys weren't around last Monday (IIRC)

 /cc @cristianonicolai

evacchi:
superseded by #1299

PR #1277, opened 2021-05-05
KOGITO-5062 Use reflection in generated models

evacchi:
The idea is to remove useless codegen and align to a single strategy over the whole platform; i.e., replace it with reflection; then use Quarkus' opt-in mechanisms where necessary to allow reflection on our models. 

This will also make it possible for users to use their own models (otherwise user-provided models should explicitly provide toMap/fromMap methods, which is inconvenient).

- In further PRs we are going also to remove generics from `Process<T>` and `ProcessInstance<T>`; making them `Process` and `ProcessInstance`.
- Instead of the `T variables()` method, we will provide a `<T> T variables(Class<T>)` method
- users will be able to supply their choice of a class to such method, and retrieve the subset of fields that match the variables
- this in turn subsumes the requirement for different codegen for Model, InputModel and OutputModel!

Example:
- variables: `name`, `age`, `school`
- `class Person { String name; int age; }`
- `class Student extends Person { String name; int age; String school; }`
- `Person p = proc.variables(Person.class)`, `Student s = proc.variables(Student.class)`

Strategy:

- add default methods to interfaces (`toMap`, `fromMap`)
- delegate such methods to a static library `Models` (that for this PR is in the `api` module; should probably move to a `shared` kind of module). Notice that for convenience, and rapid prototyping this library is hand-rolled using Java's `Introspector`; I think Drools provides such utilities, we may choose to use those after careful refactoring to a stand-alone library 
- in a future PR, the `toMap`, `fromMap` can be removed entirely, substituted for explicit invocation to the static library

TODO: ensure native image works

### DIFF
--- a/api/kogito-api/src/main/java/org/kie/kogito/MapInputId.java
+++ b/api/kogito-api/src/main/java/org/kie/kogito/MapInputId.java
@@ -18,16 +18,19 @@ package org.kie.kogito;
 import java.util.Map;
 
 /**
- * To be implemented by classes which can express its internal information as a Map
+ * To be implemented by classes which can be populated from a Map
  */
-public interface MapOutput {
+public interface MapInputId {
 
     /**
-     * Returns class representation as map
+     * Fills the class with information retrieved from the map
      * 
-     * @return non null map of data extracted from the class
+     * @param params Map containing keys which matches names of fields
+     *        in the class
      */
-    default Map<String, Object> toMap() {
-        return Models.toMap(this);
+    default void fromMap(String id, Map<String, Object> params) {
+        Models.setId(this, id);
+        Models.fromMap(this, params);
     }
+
 }
--- a/api/kogito-api/src/main/java/org/kie/kogito/Models.java
+++ b/api/kogito-api/src/main/java/org/kie/kogito/Models.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.kogito;
+
+import java.beans.BeanInfo;
+import java.beans.IntrospectionException;
+import java.beans.Introspector;
+import java.beans.PropertyDescriptor;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public class Models {
+    private Models() {
+    }
+
+    public static Map<String, Object> toMap(Object m) {
+        try {
+            Map<String, Object> map = new HashMap<>();
+            BeanInfo beanInfo = Introspector.getBeanInfo(m.getClass());
+            Map<String, PropertyDescriptor> descriptors = descriptorMap(beanInfo);
+
+            for (Map.Entry<String, PropertyDescriptor> e : descriptors.entrySet()) {
+                String k = e.getKey();
+                if (k.equals("class") || k.equals("id")) {
+                    continue;
+                }
+                if (e.getKey().startsWith("v$")) {
+                    k = k.substring(2);
+                }
+                map.put(k, e.getValue().getReadMethod().invoke(m));
+            }
+            return map;
+        } catch (IntrospectionException | IllegalAccessException | InvocationTargetException e) {
+            throw new IllegalArgumentException(e);
+        }
+    }
+
+    public static <T> T fromMap(T m, Map<String, Object> map) {
+        try {
+            BeanInfo beanInfo = Introspector.getBeanInfo(m.getClass());
+            Map<String, PropertyDescriptor> descriptors = descriptorMap(beanInfo);
+
+            for (Map.Entry<String, PropertyDescriptor> e : descriptors.entrySet()) {
+                String k = e.getKey();
+                if (e.getKey().startsWith("v$")) {
+                    k = k.substring(2);
+                }
+                if (map.containsKey(k)) {
+                    e.getValue().getWriteMethod().invoke(m, map.get(k));
+                }
+            }
+            return m;
+        } catch (IntrospectionException | IllegalAccessException | InvocationTargetException e) {
+            throw new IllegalArgumentException(e);
+        }
+    }
+
+    public static <T> T fromMap(Class<T> cls, Map<String, Object> map) {
+        try {
+            Constructor<T> constructor = cls.getConstructor();
+            T t = constructor.newInstance();
+            fromMap(t, map);
+            return t;
+        } catch (IllegalAccessException | InvocationTargetException | NoSuchMethodException | InstantiationException e) {
+            throw new IllegalArgumentException(e);
+        }
+    }
+
+    public static void setId(Object m, String id) {
+        try {
+            BeanInfo beanInfo = Introspector.getBeanInfo(m.getClass());
+            for (PropertyDescriptor pd : beanInfo.getPropertyDescriptors()) {
+                if (pd.getName().equals("id")) {
+                    pd.getWriteMethod().invoke(m, id);
+                }
+            }
+        } catch (IntrospectionException | IllegalAccessException | InvocationTargetException e) {
+            throw new IllegalArgumentException(e);
+        }
+
+    }
+
+    public static <I, O> O convert(I in, O out) {
+        fromMap(out, toMap(in));
+        return out;
+    }
+
+    private static Map<String, PropertyDescriptor> descriptorMap(BeanInfo beanInfo) {
+        return Arrays.stream(beanInfo.getPropertyDescriptors())
+                .filter(pd -> !pd.getName().equals("class"))
+                .collect(Collectors.toMap(
+                        PropertyDescriptor::getName,
+                        Function.identity()));
+    }
+}
--- a/api/kogito-api/src/main/java/org/kie/kogito/Models.java
+++ b/api/kogito-api/src/main/java/org/kie/kogito/Models.java
@@ -15,12 +15,14 @@
  */
 package org.kie.kogito;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.beans.BeanInfo;
 import java.beans.IntrospectionException;
 import java.beans.Introspector;
 import java.beans.PropertyDescriptor;
 import java.lang.reflect.Constructor;
-import java.lang.reflect.InvocationTargetException;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
@@ -28,6 +30,16 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 
 public class Models {
+    private static final Logger LOGGER = LoggerFactory.getLogger(Models.class);
+
+    private static final String CLASS_PROP = "class";
+    private static final String ID_PROP = "id";
+    /**
+     * this prefix is only used when a variable name
+     * clashes with a predefined Java keyword (e.g. `static`)
+     */
+    private static final String VAR_PREFIX = "v$";
+
     private Models() {
     }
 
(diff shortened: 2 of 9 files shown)


evacchi:
the jenkins build is green regardless what it says about skipped tests

evacchi:
jenkins retest

evacchi:
jenkins please retest this

integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-kafka-it/src/test/java/org/kie/kogito/integrationtests/springboot/PingPongMessageTest.java
```diff
@@ -50,6 +50,7 @@
 
     @Test
     void testPingPongBetweenProcessInstances() throws InterruptedException {
+        Thread.sleep(5_000); // arbitrary wait to avoid timeouts
```
evacchi, line 53: add sleep to avoid timeout, unrelated to this PR

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ModelMetaData.java
```diff
@@ -211,43 +210,13 @@ private CompilationUnit compilationUnit() {
             fd.createGetter();
             fd.createSetter();
 
-            // toMap method body
```
evacchi, line 214: the following code is no longer necessary

jbpm/jbpm-flow-builder/src/main/resources/class-templates/ModelNoIDTemplate.java
```diff
@@ -15,31 +15,17 @@
...
+import org.kie.kogito.MapOutput;
+
 import java.util.Map;
 import java.util.HashMap;
 
 import org.kie.kogito.MappableToModel;
 import org.kie.kogito.Model;
 
-public class XXXModel implements Model,
+public class XXXModel implements Model, MapInput, MapInputId, MapOutput,
```
evacchi, line 28: implement the default methods via the interfaces

evacchi:
jenkins please retest this

evacchi:
jenkins run native

evacchi:
native build failed, but for seemingly unrelated issues:

```
06:35:17,900 [INFO] --- maven-failsafe-plugin:3.0.0-M5:integration-test (default) @ integration-tests-quarkus-processes-postgresql ---
06:35:18,697 [INFO] 
06:35:18,697 [INFO] -------------------------------------------------------
06:35:18,697 [INFO]  T E S T S
06:35:18,697 [INFO] -------------------------------------------------------
06:35:18,715 [ERROR] Error: --add-modules requires modules to be specified

```

evacchi:
jenkins run native

integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-kafka-it/src/test/java/org/kie/kogito/integrationtests/springboot/PingPongMessageTest.java
danielezonca: I think you can remove this sleep, we usually have green GHA build even without while on Jenkins the problem is not really predictable

api/kogito-api/src/main/java/org/kie/kogito/Models.java
```diff
@@ -0,0 +1,113 @@
...
+            Map<String, Object> map = new HashMap<>();
+            BeanInfo beanInfo = Introspector.getBeanInfo(m.getClass());
+            Map<String, PropertyDescriptor> descriptors = descriptorMap(beanInfo);
+
+            for (Map.Entry<String, PropertyDescriptor> e : descriptors.entrySet()) {
+                String k = e.getKey();
+                if (k.equals("class") || k.equals("id")) {
+                    continue;
+                }
+                if (e.getKey().startsWith("v$")) {
```
danielezonca, line 45: When this can happen? Can you please add a comment?

api/kogito-api/src/main/java/org/kie/kogito/Models.java
```diff
@@ -0,0 +1,113 @@
...
+    public static Map<String, Object> toMap(Object m) {
+        try {
+            Map<String, Object> map = new HashMap<>();
+            BeanInfo beanInfo = Introspector.getBeanInfo(m.getClass());
+            Map<String, PropertyDescriptor> descriptors = descriptorMap(beanInfo);
+
+            for (Map.Entry<String, PropertyDescriptor> e : descriptors.entrySet()) {
+                String k = e.getKey();
+                if (k.equals("class") || k.equals("id")) {
+                    continue;
```
danielezonca, line 43: Why is `id` has a special handling?
Btw we can consider to log.debug/trace here to help future debugging

api/kogito-api/src/main/java/org/kie/kogito/Models.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    public static <T> T fromMap(T m, Map<String, Object> map) {
+        try {
+            BeanInfo beanInfo = Introspector.getBeanInfo(m.getClass());
+            Map<String, PropertyDescriptor> descriptors = descriptorMap(beanInfo);
+
+            for (Map.Entry<String, PropertyDescriptor> e : descriptors.entrySet()) {
+                String k = e.getKey();
+                if (e.getKey().startsWith("v$")) {
```
danielezonca, line 63: When this can happen? Can you please add a comment?

api/kogito-api/src/main/java/org/kie/kogito/Models.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+    }
+
+    public static <T> T fromMap(Class<T> cls, Map<String, Object> map) {
+        try {
+            Constructor<T> constructor = cls.getConstructor();
+            T t = constructor.newInstance();
+            fromMap(t, map);
+            return t;
+        } catch (IllegalAccessException | InvocationTargetException | NoSuchMethodException | InstantiationException e) {
```
danielezonca, line 82: What about handle `NoSuchMethdoException` in a different catch block with an error message like "Class xyz must have a public empty constructor"?

api/kogito-api/src/main/java/org/kie/kogito/Models.java
```diff
@@ -0,0 +1,113 @@
...
+            throw new IllegalArgumentException(e);
+        }
+    }
+
+    public static void setId(Object m, String id) {
+        try {
+            BeanInfo beanInfo = Introspector.getBeanInfo(m.getClass());
+            for (PropertyDescriptor pd : beanInfo.getPropertyDescriptors()) {
+                if (pd.getName().equals("id")) {
+                    pd.getWriteMethod().invoke(m, id);
```
danielezonca, line 92: Can you add a `idFound` boolean or similar with a check after the `for` loop to make sure the id is available? At the moment this code will silently fail if `id` is not available

api/kogito-api/src/main/java/org/kie/kogito/Models.java
```diff
@@ -0,0 +1,113 @@
...
+
+    public static Map<String, Object> toMap(Object m) {
+        try {
+            Map<String, Object> map = new HashMap<>();
+            BeanInfo beanInfo = Introspector.getBeanInfo(m.getClass());
+            Map<String, PropertyDescriptor> descriptors = descriptorMap(beanInfo);
+
+            for (Map.Entry<String, PropertyDescriptor> e : descriptors.entrySet()) {
+                String k = e.getKey();
+                if (k.equals("class") || k.equals("id")) {
```
danielezonca, line 42: The key `class` is already filtered out in `descriptorMap` method

fjtirado (approved):

fjtirado (approved):

api/kogito-api/src/main/java/org/kie/kogito/Models.java
ricardozanini: And move this to a constant documenting it.

ricardozanini (approved):
LGTM after all comments are handled.

evacchi:
jenkins run native

jbpm/jbpm-flow-builder/src/main/resources/class-templates/ModelNoIDTemplate.java
evacchi: Note: this will be dropped at some point: we should allow method-less interfaces and instead using the static library directly. This is just the first step

api/kogito-api/src/main/java/org/kie/kogito/Models.java
```diff
@@ -0,0 +1,113 @@
...
+import java.beans.PropertyDescriptor;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public class Models {
```
evacchi, line 30: I am applying the suggestions, but I'd like to replace this library with some actually shared code -- I think there may be already something like this in Drools

api/kogito-api/src/main/java/org/kie/kogito/Models.java
evacchi: I think a better way would be to just always generate `${name}_` for fields and use `get${Name}` for accessors, this way there is never any clash, thus the check would become unnecessary. This is only to preserve the same behavior of the current codegen procedure, but I'd rather change the codegen procedure itself :)

api/kogito-api/src/main/java/org/kie/kogito/Models.java
evacchi: I have no idea. This is just explicitly encoding the behavior we have in the code-generated methods.

api/kogito-api/src/main/java/org/kie/kogito/MapInputId.java
```diff
@@ -0,0 +1,36 @@
...
+public interface MapInputId {
+
+    /**
+     * Fills the class with information retrieved from the map
+     * 
+     * @param params Map containing keys which matches names of fields
+     *        in the class
+     */
+    default void fromMap(String id, Map<String, Object> params) {
+        Models.setId(this, id);
```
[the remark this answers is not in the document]
evacchi: I am adding a `Models.fromMap(T m, String id, Map<String, Object> map)`, although I'd like `Models` to be very minimal.

fjtirado:
@evacchi One thing that just come to my mind related with the reflection logic. 
Using scafolding, with generated code, when a new attribute is added to the POJO, customer was expected to also change the converters if he want that field to be included in the map. One benefit of using reflection is that customer does not longer to do that. 
But I think we need some way (annotation or transient reserved word ) to allow the customer to say that he does not want some of the fields he adds to the pojo to be converted into the map, wdyt? If you agree, we can create another JIRA to achieve that.

danielezonca (approved):

evacchi:
@fjtirado 

> Using scafolding, with generated code, when a new attribute is added to the POJO, customer was expected to also change the converters if he want that field to be included in the map. One benefit of using reflection is that customer does not longer have to do that.

yes, that was one of the points :)

> But I think we need some way (annotation or transient reserved word ) to allow the customer to establish that he does not want some of the fields he adds to the pojo to be converted into the map, wdyt? If you agree, we can create another JIRA to achieve that

definitely. The ongoing discussion on the programming model is trying to standardize on a consistent set of annotations that may be applied across the board (on all engines) to do data binding. Feel free to crate a story though, I will link the relevant documentation sections :)

fjtirado:
@evacchi 
While an agreement is reached on the right annotation (which I guess will take a while;)) for now I will just check the `transient` modifier for each field in the code that converts Map to Pojo. 
It is an easy change in your code and you will be providing an alternative in case any user ask (and it is aligned with jdk spec for serialization)

evacchi:
sure, please file me a PR for that and let's move this forward: users are still not allowed to supply their own classes, because the Java API is not for public consumption!

evacchi:
now now, native fails on apps and runtimes 🤔 interestingly enough, it works on examples. I wonder how that is on master -- must check nightlies...

fjtirado:
> sure, please file me a PR for that and let's move this forward: users are still not allowed to supply their own classes, because the Java API is not for public consumption! :)

sure, https://issues.redhat.com/browse/KOGITO-5324

evacchi:
Jenkins run native

I have rebased on 2.0. Locally, it looks like some substitution is not being applied or can't be applied:

```
[ERROR] Failed to execute goal io.quarkus:quarkus-maven-plugin:1.13.3.Final:build (default) on project integration-tests-quarkus-processes-kafka-persistence: Failed to build quarkus application: io.quarkus.builder.BuildException: Build failure: Build failed due to errors
[ERROR] 	[error]: Build step io.quarkus.deployment.pkg.steps.NativeImageBuildStep#build threw an exception: java.lang.RuntimeException: Failed to build native image
[ERROR] 	at io.quarkus.deployment.pkg.steps.NativeImageBuildStep.build(NativeImageBuildStep.java:209)
[ERROR] 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[ERROR] 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:78)
[ERROR] 	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[ERROR] 	at java.base/java.lang.reflect.Method.invoke(Method.java:567)
[ERROR] 	at io.quarkus.deployment.ExtensionLoader$2.execute(ExtensionLoader.java:920)
[ERROR] 	at io.quarkus.builder.BuildContext.run(BuildContext.java:277)
[ERROR] 	at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2415)
[ERROR] 	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1452)
[ERROR] 	at java.base/java.lang.Thread.run(Thread.java:831)
[ERROR] 	at org.jboss.threads.JBossThread.run(JBossThread.java:501)
[ERROR] Caused by: java.lang.RuntimeException: Image generation failed. Exit code: 1
[ERROR] 	at io.quarkus.deployment.pkg.steps.NativeImageBuildStep.imageGenerationFailed(NativeImageBuildStep.java:350)
[ERROR] 	at io.quarkus.deployment.pkg.steps.NativeImageBuildStep.build(NativeImageBuildStep.java:181)
[ERROR] 	... 10 more
[ERROR] -> [Help 1]
[ERROR]
```

evacchi:
jenkins run runtimes native

evacchi:
Jenkins run runtimes tests

ricardozanini:
jenkins run apps native

ricardozanini:
jenkins run apps native

evacchi:
[native apps fail on master already](https://eng-jenkins-csb-business-automation.apps.ocp4.prod.psi.redhat.com/job/KIE/job/kogito/job/nightly/job/master/job/kogito-native/42/). Opening JIRAs and a PR to disable them

evacchi:
I'll do one last rerun and then merge. The native failures on apps are unrelated

Jenkins retest this

PR #1301, opened 2021-05-13
[KOGITO-4600] Refactor jbpm-serverless-workflow

fjtirado:
This is a huge refactor of the jbpm serverless wokflow parser. Main changes were:

- Changes in V7 fluent API were incorporated into Kogito fluent API (with its ability to use concrete class after invoking done())
- Parser is now building the process from workflow using same fluent API that is used for building the process from BPMN, so ServerlessWorkflowParserFactory class has been deleted
- State handling has been delegated smaller classes, which extending StateHandler abstract class, will also be responsible to build start and end nodes when appropiated. 
- If an unsupported state is found, rather than returning empty process, the state is ignored. 
- Static initializer for ServerlessWorkflowParser


ricardozanini (approved):
LGTM! Minor nitpicks only.
I think it's a great first step towards the increase of maintainability of this module. I agree with the naming change to `jbpm-serverless-workflow-builder`. Makes sense to me.

jbpm/jbpm-flow/src/main/java/org/jbpm/ruleflow/core/factory/ActionNodeFactory.java
```diff
@@ -18,55 +18,51 @@
...
     }
 
-    public ActionNodeFactory action(String dialect,
+    public ActionNodeFactory<T> action(String dialect,
             String action,
             boolean isDroolsAction) {
         if (isDroolsAction) {
             DroolsAction droolsAction = new DroolsAction();
-            droolsAction.setMetaData(ACTION, action);
+            droolsAction.setMetaData("Action", action);
```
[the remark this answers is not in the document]
fjtirado: Problem is that constant value is "action" (which is used to generate proper calls to action() method) and metadata key is "Action" (note the first capital)
Actually, first implementation was using the constant for both and was generating errors in test, so I changed to hardcoded value for metadata

jbpm/jbpm-flow/src/main/java/org/jbpm/ruleflow/core/RuleFlowProcessFactory.java
```diff
@@ -211,19 +227,14 @@ public RuleFlowProcessFactory errorExceptionHandler(String signalType, String fa
...
-
     public RuleFlowProcessFactory validate() {
         link();
         ProcessValidationError[] errors = RuleFlowProcessValidator.getInstance().validateProcess(getRuleFlowProcess());
-        for (ProcessValidationError error : errors) {
-            logger.error(error.toString());
-        }
         if (errors.length > 0) {
-            throw new RuntimeException("Process could not be validated !");
+            if (logger.isErrorEnabled()) {
```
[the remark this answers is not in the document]
fjtirado: This was to avoid calling Arrays.toString if error log is not enabled, but I think that we do not need to print that, but include the errors in the exception that is thrown to the upper layer

jbpm/jbpm-flow/src/main/java/org/jbpm/ruleflow/core/factory/ActionNodeFactory.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2010 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2015 Red Hat, Inc. and/or its affiliates.
```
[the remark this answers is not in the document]
fjtirado: Im not sure, this project is too old ;), but yes, looks like a wrong copy paste which probably can be traced to 2015 ;), Ill change that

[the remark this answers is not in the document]
fjtirado: To be honest I do not know why I changed that, reverting to original

jbpm/jbpm-serverless-workflow/src/main/java/org/jbpm/serverless/workflow/parser/ServerlessWorkflowParser.java
```diff
@@ -46,568 +39,122 @@
...
-import io.serverlessworkflow.api.states.SwitchState;
-import io.serverlessworkflow.api.switchconditions.DataCondition;
-import io.serverlessworkflow.api.switchconditions.EventCondition;
-import io.serverlessworkflow.api.transitions.Transition;
-import io.serverlessworkflow.api.workflow.Functions;
 
 public class ServerlessWorkflowParser {
 
-    private static final Logger LOGGER = LoggerFactory.getLogger(ServerlessWorkflowParser.class);
+    private static final Logger logger = LoggerFactory.getLogger(ServerlessWorkflowParser.class);
```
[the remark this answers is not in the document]
fjtirado: Actually, it seems that both of them are used (discussion alert!!! ;))
I prefer the lowercase version, maybe we should discuss what we want to do with logger naming in a broad scope.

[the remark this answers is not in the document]
ricardozanini: Weird, IIRC the IDE settings defines statics as capitalized. @danielezonca you remember?

fjtirado:
jenkins retest this

fjtirado:
unrelated flaky test giving timeout, jenkins retest this

ricardozanini:
@danielezonca can you take a look as well, please?

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
```diff
@@ -294,11 +293,7 @@ public static String defaultProcessListenerConfigClass(String packageName) {
                 ProcessMetaData generate = execModelGen.generate();
                 processIdToMetadata.put(id, generate);
                 processExecutableModelGenerators.add(execModelGen);
-            } catch (UndefinedGeneratedRuleUnitVariable e) {
```
danielezonca, line 297: Why this change? Is this catch still possible?

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/impl/NodeContainerImpl.java
```diff
@@ -32,9 +32,10 @@
     private Map<Long, Node> nodes;
 
     public NodeContainerImpl() {
-        this.nodes = new HashMap<Long, Node>();
+        this.nodes = new LinkedHashMap<Long, Node>();
```
danielezonca, line 35: Do we have assumption related to iteration order?

jbpm/jbpm-serverless-workflow/src/main/java/org/jbpm/serverless/workflow/parser/handlers/SwitchHandler.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+    }
+
+    private EndNodeFactory<P> endNodeFactory(List<ProduceEvent> produceEvents) {
+        EndNodeFactory<P> endNodeFactory = factory.endNode(idGenerator.getId());
+        if (produceEvents == null || produceEvents.isEmpty()) {
+            endNodeFactory.terminate(true);
+        } else {
+            ServerlessWorkflowParser.sendEventNode(endNodeFactory, ServerlessWorkflowUtils.getWorkflowEventFor(workflow,
+                    produceEvents.get(0).getEventRef()));
```
danielezonca, line 132: Is it possible to have >1 elements?

danielezonca (approved):
LGTM, some minor comments

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/impl/NodeContainerImpl.java
fjtirado: yes, the unit test are doing that assumption, hence this change

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
fjtirado: UndefinedGeneratedRuleUnitVariable is a RuntimeException and is covered by the next catch. The only purpose of catching it was to print that message, which was redundant in my opinion.

jbpm/jbpm-serverless-workflow/src/main/java/org/jbpm/serverless/workflow/parser/handlers/SwitchHandler.java
fjtirado: There can be certainly more, but only the first event is used as end one. 
Actually, during the refactor I try to keep current logic intact (without questioning it), but I bet this handler will change according to specification evolution.

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
danielezonca: Are you sure it is redundant? As far as I can see this message
`LOGGER.error(e.getMessage() + "\nRemember: in this case rule unit variables are usually named after process variables.");`
has been removed and it has no alternative 🤔

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/impl/NodeContainerImpl.java
danielezonca: Ok can you please add a comment to explain why `LinkedHashMap`? If you think this is wrong/to be changed please create a ticket and add link

jbpm/jbpm-serverless-workflow/src/main/java/org/jbpm/serverless/workflow/parser/handlers/SwitchHandler.java
danielezonca: Ok do you think it makes sense to create/add ticket here to improve this handling?

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/impl/NodeContainerImpl.java
fjtirado: I think order of insertions of nodes should be kept in any case. I was just explaining why I realize it was needed.

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
fjtirado: In my opinion, this message is not needed. It is just reminding the user that may be the rule is not defined because it has the wrong name. I thing that can be inferred from the exception type anyway.

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/core/impl/NodeContainerImpl.java
ricardozanini: @fjtirado I think it's worth adding a comment, in this case, explaining the usage of `LinkedHashMap`. Your previous comment explains that pretty nicely. :)

jbpm/jbpm-serverless-workflow/src/main/java/org/jbpm/serverless/workflow/parser/handlers/SwitchHandler.java
fjtirado: https://issues.redhat.com/browse/KOGITO-4304

PR #1327, opened 2021-05-20
Add Quarkus extension deployment artifacts to the kogito-bom

aloubyansky:
Given that the kogito-bom is imported into the quarkus-universe-bom, it is expected to include the extension deployment artifacts in addition to the runtime ones.

### DIFF
--- a/kogito-bom/pom.xml
+++ b/kogito-bom/pom.xml
@@ -1156,6 +1156,11 @@
       </dependency>
       <dependency>
         <groupId>org.kie.kogito</groupId>
+        <artifactId>kogito-quarkus-deployment</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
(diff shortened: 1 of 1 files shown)


danielezonca:
Thanks @aloubyansky for the PR 👍

danielezonca (approved):

danielezonca:
ok to test

aloubyansky:
@danielezonca off topic, which version do you think will be based on Quarkus 2.x?

danielezonca:
> @danielezonca off topic, which version do you think will be based on Quarkus 2.x?

We have the PR ongoing but we still have to solve some compatibility issues. In the best case scenario I think it could be Kogito 1.8 that means in ~3 weeks from now but more probably Kogito 1.9 in 5/6 weeks from now.
This is my expectation but @evacchi might have more information

PR #1328, opened 2021-05-21
KOGITO-5174 - Add task name to retrieve schema

cristianonicolai:


cristianonicolai:
jenkins please retest this

integration-tests/integration-tests-quarkus-processes/src/test/java/org/kie/kogito/integrationtests/quarkus/TaskIT.java
```diff
@@ -59,19 +58,47 @@
 
     @Test
     void testJsonSchema() {
-        // Quarkus returns URI with "quarkus://" scheme when running via CLI and this is not compatible with
```
danielezonca, line 62: Are you sure this problem will not apply anymore?

cristianonicolai: good point, reverted back

cristianonicolai:
jenkins please retest this

cristianonicolai:
jenkins please retest this

cristianonicolai:
jenkins please retest this

cristianonicolai:
jenkins please retest this

pefernan:
Hey @cristianonicolai code looks well but trying it I'm still not able to get the form. I tried different examples starting in dev mode and looks like the json file isn't available in the classloader .

`{"message":"Cannot find file META-INF/jsonSchema/hiring_HRInterview.json in classpath"}`

cristianonicolai:
@pefernan weird, have you check if the file is generated, and if so the name matches?

cristianonicolai:
jenkins please retest this

pefernan:
@cristianonicolai yea the json files apparently are correctly generated in target `classes/META-INF/jsonSchema`

cristianonicolai:
@pefernan fixed, was failing on dev mode.

cristianonicolai:
jenkins please retest this

pefernan (approved):
works! Thanks @cristianonicolai

integration-tests/integration-tests-quarkus-processes/src/test/java/org/kie/kogito/integrationtests/quarkus/TaskIT.java
```diff
@@ -58,20 +58,60 @@
     }
 
     @Test
-    void testJsonSchema() {
+    void testJsonSchema() throws IOException {
```
fjtirado, line 61: Do we need to throw IOException?

fjtirado (approved):
Just a minor comment in one of the test.
Rest looks good to me. 
Problem was a typo when refactoring REST (thats the real change)
All other changes are to ensure we are testing the json schema instance case (which was the reason this typo was not detected before)

integration-tests/integration-tests-quarkus-processes/src/test/java/org/kie/kogito/integrationtests/quarkus/TaskIT.java
cristianonicolai: I'm just letting the test fail in case that happens when trying to read some of the json test data files.

integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-it/pom.xml
```diff
@@ -209,6 +209,9 @@
                     </plugin>
                 </plugins>
                 <testResources>
+                    <testResource>
+                        <directory>src/test/resources</directory>
```
danielezonca, line 213: Isn't this the default test resource folder? Why you have specified it explicitly?

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/ProcessServiceImpl.java
```diff
@@ -403,6 +404,6 @@ public ProcessServiceImpl(Application application) {
                 id,
                 taskId,
                 Policies.of(user, groups),
-                JsonSchemaUtil.load(this.getClass().getClassLoader(), process.id(), "$taskName$"));
+                JsonSchemaUtil.load(Thread.currentThread().getContextClassLoader(), process.id(), taskName));
```
danielezonca, line 407: In general it is not ideal to use TCCL, we might need to consider to expose it via `Application` 🤔 
Btw it is fine for this PR, this is more a general comment for the future.
Wdyt?

danielezonca (approved):

integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-it/pom.xml
cristianonicolai: its but because there is another folder specified below, this needs to be added so both can work.

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/ProcessServiceImpl.java
cristianonicolai: yeah, that probably makes sense to have the reference in `Application`, the only thing is that this seems to be the only way to get dev mode and app working.

fjtirado: what we can for the future (not for this PR) is to provide a `InputStream getResource(Path path)` method at `Application` level, hiding the classloader from application code.
That getResource can rely on a protected `getClassLoader()` method, which default use the context class loader.

MarianMacik (approved):
Looks good to me.

PR #1329, opened 2021-05-21
[KOGITO-5187] Group kie7 dependencies in a reusable internal BOM

danielezonca:
https://issues.redhat.com/browse/KOGITO-5187

This PR introduce a new bom for internal usage named `kogito-kie7-bom` that handle all kie7 dependencies.
For now kie7 dependencies were explicitly declared with the version instead of using depMan to avoid the misleading where for example `org.drools:drools-core` and `org.kie.kogito:drools-core` are both in the "general" depMan but only the kogito one should be used.
In the future, with Drools8 this bom will be probably removed because and the misleading situation described before will not happen anymore.
The only problem with current approach is that `version.org.kie7` cannot remain in `kogito-build-parent` and it is declared inside `kogito-kie7-bom` itself. This is to keep it completely self contained (and not inherit from other kogito pom).
This PR also remove the duplication of `version.org.kie7` property in `optaplanner-build-parent` that for now we manually keep in sync 

PRs:
- https://github.com/kiegroup/kogito-runtimes/pull/1329
- https://github.com/kiegroup/optaplanner/pull/1336
- https://github.com/kiegroup/kogito-apps/pull/840
- https://github.com/kiegroup/kogito-pipelines/pull/175

- [ ] Your code is properly formatted according to [this configuration](https://github.com/kiegroup/kogito-runtimes/tree/master/kogito-build/kogito-ide-config)

<details>
<summary>
</summary>

* <b>Pull Request</b>  
 
* <b>Quarkus LTS checks</b>  

* <b>Native checks</b>  

</details>


kogito-build/kogito-kie7-bom/pom.xml
```diff
@@ -0,0 +1,152 @@
...
+    <artifactId>kogito-kie7-bom</artifactId>
+    <packaging>pom</packaging>
+
+    <name>Kogito :: Kie7 BOM</name>
+    <description>Internal utility BOM used in Kogito to group kie7 artifact without add them to Kogito dependencyManagement section or kogito-bom.
+        This is to prevent misleading situation where both org.drools:drools-core and org.kie.kogito:drools-core as managed dependency</description>
+
+    <properties>
+        <version.org.kie7>7.55.0.Beta1</version.org.kie7>
+    </properties>
```
danielezonca, line 22: This BOM is intended to group and version kie7 artifacts so that the perimeter is clearly defined. This required to create an "independented" and standalone module (that can be reused also by OptaPlanner for example) that is not inheriting from `kogito-build-parent` (otherwise it will also take all dependencyManagement). 
The consequence of this is that `version.org.kie7` property is defined here.
@mareknovotny 
Wdyt?

danielezonca:
jenkins retest this please

almope (approved):

kogito-build/kogito-kie7-bom/pom.xml
Ginxo: @danielezonca @mareknovotny 
I totally understand the purpose for having `7` on the artifactId and `version.org.kie7` property, but I'm not totally sure this is right. What if we don't heritage version from parent and we force `kogito-kie-bom` to be the one from kie projects? :thinking: This way we won't have to change artifactId in case we want to just point to a different kie version.

kogito-build/kogito-kie7-bom/pom.xml
```diff
@@ -0,0 +1,152 @@
...
+                <groupId>org.drools</groupId>
+                <artifactId>drools-model-compiler</artifactId>
+                <version>${version.org.kie7}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.drools</groupId>
+                <artifactId>drools-alphanetwork-compiler</artifactId>
+                <version>${version.org.kie7}</version>
+            </dependency>
+
```
Ginxo, line 89: empty line to remove

kogito-build/kogito-kie7-bom/pom.xml
```diff
@@ -0,0 +1,152 @@
...
+                <groupId>org.kie</groupId>
+                <artifactId>kie-dmn-validation</artifactId>
+                <version>${version.org.kie7}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.kie</groupId>
+                <artifactId>kie-dmn-openapi</artifactId>
+                <version>${version.org.kie7}</version>
+            </dependency>
+
```
Ginxo, line 112: empty line to remove

kogito-build/kogito-kie7-bom/pom.xml
```diff
@@ -0,0 +1,152 @@
...
+                <groupId>org.drools</groupId>
+                <artifactId>drools-scenario-simulation-api</artifactId>
+                <version>${version.org.kie7}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.drools</groupId>
+                <artifactId>drools-scenario-simulation-backend</artifactId>
+                <version>${version.org.kie7}</version>
+            </dependency>
+
```
Ginxo, line 147: empty line to remove

kogito-build/kogito-kie7-bom/pom.xml
```diff
@@ -0,0 +1,152 @@
...
+        <groupId>org.kie.kogito</groupId>
+        <artifactId>kogito-build</artifactId>
+        <version>2.0.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>kogito-kie7-bom</artifactId>
+    <packaging>pom</packaging>
+
```
Ginxo, line 15: empty line to remove

Ginxo (changes requested):
please check my comments

kogito-build/kogito-kie7-bom/pom.xml
danielezonca: What do you mean?
Some artifacts has a different GAV (org.kie.kogito instead of org.drools) on purpose because they are designed to only work in kogito and viceversa.
Some other will be removed in the future (for example I expect `kie-api` will be removed or at least massively changed).
In general we don't want to include all kie7 artifacts because only few of them (mainly from drools.git repo) are necessary so we cannot reuse existing kie-bom or similar.
It is not clear to me what you are proposing 🤔

evacchi: I think the question is only if having a number in the artifactId ("7") is correct :) I don't think it's an issue

evacchi (approved):
Let's not forget to update the script at https://github.com/kiegroup/kogito-pipelines/blob/master/tools/update-kie7-versions.sh

kogito-build/kogito-kie7-bom/pom.xml
Ginxo: @danielezonca it's exactly what @evacchi means. It's not an issue, just wondering whether this `7` should be there or not. If you all agree it is good to keep it there I totally agree with it.

AntStephenson (approved):
LGTM

kogito-build/kogito-kie7-bom/pom.xml
```diff
@@ -0,0 +1,152 @@
...
+    <name>Kogito :: Kie7 BOM</name>
+    <description>Internal utility BOM used in Kogito to group kie7 artifact without add them to Kogito dependencyManagement section or kogito-bom.
+        This is to prevent misleading situation where both org.drools:drools-core and org.kie.kogito:drools-core as managed dependency</description>
+
+    <properties>
+        <version.org.kie7>7.55.0.Beta1</version.org.kie7>
+    </properties>
+
+    <dependencyManagement>
+        <dependencies>
```
ge0ffrey, line 25: Can we add a comment to note that this bom is a user bom, so it can not import other groupId's than org.kie and org.drools?
```suggestion
        <dependencies>
           <!-- This bom can only import dependencies from the groupId's org.kie and org.drools. -->
```

ge0ffrey (approved):

radtriste (approved):

danielezonca:
> Let's not forget to update the script at https://github.com/kiegroup/kogito-pipelines/blob/master/tools/update-kie7-versions.sh

Done https://github.com/kiegroup/kogito-pipelines/pull/175

kogito-build/kogito-kie7-bom/pom.xml
danielezonca: Done 👍

kogito-build/kogito-kie7-bom/pom.xml
danielezonca: Done 👍

kogito-build/kogito-kie7-bom/pom.xml
danielezonca: Done 👍

kogito-build/kogito-kie7-bom/pom.xml
danielezonca: Done 👍

kogito-build/kogito-kie7-bom/pom.xml
danielezonca: Done 👍

kogito-build/kogito-kie7-bom/pom.xml
mareknovotny: well, if that is really only Drools it should be mentioned there instead of kie as that is more than drools.

danielezonca: Ah ok thanks for the clarification, I would prefer to keep the `7` explicit, even because this will probably be removed in the future together with Drools8.
For example OptaPlanner already has both 7 and 8 versions so its "latest" artifacts should not be never included in this bom. That's why I think `kogito-kie-bom` name is more misleading

Ginxo (approved):
LGTM

kogito-build/kogito-kie7-bom/pom.xml
danielezonca: @mareknovotny 
It is `drools.git` repo mainly but that repo contains DMN, PMML and test scenario runner too for example so `kogito-drools-bom` is still misleading from my point of view

PR #1336, opened 2021-05-26
KOGITO-5207 Remove Generics from Process/ProcessInstance

evacchi:
The generic `T` in `Process<T>`, `ProcessInstance<T>` has never really been used for its intended purpose; instead there are a lot of casts and conversions at various places.

This PR:

- Removes generic type from `Process<T>`, `ProcessInstance<T>`, `ProcessInstances<T>`
- Inserts casts where relevant (very few places in fact, and only temporarily)

Future PR(s):

- will introduce a new API `T variables(Class<T>)` returning an instance of Class<T> with fields set to the variable values. This supersedes also toModel/ToOutputModel
- this API will use reflection, thus reflection will be opted-in for such classes in the Quarkus extension

Disclaimer:

- This PR attempts a "big-bang" change everywhere (in kogito-runtimes). It may not work; if it doesn't work, it will be closed and substituted with a set of smaller changes.
- This PR is intended to be followed up by further refinements;
- This PR is intended to address only kogito-runtimes
- Usages outside of kogito-runtimes should be backwards compatible (for now)
- If necessary an (unused) generic type may-be re-introduced for the sole purposes of making apps/examples compile for the time being
- Follow-up PR will address apps/examples if necessary!

### DIFF
--- a/api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
+++ b/api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
@@ -24,6 +24,7 @@ import java.util.Set;
 import java.util.function.Function;
 import java.util.function.Predicate;
 
+import org.kie.kogito.Model;
 import org.kie.kogito.internal.process.runtime.KogitoNodeInstance;
 import org.kie.kogito.internal.process.runtime.KogitoWorkItem;
 import org.kie.kogito.process.flexible.AdHocFragment;
@@ -31,7 +32,7 @@ import org.kie.kogito.process.flexible.Milestone;
 import org.kie.kogito.process.workitem.Policy;
 import org.kie.kogito.process.workitem.Transition;
 
-public interface ProcessInstance<T> {
+public interface ProcessInstance {
 
     int STATE_PENDING = 0;
     int STATE_ACTIVE = 1;
@@ -45,7 +46,7 @@ public interface ProcessInstance<T> {
      *
      * @return process definition of this process instance
      */
-    Process<T> process();
+    Process process();
 
     /**
      * Starts process instance
@@ -92,12 +93,12 @@ public interface ProcessInstance<T> {
      *
      * @return variables of the process instance
      */
-    T variables();
+    Model variables();
 
     /**
      * Updates process variables of this process instance
      */
-    T updateVariables(T updates);
+    Model updateVariables(Model updates);
 
     /**
      * Returns current status of this process instance
@@ -201,7 +202,7 @@ public interface ProcessInstance<T> {
      */
     Optional<ProcessError> error();
 
-    default ProcessInstance<T> checkError() {
+    default ProcessInstance checkError() {
         Optional<ProcessError> error = error();
         if (error.isPresent()) {
             throw new ProcessInstanceExecutionException(id(), error.get().failedNodeId(), error.get().errorMessage());
--- a/jbpm/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/VariableTagsTest.java
+++ b/jbpm/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/VariableTagsTest.java
@@ -118,16 +118,17 @@ public class VariableTagsTest extends JbpmBpmn2TestCase {
         Map<String, Object> params = new HashMap<>();
         params.put("approver", "john");
 
-        org.kie.kogito.process.ProcessInstance<BpmnVariables> instance = process.createInstance(BpmnVariables.create(params));
+        org.kie.kogito.process.ProcessInstance instance = process.createInstance(BpmnVariables.create(params));
         instance.start();
 
         assertEquals(STATE_ACTIVE, instance.status());
 
-        assertThat(instance.variables().toMap()).hasSize(1);
-        assertThat(instance.variables().toMap(BpmnVariables.OUTPUTS_ONLY)).hasSize(0);
-        assertThat(instance.variables().toMap(BpmnVariables.INPUTS_ONLY)).hasSize(0);
-        assertThat(instance.variables().toMap(BpmnVariables.INTERNAL_ONLY)).hasSize(0);
-        assertThat(instance.variables().toMap(v -> v.hasTag("onlyAdmin"))).hasSize(1).containsEntry("approver", "john");
+        BpmnVariables variables = (BpmnVariables) instance.variables();
+        assertThat(variables.toMap()).hasSize(1);
+        assertThat(variables.toMap(BpmnVariables.OUTPUTS_ONLY)).hasSize(0);
+        assertThat(variables.toMap(BpmnVariables.INPUTS_ONLY)).hasSize(0);
+        assertThat(variables.toMap(BpmnVariables.INTERNAL_ONLY)).hasSize(0);
+        assertThat(variables.toMap(v -> v.hasTag("onlyAdmin"))).hasSize(1).containsEntry("approver", "john");
 
         instance.abort();
 
--- a/kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessGenerator.java
+++ b/kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessGenerator.java
@@ -27,7 +27,6 @@ import javax.lang.model.SourceVersion;
 
 import org.drools.compiler.compiler.io.memory.MemoryFileSystem;
 import org.drools.core.util.KieFunctions;
-import org.drools.core.util.StringUtils;
 import org.jbpm.compiler.canonical.ProcessMetaData;
 import org.jbpm.compiler.canonical.TriggerMetaData;
 import org.jbpm.process.core.datatype.impl.type.ObjectDataType;
@@ -60,7 +59,6 @@ import com.github.javaparser.ast.expr.FieldAccessExpr;
 import com.github.javaparser.ast.expr.MethodCallExpr;
 import com.github.javaparser.ast.expr.NameExpr;
 import com.github.javaparser.ast.expr.ObjectCreationExpr;
-import com.github.javaparser.ast.expr.SimpleName;
 import com.github.javaparser.ast.expr.StringLiteralExpr;
 import com.github.javaparser.ast.expr.SuperExpr;
 import com.github.javaparser.ast.expr.ThisExpr;
@@ -157,12 +155,13 @@ public class ProcessGenerator {
                         .setType(processInstanceFQCN)
                         .setArguments(NodeList.nodeList(
                                 new ThisExpr(),
-                                new NameExpr("value"),
+                                new CastExpr().setType(modelTypeName)
+                                        .setExpression(new NameExpr("value")),
                                 createProcessRuntime())));
 
         methodDeclaration.setName("createInstance")
                 .addModifier(Modifier.Keyword.PUBLIC)
-                .addParameter(modelTypeName, "value")
+                .addParameter(Model.class.getCanonicalName(), "value")
                 .setType(processInstanceFQCN)
                 .addAnnotation(Override.class)
                 .setBody(new BlockStmt()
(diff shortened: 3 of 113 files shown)


jbpm/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/VariableTagsTest.java
```diff
@@ -118,16 +118,17 @@ public void testRequiredVariableFiltering() {
...
-        assertThat(instance.variables().toMap(BpmnVariables.OUTPUTS_ONLY)).hasSize(0);
-        assertThat(instance.variables().toMap(BpmnVariables.INPUTS_ONLY)).hasSize(0);
-        assertThat(instance.variables().toMap(BpmnVariables.INTERNAL_ONLY)).hasSize(0);
-        assertThat(instance.variables().toMap(v -> v.hasTag("onlyAdmin"))).hasSize(1).containsEntry("approver", "john");
+        BpmnVariables variables = (BpmnVariables) instance.variables();
+        assertThat(variables.toMap()).hasSize(1);
+        assertThat(variables.toMap(BpmnVariables.OUTPUTS_ONLY)).hasSize(0);
+        assertThat(variables.toMap(BpmnVariables.INPUTS_ONLY)).hasSize(0);
+        assertThat(variables.toMap(BpmnVariables.INTERNAL_ONLY)).hasSize(0);
+        assertThat(variables.toMap(v -> v.hasTag("onlyAdmin"))).hasSize(1).containsEntry("approver", "john");
```
danielezonca, line 131: Here is an example of what I mean with the previous comment about `Model`/`BpmnVariables`

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstance.java
```diff
@@ -92,12 +93,12 @@
      *
      * @return variables of the process instance
      */
-    T variables();
+    Model variables();
```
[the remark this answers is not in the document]
evacchi: > in the test we don't use generated classes?

this

also, please notice that this is a first step. As mentioned, there will be a `T variables(Class<T>)`. In the meantime, casts will be fine (this will stay internal at least until the other method is present)

evacchi:
omg I can't believe it's green 🟢 
the examples parts is obviously red because there is still a bunch of `Process<?>`s and `ProcessInstance<?>`s, that should be addressable with a single PR 🤞

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessGenerator.java
```diff
@@ -221,6 +190,37 @@ private MethodDeclaration createInstanceGenericWithBusinessKeyMethod(String proc
...
+    //                        .addArgument(new CastExpr(new ClassOrInterfaceType(null, modelTypeName), new NameExpr("value"))));
+    //
+    //        methodDeclaration.setName("createInstance")
+    //                .addModifier(Modifier.Keyword.PUBLIC)
+    //                .addParameter(String.class.getCanonicalName(), BUSINESS_KEY)
+    //                .addParameter(Model.class.getCanonicalName(), "value")
+    //                .setType(processInstanceFQCN)
+    //                .setBody(new BlockStmt().addStatement(returnStmt));
+    //        return methodDeclaration;
+    //    }
```
danielezonca, line 223: To be removed

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessGenerator.java
```diff
@@ -439,8 +438,8 @@ public ClassOrInterfaceDeclaration classDeclaration() {
                                 .addStatement(new ReturnStmt(new ObjectCreationExpr(null,
                                         new ClassOrInterfaceType(null, modelTypeName),
                                         NodeList.nodeList())))))
-                .addMember(createInstanceGenericMethod(processInstanceFQCN))
-                .addMember(createInstanceGenericWithBusinessKeyMethod(processInstanceFQCN))
+                //                .addMember(createInstanceGenericMethod(processInstanceFQCN))
+                //                .addMember(createInstanceGenericWithBusinessKeyMethod(processInstanceFQCN))
```
danielezonca, line 442: To be removed

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessResourceGenerator.java
```diff
@@ -263,8 +263,8 @@ public String generate() {
         template.findAll(MethodDeclaration.class).forEach(this::interpolateMethods);
 
         if (context.hasDI()) {
-            template.findAll(FieldDeclaration.class,
-                    CodegenUtils::isProcessField).forEach(fd -> context.getDependencyInjectionAnnotator().withNamedInjection(fd, processId));
+            //            template.findAll(FieldDeclaration.class,
+            //                    CodegenUtils::isProcessField).forEach(fd -> context.getDependencyInjectionAnnotator().withNamedInjection(fd, processId));
```
danielezonca, line 267: To be removed

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/RestResourceUserTaskSpringTemplate.java
```diff
@@ -97,7 +97,7 @@ public ResponseEntity signal(@PathVariable("id") final String id,
                                @RequestParam(value = "user", required = false) final String user,
                                @RequestParam(value = "group",
                                        required = false) final List<String> groups) {
-        return processService.getTask(process, id, taskId, user, groups, $TaskModel$::from)
+        return processService.<$TaskModel$>getTask(process, id, taskId, user, groups, wi-> ($TaskModel$) $TaskModel$.from(wi))
```
danielezonca, line 100: This is something we might need to improve in the future: here it is not a problem but when `processService` is used directly by user we should avoid the need of explicitly provide the generic `$TaskModel$`

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessInstanceGenerator.java
```diff
@@ -29,8 +30,8 @@
 import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
 import com.github.javaparser.ast.body.ConstructorDeclaration;
 import com.github.javaparser.ast.body.MethodDeclaration;
-import com.github.javaparser.ast.expr.MethodCallExpr;
-import com.github.javaparser.ast.expr.NameExpr;
+import com.github.javaparser.ast.body.VariableDeclarator;
+import com.github.javaparser.ast.expr.*;
```
danielezonca, line 34: Wildcard

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/RestResourceUserTaskSpringTemplate.java
tiagodolphine: yeah, this might change, but it would be better if we could avoid the generics and casting.

tiagodolphine (commented):
Nice work, in general it looks good to me, make the codebase simpler and with the `T variables(Class<T>)` it would avoid casting, tbf there were not many places that required casting.

fjtirado (commented):
Im sorry to say this but still I do not see this PR going in the good direction. 
Let me try to explain why. 
Currently, code is generating
TravelersProcess extends AbstractProcess<org.kie.kogito.test.TravelersModel>
So a developer can
TravellerPRocess process;
variables = process.getVariables();
and the IDE will generate the proper return type TravellersModel automatically. 
In future, what we are proposing is that the user has to
TravellesPRocess process; // does not longer now it has an associated TravellersModel
variables = process.getVariables (TravellersModel.class);
This apparently naive change for me has a huge implication, that developer has to manually search for the rigth model, and even worse he can easily cast to the WRONG model, and he will not realize till he executes the code at runtime.
So my vote is to not do this till we have really asseses the implications for developers using the API

danielezonca:
> Im sorry to say this but still I do not see this PR going in the good direction.
> Let me try to explain why.
> Currently, code is generating
> TravelersProcess extends AbstractProcess<org.kie.kogito.test.TravelersModel>
> So a developer can
> TravellerPRocess process;
> variables = process.getVariables();
> and the IDE will generate the proper return type TravellersModel automatically.
> In future, what we are proposing is that the user has to
> TravellesPRocess process; // does not longer now it has an associated TravellersModel
> variables = process.getVariables (TravellersModel.class);
> This apparently naive change for me has a huge implication, that developer has to manually search for the rigth model, and even worse he can easily cast to the WRONG model, and he will not realize till he executes the code at runtime.
> So my vote is to not do this till we have really asseses the implications for developers using the API

This is not entirely true: at the moment user has no (simple) way to access generate code so in reality he can never write `TravelersProcess` in the source code because it refers to code that is generated after the `compile` step by kogito extension. If you look at all the examples and tests, we always write something like `ProcessInstance<? extends Model>` and then we use methods from `ProcessInstance` interface to interact with it.
With variables we have the same problem, at the moment user cannot really use the typed API so the output of that method is a generic `? extends Model` instance so again only the methods of the interface `Model` are available.

The scaffolding experiment was an attempt to codegenerate these classes upfront so that user can refer them but it has many other maintenance implications: what if users change one of these classes?

Btw I see your point about the possible wrong param for `getVariables` method and in general we should revisit it because it is quite error prone. Probably at the moment I would suggest to return a "fixed" `Model` interface so that user can safely use interface methods or he will have to explicitly cast, but at this point he should "know" to check the type before the cast (like for any other cast).
Wdyt?

evacchi:
just to be clear: I am re-evaluating the need for this PR "per se", and instead provide the generic-less API side-by-side (in a separate package, and module (e.g. `kogito-api-incubation-processes` and `org.kie.kogito.incubation.processes`). Nonetheless, I still think we should drop the generic.

What @fjtirado is saying is correct, we are going to lose the typesafe version; OTOH what @danielezonca is saying summarizes why this change is for the better: 

1. we are not really using it to anywhere -- the codebase is littered with `Process<?>` 
2. we are actually converting between typed models in a number of places: `Model` vs. `InputModel` vs `OutputModel`

the split of `Model` vs. `InputModel` vs `OutputModel` means that really there is no "model" per se, but only a projection of the process variables on a typed POJO. Hence, to me, it makes more sense for the user to pull the data out (our submit it) using a typed POJO and have the mapping occur at use-site, rather then declaration site.

fjtirado:
@danielezonca
Ok, let me rephase the start of my comment : "when the user has an easier access to the generated classes..." ;). I think my point is still valid. 
@evacchi 
I think we better split the issues in order to discuss them:

1. Process knowing about the model or not, so there is not need to typecast when retrieving data from process
2. What is the model? ;)

I think my position about 1) is clear and I think the provided example is enough reason to postpone doing it till we have a conversation with the whole team in order to agree which is the direction we want to take. In my opinion, the fact that we have a lot of "?" in code that generically deals with the process is not a signal that we are doing something wrong, I would say is the right expectation. We need to look at this issue from an outsider developer perspective. I think they will prefer T getVariables rather tham T getVariables (Class<T>). And thats just an example, having information about the model concrete class in AbstractProcess automatically ensure type correctness without having to program a single line, it is all generated. 
Regarding 2), I think we are on the same page. There is not model per se, but different views, the input one, the output one, etc.

fjtirado:
@evacchi @danielezonca @cristianonicolai @tiagodolphine @ricardozanini 
There is a middle way to avoid having too many ? in general code and at the same type keep the type association between generated process and generate model. We can divide current `ProcessInstance` interface in two different classes: the existing   one, which will not not be aware of the model (remove the generic and all method that reference it) and will be used from common code; and a new child one, `ModelProcessInstance` , which will inherit from the existing one, will have the type association, will define`variables` and `updateVariables` methods, and will be inherited by generated code, but will not be used from common code.  
So developers using the concrete process still have the type information when performing `variables` call. I know is slightly  more complex to have one more class in the process instance hierarchy (not so much from my point of view), but i think is a good compromise. I honestly think completely removing type information from generated process is a mistake.

ricardozanini:
I agree about having an interface with generated code like `Type variable = process.get()`. And I think everyone in this thread agrees that this is the path to take instead of the hard cast from the developer side. It will also be a better place from the SB support point of view (where people **will** interact with generated code).

Can we take a simple step now towards this direction and continue to improve in later PRs? Is this an option now?

danielezonca:
> It will also be a better place from the SB support point of view (where people **will** interact with generated code).

@ricardozanini what do you mean with this?

PR #1359, opened 2021-06-08
[KOGITO-4714] - Introducing Quarkus Kogito Processes Extension

ricardozanini:
Signed-off-by: Ricardo Zanini <zanini@redhat.com>

See: https://issues.redhat.com/browse/KOGITO-4714

In this PR, we introduce the possibility of having a Quarkus Kogito Extension with only Process support. It should be ideal for creating simple _straight-through-process_ applications when you don't require rules, decisions, or other platform capabilities.

I also made a small adjustment (HotReload/Naming) in the Serverless Workflow extension.

Enjoy!

PS: I'm using process**es** on plural since all the other extensions follow the same pattern (also codegen).

Many thanks for submitting your Pull Request :heart:! 

Please make sure that your PR meets the following requirements:

- [x] You have read the [contributors guide](CONTRIBUTING.md)
- [x] Your code is properly formatted according to [this configuration](https://github.com/kiegroup/kogito-runtimes/tree/master/kogito-build/kogito-ide-config)
- [x] Pull Request title is properly formatted: `KOGITO-XYZ Subject`
- [x] Pull Request title contains the target branch if not targeting master: `[0.9.x] KOGITO-XYZ Subject`
- [x] Pull Request contains link to the JIRA issue
- [x] Pull Request contains link to any dependent or related Pull Request
- [x] Pull Request contains description of the issue
- [x] Pull Request does not include fixes for issues other than the main ticket

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* <b>Pull Request</b>  
  Please add comment: <b>Jenkins retest this</b>
 
* <b>Quarkus LTS checks</b>  
  Please add comment: <b>Jenkins run LTS</b>

* <b>Native checks</b>  
  Please add comment: <b>Jenkins run native</b>

* <b>Full Kogito testing</b> (with cloud images and operator BDD testing)  
  Please add comment: <b>Jenkins run BDD</b>  
  <b>This check should be used only if a big change is done as it takes time to run, need resources and one full BDD tests check can be done at a time ...</b>
</details>

PR #1366, opened 2021-06-11
[KOGITO-5302] Disable fail on empty beans

fjtirado:
Disable fail on empty beans is all cases.

### DIFF
--- a/kogito-codegen-modules/kogito-codegen-core/src/main/java/org/kie/kogito/codegen/core/ApplicationConfigGenerator.java
+++ b/kogito-codegen-modules/kogito-codegen-core/src/main/java/org/kie/kogito/codegen/core/ApplicationConfigGenerator.java
@@ -74,6 +74,10 @@ public class ApplicationConfigGenerator {
 
         generatedFiles.add(generateApplicationConfigDescriptor(configClassNames));
 
+        if (context.hasDI()) {
+            generatedFiles.add(ObjectMapperGenerator.generate(context));
+        }
+
         return generatedFiles;
     }
 
--- a/kogito-codegen-modules/kogito-codegen-core/src/main/resources/class-templates/config/GlobalObjectMapperQuarkusTemplate.java
+++ b/kogito-codegen-modules/kogito-codegen-core/src/main/resources/class-templates/config/GlobalObjectMapperQuarkusTemplate.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package $Package$;
+
+import java.util.List;
+
+import java.io.IOException;
+import java.util.List;
+
+import com.fasterxml.jackson.databind.JavaType;
+import javax.inject.Singleton;
+import javax.inject.Inject;
+
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+
+import io.quarkus.jackson.ObjectMapperCustomizer;
+
+@Singleton
+public class GlobalObjectMapper implements ObjectMapperCustomizer {
+    
+    @Inject
+    ConfigBean configBean;
+
+    public void customize(ObjectMapper mapper) {
+        if (!configBean.failOnEmptyBean()) {
+            mapper.disable(SerializationFeature.FAIL_ON_EMPTY_BEANS);
+        }
+    }
+}
--- a/api/kogito-api/src/main/java/org/kie/kogito/conf/ConfigBean.java
+++ b/api/kogito-api/src/main/java/org/kie/kogito/conf/ConfigBean.java
@@ -24,6 +24,10 @@ public interface ConfigBean extends KogitoConfig {
 
     boolean useCloudEvents();
 
+    default boolean failOnEmptyBean() {
+        return false;
+    }
+
     String getServiceUrl();
 
     Optional<KogitoGAV> getGav();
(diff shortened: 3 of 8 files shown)


tiagodolphine (approved):
LGTM, just a comment, we could use a ConfigProperty to let the user enable/disable it, and set the default to disable.

ricardozanini:
> LGTM, just a comment, we could use a ConfigProperty to let the user enable/disable it, and set the default to disable.

Can we have a JIRA to track this down?

@fjtirado many thanks for this PR, it's intended to address this issue? https://kie.zulipchat.com/#narrow/stream/232676-kogito/topic/message.20event.20parsing

ricardozanini (approved):

kogito-codegen-modules/kogito-codegen-rules/src/main/resources/class-templates/rules/KogitoObjectMapperQuarkusTemplate.java
```diff
@@ -44,6 +45,7 @@
 public class KogitoObjectMapper implements ObjectMapperCustomizer {
 
     public void customize(ObjectMapper mapper) {
+        mapper.disable(SerializationFeature.FAIL_ON_EMPTY_BEANS);
```
danielezonca, line 48: This customizer is not really global because it is part of rules extension
I think it is better to copy the whole class from your previous PR https://github.com/kiegroup/kogito-examples/pull/728/files

In addition this template is only for quarkus, is it correct? Does the problem reproduce also on spring?

cristianonicolai:
Jenkins rerun apps tests

kogito-codegen-modules/kogito-codegen-rules/src/main/resources/class-templates/rules/KogitoObjectMapperQuarkusTemplate.java
fjtirado: added to spring

fjtirado:
> LGTM, just a comment, we could use a ConfigProperty to let the user enable/disable it, and set the default to disable.

Done, I mead added config property

kogito-codegen-modules/kogito-codegen-rules/src/main/resources/class-templates/rules/KogitoObjectMapperQuarkusTemplate.java
fjtirado: but the KogitoObjectMapper generated there is included in the example without any extra dependency.

fjtirado:
> > LGTM, just a comment, we could use a ConfigProperty to let the user enable/disable it, and set the default to disable.
> 
> Can we have a JIRA to track this down?

I already added it with this one

fjtirado:
jenkins retest this

fjtirado:
jenkins retest this

fjtirado:
Jenkins rerun apps tests

fjtirado:
Jenkins rerun examples test

fjtirado:
Jenkins rerun examples tests

kogito-codegen-modules/kogito-codegen-rules/src/main/resources/class-templates/rules/KogitoObjectMapperQuarkusTemplate.java
danielezonca: This is true only for now that process examples are using platform extension that includes all other extensions. Using SW extension or the new process extension (as soon as it will be merged https://github.com/kiegroup/kogito-runtimes/pull/1359 ) this will not be true anymore and the problem will appear again

fjtirado: I was not aware of such incoming change ( I knew I should not look into this issue, a simple PR is becoming so long...)
Then, in which project we should locate the object mapper factory template so it is included in any extension?

danielezonca: I would prefer to have two different customizers if possible. Can you please verify if we can have multiple impl of `ObjectMapperCustomizer`?
I'm asking this because the existing codegenerated ObjectMapper customizer is very specific for rules and ruleUnit mashalling so I don't like the idea to move it to [common module](https://github.com/kiegroup/kogito-runtimes/tree/master/kogito-codegen-modules/kogito-codegen-core/src/main/java/org/kie/kogito/codegen/core).

fjtirado: As you thought, multiple `ObjectMapperCustomizer` are working, so I kept unmodified the rules one and moved to common module the configurable fail-on-beans template.

kogito-codegen-modules/kogito-codegen-core/src/main/java/org/kie/kogito/codegen/core/ApplicationConfigGenerator.java
```diff
@@ -74,6 +74,8 @@ public ApplicationConfigGenerator addConfigGenerator(ConfigGenerator configGener
 
         generatedFiles.add(generateApplicationConfigDescriptor(configClassNames));
 
+        generatedFiles.add(ObjectMapperGenerator.generate(context));
```
danielezonca, line 77: You probably need to update [`ApplicationGeneratorTest`](https://github.com/kiegroup/kogito-runtimes/blob/master/kogito-codegen-modules/kogito-codegen-core/src/test/java/org/kie/kogito/codegen/core/ApplicationGeneratorTest.java#L38) class 🤔

danielezonca (approved):

fjtirado:
Jenkins rerun examples tests

fjtirado:
Jenkins rerun examples tests

fjtirado:
@kiegroup/gatekeepers There is one flaky test on serverless workflow spring boot that I cannot reproduce locally, I think we can merge

kogito-codegen-modules/kogito-codegen-core/src/main/resources/class-templates/config/GlobalObjectMapperQuarkusTemplate.java
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+package $Package$;
+
+import java.util.List;
+
+import java.io.IOException;
+import java.util.List;
+
+import com.fasterxml.jackson.databind.JavaType;
```
mareknovotny, line 23: we should sort the imports as it is written in java contribution guide and to comply with https://github.com/kiegroup/kogito-runtimes/blob/master/kogito-build/kogito-ide-config/src/main/resources/eclipse.importorder

PR #1370, opened 2021-06-15
[KOGITO-5259] Changes in RestWorkItemHandlerResult

fjtirado:
Refactoring Kogito RestWorkItemHandler to work both with serverless worfkow and bpmn. 

Example PR https://github.com/kiegroup/kogito-examples/pull/733


kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
```diff
@@ -75,39 +90,77 @@ public RestWorkItemHandler(WebClient client) {
...
+        String endPoint = getParam(parameters, ENDPOINT, String.class, null);
+        if (endPoint == null) {
+            throw new IllegalArgumentException("Missing required parameter " + ENDPOINT);
+        }
+        HttpMethod method = HttpMethod.valueOf(getParam(parameters, METHOD, String.class, "GET").toUpperCase());
+        Object inputModel = getParam(parameters, PARAMETER, Object.class, Collections.emptyMap());
+        String user = getParam(parameters, USER, String.class, null);
+        String password = getParam(parameters, PASSWORD, String.class, null);
+        String hostProp = getParam(parameters, HOST, String.class, "localhost");
+        int portProp = getParam(parameters, PORT, Integer.class, 8080);
```
ricardozanini, line 106: Probably a new PR, but if I'm not mistaken, these properties are being read only in build time. It would be nice to also read them in runtime. Kubernetes use cases would map these properties in a ConfigMap to be read by the application on bootstrap. The way we are handling it today a user can't change the property to point to a different endpoint. Makes it very difficult for cloud scenarios.

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandlerUtils.java
```diff
@@ -25,21 +28,19 @@
...
-    @Override
-    public Object apply(Object inputParameter, Object node) {
-        return transform((JsonObject) node, (ObjectNode) inputParameter);
+    private RestWorkItemHandlerUtils() {
     }
 
-    private ObjectNode transform(JsonObject src, ObjectNode target) {
+    private static final Logger logger = LoggerFactory.getLogger(RestWorkItemHandlerUtils.class);
+
+    public static ObjectNode mergeJson(JsonObject src, ObjectNode target) {
```
ricardozanini, line 38: I think we have a similar method: https://github.com/kiegroup/kogito-runtimes/blob/master/kogito-workitems/kogito-openapi-workitem/src/main/java/org/kogito/workitem/openapi/JsonNodeResultHandler.java#L41

In this method, I merge the object nevertheless by handling each possible type. I think we might consider using it here.

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
fjtirado: Im afraid these properties are only read at runtime!. Their value is taken from the bpmn, take a look at the example https://github.com/kiegroup/kogito-examples/pull/733/files#diff-20cf1881c8c0aedd96227de43c0cb3a6924eb40ef5a08224ad83ee92cbe6c391

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandlerUtils.java
fjtirado: Hey, I implement this one first (it arises in the review because I moved the already existing logic to a different class, since I remove the ResultHandler), you added the open api one later and I forgot to mention in the review that a similar method already exist ;)
Lets open a different JIRA to move one of the two implementations to a common place holder ( I think mine is faster, yours shorter and probably more robust) and invoke it from both the Work Item ResultHandler and the Open APi ResultHandler.

fjtirado: I asked for a new JIRA because If we start a discussion about benefit/drawbacks of each implementation we will lost focus ;). 
From yours I like it is more concise than mine, but I do not like the conversion from ObjectMapper to JsonNode and the magic "response" string, thats the reason I made my own that merge the JsonObject obtained from the WS response into the ObjectMapper tha represents the SWF model. 
Since whatever the implementation we chose, we need to modify OpenAPI to allow it to be called from RestWorkItemHandler, better do it in a separate PR.

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
```diff
@@ -141,14 +194,23 @@ static String resolvePathParams(String endPoint, Map<String, Object> parameters,
...
+
         if (value == null) {
-            throw new IllegalArgumentException("Missing required parameter " + paramName);
-        }
-        if (!type.isAssignableFrom(value.getClass())) {
-            throw new IllegalArgumentException("Parameter paramName should be of type " + type + " but it is of type " +
-                    value.getClass());
+            value = defaultValue;
+        } else if (!type.isAssignableFrom(value.getClass())) {
+            if (type.isAssignableFrom(Integer.class) && value.getClass().isAssignableFrom(String.class)) {
```
tiagodolphine, line 203: Why just Integer is being handled in a specific way?

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
```diff
@@ -75,39 +90,77 @@ public RestWorkItemHandler(WebClient client) {
...
-        String portProp = (String) parameters.remove(PORT);
-        WorkItemHandlerResult resultHandler = getParam(parameters, RESULT_HANDLER, WorkItemHandlerResult.class);
+        String endPoint = getParam(parameters, ENDPOINT, String.class, null);
+        if (endPoint == null) {
+            throw new IllegalArgumentException("Missing required parameter " + ENDPOINT);
+        }
+        HttpMethod method = HttpMethod.valueOf(getParam(parameters, METHOD, String.class, "GET").toUpperCase());
+        Object inputModel = getParam(parameters, PARAMETER, Object.class, Collections.emptyMap());
+        String user = getParam(parameters, USER, String.class, null);
+        String password = getParam(parameters, PASSWORD, String.class, null);
```
tiagodolphine, line 104: So we are supporting Basic Authentication for the requests, that's fine, but we need to support OpenID/Oauth2 at some point, that's probably the more common use cases aligned with keycloak.. and with bpmn, don't know if we have support for these parameters in the Editor.. anyway just a comment.

tiagodolphine: So my point is, we need to have a way to choose the security/authentication mechanism depending on the parameters set by the user, not handing the security in-line in the method itself.

tiagodolphine (commented):
LGTM, just added some comments.

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
fjtirado: Note that this is just a convenient hack for a private method. If in future we need to handle more conversion (string to boolean) can be added there, but right now just string to int is needed for PORT variable.

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
fjtirado: In future we will need an Authentication type parameter. Right now, basic authentication is assumed.

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
ricardozanini: Yes, I know that. But this way we still have the configuration problem. These variables should be read from the application.properties file like we are doing in the OpenAPI:
https://github.com/kiegroup/kogito-examples/blob/stable/serverless-workflow-temperature-conversion/conversion-workflow/src/main/resources/application.properties#L7

fjtirado: hmmm, Im afraid I disagree, the process should be self contained.
The reason you are adding there in application.properties for the SWF example is because that info cannot be specified in the workflow itself, but this is not the case in bpmn.

fjtirado: In any case, in order application.properties to be taking into account for bpmn process, this is not the code to be modified, but the bpmn parser (to do the same than the swf parser, but again, I do not think thats needed, I would say it will potentially confusing)

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandlerUtils.java
fjtirado: I added more test to ensure current implementation is robust enough. Thats a job is there when we switch implementation.

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
fjtirado: After a live chat with Ricardo, we agreed that there should be a mechanism to assign values taken from external properties files to bpmn variables. 
Im going to investigate such mechanism, which is beyond the scope of this JIRA (and RestWorkItermHandler code)

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
fjtirado: I decided is better to ensure both BPMN and SWF will pass an integer here rather than having this "hack"

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandlerUtils.java
fjtirado: Create JIRA https://issues.redhat.com/browse/KOGITO-5368

fjtirado:
Jenkins run examples

ricardozanini (approved):

fjtirado:
@tiagodolphine @cristianonicolai I think this is ready to go, wdyt?

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
cristianonicolai: we definitely need to look into that, @fjtirado  should we create a follow up JIRA? just to mention, I think in v7 we used to have WIH constructor parameters that would help with this. Of course, for Kogito we want to rely on the application properties. So I think we can still have these set in the BPMN if the user wants to, but allow to override using application.properties and environment variables.

cristianonicolai (approved):

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
fjtirado: I think we should provide a mechanism to link a bpmn variable  to an env varialbe in the bpmn itself. 
What people is doing now for V7 is to use script tasks to do that, which is kind of cumbersome, I think we need a better mechanism

tiagodolphine (approved):

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
```diff
@@ -141,16 +187,23 @@ static String resolvePathParams(String endPoint, Map<String, Object> parameters,
...
         Object value = parameters.remove(paramName);
         if (value == null) {
-            throw new IllegalArgumentException("Missing required parameter " + paramName);
-        }
-        if (!type.isAssignableFrom(value.getClass())) {
-            throw new IllegalArgumentException("Parameter paramName should be of type " + type + " but it is of type " +
-                    value.getClass());
+            value = defaultValue;
+        } else if (!type.isAssignableFrom(value.getClass())) {
+            if (type.isAssignableFrom(Integer.class) && value.getClass().isAssignableFrom(String.class)) {
```
MarianMacik, line 195: Wouldn't this part make more sense with `type.isAssignableFrom(Integer.class) && String.class.isInstance(value.getClass())`? The first part is something like a consumer type (we will be assigning an Integer to it, so it has to be Integer and above) and the second one is a producer one (we need it to be String and below to be able to actually meaningfully parse it).

kogito-workitems/kogito-rest-workitem/src/test/java/org/kogito/workitem/rest/RestWorkItemHandlerUtilsTest.java
```diff
@@ -0,0 +1,88 @@
...
+        assertEquals("pepe", person.getName());
+    }
+
+    @Test
+    public void testMergeJson() throws IOException {
+        ObjectMapper objectMapper = new ObjectMapper();
+        ObjectNode objectNode = objectMapper.createObjectNode();
+        JsonObject jsonObject = new JsonObject(
+                "{\"id\":26,\"name\":\"pepe\",\"array\":[1,\"manolo\",true,3.5, 10000000000],\"objectArray\":[{\"item\":\"javierito\"}],\"miles\":3.4,\"married\":true,\"population\":1000000}");
+        objectNode = RestWorkItemHandlerUtils.mergeObject(objectNode, jsonObject);
```
MarianMacik, line 70: So this just merges empty JsonObject to a non-empty one, right? As we are mergint `objectNode` to `jsonObject`.

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandlerUtils.java
```diff
@@ -55,7 +84,7 @@ private void setValue(ObjectNode result, String key, Object value) {
         } else if (value instanceof String) {
             result.put(key, (String) value);
         } else if (value instanceof JsonObject) {
-            result.set(key, transform((JsonObject) value, result.objectNode()));
+            result.set(key, mergeJson((JsonObject) value, result.objectNode()));
```
MarianMacik, line 87: Here we are not setting the value directly because we want a deep copy?

kogito-workitems/kogito-rest-workitem/src/test/java/org/kogito/workitem/rest/RestWorkItemHandlerUtilsTest.java
fjtirado: the other way around this merges a non empty json node into an object node, the first parameter is the target, the second one the src

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
```diff
@@ -39,16 +48,21 @@
...
 public class RestWorkItemHandler implements KogitoWorkItemHandler {
 
-    public static final String REST_TASK_TYPE = "Rest Task";
-    public static final String ENDPOINT = "endpoint";
-    public static final String METHOD = "method";
-    public static final String PARAMETER = "Parameter";
+    public static final String REST_TASK_TYPE = "Rest";
+    public static final String ENDPOINT = "Url";
+    public static final String METHOD = "Method";
+    public static final String PARAMETER = "ContentData";
```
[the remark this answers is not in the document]
fjtirado: I wanted to avoid that, but lets change

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
fjtirado: you are right, acually it should be CharSequence.class.isAssignableFrom(value.getClass()). 
I also modify a test to cover that code

MarianMacik (approved):
Looks good now. Thanks!

PR #1372, opened 2021-06-16
[KOGITO-5302] Import order

fjtirado:


mareknovotny (approved):
thanks @fjtirado for fixing the import order

PR #1376, opened 2021-06-17
[KOGITO-5343] PMN project build problem on Windows

fjtirado:
Tell windows to create the directories


fjtirado:
After reviewing the code I see that createDirectories is already invoked

PR #1385, opened 2021-06-22
[KOGITO-5319] KogitoReferenceId handling

fjtirado:
As a quick fix to allow process chaining through messages, if
kogitoReferenceId does not point to an existing process instance, create
a new one.

This is a temporary solution till we migrate jBPM conversation approach
implemented with https://issues.redhat.com/browse/JBPM-9735


fjtirado:
@tiagodolphine this should be trivial, please take a look

tiagodolphine (approved):
@fjtirado in general this is ok, as a temporary solution, but let's keep track to incorporate a proper solution to the communication among processes, this is an important use case IMO for orchestration and choreography I would say.
Thanks!

fjtirado:
@kiegroup/gatekeepers please merge

PR #1401, opened 2021-06-24
KOGITO-5413 - Upgrade Java release version to v11

cristianonicolai:
Assembly:
- https://github.com/kiegroup/kogito-runtimes/pull/1401
- https://github.com/kiegroup/kogito-apps/pull/906
- https://github.com/kiegroup/kogito-examples/pull/747

### DIFF
--- a/kogito-build/kogito-build-parent/pom.xml
+++ b/kogito-build/kogito-build-parent/pom.xml
@@ -51,7 +51,9 @@
     <container.image.kafka>confluentinc/cp-kafka:${version.org.confluent.kafka}</container.image.kafka>
     <container.image.mongodb>mongo:${version.org.mongo}</container.image.mongodb>
 
-    <maven.compiler.release>8</maven.compiler.release>
+    <maven.compiler.release>11</maven.compiler.release>
+    <maven.compiler.source>11</maven.compiler.source>
+    <maven.compiler.target>11</maven.compiler.target>
     <jacoco.agent.argLine/>
     <!--
       JaCoCo coverage data file location. Using a single file for appending in the project's root directory makes it


cristianonicolai:
Jenkins rerun runtimes tests

mareknovotny:
I am ok with this, but as Quarkus did, we should very probably change the kogito major version release for this too so instead of targeting it for 1.9.0.Final, skip that version and do 2.0.0.Final.

danielezonca:
> I am ok with this, but as Quarkus did, we should very probably change the kogito major version release for this too so instead of targeting it for 1.9.0.Final, skip that version and do 2.0.0.Final.

Personally I would prefer to do it now for multiple reasons:
- Quarkus 2.0 is already jdk 11 only so keeping bytecode compatibility with 8 is just an internal limitation
- I don't think Kogito 2.0 will be released just after 1.9, personally I think we will need a bit more time
- JDK8 compatibility in Kogito is just an internal detail: we always report in doc and examples JDK 11 as required

danielezonca (approved):

mareknovotny:
you can release now 2.0, just because of that change.The thing is you are  now changing the release and target byte code of all built jars too

tiagodolphine (approved):

ricardozanini (approved):
:palms_up_together:

kogito-build/kogito-build-parent/pom.xml
```diff
@@ -51,7 +51,9 @@
     <container.image.kafka>confluentinc/cp-kafka:${version.org.confluent.kafka}</container.image.kafka>
     <container.image.mongodb>mongo:${version.org.mongo}</container.image.mongodb>
 
-    <maven.compiler.release>8</maven.compiler.release>
+    <maven.compiler.release>11</maven.compiler.release>
+    <maven.compiler.source>11</maven.compiler.source>
+    <maven.compiler.target>11</maven.compiler.target>
```
MarianMacik, line 56: Isn't the release property enough?
See https://stackoverflow.com/questions/43102787/what-is-the-release-flag-in-the-java-9-compiler

MarianMacik (approved):
Looks good, I have just 2 things:
* Isn't the release property enough?
* What was the purpose of our "internal" limitation to JDK 8 when we anyway require JDK 11 by the enforcer?

krisv:
Imho we allowed Java 1.8 for technical reasons (because Quarkus required this for the extensions) but we have always recommended a JDK 11 minimum to users / customers.  Therefore I think we should just do this without bumping the major version, as there are other things we are planning to be part of the Kogito 2.0 but these are not feasible in a short time frame, and I don't think doing another major update soon after would be recommended.

kogito-build/kogito-build-parent/pom.xml
mareknovotny: no, the defaults for source and target  is 1.6. see https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html

MarianMacik: Yes, but release will override both, right? From http://openjdk.java.net/jeps/247:
```
--release N is roughly equivalent to:

    for N < 9: -source N -target N -bootclasspath <documented-APIs-from-N>,
    for N >= 9: -source N -target N --system <documented-APIs-from-N>.
```
But who knows :)

evacchi (approved):

ricardozanini:
@mareknovotny are you ok with this PR?

PR #1405, opened 2021-06-25
[KOGITO-4537] Claim operation seems to add empty String to some HumanTask fields

elguardian:
if the variables are not in the message we need to check with the right call "hasXXX"


cristianonicolai (approved):

MarianMacik (approved):

cristianonicolai:
Jenkins rerun examples tests

PR #1425, opened 2021-07-08
fix native image when using the legacy-api

mariofusco:


evacchi (approved):
it'd be nice if we had a drools-core module equivalent that doesn't depend on drools-dynamic

evacchi:
jenkins run native

PR #1467, opened 2021-07-22
[KOGITO-5255] Process integration with Decisions - Improvements

tiagodolphine:
Related: https://github.com/kiegroup/kogito-examples/pull/809

Main changes on:
- RuleSetNodeInstance
- RestWorkItemHandler

---

- [ ] Your code is properly formatted according to [this configuration](https://github.com/kiegroup/kogito-runtimes/tree/master/kogito-build/kogito-ide-config)

<details>
<summary>
</summary>

* <b>Run all builds</b>  

* <b>Quarkus LTS checks</b>  

* <b>Native checks</b>  

</details>

### DIFF
--- a/jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/transformation/JsonResolver.java
+++ b/jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/transformation/JsonResolver.java
@@ -34,21 +34,21 @@ import com.fasterxml.jackson.databind.PropertyNamingStrategies;
 
 public class JsonResolver {
 
-    private ObjectMapper objectMapper;
+    private final ObjectMapper objectMapper;
 
     /**
      * @param objectMapper object mapper to be used when converting input items
      */
     public JsonResolver(ObjectMapper objectMapper) {
         this.objectMapper = objectMapper;
-        this.objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
-                .configure(DeserializationFeature.ACCEPT_EMPTY_STRING_AS_NULL_OBJECT, true)
-                .setPropertyNamingStrategy(PropertyNamingStrategies.LOWER_CAMEL_CASE);
-
     }
 
     public JsonResolver() {
-        this(new ObjectMapper().setSerializationInclusion(JsonInclude.Include.NON_NULL));
+        this(new ObjectMapper()
+                .setSerializationInclusion(JsonInclude.Include.NON_NULL)
+                .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+                .configure(DeserializationFeature.ACCEPT_EMPTY_STRING_AS_NULL_OBJECT, true)
+                .setPropertyNamingStrategy(PropertyNamingStrategies.LOWER_CAMEL_CASE));
     }
 
     /**
--- a/jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/transformation/JsonResolver.java
+++ b/jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/transformation/JsonResolver.java
@@ -62,7 +62,8 @@ public class JsonResolver {
 
     private boolean hasJacksonAnnotations(AnnotatedElement element) {
         Annotation[] declaredAnnotations = element.getDeclaredAnnotations();
-        Annotation[] declaringClassAnnotations = Optional.of(element).filter(Field.class::isInstance).map(Field.class::cast).map(Field::getType).map(Class::getDeclaredAnnotations).orElse(new Annotation[0]);
+        Annotation[] declaringClassAnnotations =
+                Optional.of(element).filter(Field.class::isInstance).map(Field.class::cast).map(Field::getType).map(Class::getDeclaredAnnotations).orElse(new Annotation[0]);
         if (!Stream.of(declaredAnnotations, declaringClassAnnotations).flatMap(Stream::of)
                 .noneMatch(a -> a.annotationType().getAnnotationsByType(JacksonAnnotation.class).length > 0)) {
             return true;
--- a/jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/transformation/JsonResolver.java
+++ b/jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/transformation/JsonResolver.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jbpm.process.core.transformation;
+
+import java.lang.reflect.AnnotatedElement;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+import com.fasterxml.jackson.annotation.JacksonAnnotation;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public class JsonResolver {
+
+    private ObjectMapper objectMapper;
+
+    public JsonResolver(ObjectMapper objectMapper) {
+        this.objectMapper = objectMapper;
+    }
+
+    public JsonResolver() {
+        this(new ObjectMapper().setSerializationInclusion(JsonInclude.Include.NON_NULL));
+    }
+
+    /**
+     * Receive an Input Map of items and returns a new Map containing all items from input but the values
+     * with Jackson annotations are translated into an inner Map in a recursive way. This is useful for the
+     * communication of 2 services based on json parameter values as inputs but already deserialized into java POJOs.
+     * 
+     * @param items Input items map
+     * @return a new Map containing all items from input but with the resolved values to Json if applied
+     */
+    public Map<String, Object> resolveItems(Map<String, Object> items) {
+        Map<String, Map> resolved = items.entrySet().stream()
+                .filter(v -> Objects.nonNull(v.getValue()))
+                .filter(v -> hasJacksonAnnotations(v.getValue().getClass()))
+                .collect(Collectors.toMap(Map.Entry::getKey, v -> objectMapper.convertValue(v.getValue(), Map.class)));
+        HashMap<String, Object> result = new HashMap<>(items);
+        result.putAll(resolved);
+        return result;
+    }
+
+    private boolean hasJacksonAnnotations(AnnotatedElement element) {
+        if (!Arrays.stream(element.getDeclaredAnnotations())
+                .noneMatch(a -> a.annotationType().getAnnotationsByType(JacksonAnnotation.class).length > 0)) {
+            return true;
+        }
+        if (!(element instanceof Class) || ((Class<?>) element).isPrimitive()) {
+            return false;
+        }
+        Class<?> clazz = (Class<?>) element;
+        return !Arrays.stream(clazz.getDeclaredFields()).noneMatch(f -> hasJacksonAnnotations(f));
+    }
+}
--- a/jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/transformation/JsonResolver.java
+++ b/jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/transformation/JsonResolver.java
@@ -15,12 +15,16 @@
  */
 package org.jbpm.process.core.transformation;
 
+import java.lang.annotation.Annotation;
 import java.lang.reflect.AnnotatedElement;
+import java.lang.reflect.Field;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import com.fasterxml.jackson.annotation.JacksonAnnotation;
 import com.fasterxml.jackson.annotation.JsonInclude;
@@ -57,7 +61,9 @@ public class JsonResolver {
     }
 
     private boolean hasJacksonAnnotations(AnnotatedElement element) {
-        if (!Arrays.stream(element.getDeclaredAnnotations())
+        Annotation[] declaredAnnotations = element.getDeclaredAnnotations();
+        Annotation[] declaringClassAnnotations = Optional.of(element).filter(Field.class::isInstance).map(Field.class::cast).map(Field::getType).map(Class::getDeclaredAnnotations).orElse(new Annotation[0]);
+        if (!Stream.of(declaredAnnotations, declaringClassAnnotations).flatMap(Stream::of)
                 .noneMatch(a -> a.annotationType().getAnnotationsByType(JacksonAnnotation.class).length > 0)) {
             return true;
         }
--- a/jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/transformation/JsonResolver.java
+++ b/jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/transformation/JsonResolver.java
@@ -34,6 +34,9 @@ public class JsonResolver {
 
     private ObjectMapper objectMapper;
 
+    /**
+     * @param objectMapper object mapper to be used when converting input items
+     */
     public JsonResolver(ObjectMapper objectMapper) {
         this.objectMapper = objectMapper;
     }
@@ -58,6 +61,12 @@ public class JsonResolver {
         HashMap<String, Object> result = new HashMap<>(items);
         result.putAll(resolved);
         return result;
+    }
+
+    public Map<String, Object> resolveAll(Map<String, Object> items) {
+        return items.entrySet()
+                .stream()
+                .collect(Collectors.toMap(Map.Entry::getKey, v -> objectMapper.convertValue(v.getValue(), Map.class)));
     }
 
     private boolean hasJacksonAnnotations(AnnotatedElement element) {
--- a/jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/transformation/JsonResolver.java
+++ b/jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/transformation/JsonResolver.java
@@ -28,7 +28,9 @@ import java.util.stream.Stream;
 
 import com.fasterxml.jackson.annotation.JacksonAnnotation;
 import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.PropertyNamingStrategies;
 
 public class JsonResolver {
 
@@ -39,6 +41,10 @@ public class JsonResolver {
      */
     public JsonResolver(ObjectMapper objectMapper) {
         this.objectMapper = objectMapper;
+        this.objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+                .configure(DeserializationFeature.ACCEPT_EMPTY_STRING_AS_NULL_OBJECT, true)
+                .setPropertyNamingStrategy(PropertyNamingStrategies.LOWER_CAMEL_CASE);
+
     }
 
     public JsonResolver() {
@@ -53,7 +59,10 @@ public class JsonResolver {
      * @param items Input items map
      * @return a new Map containing all items from input but with the resolved values to Json if applied
      */
-    public Map<String, Object> resolveItems(Map<String, Object> items) {
+    public Map<String, Object> resolveOnlyAnnotatedItems(Map<String, Object> items) {
+        if (Objects.isNull(items)) {
+            return null;
+        }
         Map<String, Map> resolved = items.entrySet().stream()
                 .filter(v -> Objects.nonNull(v.getValue()))
                 .filter(v -> hasJacksonAnnotations(v.getValue().getClass()))
@@ -64,9 +73,9 @@ public class JsonResolver {
     }
 
     public Map<String, Object> resolveAll(Map<String, Object> items) {
-        return items.entrySet()
-                .stream()
-                .collect(Collectors.toMap(Map.Entry::getKey, v -> objectMapper.convertValue(v.getValue(), Map.class)));
+        return Optional.ofNullable(items)
+                .map(input -> objectMapper.convertValue(items, Map.class))
+                .orElse(null);
     }
 
     private boolean hasJacksonAnnotations(AnnotatedElement element) {
--- a/jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/transformation/JsonResolver.java
+++ b/jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/transformation/JsonResolver.java
@@ -82,14 +82,14 @@ public class JsonResolver {
         Annotation[] declaredAnnotations = element.getDeclaredAnnotations();
         Annotation[] declaringClassAnnotations =
                 Optional.of(element).filter(Field.class::isInstance).map(Field.class::cast).map(Field::getType).map(Class::getDeclaredAnnotations).orElse(new Annotation[0]);
-        if (!Stream.of(declaredAnnotations, declaringClassAnnotations).flatMap(Stream::of)
-                .noneMatch(a -> a.annotationType().getAnnotationsByType(JacksonAnnotation.class).length > 0)) {
+        if (Stream.of(declaredAnnotations, declaringClassAnnotations).flatMap(Stream::of)
+                .anyMatch(a -> a.annotationType().getAnnotationsByType(JacksonAnnotation.class).length > 0)) {
             return true;
         }
         if (!(element instanceof Class) || ((Class<?>) element).isPrimitive()) {
             return false;
         }
         Class<?> clazz = (Class<?>) element;
-        return !Arrays.stream(clazz.getDeclaredFields()).noneMatch(f -> hasJacksonAnnotations(f));
+        return Arrays.stream(clazz.getDeclaredFields()).anyMatch(f -> hasJacksonAnnotations(f));
     }
 }
--- a/jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
+++ b/jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
@@ -67,12 +67,15 @@ import org.kie.dmn.api.core.DMNRuntime;
 import org.kie.internal.runtime.StatefulKnowledgeSession;
 import org.kie.kogito.decision.DecisionModel;
 import org.kie.kogito.dmn.DmnDecisionModel;
+import org.kie.kogito.dmn.rest.DMNJSONUtils;
 import org.kie.kogito.internal.process.runtime.KogitoNodeInstance;
 import org.kie.kogito.rules.RuleUnitData;
 import org.kie.kogito.rules.RuleUnitInstance;
 import org.mvel2.integration.impl.MapVariableResolverFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
 
 /**
  * Runtime counterpart of a ruleset node.
@@ -134,7 +137,10 @@ public class RuleSetNodeInstance extends StateBasedNodeInstance implements Event
                                         model))
                                 .get();
 
-                DMNContext context = modelInstance.newContext(inputs);
+                ObjectMapper objectMapper = new ObjectMapper();
+                DMNContext context = DMNJSONUtils.ctx(modelInstance, objectMapper.readValue(objectMapper.writeValueAsString(inputs), Map.class));
+                logger.info("DMN with context {}", context);
+
                 DMNResult dmnResult = modelInstance.evaluateAll(context);
 
                 if (dmnResult.hasErrors()) {
--- a/jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
+++ b/jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
@@ -137,9 +137,8 @@ public class RuleSetNodeInstance extends StateBasedNodeInstance implements Event
                                         model))
                                 .get();
 
-                ObjectMapper objectMapper = new ObjectMapper();
-
                 //Input Binding
+                ObjectMapper objectMapper = new ObjectMapper();
                 DMNContext context = DMNJSONUtils.ctx(modelInstance, objectMapper.readValue(objectMapper.writeValueAsString(inputs), Map.class));
                 logger.info("DMN with context {}", context);
                 DMNResult dmnResult = modelInstance.evaluateAll(context);
--- a/jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
+++ b/jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
@@ -16,7 +16,9 @@
 package org.jbpm.workflow.instance.node;
 
 import java.io.Serializable;
+import java.util.AbstractMap;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
@@ -75,6 +77,8 @@ import org.mvel2.integration.impl.MapVariableResolverFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.fasterxml.jackson.annotation.JacksonAnnotation;
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 /**
@@ -138,8 +142,11 @@ public class RuleSetNodeInstance extends StateBasedNodeInstance implements Event
                                 .get();
 
                 //Input Binding
-                ObjectMapper objectMapper = new ObjectMapper();
-                DMNContext context = DMNJSONUtils.ctx(modelInstance, objectMapper.readValue(objectMapper.writeValueAsString(inputs), Map.class));
+                ObjectMapper objectMapper = new ObjectMapper().setSerializationInclusion(JsonInclude.Include.NON_NULL);
+                inputs.entrySet().stream().filter(v -> hasJacksonAnnotations(v)).map(v -> new AbstractMap.SimpleEntry(v.getKey(), objectMapper.convertValue(v.getValue(), Map.class)))
+                        .forEach(v -> inputs.put((String) v.getKey(), v.getValue()));
+                DMNContext context = DMNJSONUtils.ctx(modelInstance, inputs);
+
                 logger.info("DMN with context {}", context);
                 DMNResult dmnResult = modelInstance.evaluateAll(context);
 
@@ -208,6 +215,17 @@ public class RuleSetNodeInstance extends StateBasedNodeInstance implements Event
         } catch (Exception e) {
             handleException(e);
         }
+    }
+
+    public boolean hasJacksonAnnotations(Entry<String, Object> v) {
+        //TODO refactor
+        Class<?> clazz = v.getValue().getClass();
+        if (Arrays.stream(clazz.getDeclaredAnnotations()).filter(a -> a.annotationType().getAnnotationsByType(JacksonAnnotation.class).length > 0).findAny().isPresent()) {
+            return true;
+        }
+
+        return Arrays.stream(clazz.getDeclaredFields()).flatMap(f -> Arrays.stream(f.getDeclaredAnnotations())).filter(a -> a.annotationType().getAnnotationsByType(JacksonAnnotation.class).length > 0)
+                .findAny().isPresent();
     }
 
     private void handleException(Throwable e) {
--- a/jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
+++ b/jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
@@ -16,9 +16,7 @@
 package org.jbpm.workflow.instance.node;
 
 import java.io.Serializable;
-import java.util.AbstractMap;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
@@ -40,6 +38,7 @@ import org.jbpm.process.core.context.variable.VariableScope;
 import org.jbpm.process.core.datatype.DataType;
 import org.jbpm.process.core.datatype.impl.type.ObjectDataType;
 import org.jbpm.process.core.impl.DataTransformerRegistry;
+import org.jbpm.process.core.transformation.JsonResolver;
 import org.jbpm.process.instance.ContextInstance;
 import org.jbpm.process.instance.ContextInstanceContainer;
 import org.jbpm.process.instance.context.exception.ExceptionScopeInstance;
@@ -77,10 +76,6 @@ import org.mvel2.integration.impl.MapVariableResolverFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.fasterxml.jackson.annotation.JacksonAnnotation;
-import com.fasterxml.jackson.annotation.JsonInclude;
-import com.fasterxml.jackson.databind.ObjectMapper;
-
 /**
  * Runtime counterpart of a ruleset node.
  */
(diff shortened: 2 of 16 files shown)


tiagodolphine:
@fjtirado I did some changes in the WIH to make the multiple params work.

tiagodolphine:
Jenkins run runtimes

tiagodolphine:
Jenkins run runtimes tests

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
```diff
@@ -146,7 +179,7 @@ private RestWorkItemTargetInfo getTargetInfo(KogitoWorkItem workItem) {
         Variable variable = variableScope.findVariable(varName);
         if (variable != null) {
             try {
-                return Class.forName(variable.getType().getStringType());
+                return Thread.currentThread().getContextClassLoader().loadClass(variable.getType().getStringType());
```
danielezonca, line 182: Is TCCL safe to be used? It is probably the right classLoader but we should try to use the instance provided to processRuntime so that we are sure to always use the same

danielezonca (commented):
Hi, I know it is still draft so feel free to ignore my comments if the code will change :)

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
```diff
@@ -68,16 +73,27 @@
...
-            return value instanceof WorkItemHandlerParamResolver
-                    ? ((WorkItemHandlerParamResolver) value).apply(inputModel)
-                    : value;
+            if (value instanceof WorkItemHandlerParamResolver) {
+                return ((WorkItemHandlerParamResolver) value).apply(inputModel);
+            }
+            if (value instanceof String && MVELProcessHelper.validateExpression((String) value).isEmpty()) {
+                //evaluate the expression with Variables resolution
+                try {
+                    return MVELProcessHelper.evaluator().eval((String) value, new NodeInstanceResolverFactory((NodeInstance) workItem.getNodeInstance()));
```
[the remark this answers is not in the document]
evacchi: yes, AFAIR that's how it's done now

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
```diff
@@ -126,7 +129,10 @@ public void internalTrigger(KogitoNodeInstance from, String type) {
                                         model))
                                 .get();
 
-                DMNContext context = modelInstance.newContext(inputs);
+                //Input Binding
+                DMNContext context = DMNJSONUtils.ctx(modelInstance, jsonResolver.resolveItems(inputs));
```
tarilabs, line 133: Is this change to support when an `inputs` k-v has "person" and a composite object for value as a `Map`,
now instead would be converted to the expected (on jBPM side) `Person.class`?

For example, Before:
```
inputs = Map.of("person", Map.of("name", "John Doe", "age", 47))
```

After this code change, now instead `input` would be passed to DMN as
```
inputs = Map.of("person", Person.class@12345)
```

Am I understanding correctly?

This change is not _necessary_ for DMN (strictly speaking), but it is welcome if it's idiomatic for jBPM on kogito.

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
```diff
@@ -126,7 +129,10 @@ public void internalTrigger(KogitoNodeInstance from, String type) {
...
+
+                logger.info("DMN with context {}", context);
```
tarilabs, line 135: I would recommend to change this by trimming the toString of `context` which could be quite bloated, or alternatively change this to debug level before merging.

tarilabs (commented):
@danielezonca 👇

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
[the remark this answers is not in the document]
tiagodolphine: good point, in the expression evaluation, does not work in native...  that might be an issue... the thing we need to solve here is basically when the user defines an expression to be mapped to and from a process variable... so probably we would need some codegen at this point... wdyt? @danielezonca @evacchi @fjtirado ^

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
```diff
@@ -192,17 +225,7 @@ static String resolvePathParams(String endPoint, Map<String, Object> parameters,
...
-                    value = Integer.parseInt(value.toString());
-                } catch (NumberFormatException ex) {
-                    value = defaultValue;
-                }
-            } else {
-                throw new IllegalArgumentException("Parameter paramName should be of type " + type +
-                        " but it is of type " +
-                        value.getClass());
-            }
+            value = MVELProcessHelper.evaluator().eval((String) value, type);
```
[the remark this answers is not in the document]
tiagodolphine: yeah, same issue

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
tiagodolphine: the thing is, for example, a scenario where we have in the process:
```
class Person
   @JsonProperty("User Name")
   String userName
```
 
in DMN we have a data type:
```
Person
 - "User Name" : string
```
 
So, we need to map the input for the DMN context from the process variable defined in the Person class, in case the attribute names contain spaces or whatever, not a simple java attribute name identifier. See https://github.com/kiegroup/kogito-examples/blob/ebe5c4c1d658980cd8d58055474b44a9a90c5a5a/process-decisions-quarkus/src/main/java/org/kie/kogito/traffic/Violation.java in this PR https://github.com/kiegroup/kogito-examples/pull/809

WDYT? ideas? this was just the first solution that came to me, like doing json translation to Map with Jackson in case we find JacksonAnnotations in the Pojo class.

@tarilabs @danielezonca ^

tarilabs: Maybe this is better discussed in a quick call, because I'm a little lost sorry :)

When feeding into DMN input side, `DMNJSONUtils.ctx()` should be able to deal with Maps of Maps etc.
This is especially needed when in an incoming REST payload the Jackson would deserialize as a String, the value in a Map corresponding to a `FEEL:date` for ItemDefinition.
The DMNJSONUtils is aware of the ItemDefinition and the typeRef of the InputData node, and deals with creating a more adherently typed Map of Map (and instead of the String, convert to a `java.time.Date` etc as expected).

So my comment is that I'm not entirely following why modification on input side BPMN->DMN is needed for `jsonResolver.resolveItems`, beyond actually using DMNJSONUtils which is very welcome change.

What you might also want to look at, is how to deal with the other direction, DMN returning back to BPMN, since most of the time on BPMN you have a Java FQCN you expect.
I believe you might be interested to apply here the same code changes related to https://issues.redhat.com/browse/BAPL-1963

As mentioned, always available for a quick call if preferred/more suitable!

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
tiagodolphine: sure, I can change it

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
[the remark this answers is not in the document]
danielezonca: @lucamolteni 
Is it possible to use drools-mvel-compiler at compile time to convert the expression in a Java Lambda?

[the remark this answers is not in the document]
lucamolteni: Yes, take a look at the examples in https://github.com/kiegroup/drools/blob/main/drools-model/drools-mvel-compiler/src/test/java/org/drools/mvelcompiler/MvelCompilerTest.java 

`mvel-compiler` is self contained and has no specific Drools code inside

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
tarilabs: Thank you @tiagodolphine @evacchi for explaining me in more details the purpose of these line of code.
I was wrong on my previous https://github.com/kiegroup/kogito-runtimes/pull/1467#discussion_r678035883 I understand now, sorry.

As discussed, on Kie v7 the idiomatic way was to annotate with `@FEELProperty` the desired getter when the identifier was composed on non-Java-compatible characters following the POJO naming convention.

I think the current code highlight is working as expected on Kogito, without any need for  `@FEELProperty` or additional annotation beyond Jackson, and idiomatic of Kogito 👍

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/transformation/JsonResolver.java
```diff
@@ -0,0 +1,77 @@
...
+        if (!Stream.of(declaredAnnotations, declaringClassAnnotations).flatMap(Stream::of)
+                .noneMatch(a -> a.annotationType().getAnnotationsByType(JacksonAnnotation.class).length > 0)) {
+            return true;
+        }
+        if (!(element instanceof Class) || ((Class<?>) element).isPrimitive()) {
+            return false;
+        }
+        Class<?> clazz = (Class<?>) element;
+        return !Arrays.stream(clazz.getDeclaredFields()).noneMatch(f -> hasJacksonAnnotations(f));
+    }
```
tarilabs, line 76: As mentioned accordingly to https://github.com/kiegroup/kogito-runtimes/pull/1467#discussion_r678460272 
I **personally** don't think this is needed.

Here is about Kogito, and if a user defines a Domain Model class, that is expected to be serializable, and for Kogito the serialization framework of choice is already Jackson.

So if a BPMN2 process variable is indicated to be a particular class, that is _expected_ to be serializable, I believe.

my2c

tiagodolphine: yeah, I agree, after our conversation, I don't see the need to check if each item contains Jackson annotations, we can assume all items are serializable. So we can remove all this reflection code to check it.

danielezonca:
@triceo @rsynek 
OptaPlanner build is failed with
`kiegroup_optaplanner/optaplanner/optaplanner-core/src/test/java/org/optaplanner/core/impl/heuristic/thread/OrderByMoveIndexBlockingQueueTest.java:135 | java.lang.IllegalStateException: The move thread with moveThreadIndex (1) has thrown an exception. Relayed here in the parent thread. 	at org.optaplanner.core.impl.heuristic.thread.OrderByMoveIndexBlockingQueue.take(OrderByMoveIndexBlockingQueue.java:147)`
https://github.com/kiegroup/kogito-runtimes/pull/1467/checks?check_run_id=3174754429

I don't see how this PR can impact so it could be a flaky test, can you please have a look?

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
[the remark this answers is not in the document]
tiagodolphine: I spoke with @evacchi about this, and the `mvel-compiler` could be a possibility but we would need to do it in a different place, during the compile / codegen phase. At this point the work item handler is more runtime, so a quick win (for now) to allow the easily integration with Decisions when doing the data mapping for the REST requests, and a restriction it won't work in native mode, and we open a JIRA to properly handle expressions not only for this case, but but all other places we are able to define expressions in processes.

tiagodolphine:
jenkins retest this

rsynek:
> @triceo @rsynek
> OptaPlanner build is failed with
> `kiegroup_optaplanner/optaplanner/optaplanner-core/src/test/java/org/optaplanner/core/impl/heuristic/thread/OrderByMoveIndexBlockingQueueTest.java:135 | java.lang.IllegalStateException: The move thread with moveThreadIndex (1) has thrown an exception. Relayed here in the parent thread. at org.optaplanner.core.impl.heuristic.thread.OrderByMoveIndexBlockingQueue.take(OrderByMoveIndexBlockingQueue.java:147)`
> https://github.com/kiegroup/kogito-runtimes/pull/1467/checks?check_run_id=3174754429
> 
> I don't see how this PR can impact so it could be a flaky test, can you please have a look?

Thanks for reporting. This test focuses on the internals of multi-threaded solving, so there is some room for instability. However, I haven't seen it failing ever since I remember. The line number OrderByMoveIndexBlockingQueueTest.java:135 does not make much sense; there should not be any failure. Unfortunately, the GHA run has been already replaced by a new one.

tiagodolphine:
jenkins retest this

cristianonicolai:
Jenkins rerun apps tests

cristianonicolai:
Jenkins rerun examples tests

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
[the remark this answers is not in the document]
cristianonicolai: @tiagodolphine I guess that's fine and we need to review expressions anyway, lets just highlight in the docs somehow that this is a known limitation in native for now.

tiagodolphine:
jenkins retest this

tiagodolphine:
Updates:
- Removed the MVEL evaluation from the RestWorkItemHandler
- Adding BodyBuilder to allow to be parametrized with Strings (the simple class name)
- Update the RuleSetNodeInstance to resolve the inputs as Json for DMN

tiagodolphine:
jenkins retest this

cristianonicolai:

cristianonicolai (approved):

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
```diff
@@ -93,6 +102,9 @@ public void executeWorkItem(KogitoWorkItem workItem, KogitoWorkItemManager manag
         logger.debug("Using target {}", targetInfo);
         //retrieving parameters
         Map<String, Object> parameters = new HashMap<>(workItem.getParameters());
+        //removing unnecessary parameter
```
fjtirado, line 105: This logic should be in body builder implementation

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
fjtirado: is this change needed?, whats the advantage of use TTCL over just Class.forName?

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
```diff
@@ -106,23 +118,50 @@ public void executeWorkItem(KogitoWorkItem workItem, KogitoWorkItemManager manag
...
 
         logger.debug("Filtered parameters are {}", parameters);
         // create request
         UnaryOperator<Object> resolver = new RestUnaryOperator(inputModel);
         endPoint = resolvePathParams(endPoint, parameters, resolver);
-        URI uri = URI.create(endPoint);
-        String host = uri.getHost() != null ? uri.getHost() : hostProp;
-        int port = uri.getPort() > 0 ? uri.getPort() : portProp;
-        HttpRequest<Buffer> request = client.request(method, port, host, uri.getPath());
+        Optional<URL> url = getUrl(endPoint);
```
fjtirado, line 127: Honestly, I do not see the need for that change.
Does uri.getPath() the enconding problem we are trying to fix with a hardcoded replace?

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
```diff
@@ -106,23 +118,50 @@ public void executeWorkItem(KogitoWorkItem workItem, KogitoWorkItemManager manag
...
+    public RestWorkItemHandlerBodyBuilder getBodyBuilder(Map<String, Object> parameters) {
+        Object param = parameters.get(BODY_BUILDER);
+        if (Objects.isNull(param)) {
+            return DEFAULT_BODY_BUILDER;
+        }
+        if (param instanceof RestWorkItemHandlerBodyBuilder) {
+            return (RestWorkItemHandlerBodyBuilder) param;
+        }
+        if (param instanceof String) {
+            return BODY_BUILDERS.get(param);
```
fjtirado, line 151: why not creating the instance from the class name here?

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
tiagodolphine: I was getting class not found in the Pojo Classes defined in the process.

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
```diff
@@ -63,6 +69,9 @@
     private static final Logger logger = LoggerFactory.getLogger(RestWorkItemHandler.class);
     private static final RestWorkItemHandlerResult DEFAULT_RESULT_HANDLER = new DefaultRestWorkItemHandlerResult();
     private static final RestWorkItemHandlerBodyBuilder DEFAULT_BODY_BUILDER = new DefaultWorkItemHandlerBodyBuilder();
+    private static final Map<String, RestWorkItemHandlerBodyBuilder> BODY_BUILDERS = Stream
```
fjtirado, line 72: I think this is too much overhead for creating a mapo with just two keys. 
In any case, I think this map can be used in a different way
I think it shoul be empty initially and then lazily populate it on demand by instantiating the class if the user specify class name

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandlerUtils.java
```diff
@@ -62,14 +65,37 @@ private static void setValue(PropertyDescriptor[] propertyDescritors, Object tar
...
         }
-        logger.warn("No property found for name {}", name);
+        logger.debug("No property found for name {}", name);
+    }
+
+    private static Object convert(Object value, Class<?> targetClass) {
+        if (!targetClass.isAssignableFrom(value.getClass())) {
+            if (Date.class.isAssignableFrom(targetClass)) {
+                if (value instanceof String) {
+                    return Date.from(Instant.from(DateTimeFormatter.ISO_DATE_TIME.parse(value.toString())));
```
fjtirado, line 82: Just use  DateTimeFormatter.ISO_DATE_TIME.parse(value.toString()

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
fjtirado: I mean bodyBuilder.computeIfAbsent(x -> Class.forName((String)param))

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
fjtirado: I mean I see you do not like null checks in code ;), but lets use URI rather that URL in the optional.

fjtirado (commented):
Some minor comments, but looks good to me.  
I like the improvement that allow body builder to be an instance or the className. 
On top of that, I think we can change default body builder to use parameters (in case content_data is null), so the examples wont need to specify body builder parameter. 
Also, SWF wont have to specify a different hander
I can do this changes once this is merger with a followup JIRA, wdyt?

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
tiagodolphine: URI.create with a path that contains spaces throws an exception.

`Exception in thread "main" java.lang.IllegalArgumentException: Illegal character in path at index 29: http://localhost:8080/Traffic Violation`

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
tiagodolphine: TaskName is set in the parameters and is not used at all, so the body was being serialized with it. We can just filter and generate a new map instead of removing.. from original map.

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
tiagodolphine: IMO instantiating a new instance in every request is way worse, IMO better and more extensible to keep a single instance of each and get by a hash, in case we add more body builders impl over time.

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandlerUtils.java
tiagodolphine: I'll test again, but IIRC I tested it and get an issue... let me check again.

tiagodolphine: The `DateTimeFormatter.ISO_DATE_TIME.parse` returns a TemporalAccessor, where it should ba Date.. so a type mismatch... just tested it again.

```
2021-08-09 15:14:28,729 ERROR [org.jbp.wor.ins.imp.WorkflowProcessInstanceImpl] (executor-thread-0) Unexpected error (id d3747eef-e6f9-4237-a59c-0b3accf4a15e) while executing node LicenseValidation
DRL in process instance d6ec6632-8b89-406e-a9b6-5ae90b7067fd: org.jbpm.workflow.instance.WorkflowRuntimeException: [traffic_wih:d6ec6632-8b89-406e-a9b6-5ae90b7067fd - LicenseValidation
DRL:2] -- argument type mismatch
```

fjtirado: ok, sorry for my confusion, I mistake DateTimeFormatter with DateTimeFormat (;

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
tiagodolphine: I already tested it and it does not work..  but... 
`System.out.println(URI.create(URLEncoder.encode("http://localhost:8080/Traffic Violation", "UTF-8")));`
>`http%3A%2F%2Flocalhost%3A8080%2FTraffic+Violation`

and

`System.out.println(URLEncoder.encode(URI.create("http://localhost:8080/Traffic Violation").getPath(), "UTF-8"));`
 ```
java.lang.IllegalArgumentException: Illegal character in path at index 29: http://localhost:8080/Traffic Violation
	at java.base/java.net.URI.create(URI.java:906)
```
....
To be honest the current changes are fine IMO.

File: kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java (hunk 7 above)
Comment (tiagodolphine) [reply kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java:105]: the `parameters` is already a copy from the workItem, so I think that's fine to filter the "TaskName" that is not used.

**Comment by tiagodolphine** [comment]
@fjtirado updated the PR with some changes related to the comments.

**Comment by tiagodolphine** [comment]
Jenkins run apps tests

**Comment by tiagodolphine** [comment]
@cristianonicolai @fjtirado @danielezonca any other consideration? just pinging because we are close to the freeze and we need this merged. 
Thanks !!

**Comment by cristianonicolai** [comment]
all good from my side

File: kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java [hunk 12]
```diff
@@ -93,36 +98,78 @@ public void executeWorkItem(KogitoWorkItem workItem, KogitoWorkItemManager manag
...
         if (user != null && !user.trim().isEmpty() && password != null && !password.trim().isEmpty()) {
             request.basicAuthentication(user, password);
         }
         HttpResponse<Buffer> response = method == HttpMethod.POST || method == HttpMethod.PUT ? request.sendJsonAndAwait(bodyBuilder.apply(inputModel, parameters, resolver)) : request.sendAndAwait();
         manager.completeWorkItem(workItem.getStringId(), targetInfo != null ? Collections.singletonMap(RESULT,
                 resultHandler.apply(targetInfo, response)) : Collections.emptyMap());
+
+    }
+
+    public RestWorkItemHandlerBodyBuilder getBodyBuilder(Map<String, Object> parameters) {
```
danielezonca, line 138: Can you add a small comment for each `if`? It is not clear to me why/when it is possible that the `BODY_BUILDER` param is null or directly the Class instance or also a string to be used as FQCN to load it 🤔

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandlerUtils.java
```diff
@@ -62,14 +65,37 @@ private static void setValue(PropertyDescriptor[] propertyDescritors, Object tar
...
                 }
                 return;
             }
         }
-        logger.warn("No property found for name {}", name);
+        logger.debug("No property found for name {}", name);
+    }
+
+    private static Object convert(Object value, Class<?> targetClass) {
+        if (!targetClass.isAssignableFrom(value.getClass())) {
```
danielezonca, line 79: Is it possible that `value` is `null`? If so it will throw NPE

danielezonca:
@tiagodolphine 
I added some other minor comments, mainly clarification or minor improvements. Overall the PR looks good to me.

I also had a look to previous comments:
- [JsonResolver native](https://github.com/kiegroup/kogito-runtimes/pull/1467#discussion_r678024666): I asked about native build, have you checked if it is working? If not we probably need a note in the doc
- [JsonResolver Jackson annotations](https://github.com/kiegroup/kogito-runtimes/pull/1467#discussion_r678550965): in this comment it seems that you and Matteo agreed about removing some reflection Jackson checks but the code now it is still there so I'm not sure if this comment is solved or not

Can you please have a look?

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/transformation/JsonResolver.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Objects;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import com.fasterxml.jackson.annotation.JacksonAnnotation;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public class JsonResolver {
```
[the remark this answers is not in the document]
tiagodolphine: It works because the classes are all used in the process itself as the process variables, same used in the rest endpoints, so there is no need to mark them for reflection, there are examples in this PR where there are native image tests as well https://github.com/kiegroup/kogito-examples/pull/809. 
I think there would be issues in native in case we try to convert a json to a type where there wasn't any reference in the process, so probably they would be excluded during the aot compilation and we might add them for reflection.

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/transformation/JsonResolver.java
```diff
@@ -0,0 +1,95 @@
...
+    public JsonResolver(ObjectMapper objectMapper) {
+        this.objectMapper = objectMapper;
+        this.objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+                .configure(DeserializationFeature.ACCEPT_EMPTY_STRING_AS_NULL_OBJECT, true)
+                .setPropertyNamingStrategy(PropertyNamingStrategies.LOWER_CAMEL_CASE);
+
+    }
+
+    public JsonResolver() {
+        this(new ObjectMapper().setSerializationInclusion(JsonInclude.Include.NON_NULL));
```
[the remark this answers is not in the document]
tiagodolphine: The ideal case would be to inject a common objectMapper that is used by the application, and avoid the customization that were added, I'll try to improve it, and about the setSerializationInclusion you are right it should be moved to the other constructor

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
tiagodolphine: +1

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandlerUtils.java
tiagodolphine: +1 adding a null check

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandlerUtils.java
```diff
@@ -62,14 +65,37 @@ private static void setValue(PropertyDescriptor[] propertyDescritors, Object tar
...
+                return Integer.parseInt(value.toString());
+            } else if (Long.class.isAssignableFrom(targetClass) && value instanceof String) {
+                return Long.parseLong(value.toString());
+            } else if (Boolean.class.isAssignableFrom(targetClass) && value instanceof String) {
+                return Boolean.parseBoolean(value.toString());
+            } else if (Float.class.isAssignableFrom(targetClass) && value instanceof String) {
+                return Float.parseFloat(value.toString());
+            } else if (Double.class.isAssignableFrom(targetClass) && value instanceof String) {
+                return Double.parseDouble(value.toString());
+            }
```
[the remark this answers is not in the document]
tiagodolphine: I think the original idea is to return the original value in case there is no match, @fjtirado is this right?

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/bodybuilders/RestWorkItemHandlerBodyBuilder.java
```diff
@@ -17,7 +17,12 @@
 
 import java.util.Map;
 import java.util.function.UnaryOperator;
+import java.util.stream.Collectors;
 
 public interface RestWorkItemHandlerBodyBuilder {
     Object apply(Object contentData, Map<String, Object> parameters, UnaryOperator<Object> resolver);
+
+    default Map<String, Object> buildMap(Map<String, Object> parameters, UnaryOperator<Object> resolver) {
```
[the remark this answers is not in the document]
tiagodolphine: +1 it can be static

quarkus/extensions/kogito-quarkus-extension-common/kogito-quarkus-common-deployment/src/main/java/org/kie/kogito/quarkus/common/deployment/KogitoAssetsProcessor.java
```diff
@@ -134,6 +134,11 @@ public ReflectiveClassBuildItem reflectionJobsManagement() {
         return new ReflectiveClassBuildItem(true, true, "org.kie.kogito.jobs.api.Job");
     }
 
+    @BuildStep
+    public ReflectiveClassBuildItem reflectionRestWorkItemHandler() {
+        return new ReflectiveClassBuildItem(true, true, "org.kogito.workitem.rest.bodybuilders.ParamsRestWorkItemHandlerBodyBuilder");
```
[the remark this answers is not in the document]
tiagodolphine: yeah, it can be moved, I just followed the same pattern because all other ReflectiveClassBuildItem were added here, but we can move to the ProcessesAssetProcessor.

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/transformation/JsonResolver.java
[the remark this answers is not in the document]
tiagodolphine: I'll file a JIRA to improve this, because it could be merged into the recently created https://github.com/kiegroup/kogito-runtimes/blob/fdc27397ec5fae2ee88576d90975481f409848eb/jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/util/TypeTransformer.java where they have similar role let's say, we could unify the classes and inject a common `objectMapper` used by the whole application.
For now, I think it is better to keep the current approach due to the deadline for the freeze, basically, we need to merge this ASAP.

tiagodolphine:
> @tiagodolphine
> I added some other minor comments, mainly clarification or minor improvements. Overall the PR looks good to me.
> 
> I also had a look to previous comments:
> 
> * [JsonResolver native](https://github.com/kiegroup/kogito-runtimes/pull/1467#discussion_r678024666): I asked about native build, have you checked if it is working? If not we probably need a note in the doc
> * [JsonResolver Jackson annotations](https://github.com/kiegroup/kogito-runtimes/pull/1467#discussion_r678550965): in this comment it seems that you and Matteo agreed about removing some reflection Jackson checks but the code now it is still there so I'm not sure if this comment is solved or not
> 
> Can you please have a look?

- JsonResolver native: I gave some details in the comment.

- JsonResolver Jackson annotations: I removed the call that uses reflection but kept the method in the JsonResolver because it could be useful and there are some tests for it already done, but if you think it is better I can remove np, we are not using it at the moment.

danielezonca (approved):
Thanks for the answers, approving 👍

tiagodolphine:
jenkins retest this

tiagodolphine:
Just waiting for the green build to merge.

tiagodolphine:
Jenkins run apps tests

tiagodolphine:
Jenkins run examples tests

cristianonicolai:
Jenkins rerun examples tests

cristianonicolai:
Jenkins rerun apps tests

tiagodolphine:
Jenkins run examples tests

quarkus/extensions/kogito-quarkus-processes-extension/kogito-quarkus-processes-deployment/src/main/java/org/kie/kogito/quarkus/processes/deployment/ProcessesAssetsProcessor.java
```diff
@@ -115,6 +115,11 @@ IndexDependencyBuildItem addProtoDependencies() {
         return new IndexDependencyBuildItem("com.google.protobuf", "protobuf-java");
     }
 
+    @BuildStep
+    public ReflectiveClassBuildItem reflectionRestWorkItemHandler() {
+        return new ReflectiveClassBuildItem(true, true, "org.kogito.workitem.rest.bodybuilders.ParamsRestWorkItemHandlerBodyBuilder");
```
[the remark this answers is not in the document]
tiagodolphine: now it is possible to instantiate a class by the FQCN in the RestWorkItemHandler, in the native build if the class is not marked for reflection it won't work.

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandlerUtils.java
```diff
@@ -62,14 +66,37 @@ private static void setValue(PropertyDescriptor[] propertyDescritors, Object tar
...
                 return;
             }
         }
-        logger.warn("No property found for name {}", name);
+        logger.debug("No property found for name {}", name);
+    }
+
+    private static Object convert(Object value, Class<?> targetClass) {
+        if (Objects.nonNull(value) && !targetClass.isAssignableFrom(value.getClass())) {
+            if (Date.class.isAssignableFrom(targetClass)) {
```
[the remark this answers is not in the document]
tiagodolphine: makes sense, the targetClass should be Date or superclass of Date and not a subclass.

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/transformation/JsonResolver.java
```diff
@@ -0,0 +1,95 @@
...
+        Annotation[] declaringClassAnnotations =
+                Optional.of(element).filter(Field.class::isInstance).map(Field.class::cast).map(Field::getType).map(Class::getDeclaredAnnotations).orElse(new Annotation[0]);
+        if (!Stream.of(declaredAnnotations, declaringClassAnnotations).flatMap(Stream::of)
+                .noneMatch(a -> a.annotationType().getAnnotationsByType(JacksonAnnotation.class).length > 0)) {
+            return true;
+        }
+        if (!(element instanceof Class) || ((Class<?>) element).isPrimitive()) {
+            return false;
+        }
+        Class<?> clazz = (Class<?>) element;
```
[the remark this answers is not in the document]
tiagodolphine: the idea was to return true if the root class, the fields of it, or the fields classes contain jackson annotation (in arecursive way). But to be honest, we decided to not use reflection, this method was kept just for convenience in case we need it in the future, but we can remove it since it is not being used.

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
tiagodolphine: I'd say this is reserved work, in the same way as others.. if we want to allow the usage of a reserved work as a custom parameter we need to think of a solution for it, like using some prefix or whatever... anyway for now IMO this is not a blocker it is more an edge case. @fjtirado wdyt?
```
    public static final String REST_TASK_TYPE = "Rest";
    public static final String URL = "Url";
    public static final String METHOD = "Method";
    public static final String CONTENT_DATA = "ContentData";
    public static final String RESULT = "Result";
    public static final String USER = "Username";
    public static final String PASSWORD = "Password";
    public static final String HOST = "Host";
    public static final String PORT = "Port";
    public static final String RESULT_HANDLER = "ResultHandler";
    public static final String BODY_BUILDER = "BodyBuilder";
```

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandlerUtils.java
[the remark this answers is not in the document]
MarianMacik: @tiagodolphine I see you updated it, but shouldn't it be the same for all other branches?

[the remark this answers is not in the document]
tiagodolphine: IMO it should the be same, @fjtirado do you see any issue changing this? just asking because this was aligned with you previously.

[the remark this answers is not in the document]
tiagodolphine: @MarianMacik updated all, I think that's fine

cristianonicolai:
Jenkins rerun examples tests

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/transformation/JsonResolver.java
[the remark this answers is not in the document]
MarianMacik: OK, up to you. Just saying that it may not be completely recursive at least based on my understanding. But if it is there to catch the idea, I am fine with it.

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
MarianMacik: OK, no problem with that, makes sense.

MarianMacik (approved):
Looks good to me, thanks for the fixes!

tiagodolphine:
Jenkins run examples tests

tiagodolphine:
Jenkins run examples tests

tiagodolphine:
Jenkins run examples tests

cristianonicolai:
Jenkins rerun examples tests

cristianonicolai:
Jenkins rerun examples tests

tiagodolphine:
finally green

PR #1479, opened 2021-07-27
[KOGITO-5578] Changing listener approach

fjtirado:
QuarkusMultiCloudEventReceiver is not longer needed once we are
generating the listeners, by just adding annotation


danielezonca (approved):

ricardozanini (approved):

kogito-codegen-modules/kogito-codegen-core/src/main/resources/class-templates/config/GlobalObjectMapperQuarkusTemplate.java
```diff
@@ -25,6 +25,7 @@
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
 
+import io.cloudevents.jackson.JsonFormat;
```
ricardozanini, line 28: This dependency only comes with the `kogito-addons-cloudevents`. Won't be a problem? How about injecting the `GlobalObjectMapper` into the Quarkus CloudEvents addon and register the module there?

ricardozanini (commented):
I think we might have a problem with CE module. It's not on the classpath without the addon. :thinking: Or it should not..

fjtirado (commented):
Opened https://issues.redhat.com/browse/KOGITO-5590 to deal with the point raised by @ricardozanini

kogito-codegen-modules/kogito-codegen-core/src/main/resources/class-templates/config/GlobalObjectMapperQuarkusTemplate.java
fjtirado: Opened https://issues.redhat.com/browse/KOGITO-5590 to deal with that

fjtirado:
jenkins retest app examples

fjtirado:
jenkins run app examples

fjtirado:
jenkins rerun examples tests

fjtirado:
jenkins rerun examples tests

PR #1480, opened 2021-07-27
[KOGITO-5343] BPMN project build problem on Windows

fjtirado:
Avoid using "/", which is not OS independent.

https://issues.redhat.com/browse/KOGITO-5343

### DIFF
--- a/kogito-codegen-modules/kogito-codegen-api/src/main/java/org/kie/kogito/codegen/api/GeneratedFile.java
+++ b/kogito-codegen-modules/kogito-codegen-api/src/main/java/org/kie/kogito/codegen/api/GeneratedFile.java
@@ -21,16 +21,17 @@ import java.util.Objects;
 
 public class GeneratedFile {
 
-    private final String relativePath;
+    private final Path path;
+    private final String pathAsString;
     private final byte[] contents;
     private final GeneratedFileType type;
 
-    public GeneratedFile(GeneratedFileType type, Path relativePath, String contents) {
-        this(type, relativePath, contents.getBytes(StandardCharsets.UTF_8));
+    public GeneratedFile(GeneratedFileType type, Path path, String contents) {
+        this(type, path, contents.getBytes(StandardCharsets.UTF_8));
     }
 
-    public GeneratedFile(GeneratedFileType type, Path relativePath, byte[] contents) {
-        this(type, relativePath.toString(), contents);
+    public GeneratedFile(GeneratedFileType type, Path path, byte[] contents) {
+        this(type, path, path.toString(), contents);
     }
 
     public GeneratedFile(GeneratedFileType type, String relativePath, String contents) {
@@ -38,13 +39,22 @@ public class GeneratedFile {
     }
 
     public GeneratedFile(GeneratedFileType type, String relativePath, byte[] contents) {
+        this(type, Path.of(relativePath), relativePath, contents);
+    }
+
+    private GeneratedFile(GeneratedFileType type, Path path, String pathAsString, byte[] contents) {
         this.type = type;
-        this.relativePath = relativePath;
+        this.path = path;
+        this.pathAsString = pathAsString;
         this.contents = contents;
     }
 
     public String relativePath() {
-        return relativePath;
+        return pathAsString;
+    }
+
+    public Path path() {
+        return path;
     }
 
     public byte[] contents() {
@@ -63,7 +73,7 @@ public class GeneratedFile {
     public String toString() {
         return "GeneratedFile{" +
                 "type=" + type +
-                ", relativePath='" + relativePath + '\'' +
+                ", path='" + path + '\'' +
                 '}';
     }
 
@@ -76,11 +86,11 @@ public class GeneratedFile {
             return false;
         }
         GeneratedFile that = (GeneratedFile) o;
-        return Objects.equals(relativePath, that.relativePath);
+        return Objects.equals(path, that.path);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(relativePath);
+        return Objects.hash(path);
     }
 }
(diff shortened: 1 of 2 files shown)


kogito-codegen-modules/kogito-codegen-api/src/main/java/org/kie/kogito/codegen/api/GeneratedFile.java
```diff
@@ -21,30 +21,48 @@
...
 
+    /**
+     * @deprecated use version of the constructor that accepts <code>Path</code> rather than <code>String</code>
+     */
+    @Deprecated
     public GeneratedFile(GeneratedFileType type, String relativePath, String contents) {
         this(type, relativePath, contents.getBytes(StandardCharsets.UTF_8));
     }
 
+    /**
```
fjtirado, line 45: I think we should deprecate the String constructor to prevent `generator` classes to build paths that might content OS dependant characters. 
Problem of doing so is that it will generate ton of warnings since all generator has a method (by convention) of the form `String generatedFilePath()`

cristianonicolai (approved):

cristianonicolai:
Jenkins rerun examples tests

cristianonicolai:
Jenkins rerun runtimes tests

fjtirado:
Jenkins rerun examples tests

kogito-codegen-modules/kogito-codegen-api/src/main/java/org/kie/kogito/codegen/api/GeneratedFile.java
ricardozanini: Maybe we could recreate the string path using the Path API then?

ricardozanini (approved):
Many thanks for handling this issue!

fjtirado:
Jenkins rerun examples tests

kogito-codegen-modules/kogito-codegen-core/src/main/java/org/kie/kogito/codegen/core/utils/GeneratedFileWriter.java
```diff
@@ -136,14 +136,13 @@ private void writeGeneratedFile(GeneratedFile f, Path location) throws IOExcepti
         if (location == null) {
             return;
         }
-        // verify if this is still needed https://issues.redhat.com/browse/KOGITO-3085
-        String generatedClassFile = f.relativePath().replace("src/main/java", "");
+        Path srcPath = Path.of("src", "main", "java");
```
danielezonca, line 139: To be honest this ticket is very old, I would suggest to remove it and see if it is green the build 🤔

kogito-codegen-modules/kogito-codegen-api/src/main/java/org/kie/kogito/codegen/api/GeneratedFile.java
danielezonca: I don't like the idea to deprecate the method in this way: user will see many warning while we should just apply the change if we thing (and I agree) that `Path` is a better abstraction. This part of the codebase is not huge so feel free to apply this change now or postpone the change to another ticket, in any case I would avoid the deprecation as "intermediate" step

fjtirado: yes, good point, Ill remove the deprecation

kogito-codegen-modules/kogito-codegen-core/src/main/java/org/kie/kogito/codegen/core/utils/GeneratedFileWriter.java
fjtirado: I prefer to be conservative and keep that just in case. I changed to static so there is just one src patch object to check againts and this safeguard does not really interfere with codegen in case is not needed

ricardozanini:
Jenkins rerun examples tests

cristianonicolai:
Jenkins rerun examples tests

kogito-codegen-modules/kogito-codegen-core/src/main/java/org/kie/kogito/codegen/core/utils/GeneratedFileWriter.java
danielezonca: @evacchi @cristianonicolai 
Do you remember the reason of this replace? Is there anything in particular to verify to see if it is still necessary?

fjtirado:
Jenkins rerun examples tests

cristianonicolai:
Jenkins rerun examples tests

kogito-codegen-modules/kogito-codegen-core/src/main/java/org/kie/kogito/codegen/core/utils/GeneratedFileWriter.java
cristianonicolai: sorry, I dont recall that

fjtirado: I removed the line and the build was green, so I guess it was not used

fjtirado:
Jenkins rerun examples tests

fjtirado:
Jenkins rerun apps tests

fjtirado:
Jenkins rerun examples tests

kogito-codegen-modules/kogito-codegen-core/src/main/java/org/kie/kogito/codegen/core/utils/GeneratedFileWriter.java
evacchi: originally the files that were built by the drools compiler needed to be in that path IIRC

ricardozanini:
Jenkins rerun examples tests

kogito-codegen-modules/kogito-codegen-core/src/main/java/org/kie/kogito/codegen/core/utils/GeneratedFileWriter.java
fjtirado: @evacchi should we have this into account and remove that path segment to be safe?

evacchi: no, I was just answering @danielezonca 's question. If it's all green then I guess it's no longer needed (probably we have changed something in the meantime that makes that irrelevant)

PR #1481, opened 2021-07-27
[KOGITO-5592]  REST WorkItemHandler for SpringBoot

fjtirado:
Quick fix for generating the Vertx instance is not present in the
injection context

### DIFF
--- a/jbpm/jbpm-flow-builder/src/main/resources/class-templates/RestWorkItemHandlerTemplate.java
+++ b/jbpm/jbpm-flow-builder/src/main/resources/class-templates/RestWorkItemHandlerTemplate.java
@@ -18,6 +18,7 @@ package org.jbpm.process.codegen;
 import io.vertx.mutiny.ext.web.client.WebClient;
 import io.vertx.mutiny.core.Vertx;
 import org.kogito.workitem.rest.RestWorkItemHandler;
+import static org.kogito.workitem.rest.RestWorkItemHandlerUtils.vertx;
 
 public class xxxRestWorkItemHandler extends RestWorkItemHandler {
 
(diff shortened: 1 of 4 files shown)


fjtirado:
jenkins run examples tests

jbpm/jbpm-flow-builder/src/main/resources/class-templates/RestWorkItemHandlerTemplate.java
```diff
@@ -26,7 +26,8 @@ public xxxRestWorkItemHandler() {
     }
 
     public xxxRestWorkItemHandler(Vertx vertx) {
-        super(WebClient.create(vertx));
+        
+        super(WebClient.create(vertx == null ? Vertx.vertx() : vertx));
```
tiagodolphine, line 30: my only concern is, it will create a vertx instance for every rest wih, and probably they could share a single vertx instance... in case a process or different processes in the same application have multiple rest with.
So having a producer with @Bean and Injecting the instance in all would be better IMO, wdyt? makes sense?

fjtirado: injection will require an additional dependency for sprinboot in all examples (the dependency that will create the bean)

fjtirado: so, to avoid the additional dependency, Ill be using singleton approach, good catch!!!

fjtirado:
@tiagodolphine ready to merge?

fjtirado:
jenking rerun test examples

fjtirado:
jenkins run test examples

fjtirado:
Jenkins rerun examples tests

fjtirado:

fjtirado:
Jenkins rerun examples tests

fjtirado:
@tiagodolphine ready to go?

tiagodolphine (approved):

PR #1483, opened 2021-07-28
[KOGITO-5500] fix in memory compilation under Windows

mariofusco:


quarkus/extensions/kogito-quarkus-extension-common/kogito-quarkus-common-deployment/src/main/java/org/kie/kogito/quarkus/common/deployment/KogitoQuarkusResourceUtils.java
```diff
@@ -220,7 +220,7 @@ public static String toClassName(String sourceName) {
         } else if (sourceName.endsWith(".class")) {
             sourceName = sourceName.substring(0, sourceName.length() - 6);
         }
-        return sourceName.replace('/', '.');
+        return sourceName.replace('/', '.').replace('\\', '.');
```
gitgabrio, line 223: @mariofusco 
I'm sure this fix the immediate bug. At the same time, I think we should strive to improve the codebase whenever we have to make a bug fix, and this modification IMO simply hides a bigger issue
1) we should not manage files as string, because this is extremely error-prone (as all those bug-fixes are demonstrating)
2) This method is invoked from [InMemoryCompiler.compile](https://github.com/kiegroup/kogito-runtimes/blob/f426a091e7d627a55e94bb965333d2871fecbe20/quarkus/extensions/kogito-quarkus-extension-common/kogito-quarkus-common-deployment/src/main/java/org/kie/kogito/quarkus/common/deployment/InMemoryCompiler.java#L82)
    `    String fileName = toRuntimeSource(toClassName(generatedClassFile));`
   soon before that invocation, there is 
  `       String generatedClassFile = relativePath.replace("src/main/java/", "");` that completely ignores the OS
 and soon after the [toRuntimeSource](https://github.com/kiegroup/kogito-runtimes/blob/f426a091e7d627a55e94bb965333d2871fecbe20/quarkus/extensions/kogito-quarkus-extension-common/kogito-quarkus-common-deployment/src/main/java/org/kie/kogito/quarkus/common/deployment/InMemoryCompiler.java#L117) is invoked
 `       return "src/main/java/" + className.replace('.', '/') + ".java"; ` that  a) put back the "src/main/java" that was rmoved two lines before b) again, completely ignores OS

3) The issue I see in this particular situation is that InMemoryCompiler.compile iterates over `Collection<GeneratedFile>` to retrieve information about path, but `GeneratedFile` stores and returns it as String. GeneratedFile should keep path information as `Path`, and not translated to String - this would make it much less error prone - and then it would be possible to bring around a `Path` to the file, and not a string representing it

gitgabrio (commented):
@mariofusco 
I would prefer a more substantial modification, but it is just my personal opinion.

gitgabrio (approved):
Thanks for PR.

LGTM

pibizza (approved):

PR #1484, opened 2021-07-28
[KOGITO-3922] Preparing the engine for swf extension

fjtirado:
This is the first step to be able to add a context for SWF that handles
the exception


fjtirado:
Wil use a completely different approach to implement that, base on already existing BoundaryEvent Error functionality

PR #1495, opened 2021-08-03
[KOGITO-5320] protobuf data type lookup fix

uteegozi:
<details>
<summary>

**problem description**: please see Jira https://issues.redhat.com/browse/KOGITO-5320

**problem solution**: resolved by a) making the lookup name a full-qualified name and b) ensuring that the kogito-types.proto is present as dependency when the application-types.proto is processed

</summary>

</details>

### DIFF
--- a/jbpm/process-serialization-protobuf/src/main/java/org/kie/kogito/serialization/process/protobuf/ProtostreamProtobufAdapterTypeProvider.java
+++ b/jbpm/process-serialization-protobuf/src/main/java/org/kie/kogito/serialization/process/protobuf/ProtostreamProtobufAdapterTypeProvider.java
@@ -25,7 +25,6 @@ import java.util.Collection;
 import java.util.Comparator;
 import java.util.EnumSet;
 import java.util.List;
-import java.util.Map;
 import java.util.stream.Collectors;
 
 import org.infinispan.protostream.FileDescriptorSource;
@@ -67,15 +66,18 @@ public class ProtostreamProtobufAdapterTypeProvider implements ProtobufTypeProvi
         return fd != null && "kogito".equals(fd.getPackage());
     }
 
+    protected List<FileDescriptor> sortFds(Collection<FileDescriptor> descriptors){
+        Comparator<FileDescriptor> fdComparator = (fd1, fd2) -> isKogitoPackage(fd1) ? (isKogitoPackage(fd2) ? 0 : -1) : (isKogitoPackage(fd2) ? 1 : 0);
+        return descriptors.stream().sorted(fdComparator).collect(Collectors.toList());
+    }
+
     // we transform protostream to protobuf descriptors
     private List<com.google.protobuf.Descriptors.FileDescriptor> build() throws IOException, DescriptorValidationException {
         SerializationContextImpl context = buildSerializationContext();
 
         List<com.google.protobuf.Descriptors.FileDescriptor> protos = new ArrayList<>();
-        Map<String, FileDescriptor> descriptors = context.getFileDescriptors();
         // make sure kogito-types is processed first or else will be missing as dependency for the application types
-        Comparator<FileDescriptor> fdComparator = (fd1, fd2) -> isKogitoPackage(fd1) ? -1 : (isKogitoPackage(fd2) ? 1 : 0);
-        List<FileDescriptor> descriptorsSorted = descriptors.values().stream().sorted(fdComparator).collect(Collectors.toList());
+        List<FileDescriptor> descriptorsSorted = sortFds(context.getFileDescriptors().values());
         for (FileDescriptor entry : descriptorsSorted) {
             com.google.protobuf.Descriptors.FileDescriptor[] dependencies = protos.toArray(new com.google.protobuf.Descriptors.FileDescriptor[protos.size()]);
             protos.add(Descriptors.FileDescriptor.buildFrom(buildMessageTypes(entry), dependencies));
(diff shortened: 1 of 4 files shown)


cristianonicolai:
ok to test

cristianonicolai:
@uteegozi would be possible to add some test?

uteegozi:
> @uteegozi would be possible to add some test?

will do

jbpm/process-serialization-protobuf/src/main/java/org/kie/kogito/serialization/process/protobuf/ProtostreamProtobufAdapterTypeProvider.java
```diff
@@ -19,12 +19,7 @@
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.EnumSet;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
cristianonicolai, line 22: @uteegozi we should probably avoid using import *

uteegozi: @cristianonicolai done

jbpm/process-serialization-protobuf/src/main/java/org/kie/kogito/serialization/process/protobuf/ProtostreamProtobufAdapterTypeProvider.java
```diff
@@ -62,16 +63,23 @@
...
 
-        // we transform protostream to protobuf descriptors
         List<com.google.protobuf.Descriptors.FileDescriptor> protos = new ArrayList<>();
         Map<String, FileDescriptor> descriptors = context.getFileDescriptors();
-        for (Map.Entry<String, FileDescriptor> entry : descriptors.entrySet()) {
-            com.google.protobuf.Descriptors.FileDescriptor[] deps = protos.toArray(new com.google.protobuf.Descriptors.FileDescriptor[protos.size()]);
-            protos.add(Descriptors.FileDescriptor.buildFrom(buildMessageTypes(entry.getValue()),deps));
-            protos.add(Descriptors.FileDescriptor.buildFrom(buildEnumTypes(entry.getValue()),deps));
+        // make sure kogito-types is processed first or else will be missing as dependency for the application types
+        Comparator<FileDescriptor> fdComparator = (fd1, fd2) -> isKogitoPackage(fd1) ? -1 : (isKogitoPackage(fd2) ? 1 : 0);
```
MarianMacik, line 77: This comparator is not symmetrical, e.g. compare(x,y) = - compare(y,x) doesn't hold. Based on the [JavaDoc](https://docs.oracle.com/javase/8/docs/api/java/util/Comparator.html#compare-T-T-):
`The implementor must ensure that sgn(compare(x, y)) == -sgn(compare(y, x)) for all x and y.`
For fd1 and fd2 both being Kogito packages it returns -1 every time. Though not sure it may pose any issues in some corner cases or if sorting implementation of Java changes in future. But maybe it would be better to stay on the safe side.

uteegozi: @MarianMacik  Updated sort plus added tests
Note: Today the code assumes that we only have the two .proto files and that at the most the application-types can be depending on the kogito-types. If the .proto files list to be converted would change/grow this sort would need to be revised as new .proto files might have different dependencies which would require a different order of processing hence a different sort.

cristianonicolai:
jenkins please retest this

cristianonicolai:
@elguardian can you take a quick look, thanks

jbpm/process-serialization-protobuf/src/main/java/org/kie/kogito/serialization/process/protobuf/ProtostreamProtobufAdapterTypeProvider.java
```diff
@@ -152,7 +162,12 @@ private FieldDescriptorProto buildFieldDescriptor(FieldDescriptor descriptor) {
         fieldBuilder.setType(buildFieldTypeDescriptor(descriptor.getType()));
         EnumSet<FieldDescriptorProto.Type> set = EnumSet.of(FieldDescriptorProto.Type.TYPE_ENUM, FieldDescriptorProto.Type.TYPE_MESSAGE);
         if(set.contains(fieldBuilder.getType())) {
-            fieldBuilder.setTypeName(descriptor.getTypeName());
+            if(isKogitoPackage(descriptor.getMessageType().getFileDescriptor())) {
+                fieldBuilder.setTypeName("." + descriptor.getTypeName());
+            }
+            else {
+                fieldBuilder.setTypeName(descriptor.getTypeName());
+            }
```
elguardian, line 170: This is a good catch but probably no the optimum approach. when the "." notation is used is for crosslinking signaling that it should look into another file so you need to detect whether the current type is in the file or not. what about something like (getting all the types in the file and comparing to the current type being modified (probably needs some twaking for enums ?) but it could be less "hardcoded"

'''
            if(descriptor.getFileDescriptor().getTypes().containsKey(descriptor.getMessageType().getFullName())) {
                fieldBuilder.setTypeName(descriptor.getTypeName());
            } else {
                fieldBuilder.setTypeName("." + descriptor.getTypeName());
            }
'''

uteegozi: you are right - adapting also for cases with messages using enum fields - from an imported and the same ,proto file

uteegozi: @elguardian implemented as suggested - plus additional support for enums

elguardian (approved):
LGTM

PR #1496, opened 2021-08-03
[KOGITO-5634] Merging cloud event quarkus addons

fjtirado:
With this PR, minimal changes to support existing examples with  just one quarkus cloud event addon are performed. 
The most relevant class is the ChannelMappingStrategy which contains the logic to map Consumers/Producers (triggers) with Smallrye channels defined by user in configuration. 
Although not required, it would be nice to merge with https://github.com/kiegroup/kogito-examples/pull/834 (remove multi addon deprecated dependency  from examples)


cristianonicolai (approved):

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
```diff
@@ -382,13 +388,20 @@ public static String defaultProcessListenerConfigClass(String packageName) {
 
             if (metaData.getTriggers() != null) {
 
+                if (context().getAddonsConfig().useCloudEvents()) {
+                    // TODO change that to make it configurable
```
ricardozanini, line 392: Jira?

ricardozanini (approved):
Looks good!

kogito-codegen-modules/kogito-codegen-api/src/main/java/org/kie/kogito/codegen/api/utils/AddonsConfigDiscovery.java
```diff
@@ -55,7 +55,6 @@ public static AddonsConfig discover(Predicate<String> classAvailabilityResolver)
         boolean useCloudEvents = classAvailabilityResolver.test(QUARKUS_RESOURCE_CLASS);
         boolean useExplainability = classAvailabilityResolver.test(QUARKUS_EXPLAINABILITY) || classAvailabilityResolver.test(SPRING_EXPLAINABILITY);
         boolean useProcessSVG = classAvailabilityResolver.test(QUARKUS_PROCESS_SVG) || classAvailabilityResolver.test(SPRING_PROCESS_SVG);
-        boolean useMultiChannel = classAvailabilityResolver.test(QUARKUS_MULTI_CLOUD_EVENTS);
```
danielezonca, line 58: Please remove also the `QUARKUS_MULTI_CLOUD_EVENTS` field

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/MessageConsumerQuarkusTemplate.java
```diff
@@ -31,7 +31,6 @@
 
 
 @io.quarkus.runtime.Startup
-@RegisterForReflection
```
danielezonca, line 34: Are you sure about this change? Have you checked in native mode?

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
```diff
@@ -275,6 +281,24 @@ public static String defaultProcessListenerConfigClass(String packageName) {
...
+            EventGenerator eventGenerator = eventGenerators.get(channel);
+            if (eventGenerator != null) {
+                beanName = eventGenerator.getBeanName();
+            }
+        }
+        return Optional.ofNullable(beanName);
+    }
+
+    private static ChannelMappingStrategy getChannelMappingStrategy() {
+        //TODO service loader mechanism is not the ideal approach to allow advance user to override default mapping strategy, need to reevaluate it
```
danielezonca, line 297: Moreover I don't think user should depend on codegen modules, if this class is intended to be pluggable we should probably move it to kogito-api module

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
```diff
@@ -297,6 +320,8 @@ public static String defaultProcessListenerConfigClass(String packageName) {
         Map<String, List<UserTaskModelMetaData>> processIdToUserTaskModel = new HashMap<>();
         Map<String, ProcessMetaData> processIdToMetadata = new HashMap<>();
 
+        Map<String, EventGenerator> eventGenerators = new LinkedHashMap<>();
```
danielezonca, line 323: Why Linked? Is ordering important?

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
```diff
@@ -531,6 +546,10 @@ public static String defaultProcessListenerConfigClass(String packageName) {
         return generatedFiles;
     }
 
+    private EventGenerator buildEventGenerator(Entry<TriggerMetaData, String> entry) {
+        return new EventGenerator(context(), entry.getValue(), entry.getKey().getType() == TriggerType.ConsumeMessage ? "EventReceiver" : "EventEmitter");
```
danielezonca, line 550: Are EventEmitter and EventReceiver constant? Are them name of classes?

danielezonca (commented):
Looks good to me, just some comments

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
fjtirado: yes, the default fallback mechanism in case no mapping is defined (which is important to be backward compatible with existing examples) is to use the first registered channel

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
fjtirado: They are the key for loading the proper template and the suffix being used to generate both class and bean name.

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
fjtirado: outdated

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/MessageConsumerQuarkusTemplate.java
fjtirado: yes, this was there because multi channel was not using injection but manual registration, and manual registration requires to pass a Method to mediator configuration. When using generation this is not longer needed

kogito-codegen-modules/kogito-codegen-api/src/main/java/org/kie/kogito/codegen/api/utils/AddonsConfigDiscovery.java
fjtirado: done, thanks

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
fjtirado: ok, I did not do that because this class is only used during code generation for processes (so codegen processes looks the natural place to me). 
Should it not be better to move it to codegen-api? 
In any case, tonight I though a better approach, rather than allow user customizing this, we should just use this to implement the most likely strategies and let the user decide which one he/she wants to use by settinga system property with the strategy name 
The only one we are implementing right now is just mapping trigger name to channel name (if channel exist) and, if there is no smallrye channel with that name, map trigger name to default channel name (as were defined by former single add on). If users do not set up a smallrye channel (believe or not, we have current working example that do that and still works), I implemented a fallback mechanism to take just the first channel defined (thats why order is important in the map), and if no one is defined, create a default channel (which wont be used and cause smallrye to print a warning, but at least will allow the example to run) 
We can implement other mapping strategies (now or in the future), for example, one that maps data type name (no message name) to channel name. And so on. 
So if you agree with that view, what I can do is define a ChannalMappingStrategy that returns different implemented strategies based on property value, wdyt?

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
danielezonca: Can you add a comment to prevent breaking changes? Is it possible to add a unit test for that?

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
danielezonca: Can you move to constant (or at least local variables) to make this more clear? 
Something like `String eventReceiverGeneratedClassName = "EventReceiver"`

fjtirado: hmmm, I do not think strings that are used just one time needs to be put as constant. It does not improve readability (you have to jump to the start of the class) and does not clarify usage.

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
danielezonca: I'm not sure about making `codegen-api` "public", for example that modules depends on JavaParser and this means user will have JavaParser dependency at runtime that is not necessary and is also "dangerous" because, if used, will probably break native compilation.
That's why probably it should be better to consider `kogito-api` as alternative, or even a new module if necessary. This can help also to force ourselves to avoid to expose internal details

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
fjtirado: Changed to use inheritance, which avoid the constant in processcodegen

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
fjtirado: ok,  agree about codegen-api
what about not making this class public and just provide users with different implementations that they can select through configuration, as I was proposing in my latest comment

jbpm/jbpm-flow-builder/src/main/resources/org/jbpm/process/TaskInputTemplate.java
```diff
@@ -13,11 +13,14 @@
...
-public interface ChannelResolver {
 
-    Set<String> getOutputChannels();
-}
+public class XXXTaskInput {
+
+    public static XXXTaskInput fromMap (Map<String,Object> params) {
+        
+    }
+}
```
MarianMacik, line 26: Any reason for this file? I cannot find it used in this PR.

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
danielezonca: Yes if we expect the strategies to be fixed this is another good option

danielezonca (approved):

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/DefaultChannelMappingStrategy.java
```diff
@@ -0,0 +1,104 @@
...
+        if (!missingTriggers.isEmpty()) {
+            String defaultIncoming = KogitoEventStreams.INCOMING;
+            String defaultOutgoing = KogitoEventStreams.OUTGOING;
+            for (Entry<TriggerMetaData, String> entry : map.entrySet()) {
+                if (entry.getKey().getType() == TriggerType.ConsumeMessage) {
+                    defaultIncoming = entry.getValue();
+                } else {
+                    defaultOutgoing = entry.getValue();
+                }
+            }
```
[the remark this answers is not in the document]
fjtirado: I slightly change handling of missing channels in latest versions. 
As you proposed, if value is null, I'm forcing default

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
fjtirado: Ok, I wrote a comment and just create the default implementation. 
If in future we decide to add more strategies, we should modify this methods and introduce system property to select between different options

quarkus/addons/cloudevents/implementation/src/main/java/org/kie/kogito/addon/cloudevents/quarkus/QuarkusCloudEventEmitter.java
```diff
@@ -15,39 +15,51 @@
...
+    }
 
     @Override
-    public <T> CompletionStage<Void> emit(T e, String type, Optional<Function<T, Object>> processDecorator) {
-        logger.debug("publishing event {} for type {}", e, type);
-        final Message<String> message = processMessage(e, processDecorator);
-        emitter.send(message);
-        return message.getAck().get();
+    protected void emit(Message<String> message) {
+        Emitter<String> emitter = (Emitter<String>) channelRegistry.getEmitter(KogitoEventStreams.OUTGOING);
```
tiagodolphine, line 58: can't you get/set the emitter in the `initialize` instead of doing this in every call to the `emit`

tiagodolphine (approved):

quarkus/addons/cloudevents/implementation/src/main/java/org/kie/kogito/addon/cloudevents/quarkus/QuarkusCloudEventEmitter.java
fjtirado: at that moment the emitter does not exist yes, but I can certainly use lazy initialization here, thanks for pointing it out, good catch!!!

MarianMacik (approved):
Looks good to me, thanks!

fjtirado:
jenkins run runtimes tests

fjtirado:
jenkins run Apps tests

fjtirado:
jenkins run Examples tests

cristianonicolai:
Jenkins rerun examples tests

fjtirado:
Jenkins rerun examples tests

PR #1533, opened 2021-08-17
KOGITO-5712 Include process related examples to productized profile

cristianonicolai:


ricardozanini (approved):
Many thanks man!

addons/common/pom.xml
```diff
@@ -19,7 +19,6 @@
     <module>events</module>
```
tiagodolphine, line 19: should events be moved to default? I think we just need the messaging one, right?

api/pom.xml
```diff
@@ -16,35 +16,25 @@
...
-        <updatePolicy>daily</updatePolicy>
-      </snapshots>
-    </repository>
-  </repositories>
-
   <modules>
     <module>kogito-events-api</module>
     <module>kogito-api</module>
     <module>kogito-services</module>
-    <module>kogito-timer</module>
```
tiagodolphine, line 46: `kogito-timer` module is not related to the job-service, it is used by internal timers in the engine level as well.. but don't know if this makes difference for now... because we don't support timer events for STP, right?

springboot/addons/events/pom.xml
```diff
@@ -16,6 +16,7 @@
 
   <modules>
     <module>decisions</module>
+    <module>kafka</module>
```
tiagodolphine, line 19: the same, I think we don't need any events addon, because this would be related to publishing runtime events to data-index, right?

mareknovotny:
thanks @cristianonicolai !

mareknovotny:
i will let approving individual modules for others, but i confirm the way of adding into default/productized profile is what I asked for

api/pom.xml
cristianonicolai: this module needs to be in anyway as _kogito-api_ depends on it

springboot/addons/events/pom.xml
cristianonicolai: yeah, need to review these ones

cristianonicolai: done

api/pom.xml
tiagodolphine: +1

cristianonicolai:
Jenkins rerun apps tests

cristianonicolai:
Jenkins run native

addons/deprecated/pom.xml
```diff
@@ -12,26 +12,42 @@
...
     <module>monitoring-prometheus-common</module>
     <module>monitoring-prometheus-quarkus-addon</module>
     <module>monitoring-prometheus-springboot-addon</module>
   </modules>
+
+  <profiles>
+    <profile>
+      <id>default</id>
+      <activation>
+        <property>
```
fjtirado, line 34: so, if productized is true, these addons are not deprecated?

fjtirado (approved):

addons/deprecated/pom.xml
cristianonicolai: If productized is true these are excluded from the build. The modules list above should be added to prod profile only to keep compatibility with the previous release where they were included.

cristianonicolai:
Jenkins rerun native apps tests

quarkus/extensions/kogito-quarkus-extension/pom.xml
```diff
@@ -1,24 +1,37 @@
...
+    <profile>
+      <id>default</id>
+      <activation>
+        <property>
+          <name>!productized</name>
+        </property>
+      </activation>
+      <modules>
         <module>kogito-quarkus-integration-test</module>
         <module>kogito-quarkus-integration-test-maven-devmode</module>
```
MarianMacik, line 32: Up until now we have been executing tests for product with `-Dproductized`, so this will effectively eliminate these from execution. I think we should leave all product-related modules including tests in productized profiles.

MarianMacik (commented):
I think we should leave product-related tests in productized profiles as well.

pom.xml
```diff
@@ -142,8 +141,10 @@
         </property>
       </activation>
       <modules>
+        <module>kogito-test-utils</module>
         <module>kogito-cloud-services</module>
         <module>integration-tests</module>
+        <module>kogito-serverless-workflow</module>
```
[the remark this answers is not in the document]
cristianonicolai: thanks, fixed

quarkus/extensions/kogito-quarkus-extension/pom.xml
cristianonicolai: @MarianMacik reverted here, tests continue to be in prod profile

addons/common/pom.xml
danielezonca: `events` is necessary because it contains `decisions` that is already included in the product since 7.11. I suggest to exclude only `mongodb`

tiagodolphine: yeah, great point

pom.xml
```diff
@@ -142,6 +141,7 @@
         </property>
       </activation>
       <modules>
+        <module>kogito-test-utils</module>
```
danielezonca, line 144: If we expect product build to run all integration tests I think this module is required 🤔

kogito-codegen-modules/pom.xml
```diff
@@ -1,44 +1,43 @@
...
+  <profiles>
+    <profile>
+      <id>default</id>
+      <activation>
+        <property>
+          <name>!productized</name>
+        </property>
+      </activation>
+      <modules>
+        <module>kogito-codegen-integration-tests</module>
```
danielezonca, line 37: Same for this module, it contains many process integration tests so if we expect to run all the tests with product build this is probably necessary

quarkus/extensions/pom.xml
```diff
@@ -33,11 +33,24 @@
...
+  <profiles>
+    <profile>
+      <id>default</id>
+      <activation>
+        <property>
+          <name>!productized</name>
+        </property>
+      </activation>
+      <modules>
+        <module>kogito-quarkus-test-list</module>
```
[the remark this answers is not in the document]
aloubyansky: Right, this artifact provides tests that should be executed in the platform CI, see https://github.com/quarkusio/quarkus-platform/blob/main/pom.xml#L244
Right now we assume the version of this artifact matches the version of the Kogito BOM imported in the platform. It doesn't really have to be the same version but it was simpler that way.

addons/common/pom.xml
cristianonicolai: yep, its included because of that.

kogito-codegen-modules/pom.xml
cristianonicolai: +1

quarkus/extensions/pom.xml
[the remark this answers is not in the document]
cristianonicolai: @danielezonca I will include this module but we will need to tweak the pom, as it mentions persistence and SW. Will do a follow-up PR for that.

springboot/starters/pom.xml
```diff
@@ -23,9 +23,22 @@
...
+  <profiles>
+    <profile>
+      <id>default</id>
+      <activation>
+        <property>
+          <name>!productized</name>
+        </property>
+      </activation>
+      <modules>
+        <module>kogito-serverless-workflow-spring-boot-starter</module>
```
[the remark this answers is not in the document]
cristianonicolai: nope

quarkus/extensions/pom.xml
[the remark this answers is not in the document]
aloubyansky: Sorry, just curious, what kind of tweaking did you have in mind @cristianonicolai?

pom.xml
cristianonicolai: agree, but I still discussing with @MarianMacik and @jstastny-cz about what should be included, agree we should included most tests but for instance, integration-tests module was not included in the previous release. I'll tweak as necessary in a follow up pr

quarkus/extensions/pom.xml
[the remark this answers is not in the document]
danielezonca: This is a good point, some of the integration tests includes features that are not going to be included in next version of the product so we might need to remove them from this list 🤔 

Moreover the problem is probably even bigger 🤔 : the list of extension that Kogito provides in the product will be different (no serverless-workflow) compared to community so this could have also a larger impact at [quarkus-platform](https://github.com/quarkusio/quarkus-platform/blob/bac48b5c4af168abbb85e17cc6b2cec3e1e692f1/generated-platform-project/quarkus-kogito/bom/pom.xml#L1340-L1361) level 🤔 

/cc @evacchi FYI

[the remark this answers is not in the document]
cristianonicolai: @aloubyansky problem is that we're not including serverless workflow and persistence in this release, so items like the above would need to be excluded. I'm guessing that will be an issue?

https://github.com/kiegroup/kogito-runtimes/blob/main/quarkus/extensions/kogito-quarkus-test-list/pom.xml#L53
https://github.com/kiegroup/kogito-runtimes/blob/main/quarkus/extensions/kogito-quarkus-test-list/pom.xml#L70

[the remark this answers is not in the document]
aloubyansky: I suggest you to not do anything about it. And **not** remove non-supported extensions and other non-supported artifacts from the BOMs and this list. We are not doing it in the core `quarkus-bom` and no other member is expected to be doing that.
The idea is that you take the community BOMs (and other artifacts that are listing artifacts) and simply replace the community artifact versions with the productized ones where they are available.
We are still interested in running the non-productized community tests to make sure the product is compatible with the upstream version. Because customers still should have a choice to use non-supported extensions in their apps on their own risk.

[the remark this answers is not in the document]
cristianonicolai: @aloubyansky thanks for the clarification, makes sense to me, will beep it as is.

tiagodolphine (approved):

PR #1536, opened 2021-08-17
[KOGITO-3922] SWF error handling

fjtirado:


kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/handlers/CompositeContextNodeHandler.java
```diff
@@ -222,6 +229,47 @@ protected CompositeContextNodeHandler(S state, Workflow workflow, RuleFlowNodeCo
...
+        } else {
+            return new ObjectDataType();
+        }
+
+    }
+
+    private Object processWorkItemValue(JsonNode jsonNode, String paramName) {
+        if (jsonNode.isTextual()) {
+            // assume a string will be a json path expresion
+            return new JsonPathExprSupplier(jsonNode.asText(), paramName);
```
ricardozanini, line 259: Maybe we could add a check here just to make sure? The jsonpath expressions is within `${  }`. Just because the `inject` state can have plain strings added to the workflow data, not expressions.

ricardozanini (commented):
So in this PR we can catch any Java Exception by defining `code: <expression qualified name>`? Can you also please send a doc update to this section: https://docs.jboss.org/kogito/release/latest/html_single/#chap-kogito-orchestrating-serverless (probably a 6.4 section - Error Handling) cc @hmanwani-rh 

Thanks for the implementation and for the example!

kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/handlers/CompositeContextNodeHandler.java
fjtirado: sure, done

ricardozanini (approved):

jbpm/jbpm-flow/src/main/java/org/jbpm/ruleflow/core/RuleFlowProcessFactory.java
```diff
@@ -337,18 +309,16 @@ protected void linkBoundarySignalEvent(Node node, String attachedTo) {
...
-                            compositeNode.addContext(scope);
-                            compositeNode.setDefaultContext(scope);
-                            return scope;
-                        });
-        final Boolean hasErrorCode = (Boolean) node.getMetaData().get(HAS_ERROR_EVENT);
         final String errorCode = (String) node.getMetaData().get(ERROR_EVENT);
+        final ContextResolver compositeNode = (ContextResolver) attachedNode;
+        ExceptionScope exceptionScope = (ExceptionScope) compositeNode.resolveContext(EXCEPTION_SCOPE, errorCode);
+        if (exceptionScope == null) {
+            ContextContainer contextContainer = (ContextContainer) (node instanceof ContextContainer ? node : ((KogitoNode) node).getParentContainer());
```
tiagodolphine, line 316: so just to clarify, this change was needed because the exceptionScope should be set in the parent container in case the node itself does is not a ContextContainer.

kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/handlers/CompositeContextNodeHandler.java
```diff
@@ -222,6 +229,51 @@ protected CompositeContextNodeHandler(S state, Workflow workflow, RuleFlowNodeCo
...
+            return new StringDataType();
+        } else if (value instanceof Boolean) {
+            return new BooleanDataType();
+        } else if (value instanceof Integer) {
+            return new IntegerDataType();
+        } else if (value instanceof Number) {
+            return new FloatDataType();
+        } else {
+            return new ObjectDataType();
+        }
```
tiagodolphine, line 252: Minor, but this seems to be a common function, I mean that is spread in different parts of the codebase, probably we can centralize this to avoid replication, just something I realized.
See some examples, like based on the class names:
- https://github.com/kiegroup/kogito-runtimes/blob/f7647d5e0cfa1ec1c183888154bf7b962aad1232/jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/AbstractNodeHandler.java#L491-L508
- https://github.com/kiegroup/kogito-runtimes/blob/f855f1d0d21e1f6c0fa1d4bfcf289fe5cb0956ec/jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/DataObjectHandler.java#L90-L108
- https://github.com/kiegroup/kogito-runtimes/blob/f7647d5e0cfa1ec1c183888154bf7b962aad1232/jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/DefinitionsHandler.java#L176-L194
- https://github.com/kiegroup/kogito-runtimes/blob/aee072ff0de55fc0820437fe65b4f8f2e6e707fd/jbpm/process-serialization-protobuf/src/main/java/org/kie/kogito/serialization/process/impl/ProtobufVariableReader.java#L67-L85

tiagodolphine (approved):
LGTM just left some comments.
Thanks, @fjtirado, well done.

kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/handlers/CompositeContextNodeHandler.java
fjtirado: yes, I thought the same, that the same code should have been already used somewhere else. In fact, I made the search, but my faulty IDE does not return as many case as the ones you are pointing out...
I have opened https://issues.redhat.com/browse/KOGITO-5777 to fix this.

jbpm/jbpm-flow/src/main/java/org/jbpm/ruleflow/core/RuleFlowProcessFactory.java
fjtirado: yes, that's it, this change is enough to add "support" for boundary events over action nodes
Although the bpmn parser still has a validation check that prevents to associate a boundary error event to a script node (which is an action), this change is enough for SWF, which uses a different parser that do not have that check.

PR #1546, opened 2021-08-23
KOGITO-5747 Public API: Shared Modules (Incubation)

evacchi:
https://issues.redhat.com/browse/KOGITO-5747

Blocks:
- #1547 
- #1549 
- #1550 
- #1551 

## Description

These shared modules make available the `common` component and the `application` component. 

#### General Expected Usage Example

Every component should provide one or more service implementations. Each service takes an identifier (a Path from a root) and usually a `DataContext`. The generic form of interaction should follow the pattern:

```java
   LocalId id = appRoot.get(MyComponent.class).get("componentName").subs().get("someSubComponent")
   DataContext result = svc.evaluationMethod(id, dataContext);

```

for instance, evaluating a DMN may look like this:

```java
   LocalDecisionId decisionId = appRoot.get(DecisionIds.class).get(namespaceString, nameString);
   DataContext result = svc.evaluate(decisionId, dataContext);

```

When a component admits subcomponents (e.g. tasks in processes, decision services in DMN, queries in rules, etc.), then there should be a way to construct that path with a fluent API:

```java
   LocalDecisionId decisionServiceId = appRoot.get(DecisionIds.class).get(namespaceString, nameString).services().get(serviceString);
   DataContext result = svc.evaluate(decisionServiceId, dataContext);

```

If the types make sense, then a single `evaluate` method should be able to "parse" the provided ID, even for different paths; i.e. it is perfectly acceptable, even preferrable:

```java
   DataContext result = svc.evaluate(serviceId, inputDataContext);
   DataContext serviceResult = svc.evaluate(decisionServiceId, inputDataContext);

``` 

However, component developers are free to provide their own methods; multiple methods should be returned, especially if they represent different actions; e.g.:

 
```java
   LocalProcessInstanceId lpid = svc.createProcess(processId, inputDataContext);
   DataContext started = svc.start(lpid);
   DataContext aborted = svc.abort(lpid);
``` 

### Application and Paths

The `application` component provides the root of the typed builder API for paths (`appRoot`); 

i.e. paths such as `/processes/my.process` or `/predictions/my.prediction` are constructed via `appRoot.get(ProcessIds.class).get("my.process")` or `appRoot.get(PredictionIds.class).get("my.prediction")` respectively. The return type of this chain of methods is always an `Id` or a `LocalId`.

The `<Component>Ids` class should implement `ComponentRoot`.

paths admit subpaths if the Component admits subcomponents; e.g. a DMN admits services, so the path 

     /decisions/$id/services/$serviceId

should be allowed. The method chain should reflect the structure of that path:

    appRoot.get(DecisionIds.class).get(namespace, name).services().get(serviceId)

A `RemoteId` is also available to "install" the path on a hostname:port pair. This is not useful at the moment, it will be useful in a distributed setting.

In the future such a path will be also parsable starting from a String (e.g. `LocalId.parse(String)`)

Because of their construction, paths can be always rendered as a string, and can always be parsed from a string into a typed representation (the `/prefix` and intermediated parts (such as `/services` specify the subtype). This makes them trivially serializable for sending over the wire and trivially unserializable into their typed representation.

### Main API (`common`)

The `common` module provides basic interfaces so that each component may specialize one or more `LocalId` definitions and provide one or more service implementation. 

Every component provides a pair of modules:

- `kogito-api-incubation-<component>`: Paths and Identifiers
- `kogito-api-incubation-<component>-services`: Service Interfaces

#### Paths and Identifiers

The first module contains specializations of `LocalId` (e.g. `DecisionId`) and/or subpaths. For instance:

- a "decision service ID" should be addressable, and that service Id is a subcomponent of a decision. Then `DecisionId` should expose the method `services()` and `appRoot.get(DecisionIds.class).get("my.decision").services().get("my.service.id")` should be the path constructor; it would create a `DecisionServiceId`

#### DataContext

- A `DataContext` is a computational context. It contains all the data that is required by a service to perform its computation.
- A `DataContext` can be always converted into another type (or another DataContext) using the method `DataContext#as(Class<T extends DataContext>)`.

A DataContext can be always converted into another sub-type of DataContext. Converting a DataContext into the same DataContext is a no-op just like (Object) new Object() is a no-op. You are always able to convert a DataContext into an arbitrary sub-type; e.g. a MapDataContext into a POJO that implements DataContext.

```java
class Person implements DataContext, DefaultCastable { String name; }
MapDataContext ctx = MapDataContext.create();
ctx.set("name", "Paul");
Person p = ctx.as(Person.class);
String name = p.name; // "Paul"
```

`MapDataContext` provides methods `get(), set()` and wraps a map. This is the simplest and convenient way to pass data. However, it is suggested to create your own class. In this case your POJO should extends `DataContext` and `DefaultCastable` (the latter provides a default implementation of the `as()` method that is usually good for most cases)

Note: `DefaultCastable#as` uses Jackson internally, but it's an implementation detail and it may change in the future if we decide so.

#### Service Interfaces

Service interfaces should follow the generic pattern :

    RETURN-TYPE VERB(IDENTIFIER [, optional DATACONTEXT])

e.g.

```java
   LocalId id = appRoot.get(MyComponent.class).get("componentName").subs().get("someSubComponent")
   DataContext result = svc.evaluationMethod(id, dataContext);

```

- Service interfaces do not extend a specific interface; component developers may define their own methods, although a certain degree of consistency is expected.
- **RETURN-TYPE** should be always either an IDentifier or a DataContext. Implementations should usually take the most generic implementation `DataContext` as a parameter, and should usually return `DataContext` as a return type. It is allowed (as per Java spec) to return a subtype of DataContext, but this may not be required, as users can always convert into a more specific type.
- When a method *evaluates* the resource pointed by the IDENTIFIER and **completes** (i.e. stateless evaluation), then RETURN-TYPE should always be DataContext.
- When a method *starts the evaluation* of the resource pointed by the IDENTIFIER but it may not **complete**, then RETURN-TYPE may be an identifier.  (e.g. `processService.create(id)` would return a `processInstanceId`)

Asyncronous evaluation is allowed. In this case RETURN-TYPE should express the fact that evaluation is async. For instance:

```java
   LocalId id = appRoot.get(MyComponent.class).get("componentName").subs().get("someSubComponent")
   CompletableFuture<DataContext> result = asyncSvc.evaluationMethod(id, dataContext);

```

i.e. a Process engine may provide `AsyncProcessService`

```java
   CompletableFuture<LocalProcessInstanceId> futureLpid = asyncSvc.createProcess(processId, inputDataContext);
   CompletableFuture<DataContext> futureStarted = futureLpid.andThen(lpid -> asyncSvc.start(lpid));
   ...
``` 

### Implementation

Currently only for Quarkus. The implementation is being shipped as part of the extension for convenience (it may be refactored to its own extension in the future). Implementation should provide:

- injectable version of their component class -- used for wiring in `appRoot.get(MyComponent.class)`
- injectable versions of their services

Currently services delegate their implementation to the internal hidden API that is used for codegen. The internal API should be considered deprecated and in the future codegen should be moved to the new API.

Refer to the related PRs for more details.

### Caveats

- **What about listeners?**

Currently listeners are plugged through annotation as we currently do already in Kogito. Since internally the APIs that we use in codegen are being used, those will work automatically.

- **What about metadata?** The API only deals with data, but what about meta-data? e.g. directives that should be passed to the engine, that are not part of the Payload? (e.g. HTTP headers that should be passed to the engine to configure a specific request)

  ```java
     var result = svc.evaluate(id, payload [ metadata ???] ) 
  ```

  This is currently an open question. 

  * A proposal is to allow the DataContext to also include metadata 
  * Another proposal is to have some is that the `as` method does not just convert but it's a sort of "view"; so `ctx.as(MetaDataContext.class)` may show the metadata and _not_ the plain data -- I like this but I am not sure how to implement it.

<details>
<summary>
</summary>

* <b>Run all builds</b>  

* <b>Quarkus LTS checks</b>  

* <b>Native checks</b>  

</details>

### DIFF
--- a/api/kogito-api-incubation-application/pom.xml
+++ b/api/kogito-api-incubation-application/pom.xml
@@ -1,0 +1,83 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2021 Red Hat, Inc. and/or its affiliates.
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~       http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.kie.kogito</groupId>
+        <artifactId>kogito-api-parent</artifactId>
+        <version>2.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>kogito-api-incubation-application</artifactId>
+    <packaging>jar</packaging>
+
+    <name>Kogito :: API :: Public API :: Incubation :: Application</name>
+    <description>The Kogito Public Application API (Incubation).</description>
+
+    <properties>
+        <java.module.name>org.kie.kogito.api.incubation.application</java.module.name>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.kie.kogito</groupId>
+            <artifactId>kogito-api-incubation-common</artifactId>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-enforcer-plugin</artifactId>
+                <dependencies>
+                    <dependency>
+                        <groupId>de.skuzzle.enforcer</groupId>
+                        <artifactId>restrict-imports-enforcer-rule</artifactId>
+                        <version>${version.de.skuzzle.enforcer}</version>
+                    </dependency>
+                </dependencies>
+                <executions>
+                    <execution>
+                        <id>check-restricted-imports</id>
+                        <phase>process-sources</phase>
+                        <goals>
+                            <goal>enforce</goal>
+                        </goals>
+                        <configuration>
+                            <rules>
+                                <restrictImports implementation="de.skuzzle.enforcer.restrictimports.rule.RestrictImports">
+                                    <reason>Allow only module imports</reason>
+                                    <bannedImports>
+                                        <bannedImport>org.kie.kogito.**</bannedImport>
+                                    </bannedImports>
+                                    <allowedImports>
+                                        <!-- we only allow common.*  -->
+                                        <allowedImport>org.kie.kogito.incubation.common.**</allowedImport>
+                                    </allowedImports>
+                                </restrictImports>
+                            </rules>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+</project>
--- a/api/kogito-api-incubation-application/src/main/java/org/kie/kogito/incubation/application/AppRoot.java
+++ b/api/kogito-api-incubation-application/src/main/java/org/kie/kogito/incubation/application/AppRoot.java
@@ -15,11 +15,10 @@
  */
 package org.kie.kogito.incubation.application;
 
-import java.nio.file.Paths;
-
 import org.kie.kogito.incubation.common.ComponentRoot;
 import org.kie.kogito.incubation.common.Id;
-import org.kie.kogito.incubation.common.PathLocalId;
+import org.kie.kogito.incubation.common.LocalUri;
+import org.kie.kogito.incubation.common.LocalUriId;
 
 /**
  * Root path of an application.
@@ -32,12 +31,12 @@ import org.kie.kogito.incubation.common.PathLocalId;
  * Each component provides a fluent builder for an {@link Id} that is specific
  * to that component.
  */
-public abstract class AppRoot extends PathLocalId implements Id {
+public abstract class AppRoot extends LocalUriId implements Id {
 
     private final String name;
 
     protected AppRoot(String name) {
-        super(Paths.get("/"));
+        super(LocalUri.Root);
         this.name = name;
     }
 
--- a/api/kogito-api-incubation-application/src/main/java/org/kie/kogito/incubation/application/AppRoot.java
+++ b/api/kogito-api-incubation-application/src/main/java/org/kie/kogito/incubation/application/AppRoot.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.kogito.incubation.application;
+
+import java.nio.file.Paths;
+
+import org.kie.kogito.incubation.common.ComponentRoot;
+import org.kie.kogito.incubation.common.Id;
+import org.kie.kogito.incubation.common.PathLocalId;
+
+/**
+ * Root path of an application.
+ *
+ * An application has a name. Many applications may be created. The name creates
+ * a name space for each application, so that components may be individually
+ * addressed across application instances.
+ *
+ * An AppRoot retrieves "components" through {@link ComponentRoot}s.
+ * Each component provides a fluent builder for an {@link Id} that is specific
+ * to that component.
+ */
+public abstract class AppRoot extends PathLocalId implements Id {
+
+    private final String name;
+
+    protected AppRoot(String name) {
+        super(Paths.get("/"));
+        this.name = name;
+    }
+
+    /**
+     * subclasses should override this using the appropriate
+     * DI/ServiceLoading mechanism to allow the pattern <code>appRoot.get(Components.class)...</code>;
+     * e.g. <code>appRoot.get(ProcessIds.class).get("my.process.id).tasks().get("my.task")</code>
+     */
+    abstract public <T extends ComponentRoot> T get(Class<T> providerId);
+
+    /**
+     * Name is only used to differentiate multiple applications.
+     * Mostly useful in a distributed context, with RemoteIds.
+     */
+    public String name() {
+        return name;
+    }
+
+}
--- a/api/kogito-api-incubation-application/src/main/java/org/kie/kogito/incubation/application/ReflectiveAppRoot.java
+++ b/api/kogito-api-incubation-application/src/main/java/org/kie/kogito/incubation/application/ReflectiveAppRoot.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.incubation.application;
+
+import org.kie.kogito.incubation.common.ComponentRoot;
+
+/**
+ * Useful for testing. Creates Components reflectively.
+ */
+public class ReflectiveAppRoot extends AppRoot {
+    public ReflectiveAppRoot(String name) {
+        super(name);
+    }
+
+    public ReflectiveAppRoot() {
+        super("kogito-app");
+    }
+
+    @Override
+    public <T extends ComponentRoot> T get(Class<T> providerId) {
+        try {
+            return providerId.getConstructor().newInstance();
+        } catch (ReflectiveOperationException e) {
+            throw new IllegalArgumentException(e);
+        }
+    }
+}
--- a/api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/Castable.java
+++ b/api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/Castable.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.incubation.common;
+
+/**
+ * Represents a type that can be converted into a {@link DataContext}
+ */
+public interface Castable {
+    /**
+     * Converts this object into the given type.
+     */
+    <T extends DataContext> T as(Class<T> type);
+}
--- a/api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/DataContext.java
+++ b/api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/DataContext.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.incubation.common;
+
+/**
+ * A type that represents a computational context.
+ *
+ * <ul>
+ * <li>A DataContext is a computational context. It contains all the data that
+ * is required by a service to perform its computation.</li>
+ * <li>A DataContext can be always converted into another type
+ * (or another DataContext) using the method {@link DataContext#as(Class)}</li>
+ * <li></li>A DataContext can be always converted into another sub-type of DataContext.</li>
+ * <li>Converting a DataContext into the same DataContext is a no-op just like (Object) new Object() is a no-op.
+ * <li>You are always able to convert a DataContext into an arbitrary sub-type; e.g. a {@link MapDataContext}
+ * into a class that implements DataContext.</li>
+ * </ul>
+ * <p>
+ * Example:
+ *
+ * <code><pre>
+ * class Person implements DataContext, DefaultCastable { String name; }
+ * MapDataContext ctx = MapDataContext.create();
+ * ctx.set("name", "Paul");
+ * Person p = ctx.as(Person.class);
+ * String name = p.name; // "Paul"
+ * </pre></code>
+ * <p>
+ * {@see MapDataContext}
+ */
+public interface DataContext extends Castable {
+}
--- a/api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/DefaultCastable.java
+++ b/api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/DefaultCastable.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.incubation.common;
+
+/**
+ * Utility interface, useful to mix-in to get a default `as` implementation.
+ * <p>
+ * Provides a default implementation for the {@link #as(Class)} method,
+ * delegating to {@link InternalObjectMapper#convertValue(Object, Class)}
+ */
+public interface DefaultCastable extends Castable {
+
+    default <T extends DataContext> T as(Class<T> type) {
+        return (T) InternalObjectMapper.convertValue(this, type);
+    }
+}
--- a/api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/InternalObjectMapper.java
+++ b/api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/InternalObjectMapper.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.incubation.common;
+
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.PropertyAccessor;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * For internal use only.
+ * Provides a method to convert an object into a given type.
+ * This is an implementation detail. We may move this to a separate module in the future.
+ */
+public class InternalObjectMapper {
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+
+    static {
+        objectMapper.setVisibility(PropertyAccessor.ALL, JsonAutoDetect.Visibility.NONE);
+        objectMapper.setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY);
+    }
+
+    public static <T> T convertValue(Object self, Class<T> type) {
+        if (MapLikeDataContext.class == type || MapDataContext.class == type) {
+            return (T) MapDataContext.of(objectMapper.convertValue(self, Map.class));
+        }
+        return objectMapper.convertValue(self, type);
+    }
+
+    private InternalObjectMapper() {
+    }
+}
--- a/api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/LocalId.java
+++ b/api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/LocalId.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.incubation.common;
+
+import java.nio.file.Path;
+
+/**
+ * An identifier for a local resource
+ */
+public interface LocalId extends Id {
+    Path asPath();
+}
--- a/api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/LocalId.java
+++ b/api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/LocalId.java
@@ -16,11 +16,9 @@
 
 package org.kie.kogito.incubation.common;
 
-import java.nio.file.Path;
-
 /**
  * An identifier for a local resource
  */
 public interface LocalId extends Id {
-    Path asPath();
+    LocalUri asLocalUri();
 }
--- a/api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/MapDataContext.java
+++ b/api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/MapDataContext.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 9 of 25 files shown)


api/kogito-api-incubation-application/src/main/java/org/kie/kogito/incubation/application/ReflectiveAppRoot.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public class ReflectiveAppRoot extends AppRoot {
+    public ReflectiveAppRoot() {
+        super("kogito-app");
+    }
+
+    @Override
+    public <T extends ComponentRoot> T get(Class<T> providerId) {
+        try {
+            return providerId.getConstructor().newInstance();
```
evacchi, line 30: however, the proper way to do this would be via service loading -- we may decide to do it; however this is not intended to be used outside tests

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/InternalObjectMapper.java
```diff
@@ -0,0 +1,56 @@
...
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.PropertyAccessor;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * For internal use only (package-private).
+ * Provides a method to convert an object into a given type.
+ * This is an implementation detail. We may move this to a separate module in the future.
```
evacchi, line 29: another candidate for service loading

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/Castable.java
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.incubation.common;
+
+public interface Castable {
```
gitgabrio, line 19: Hi @evacchi (copy)
Could you explain the needs of this ? IIUC DataContext extends Castable; so it seems the parent interface is declaring a method that can actually works only on a specific "son" (i.e. another different interface extending "Castable" should always return a "DataContext", and not itself): is my question clear ? Am I misreading something ?

I read your [answer](https://github.com/kiegroup/kogito-runtimes/pull/1549/files/5fe38d63de228823afb0ddfb6dcc6cac3bd18af2#r698337490) but IMO this does not justify that design, that - again IMO - is wrong: a "parent" interface (Castable) has knowledge of and reference to a children one (DataContext): I think this is an antipattern in OOP

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/DataContext.java
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.incubation.common;
+
+public interface DataContext extends Castable {
```
gitgabrio, line 19: @evacchi (copy)
Could you please some javadoc explaining what this interface actually is for ?

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/DefaultCastable.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.incubation.common;
+
+/**
+ * Utility interface, useful to mix-in to get a default `as` implementation
+ */
+public interface DefaultCastable extends Castable {
```
gitgabrio, line 22: @evacchi
Could you please some javadoc explaining what this interface actually is for ? The provided one seems more related to "Castable" itself, and not specific to this one....

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/DefaultCastable.java
```diff
@@ -0,0 +1,27 @@
...
+
+    default <T extends DataContext> T as(Class<T> type) {
```
gitgabrio, line 24: @evacchi
See comment above: apparently, there is no direct relationship between DefaultCastable and DataContext: IIUC, an actual implementation of the former would have a different inheritance tree of an actual implementation of the latter. I have the impression this is a little bit worse then the previous comment. Here, one extending interface should have knowledge of and reference to a different extending interface of the same parent

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/MapDataContext.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.Objects;
+
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import com.fasterxml.jackson.annotation.JsonAnySetter;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
+/**
+ * A data context that wraps a <code>Map<String, Object></code>.
+ */
+public class MapDataContext implements MapLikeDataContext, MetaDataContext {
```
gitgabrio, line 30: @evacchi ([see](https://github.com/kiegroup/kogito-runtimes/pull/1549/files/5fe38d63de228823afb0ddfb6dcc6cac3bd18af2#r698317961) for reference)

gitgabrio (changes requested):
@evacchi - Thanks for the PR. Copied comments I made [here](https://github.com/kiegroup/kogito-runtimes/pull/1549) and kept discussion about interface inheritance usage

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/Castable.java
evacchi: The reason why DataContext extends Castable is because all DataContexts should be Castables, but not all Castables may be DataContexts.The idea would be that you may make any class Castable to a DataContext without being a DataContext, by implementing the `DefaultCastable` interface (or `Castable` itself).

i.e. suppose that you want to provide a data object and make it possible to convert it into a DataContext; but you don't want it to be passed as a DataContext itself.

For instance you may have an object with some extra fields that you don't want to be available to the evaluation. By implementing the `as` method yourself you'd be able to customize what fields should be made available.

TBF I am not necessarily stronly opinionated about this one, I just wanted to keep interface tiny.
I am not strongly opionionated about `as` being an instance method (it may be very constraining). Maybe we will turn that into into some sort of static method; in that case it may make sense for the two interfaces to be separate; e.g.:

     Castable.of(myContext).as(...)

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/MapDataContext.java
evacchi: @gitgabrio 
> IIUC MetaDataContext extends MapLikeDataContext, so this seems a little bit redundant

yeah it's useless without an example, the idea would be that this allows to get "metadata" when you do `.as(MetaDataContext.class)` instead of "regular data" (e.g. params). This interface is not really useful at the moment though; btw still relates to #1546

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/Castable.java
gitgabrio: Hi @evacchi I think I've understood the idea, but my remark is about the implementation.
We want to separate two concepts (Castable and DataContext) but then we merge them in a single one, because, how it is written, "Castable" actually is "CastableToDataContext". This means that a "Castable" may be cast only to a DataContext, but a "DataContext" is already a "DataContext" - what's the meaning of _casting_ a DataContext to a DataContext ?
If, on the other side, you want any given "Castable" to returns a DataContext (as the method and your explanation suggest), maybe composition is the pattern to use, not inheritance.

Anyway, beside all this academical reasoning, I think that referring a "son" in an interface/class is something better to avoid. If you find a way to change that, like the static method or whatever, I think it would be better.

evacchi: ok I see your point. The use case is to be able convert into a *subclass* of `DataContext`. Converting a `DataContext` into the same `DataContext` is a no-op just like `(Object) new Object()` is a no-op :-) 

the idea is you are always able to convert a `DataContext` into an arbitrary sub-type; e.g. a `MapDataContext` into a POJO that implements `DataContext`.

e.g.

```
class Person implements DataContext, DefaultCastable { String name; }
MapDataContext ctx = MapDataContext.create();
ctx.set("name", "Paul");
Person p = ctx.as(Person.class);
p.name // "Paul"
```

This is generalizing e.g. the DMNContext to a more "generic" interface that allows conversion of classes from map-like structures (JSON-izable if you will) into arbitrary typed POJOs 

So, both Person and MapDataContext are DataContexts, but they are not of the same type. You convert them at the use-site in order to use typed vs untyped API

gitgabrio: @evacchi 
I'll keep commenting here, but feel free to ping me - maybe by voice would be easier.
It seems we want that
1) a DataContext should be able to be "casted" to a (different) DataContext
2) a Castable should be able to be "casted" to a  DataContext

My personal opinion is that interfaces, at their core, should actually define "behaviors". Looking that way, it seems to me that  those two interfaces are actually the same. I understand that the "Castable" concept should be slightly different, but the implementation lead to that. Does this make sense ? 
I understand your explanations and requirements, but my point is that the root of current the solution is (again, IMO) wrong. So, I would prefer to have a different design/implementation to fulfill them, even if more verbose/contrived.

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/DataContext.java
evacchi: added

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/DefaultCastable.java
evacchi: added

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/MapDataContext.java
evacchi: I have removed `MetaDataContext.class` for now. I'll have to rebase the other PRs but that's only because it was used in tests -- not necessary for now.

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/DataContext.java
```diff
@@ -16,5 +16,31 @@
 
 package org.kie.kogito.incubation.common;
 
+/**
+ * A type that represents a computational context.
+ *
+ * <ul>
+ * <li>A DataContext is a computational context. It contains all the data that
+ * is required by a service to perform its computation.</li>
+ * <li>A DataContext can be always converted into another type
```
gitgabrio, line 25: @evacchi 
I add this comment here as remember

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/UriRemoteId.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+public final class UriRemoteId implements RemoteId {
+    private final URI uri;
+    private final LocalId localId;
+
+    public UriRemoteId(String appId, String host, int port, LocalId localId) {
+        this.localId = localId;
+        try {
+            // the explicit path parameter url-encodes again; this is the "better" way, apparently...
+            this.uri = new URI(String.format("kogito://%s@%s:%d%s", appId, host, port, localId.asPath()));
```
ricardozanini, line 33: Maybe this is a `KogitoUriRemoteId`? Given that we will support `kubernetes` or maybe `http|s` in the future? Either way, we should be able to scale the implementation, no?

quarkus/extensions/kogito-quarkus-extension-common/kogito-quarkus-common/pom.xml
```diff
@@ -92,6 +96,18 @@
                     </annotationProcessorPaths>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.jboss.jandex</groupId>
+                <artifactId>jandex-maven-plugin</artifactId>
```
ricardozanini, line 101: I think this plugin can be inherited from the parent extension module. IIRC, all extensions modules require it.

ricardozanini (commented):
Looks great for a first step, and the API seems robust enough for an early stage. Many thanks for the documentation. It's well written and informative.

Regarding the `DataContext` and `Castable` interfaces, I tend to agree with @gitgabrio. I think after we settle this discussion, we can move forward.

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/InternalObjectMapper.java
tiagodolphine: yeah, and I'd say we probably can share internal object mapper that already exists in other places, mainly to do data conversion, see 
https://github.com/kiegroup/kogito-runtimes/blob/3283958b111b29bf206071eb531cf82bec52ebfb/jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/transformation/JsonResolver.java#L37 
https://github.com/kiegroup/kogito-runtimes/blob/fdc27397ec5fae2ee88576d90975481f409848eb/jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/util/TypeTransformer.java#L36
https://github.com/kiegroup/kogito-runtimes/blob/22195a19dbf27e1bfdc2855bd62a46d896778964/jbpm/jbpm-flow/src/main/java/org/jbpm/util/JsonSchemaUtil.java#L43
https://github.com/kiegroup/kogito-runtimes/blob/f855f1d0d21e1f6c0fa1d4bfcf289fe5cb0956ec/kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/persistence/proto/AbstractProtoGenerator.java#L36
..... 
among others... so I'd say this is not the scope of this PR but we could open a Jira to track this, otherwise, we may come with tons of objectMapper instances

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/UriRemoteId.java
tiagodolphine: and it would be great to allow a way for customization...

quarkus/extensions/kogito-quarkus-extension-common/kogito-quarkus-common/src/main/java/org/kie/kogito/incubation/application/quarkus/QuarkusAppRoot.java
```diff
@@ -0,0 +1,39 @@
...
+import org.kie.kogito.incubation.application.AppRoot;
+import org.kie.kogito.incubation.common.ComponentRoot;
+
+@ApplicationScoped
+public class QuarkusAppRoot extends AppRoot {
+    @Inject
+    private Instance<ComponentRoot> componentRoots;
+
+    QuarkusAppRoot() {
+        super("kogito-app");
```
tiagodolphine, line 32: the same.. should we allow customization of the `name`?

tiagodolphine (approved):
LGTM, great job! Just some minor comments.

ricardozanini:
@tiagodolphine regarding ObjectMapper, I have the same opinion, but I let it go since we have it tracked here: https://issues.redhat.com/browse/KOGITO-5618

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/InternalObjectMapper.java
evacchi: +1 https://issues.redhat.com/browse/KOGITO-5807

quarkus/extensions/kogito-quarkus-extension-common/kogito-quarkus-common/pom.xml
evacchi: I'll create a JIRA: testing this setup is complicated with broken-down separate PRs; better refactor this all at once when things are more settled https://issues.redhat.com/browse/KOGITO-5810

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/UriRemoteId.java
evacchi: Yeah, this is basically here for demo purposes... I think I should just remove it for the time being. It has no real use at the moment.

evacchi:
@ricardozanini : 
@gitgabrio is suggesting the following refactoring:

```diff
- interface DataContext implements Castable {
+ interface DataContext {
+    <T extends DataContext> T as(Class<T> type);
}
```

so remove the `implements` clause and redeclare the `as` method explicitly. I am not strongly opinionated about this one. The `as()` method is there only to have the fluent:

```
myObject.as(SomeType.class)
```

as mentioned, in order to make it pluggable we may provide a different API, and *static*, instead of instance methods. Or even provide a way to use both strategies (e.g. static method + default method on interface -- the main question: shall we make it pluggable? and how? and when?)

e.g. specific services may want to provide their own DataContext implementation with an `as` method that exposes a specific subset of fields. By using Jackson's ObjectMapper we may be able to define Jackson Modules for each DataContext subtype, but we'll need a loading mechanism to plug them (DI for Quarkus/Spring -- what about "neither" ? ServiceLoaders ?)

in other words: every component should be able to define its own behavior, and that needs some further thinking

api/kogito-api-incubation-application/src/main/java/org/kie/kogito/incubation/application/ReflectiveAppRoot.java
evacchi: https://issues.redhat.com/browse/KOGITO-5815

api/kogito-api-incubation-application/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.kie.kogito</groupId>
+        <artifactId>kogito-api-parent</artifactId>
+        <version>2.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>kogito-api-incubation-application</artifactId>
```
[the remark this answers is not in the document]
evacchi: I am respecting the name convention of this sub-root. They are all `kogito-api-<something>`

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/UriRemoteId.java
evacchi: removed

cristianonicolai (approved):

quarkus/extensions/kogito-quarkus-extension-common/kogito-quarkus-common/src/main/java/org/kie/kogito/incubation/application/quarkus/QuarkusAppRoot.java
danielezonca: I agree and also `kogito-app` is quite misleading with kogito-apps repo (at least for us

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/LocalId.java
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.incubation.common;
+
+import java.nio.file.Path;
+
+/**
+ * An identifier for a local resource
```
danielezonca, line 22: What about `RelativeId` instead of local? Considering that every Id has a "local" part maybe it is easier to understand 🤔

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/LocalId.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+public interface LocalId extends Id {
+    Path asPath();
```
danielezonca, line 25: Are you sure about this? Is this "reasonable"?
Every Id can be converted to LocalId using `toLocalId()` method so this means that also a RemoteId can be converted as Path and this is impossible 🤔 
Probably URI or similar is more generic or we can consider to remove `toLocalId` method

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/DataContext.java
```diff
@@ -0,0 +1,46 @@
...
+ * class Person implements DataContext, DefaultCastable { String name; }
+ * MapDataContext ctx = MapDataContext.create();
+ * ctx.set("name", "Paul");
+ * Person p = ctx.as(Person.class);
+ * String name = p.name; // "Paul"
+ * </pre></code>
+ * <p>
+ * {@see MapDataContext}
+ */
+public interface DataContext extends Castable {
```
danielezonca, line 45: As we already discussed I think we should have a codepath that doesn't require to implement an interface so that it is possible for users to reuse their existing bean without any need of customisation/rewrite.
Said diffently having the possibility to only use composition instead of inheritance (there are already similar comment from others)
We can do it with a different PR of course

api/kogito-api-incubation-application/src/main/java/org/kie/kogito/incubation/application/AppRoot.java
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * An application has a name. Many applications may be created. The name creates
+ * a name space for each application, so that components may be individually
+ * addressed across application instances.
+ *
+ * An AppRoot retrieves "components" through {@link ComponentRoot}s.
+ * Each component provides a fluent builder for an {@link Id} that is specific
+ * to that component.
+ */
+public abstract class AppRoot extends PathLocalId implements Id {
```
danielezonca, line 35: I think another missing bit to have a full and comprehensive API is to provide an alternative/equivalent of the old KnoledgeBuilder to let user to have a full library use case with no DI or in general explicit resource wiring.
Is this already in the plan? Without a similar feature it is sort of impossible to apply this API for all the use cases.
Currently `AppRoot` is not in charge at all for resources wiring and each component/service is expected to have them "automagically"

danielezonca (commented):
Overall looks good to me, just some comments

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/LocalId.java
evacchi: > this is impossible

it won't be "local" but it would still be a path. Why do you think this is impossible?

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/LocalId.java
evacchi: it is not relative because it always starts at `/`

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/LocalId.java
danielezonca: Ok maybe I misunderstood, let me do an example to double check:

I have a RemoteId `kogito://decisions/myName/myNamespace`, if I use `toLocalId()` it becomes `/decisions/myName/myNamespace`.
Then if finally I call `asPath()`, this method "verify" if I have the actual file locally (like a "file discovery service") and returns `/path/to/my.dmn`?

Or it only "convert" the localId as a path even if it is not a real path? (Something like `new Path("/decisions/myName/myNamespace")`?

Because if it is this last option, this will ~always generate a Path that is not valid (aka referring to a physical file) and as soon as user try to use it as real Path it will fail

api/kogito-api-incubation-application/src/main/java/org/kie/kogito/incubation/application/AppRoot.java
evacchi: Well, I'd say this goes well beyond "APIs", and it is more about mechanism. Will this API allow interaction with something that is generated without DI? Well I can't see why not. 

`AppRoot` is not in charge of wiring because its role is to construct a path, regardless of where the resource is actually located. Validation of such a path should not occur at Path construction, but at the use-site (i.e. in the service). This allows for more flexible services (e.g. use cases where the service does not physically reside in the local app). 

At the same time, a local non-DI service implementation could walk the filesystem (or the classpath, or whatever) and expose some resources through some `Id`, just the same as a DI-wired instance. Construction and instantiation of the service could be done through some other mechanism (possibly an equivalent of the KnowledgeBuilder, possibly explicit instantiation), because it is a separate concern.

the "service" +ID idea is that the ID represents an address to a resource, and the service replies to a request for that address. A service may always reply for some ID regardless if it is local, hand-wired, auto-wired, etc

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/LocalId.java
evacchi: let me explain. There is always a "local" part that can be extracted. It may not be local to your machine, but it would be local to "some" machine. On that machine, it will address the local resource.

evacchi: apparently I posted the clarification but your reply hadn't show up yet. This is correct:

> Or it only "convert" the localId as a path even if it is not a real path? (Something like `new Path("/decisions/myName/myNamespace")`?

a Path in this context is just a "typed"/"structured" string if you will. Validation will occur at the use-site, i.e. in the service.

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/DataContext.java
evacchi: - [KOGITO-5839 Public API: Separate DataContext conversion concern from `as()` instance method](https://issues.redhat.com/browse/KOGITO-5839)
- [KOGITO-5840 Public API: Provide DataContext wrapper for user classes](https://issues.redhat.com/browse/KOGITO-5840)

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/DefaultCastable.java
evacchi: after our conversation I have 
(1) separated the two hierarchies as you suggested
(2) still created the following JIRAs, as I believe it will be better design overall

- [KOGITO-5839 Public API: Separate DataContext conversion concern from `as()` instance method](https://issues.redhat.com/browse/KOGITO-5839)
- [KOGITO-5840 Public API: Provide DataContext wrapper for user classes](https://issues.redhat.com/browse/KOGITO-5840)

evacchi:
asking @ricardozanini @ricardozanini @gitgabrio for approval :)

Also adding @radtriste for approval

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/LocalId.java
evacchi: further clarifications: indeed Path is meant for a local file-system Path; it is not necessary for the Path to exist; however there is a different issue that may bite us: file separators! on Windows, `toString()` for `Path("a/b/c")` prints `Path("a\\b\\c")` so indeed Path is not ideal for this.

I am investigating what to use instead (URI is very limited and quite cumbersome when it comes to appending relative paths) but I would still merge this in the meantime; JIRA: https://issues.redhat.com/browse/KOGITO-5844

gitgabrio: Hi @evacchi 
Path is meant to represent a system dependent file path to locate a file in a file system.
It is true that the path may not exists for some reason, but this does not mean that we should use it for something that we already know will never exists (IIUC the discussion).
IMO, as a rule of thumb, we should not use classes for something different then their original scope. This lead to hard-to-find bugs and unexpected behavior (at least, from the "user" point of view).
To give an example:
`protected AppRoot(String name) {
        super(Paths.get("/"));
        this.name = name;
    }`
this is extremely suspicious.

gitgabrio (commented):
HI @evacchi 
I understand the idea of delivering something quickly to speed up the iterations, but I'm not convinced on pushing/delivering something (e.g. the Path usage) that we already know it is "wrong". Even if it is an "incubation" attempt, it does not mean that it may contain anything just for the sake of speed. M2C

danielezonca (approved):

gitgabrio (changes requested):
@evacchi 
Please, remove that usage of Path. Beside being an anti-pattern to use something for a different scope is meant to fulfill, it also introduce a clear bug, making this not working  on Windows and excluding all windows user from this "incubation" iteration.

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/LocalId.java
evacchi: yeah you guys are right. I opened a JIRA and I am working on an alternative; will fix in the next iteration

evacchi:
> @evacchi
> Please, remove that usage of Path. Beside being an anti-pattern to use something for a different scope is meant to fulfill, it also introduce a clear bug, making this not working on Windows and excluding all windows user from this "incubation" iteration.

I have started writing https://github.com/evacchi/kogito-runtimes/commit/01ba04f89aaa27e273aba38a90e981cdca7b86dc WDYT ? shall I bring this in? (name HierarchicalPath is temporary. Maybe "LocalUri" ?)

evacchi:
added a84426889d with the LocalUri API

gitgabrio (changes requested):
LGTM @evacchi Thanks for modification

ricardozanini (approved):

quarkus/extensions/kogito-quarkus-extension-common/kogito-quarkus-common/src/main/java/org/kie/kogito/incubation/application/quarkus/QuarkusAppRoot.java
radtriste: `kogito-runtime` ?

radtriste (approved):

PR #1550, opened 2021-08-24
KOGITO-5749 Public API: DMN Modules (Incubation)

evacchi:
Notes:
- see #1546 for an architectural overview
- The names of the classes and modules (e.g. Decisions vs DMN) are for demo purposes only, and can be changed

<details>
<summary>
</summary>

* <b>Run all builds</b>  

* <b>Quarkus LTS checks</b>  

* <b>Native checks</b>  

</details>

### DIFF
--- a/api/kogito-api-incubation-decisions-services/src/main/java/org/kie/kogito/incubation/decisions/services/DecisionService.java
+++ b/api/kogito-api-incubation-decisions-services/src/main/java/org/kie/kogito/incubation/decisions/services/DecisionService.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.incubation.decisions.services;
+
+import org.kie.kogito.incubation.common.DataContext;
+import org.kie.kogito.incubation.common.LocalId;
+
+public interface DecisionService {
+    DataContext evaluate(LocalId decisionId, DataContext inputContext);
+}
--- a/api/kogito-api-incubation-decisions/src/main/java/org/kie/kogito/incubation/decisions/DecisionIds.java
+++ b/api/kogito-api-incubation-decisions/src/main/java/org/kie/kogito/incubation/decisions/DecisionIds.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.incubation.decisions;
+
+import org.kie.kogito.incubation.common.ComponentRoot;
+
+public class DecisionIds implements ComponentRoot {
+    public LocalDecisionId get(String namespace, String name) {
+        return new LocalDecisionId(namespace, name);
+    }
+}
--- a/quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions-integration-test/src/test/java/org/kie/kogito/quarkus/dmn/CustomEndpointIT.java
+++ b/quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions-integration-test/src/test/java/org/kie/kogito/quarkus/dmn/CustomEndpointIT.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.quarkus.dmn;
+
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.junit.QuarkusTest;
+import io.restassured.http.ContentType;
+
+import static io.restassured.RestAssured.given;
+
+@QuarkusTest
+public class CustomEndpointIT {
+
+    @Test
+    public void testHelloEndpoint() {
+        given()
+                .contentType(ContentType.JSON)
+                .accept(ContentType.JSON)
+                .body(DMNIT.ADULT_PAYLOAD)
+                .post("/dmnModel")
+                .then()
+                .statusCode(200)
+                .body("d.Hello", Matchers.is("Hello Luca"));
+    }
+
+}
--- a/api/kogito-api-incubation-decisions-services/pom.xml
+++ b/api/kogito-api-incubation-decisions-services/pom.xml
@@ -1,0 +1,106 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2021 Red Hat, Inc. and/or its affiliates.
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~       http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.kie.kogito</groupId>
+        <artifactId>kogito-api-parent</artifactId>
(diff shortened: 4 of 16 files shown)


tarilabs:
```
Examples Integration Tests / testProcess – org.acme.travel.MultiMessagingIT
13s
Stacktrace
org.opentest4j.AssertionFailedError: expected: <0> but was: <3>
	at org.acme.travel.MultiMessagingIT.testProcess(MultiMessagingIT.java:95)
```

Jenkins rerun examples tests

tarilabs (commented):
I've focused my attention to commits b7b6caa and ef9c5e3 as asked, on separate tabs.
Overall looks ok to me and thank you for taking care of this baseline implementation! 👍  good job!!

I believe some points are still open/not addressed, and while I don't believe they _need_ to be addressed for the merge, I believe worthy of being tracked.

A general comment is that with this review, I realised that possibly I misunderstood in part the scope of this.
What is called here "Kogito API", can otherwise be more precisely called "Kogito _runtime_ API", and being based on the (common) SPI for runtime services you designed.
That is, it pertains to _Runtime_ for evaluation/execution, and it seems to me it is not in the scope of this SPI/API:

- configuration
- build/"compile"

Another general comment as you know already, this is being marked as `incubation`, but is not clear what is the roadmap to eventually move out of incubation.

Also, please consider that more in general -as you rightly point out- in the overall scheme of naming these services. At a point, the leads decided it was prohibited to use technologies/standards names to avoid "historical bias" and to keep the design open to multiple implementations.
For `Processes`, I wonder, was this choice effective to support both BPMN2 and SW and the users are using it interchangeably?
I believe this time is an opportunity to evaluate the results of that choice; if it played well for `Processes` I'm happy to carry on with a general `Decisions` for DMN, otherwise I believe it is worthy to reconsider _that_ choice.

As you rightly pointed out already again, this specifically for DMN does not allow:

* specify Listener add/remove
* specify metadata 
  * the context in/out metadata (used by TrustyAI)
  * the result metadata of DMN (used for DMN, was decision x evaluated?)

For this latter point please ref specific comment.

As we agreed, I've noted all the comments in this PR page, but I reiterate to you that I remain available for quickcalls etc to discuss as needed.
For my perspective you can merge this PR; while the points raised are worthy being tracked.

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/InternalObjectMapper.java
```diff
@@ -0,0 +1,47 @@
...
+import com.fasterxml.jackson.annotation.PropertyAccessor;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * For internal use only.
+ * Provides a method to convert an object into a given type.
+ * This is an implementation detail. We may move this to a separate module in the future.
+ */
+public class InternalObjectMapper {
+    private static final ObjectMapper objectMapper = new ObjectMapper();
```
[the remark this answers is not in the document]
evacchi: memo for self: in the other PR ( https://github.com/kiegroup/kogito-runtimes/pull/1546#issuecomment-910210189 ) I was noticing that there is a mechanism to declaratively define "serializers" I wonder if there is a similar mechanism for registering modules on a per-type mechanism so that the different configuration do not "conflict" with each other 🤔 

more generally: I definitely agree that we must make this behavior pluggable

api/kogito-api-incubation-decisions-services/src/main/java/org/kie/kogito/incubation/decisions/services/DecisionService.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package org.kie.kogito.incubation.decisions.services;
+
+import org.kie.kogito.incubation.common.DataContext;
+import org.kie.kogito.incubation.common.LocalId;
+
+public interface DecisionService {
+    DataContext evaluate(LocalId decisionId, DataContext inputContext);
```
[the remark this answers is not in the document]
evacchi: +1 we were discussing this with @gitgabrio as well, because the PMML has a similar "issue" where there is both a way to retrieve the "main" result, and get the "full" version (called "descriptive") that includes "meta-data" (so to speak).

I think that we can definitely allow to return a specific subtype; @gitgabrio was suggesting to make this even more "opinionated" in a way; i.e. have a separate service definition;

i.e.

```java
interface DescriptivePredictionService {
    PMMLDataContext  evaluate(Id, DataContext)
}
// vs
interface PredictionService {
    DataContext  evaluate(Id, DataContext)
}
```

(this would easily translate on DMN in the way you can imagine)

I like this idea but I fear it may lead to an explosion in terms of number of interfaces (i.e. in the case of DMN and PMML there is probably only 1 `evaluate(Id, ctx)` method (it's only one because ID may point to the model but also to a specific decision context); on the other hand, e.g. processes may have more "commands" (create/start/abort etc.)

I am still not sure about having the subtype in the context (but open to discuss) because the thinking here is that as a end user you should not care about what type is returned, since you can always "convert" it into the shape you expect;

e.g. `result.as(DMNResultContext.class);`

although this is redundant I find that this may be useful to stay opaque.

For instance, if we translate the context to a distributed setting, where you would get a serialized payload from a remote service and you may want to unserialize only "at some point". Putting "DMNResultContext" into the signature would mean that the serialization would have to occur "right there". If you return "DataContext" you are potentially moving not only data conversion but possibly serialization "inside" the `as` method.

I hope I made myself clear, but we can also have a 1:1 so that I can explain better the high-level idea :-)

evacchi:
> I've focused my attention to commits [b7b6caa](https://github.com/kiegroup/kogito-runtimes/commit/b7b6caa1ca79ef3cb5cacde8e10eb1186e68c630) and [ef9c5e3](https://github.com/kiegroup/kogito-runtimes/commit/ef9c5e377a04477966be4de6e36af3b95eb8d2da) as asked, on separate tabs.
> Overall looks ok to me and thank you for taking care of this baseline implementation! 👍 good job!!

thanks, very appreciated :)

> 
> I believe some points are still open/not addressed, and while I don't believe they _need_ to be addressed for the merge, I believe worthy of being tracked.
> 
> A general comment is that with this review, I realised that possibly I misunderstood in part the scope of this.
> What is called here "Kogito API", can otherwise be more precisely called "Kogito _runtime_ API", and being based on the (common) SPI for runtime services you designed.
> That is, it pertains to _Runtime_ for evaluation/execution, and it seems to me it is not in the scope of this SPI/API:
> 
> * configuration
> * build/"compile"

Correct. Build+compile, for the time being is addressed through Quarkus extensions and generic codegen (which also works on Spring). This does not prevent us from designing a mechanism for building and compiling that works independently from those, but it is indeed a separate concern.

Codegen "APIs" are being investigated as a separate effort. In that effort we will probably revisit the concept of build phases (validation, etc.)

As for configuration, this is currently out of scope; the goal for this "Runtime API" was to deliver something quick that would bring immediate value to end users, i.e. writing custom REST endpoints.

> 
> Another general comment as you know already, this is being marked as `incubation`, but is not clear what is the roadmap to eventually move out of incubation.

Right. We are working on a roadmap, and will share ASAP.

> 
> Also, please consider that more in general -as you rightly point out- in the overall scheme of naming these services. At a point, the leads decided it was prohibited to use technologies/standards names to avoid "historical bias" and to keep the design open to multiple implementations.
> For `Processes`, I wonder, was this choice effective to support both BPMN2 and SW and the users are using it interchangeably?
> I believe this time is an opportunity to evaluate the results of that choice; if it played well for `Processes` I'm happy to carry on with a general `Decisions` for DMN, otherwise I believe it is worthy to reconsider _that_ choice.

I agree, this is the right time to reconsider. My main concern is about consistency. I'd prefer that each component either was named after a *standard*, or named after a *capability*.

However, this API was precisely designed with *flexibility* in mind: i.e., each component may "choose their own adventure" :-)

So even if we won't reach an agreement, if you deem that "DMN" is the right name for this component, we will call it "kogito-api-dmn-*" etc. 

Please notice that because these service APIs are quite tiny and lightweight, each component may even provide more than one (e.g. `BpmnService`, `SwService`, but also `ProcessService`). However we still have to keep in mind that the other goal is to possibly provide "async" and "distributed" equivalent where it makes sense. So there are at least 2 coordinates:

|                | ProcessService | BpmnService | SwService |
|----------------|----------------|-------------|-----------|
| sync           |        x       |      x      |      x    |  
| async          |        ?       |      ?      |      x    |
| message-based  |        ?       |      ?      |      ?    |
| .......        |    ........    |   ........  |  ........ |


this may easily become unwieldly if a service provides more methods (e.g. in the case of stateful processes, we won't have just an `evaluate()` method, but soon `create(), start(), abort()` etc. 


> 
> As you rightly pointed out already again, this specifically for DMN does not allow:
> 
> * specify Listener add/remove
> * specify metadata
>   
>   * the context in/out metadata (used by TrustyAI)

Good point. Only partially aware of this concern, will have to learn more. 

>   * the result metadata of DMN (used for DMN, was decision x evaluated?)

In both cases it's definitely necessary to allow extra-metadata to be passed and retrieve. I don't exclude we end up realizing an equivalent of the DMNContext/DMNResultContext -- I just want to come up with a design that is shared across the board, and thus satisfies all constraints. Current design is the most limited, but it is then only useful for limited use cases.

> 
> For this latter point please ref specific comment.
> 
> As we agreed, I've noted all the comments in this PR page, but I reiterate to you that I remain available for quickcalls etc to discuss as needed.
> For my perspective you can merge this PR; while the points raised are worthy being tracked.


Great. Will create more JIRAs.

tarilabs (approved):
approving per https://github.com/kiegroup/kogito-runtimes/pull/1550#pullrequestreview-745487051 , per requested, and following the upstream PR merged which happened in the meantime

evacchi:
Jenkins rerun examples tests

api/kogito-api-incubation-decisions/src/main/java/org/kie/kogito/incubation/decisions/DecisionIds.java
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.kogito.incubation.decisions;
+
+import org.kie.kogito.incubation.common.ComponentRoot;
+
+public class DecisionIds implements ComponentRoot {
```
radtriste, line 21: reason why it is not called `LocalDecisionIds` and `LocalDecisionServiceIds` ?

quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions-integration-test/src/test/java/org/kie/kogito/quarkus/dmn/CustomEndpointIT.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.junit.QuarkusTest;
+import io.restassured.http.ContentType;
+
+import static io.restassured.RestAssured.given;
+
+@QuarkusTest
```
radtriste, line 27: native ?

api/kogito-api-incubation-decisions/src/main/java/org/kie/kogito/incubation/decisions/DecisionIds.java
evacchi: it's for the fluent API. This is usually called with `appRoot.get(DecisionIds.class).get(...)` and will return a `LocalDecisionId`. TBF I am not 100% satisfied with naming conventions but I would like to align them to be more consistent. (now that I think about it, I think `ProcessId` should be `LocalProcessId`)

evacchi: https://issues.redhat.com/browse/KOGITO-5925

quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions-integration-test/src/test/java/org/kie/kogito/quarkus/dmn/CustomEndpointIT.java
evacchi: https://issues.redhat.com/browse/KOGITO-5924

radtriste (approved):

PR #1574, opened 2021-09-01
KOGITO-5808 - Load files using classloader

cristianonicolai:
https://issues.redhat.com/browse/KOGITO-5808

### DIFF
--- a/kogito-codegen-modules/kogito-codegen-integration-tests/src/test/java/org/kie/kogito/codegen/process/ProcessGenerationIT.java
+++ b/kogito-codegen-modules/kogito-codegen-integration-tests/src/test/java/org/kie/kogito/codegen/process/ProcessGenerationIT.java
@@ -15,11 +15,9 @@
  */
 package org.kie.kogito.codegen.process;
 
-import java.io.File;
-import java.io.IOException;
+import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -28,9 +26,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.Set;
 import java.util.function.BiConsumer;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.assertj.core.api.Assertions;
@@ -58,6 +54,7 @@ import org.jbpm.workflow.core.node.StartNode;
 import org.jbpm.workflow.core.node.StateBasedNode;
 import org.jbpm.workflow.core.node.Trigger;
 import org.jbpm.workflow.core.node.WorkItemNode;
+import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.MethodSource;
@@ -69,6 +66,7 @@ import org.kie.kogito.codegen.AbstractCodegenIT;
 import org.kie.kogito.process.Processes;
 import org.kie.kogito.process.impl.AbstractProcess;
 
+import static java.util.Collections.singleton;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.jbpm.ruleflow.core.Metadata.ACTION;
 import static org.jbpm.ruleflow.core.Metadata.TRIGGER_REF;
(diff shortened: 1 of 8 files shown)


ricardozanini (approved):

kogito-codegen-modules/kogito-codegen-integration-tests/src/test/java/org/kie/kogito/codegen/process/ProcessGenerationIT.java
```diff
@@ -124,7 +123,10 @@ public void testProcessGeneration(String processFile) throws Exception {
         // for some tests this needs to be set to true
         System.setProperty("jbpm.enable.multi.con", "true");
         List<org.kie.api.definition.process.Process> processes = ProcessCodegen.parseProcesses(Stream.of(processFile)
-                .map(resource -> new File(BASE_PATH.toString(), resource))
+                .map(resource -> {
+                    String file = Thread.currentThread().getContextClassLoader().getResource(resource).getFile();
```
jstastny-cz, line 127: I think we need to go further and use `getResourceAsStream` and storing to some accessible location (like a tmp file or sth). Otherwise we're still working with path within the maven repository (or where the JAR resides), similar to `file:///home/jstastny/.m2/repository/org/kie/kogito/kogito-codegen-integration-tests/2.0.0-SNAPSHOT/kogito-codegen-integration-tests-2.0.0-SNAPSHOT-tests.jar!/links/MultipleCatchLinkProcess.bpmn2`

ricardozanini: Why? What's the motivation for this change in the first place, and what's the use case we are talking about here? @cristianonicolai, can you update the PR with the JIRA and maybe adding more context? Based on the @jstastny-cz comment, I believe there's more under the surface I'm not aware of.

jstastny-cz: Basically, in quarkus-platform the tests are being run for test-jars. This requires writing tests to load resources from classpath instead of relying on filesystem location (There's no src/test/resources when running tests using test-jar).

jstastny-cz: I've made a WIP (It took me some time and it's brute force) - This might be the way to go, ProcessGenerationIT works now in both cases.
https://github.com/jstastny-cz/kogito-runtimes/commit/59db9d0a2687f33d145ef48c3d3afb0b74cef087

cristianonicolai: @jstastny-cz can you try this new version, I got rid of the src/test/resources part, maybe that's enough?

jstastny-cz: We probably need to use getResourceAsStream, otherwise, with getResource() we're getting path pointing at the JAR entry, which causes error further more:
```
Caused by: java.io.FileNotFoundException: file:/home/jstastny/.m2/repository/org/kie/kogito/kogito-codegen-integration-tests/2.0.0-SNAPSHOT/kogito-codegen-integration-tests-2.0.0-SNAPSHOT-tests.jar!/links/MultipleCatchLinkProcess.bpmn2 (No such file or directory)
```

jstastny-cz: I am getting almost there
```
Tests run: 307, Failures: 1, Errors: 1, Skipped: 0
```

The approach is to copy the resources before tests into a filesystem location to mimic the original state. But I'm failing in case of doing so with .java class (obviously), the solution there would be to duplicate the class in src/test/resources too :/ .

jstastny-cz: I have now two approaches, both rely on copying file somewhere.
1. In case of test-jar gets resources from jar and extract them to src/test/resources: https://github.com/cristianonicolai/kogito-runtimes/compare/KOGITO-5808...jstastny-cz:KOGITO-5808-WIP2?expand=1
2. In both cases take resources from jar and extract into target/test-classes:
https://github.com/cristianonicolai/kogito-runtimes/compare/KOGITO-5808...jstastny-cz:KOGITO-5808-WIP3?expand=1

I probably like option 2. more, as it does not pollute other dir than target. 

I don't see any nicer way as the tests and api is heavily File focused (no inputstreams options etc).

jstastny-cz (commented):
A few inline comments, generally about the need to keep Paths as relative all the time and later readResourceAsStream using classloader.

ProcessCloudEventMetaFactoryGeneratorTest fails due to the index starting with `/` as in `"/messageevent/IntermediateCatchEventMessage.bpmn2"`, so we need `"messageevent/IntermediateCatchEventMessage.bpmn2"`.

Plus to get ProcessGenerationIT working, we need to build sort of an file index listing all the resources, I did it in one of my branches using ant task in pom.xml which also makes sure it's inserted into resulting test-jar (and can be read from classloader):
```
    <build>
        <plugins>
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-antrun-plugin</artifactId>
                <executions>
                    <execution>
                        <goals>
                            <goal>run</goal>
                        </goals>
                        <phase>generate-test-resources</phase>
                        <configuration>
                            <target>
                                <fileset id="dist.contents" dir="src/test/resources" includes="**/*"/>
                                <property name="prop.dist.contents" refid="dist.contents"/>
                                <loadresource property="lines">
                                    <propertyresource name="prop.dist.contents"/>
                                    <filterchain>
                                        <tokenfilter>
                                            <filetokenizer/>
                                            <replacestring from=";" to="${line.separator}"/>
                                        </tokenfilter>
                                    </filterchain>
                                </loadresource>
                                <echo file="${project.build.directory}/test-classes/test-resource-index.txt">${lines}
                                </echo>
                            </target>
                        </configuration>
                    </execution>
                </executions>
            </plugin>
        </plugins>
    </build>
```
and then processProvider method in ProcessGenerationIT would look like:
```
    static Stream<Path> processesProvider() throws Exception {
        List<String> ignoredFiles = readLinesFromResource("org/kie/kogito/codegen/process/process-generation-test.skip.txt");
        List<String> testResources = readLinesFromResource("test-resource-index.txt");
        return testResources.stream()
                .filter(it -> ProcessCodegen.SUPPORTED_BPMN_EXTENSIONS.stream().anyMatch(ext -> it.endsWith(ext))
                        || ProcessCodegen.SUPPORTED_SW_EXTENSIONS.keySet().stream().anyMatch(ext -> it.endsWith(ext)))
                .filter(p -> ignoredFiles.stream().noneMatch(ignored -> p.startsWith(ignored))).map(Path::of);
    }
```

With all this, I end up with a few broken tests (regardless of whether run from filesystem or test-jar, so this needs investigation, I am getting stack traces from deep within the implementing code and I can't assess what's a valid refactor and what not).
```
[INFO] Results:
[INFO] 
[ERROR] Errors: 
[ERROR]   AnnotatedRuleUnitCompilerIT.testAnnotatedRuleUnit:33->AbstractCodegenIT.generateRulesFromJava:165->AbstractCodegenIT.generateCode:181->AbstractCodegenIT.lambda$static$3:134->AbstractCodegenIT.toCollectedResources:140 » IllegalArgument
[ERROR]   BusinessRuleTaskIT.testDecision:147->AbstractCodegenIT.generateCode:173->AbstractCodegenIT.generateCode:185 » UncheckedIO
[INFO] 
[ERROR] Tests run: 307, Failures: 0, Errors: 2, Skipped: 0
[INFO] 
[INFO] 
[INFO] --- maven-failsafe-plugin:3.0.0-M5:verify (integration-test) @ kogito-codegen-integration-tests ---
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:34 min
[INFO] Finished at: 2021-09-03T12:05:27+02:00
[INFO] ------------------------------------------------------------------------
```

PR #1547, opened 2021-08-23
KOGITO-5748 Public API: Straight-Through Process Modules (Incubation)

evacchi:
Blocked by:
- [ ] #1546 

- see #1546 for an architectural overview

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* <b>Run all builds</b>  
  Please add comment: <b>Jenkins retest this</b>

* <b>Run (or rerun) specific test(s)</b>  
  Please add comment: <b>Jenkins (re)run [runtimes|optaplanner|apps|examples] tests</b>
 
* <b>Quarkus LTS checks</b>  
  Please add comment: <b>Jenkins run LTS</b>

* <b>Run (or rerun) LTS specific test(s)</b>  
  Please add comment: <b>Jenkins (re)run [runtimes|optaplanner|apps|examples] LTS</b>

* <b>Native checks</b>  
  Please add comment: <b>Jenkins run native</b>

* <b>Run (or rerun) native specific test(s)</b>  
  Please add comment: <b>Jenkins (re)run [runtimes|optaplanner|apps|examples] native</b>

* <b>Full Kogito testing</b> (with cloud images and operator BDD testing)  
  Please add comment: <b>Jenkins run BDD</b>  
  <b>This check should be used only if a big change is done as it takes time to run, need resources and one full BDD tests check can be done at a time ...</b>
</details>

PR #1575, opened 2021-09-01
[KOGITO-5394] WorkItem retry implementation 

fjtirado:
This JIRA can be split in two parts

- First one is a carry forward from v7 to support retries in ProcessWorkItemHandlerException, which just involve ```workItemNodeInstance```
- Second one is to change the engine to support createProcessInstance of subprocess for error handling in Kogito (by hacking  ```DummyKnowledgeImplementation```).  This second one is tricky. I just add a Map to be able to access process definition from process id (the InternalProcessRuntime  instance can only start same type of process than the catching one), but I feel we should refactor the engine core in next release. There are too many layers that might be simplified

### DIFF
--- a/jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/DummyKnowledgeRuntime.java
+++ b/jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/DummyKnowledgeRuntime.java
@@ -23,7 +23,6 @@ import org.drools.core.common.InternalAgenda;
 import org.drools.core.common.InternalKnowledgeRuntime;
 import org.drools.core.common.WorkingMemoryAction;
 import org.drools.core.impl.EnvironmentImpl;
-import org.drools.core.runtime.process.InternalProcessRuntime;
 import org.drools.core.time.TimerService;
 import org.jbpm.workflow.instance.impl.CodegenNodeInstanceFactoryRegistry;
 import org.kie.api.KieBase;
@@ -103,7 +102,7 @@ class DummyKnowledgeRuntime implements InternalKnowledgeRuntime, KogitoProcessRu
 
     @Override
     public KogitoProcessEventSupport getProcessEventSupport() {
-        return ((org.jbpm.process.instance.InternalProcessRuntime) processRuntime).getProcessEventSupport();
+        return processRuntime.getProcessEventSupport();
     }
 
     @Override
@@ -116,6 +115,7 @@ class DummyKnowledgeRuntime implements InternalKnowledgeRuntime, KogitoProcessRu
         return environment;
     }
 
+    @Override
     public JobsService getJobsService() {
         return null;
     }
@@ -273,7 +273,7 @@ class DummyKnowledgeRuntime implements InternalKnowledgeRuntime, KogitoProcessRu
 
     @Override
     public KogitoProcessInstance createProcessInstance(String processId, Map<String, Object> parameters) {
-        return null;
+        return (KogitoProcessInstance) processRuntime.createProcessInstance(processId, null, parameters);
     }
 
     @Override
@@ -282,10 +282,12 @@ class DummyKnowledgeRuntime implements InternalKnowledgeRuntime, KogitoProcessRu
 
     }
 
+    @Override
     public KogitoProcessInstance startProcessInstance(String processInstanceId) {
         return null;
     }
 
+    @Override
     public KogitoProcessInstance startProcessInstance(String processInstanceId, String trigger) {
         return null;
     }
@@ -301,6 +303,7 @@ class DummyKnowledgeRuntime implements InternalKnowledgeRuntime, KogitoProcessRu
 
     }
 
+    @Override
     public void signalEvent(String type, Object event, String processInstanceId) {
 
     }
@@ -338,14 +341,17 @@ class DummyKnowledgeRuntime implements InternalKnowledgeRuntime, KogitoProcessRu
         return (WorkItemManager) getKogitoWorkItemManager();
     }
 
+    @Override
     public KogitoProcessInstance getProcessInstance(String processInstanceId) {
         return null;
     }
 
+    @Override
     public KogitoProcessInstance getProcessInstance(String processInstanceId, boolean readonly) {
         return null;
     }
 
+    @Override
     public void abortProcessInstance(String processInstanceId) {
 
     }
--- a/jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/LightProcessRuntimeContext.java
+++ b/jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/LightProcessRuntimeContext.java
@@ -16,24 +16,28 @@
 package org.jbpm.process.instance;
 
 import java.util.Collection;
-import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
 import org.drools.core.common.WorkingMemoryAction;
 import org.jbpm.process.core.ContextContainer;
+import org.jbpm.process.core.ProcessSupplier;
 import org.jbpm.process.core.context.variable.VariableScope;
 import org.jbpm.process.instance.context.variable.VariableScopeInstance;
 import org.jbpm.ruleflow.instance.RuleFlowProcessInstance;
 import org.kie.api.definition.process.Process;
 import org.kie.api.event.rule.DefaultAgendaEventListener;
 import org.kie.internal.process.CorrelationKey;
+import org.kie.kogito.Application;
+import org.kie.kogito.process.Processes;
 
 public class LightProcessRuntimeContext implements ProcessRuntimeContext {
 
-    private final List<Process> processes;
+    private Processes allProcesses;
+    private Collection<Process> processes;
 
-    public LightProcessRuntimeContext(List<Process> processes) {
+    public LightProcessRuntimeContext(Application app, Collection<Process> processes) {
+        this.allProcesses = app != null ? app.get(Processes.class) : null;
         this.processes = processes;
     }
 
@@ -44,7 +48,11 @@ public class LightProcessRuntimeContext implements ProcessRuntimeContext {
 
     @Override
     public Optional<Process> findProcess(String id) {
-        return processes.stream().filter(p -> p.getId().equals(id)).findFirst();
+        Optional<Process> result = processes.stream().filter(p -> p.getId().equals(id)).findAny();
+        if (result.isEmpty() && allProcesses != null) {
+            result = Optional.ofNullable(((ProcessSupplier) allProcesses.processById(id)).get());
+        }
+        return result;
     }
 
     @Override
--- a/jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcess.java
+++ b/jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcess.java
@@ -19,12 +19,14 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
+import org.jbpm.process.core.ProcessSupplier;
 import org.jbpm.process.core.timer.DateTimeUtils;
 import org.jbpm.process.core.timer.Timer;
 import org.jbpm.process.instance.InternalProcessRuntime;
 import org.jbpm.process.instance.LightProcessRuntime;
-import org.jbpm.process.instance.LightProcessRuntimeContext;
 import org.jbpm.process.instance.LightProcessRuntimeServiceProvider;
 import org.jbpm.process.instance.ProcessRuntimeServiceProvider;
 import org.jbpm.workflow.core.impl.WorkflowProcessImpl;
@@ -51,7 +53,7 @@ import org.kie.kogito.process.ProcessInstancesFactory;
 import org.kie.kogito.process.Signal;
 
 @SuppressWarnings("unchecked")
-public abstract class AbstractProcess<T extends Model> implements Process<T> {
+public abstract class AbstractProcess<T extends Model> implements Process<T>, ProcessSupplier {
 
     protected final ProcessRuntimeServiceProvider services;
     protected ProcessInstancesFactory processInstancesFactory;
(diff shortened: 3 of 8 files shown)


ricardozanini:
> but I feel we should refactor the engine core in next release. There are too many layers that might be simplified

Agreed. Do you mind opening a JIRA? Also, I believe it's a great addition to our tech debt grooming.

ricardozanini (approved):

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/DummyKnowledgeRuntime.java
```diff
@@ -273,7 +274,9 @@ public ProcessInstance startProcessFromNodeIds(String s, Map<String, Object> map
 
     @Override
     public KogitoProcessInstance createProcessInstance(String processId, Map<String, Object> parameters) {
-        return null;
+
+        org.kie.api.definition.process.Process process = AbstractProcess.findById(processId).orElseThrow(() -> new IllegalArgumentException("Cannot find process " + processId));
```
tiagodolphine, line 278: I personally don't think using AbstractProcess and with a static method to find the process by id is the best option, why not injecting the `ProcessRuntimeContext` in here and from there you can do `processRuntimeContext.findProcess(id)`, or even injecting any other class that already holds the processes. Or Am I missing some point?

tiagodolphine (commented):
Take a look at the comments I did, my main concern is about holding processes in the AbstractProcess class and to be accessed in a static way.

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcess.java
```diff
@@ -64,6 +70,12 @@
     protected List<String> startTimerInstances = new ArrayList<>();
     protected KogitoProcessRuntime processRuntime;
 
+    private static Map<String, org.kie.api.definition.process.Process> processes = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
fjtirado: org.kie.kogito.process.Process hold in Processes is not the same and is not in the same hierarchy than  org.kie.api.definition.process.Process,
Reason Im putting the Map in AbstractProcess is because this is the place where org.kie.api.definition.process.Process is generated by the children generated classes

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/DummyKnowledgeRuntime.java
fjtirado: I agree this is far from ideal (thats why Im proposing a refactor of the engine in next release) but If I implemented this hack  (which has the side benefit of generating the proces definition just once, and not in every create process instance as is currently happening) is because I do not see any other better alternative
Injection is ruled out because this class is not in cdi (no class in the engine is on cdi)
The problem is that we are trying to start a subprocess for error handling using the KnowledgeRuntime (we have a curious relation with ProcessRuntime, since it is created by it, but it also holds a reference to it) of the parent, this uses a  org.jbpm.process.instance.InternalProcessRutime, which is created by AbstractProcess and internally creates a ProcessRuntimeContext aassociates just to the org.kie.api.definition.process.Process the  AbstractProcess child defines. 
So that InternalProcessRuntime will fail. Hece I add a new method to InternalProcessRuntime than rather that accepting the process id accepts the process, so I can skip the check. But still I need some places to obtaing the Process definition from, and that place has to be AbstractProcess. 
So, you can say, why do not obtain a DummyKnowledge for the subprocess. Well, thats currently imposible because DummyKnowledge needs to point to a ProcessRuntime, hence it only exist associated to a process instance. And here we are trying to create one. 
The other approach will be to use ProcessService, but this belong to service layer, not flow, so it will be even worse. 
Current approach remains isolated in flow (AbstractProcess being present in flow is not accident) and it has the benefit of creating org.kie.api.definition.process.Process instance just once for all create process instances calls.

fjtirado:
> > but I feel we should refactor the engine core in next release. There are too many layers that might be simplified
> 
> Agreed. Do you mind opening a JIRA? Also, I believe it's a great addition to our tech debt grooming.

I have opened a point for discussion here https://docs.google.com/document/d/17N35EshgAlDQ3nT_39Zd9w1_MFsdZnaNkvlG05q2WrY/edit#

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/DummyKnowledgeRuntime.java
tiagodolphine: to be honest, I'm not sure yet that AbstractProcess is the place to keep/find processes definitions and mainly accessed as a public static method, IMO we should use a class that has this specific role, like https://github.com/kiegroup/kogito-runtimes/blob/f855f1d0d21e1f6c0fa1d4bfcf289fe5cb0956ec/api/kogito-api/src/main/java/org/kie/kogito/process/Processes.java that has implementations done during codegen, using DI to find for all processes and keep them there.
^ @cristianonicolai @danielezonca are you aware of any other possibility?

tiagodolphine: @evacchi pinging you because in the new API we kind of have to find the processes definitions as well and AFAIK we use the `Processes` class in the QuarkusStraightThroughProcessService, right?

evacchi: yes, although that's an implementation detail. TBF in the new API you won't have to "find" the process ID. You will just "send a message" to that ID. So in this case I imagine something along the lines of:

```
Id instanceId = processSvc.createInstance(id);
processSvc.startInstance(instanceId);
``` 

etc. this way we will no longer have to expose these routines to "find" processes; either it's there or it is not there. see #1546 #1547

evacchi: as for the problem at hand, can't you find a way to (for instance) pass the `Processes` instance to the Dummy runtime instead? a static method with a static concurrent hashmap... that's worrisome

fjtirado: @evacchi  Thats the issue, if you look deep enough in the code,  the only place where  org.kie.api.definition.process.Process (do not mistake with org.kie.kogito.process.Process ) can be obtainer is AbstractProcess. There, when process instance is created from REST (not from a parent process), the InternalRuntime and DummyKnowledge are created with the process definition. 

Ill change the approach and do not add the map to AbstractProcess, but define a BPMNProcesses class that is populated together with Processes one

Problem, though, is that org.kie.kogito.process.Process interface does not have a method to retrieve org.kie.api.definition.process.Process, so either I cast to AbstractProcess or changes an API interface exposing a detail. I hope now you see why I do this in AbstractProcess, keeping it as an implementation detail and ask for a core engine refactor.

fjtirado: @tiagodolphine Map moved to a new class (which I do not like is that it is populated from AbstractProcess, but I do not see an alternative to that without changing Kogito process interface or do a ugly castins in Processes)

fjtirado: In order to take benefit of the introduction of the new ProcessDefinition class, I fixed what for me is a conceptual mistake in ProcessRuntime  relation with AbstractProcess, which was instantiating the process definition every time a new process instance is created. Now,   LightPRocessRuntime does not longer expect the process in the constructor, but uses indirectly the ProcessDefinitions new class, through  ProcessRuntimeContext instance. DummyKnowledgeRuntime just invoke ProcessRuntime to create the new instance. 
In future we should (by changing generator):
- Make Processes implementation not generated (there is nothing really generated there)
- Populate ProcessDefintions from Processes or use Processes in ProcessDefinitions. 
- Instantiate process() in AbstractProcess just one

fjtirado: As per Tiago suggestion, I tries by passing Application to LightProcessRuntime. That way we avoid the additional Map (at the cost of polluting bpmn2 test by passing null and also check for null in the context code)

cristianonicolai (approved):

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcess.java
```diff
@@ -201,10 +203,18 @@ protected ExpirationTime configureTimerInstance(Timer timer) {
         }
     }
 
+    @Override
+    public org.kie.api.definition.process.Process get() {
+        if (process == null) {
```
tiagodolphine, line 208: Just thinking about concurreny, acessing the process at the same time, this is not thread safe.
One option would be to holp the process in a AtomicReference, instead of a process itself.... like

```
private final AtomicReference<Process> process = new AtomicReference<>();
...
process.compareAndSet(null, process());
...
process.get();

```
anyway just and idea.

tiagodolphine (commented):
Nice, thanks for the changes, it is better with this approach IMO, I just made one comment, take a look to see if it makes sense.

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcess.java
fjtirado: hmmm, you are right (I was wrongly assumed this is going to be invoked from just one thread at a time, but this is not longer true), but Im afraid the solution is our old syncrhonized friend  (to avoid double initializtion, compareAndSet will invoke process even if later it discard it)

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/LightProcessRuntimeContext.java
```diff
@@ -16,24 +16,28 @@
...
+import org.kie.kogito.process.Processes;
 
 public class LightProcessRuntimeContext implements ProcessRuntimeContext {
 
-    private final List<Process> processes;
+    private Processes allProcesses;
+    private Collection<Process> processes;
 
-    public LightProcessRuntimeContext(List<Process> processes) {
+    public LightProcessRuntimeContext(Application app, Collection<Process> processes) {
```
danielezonca, line 39: In the future I expect `processes` to be removed: application "contains" all resources so it should not be necessary to pass other variables, btw this probably requires to change codegen and rewrite many tests too so it is way out of scope for this PR.
My comment is mainly to be sure we are on the same page.
Wdyt?

danielezonca (commented):
Overall looks good to me, good refactor, just some minor comments 👍

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcess.java
```diff
@@ -64,6 +64,8 @@
     protected List<String> startTimerInstances = new ArrayList<>();
     protected KogitoProcessRuntime processRuntime;
 
+    private org.kie.api.definition.process.Process process;
```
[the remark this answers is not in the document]
fjtirado: The real impl of process() is pretty heavy and inmutable. It basically creates the process defintion from scrath. Heavy and inmutable is a good candidate for lazy singleton initialization. 
I thought of changing process() to protected with this JIRA, but it is used in a lot of places, so I just changed the calls to process inside this class.

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/LightProcessRuntimeContext.java
fjtirado: yes, we are on the same page. The reason I leave processes (which as far as I checked is just the process that creates the processruntimecontext) is to minimize the impact of this JIRA. Because processes is used for initialize listeners by processrunitme, so I prefer to keep what we have. 
In future, I  envision ProcessRuntimeContext as a singleton which access all processes deployed by application, but this is part of the refactor discussion for grooming.

fjtirado:
jenkins retest this

fjtirado:
jenkins run examples tests

fjtirado:
jenkins run examples tests

fjtirado:
jenkins run examples tests

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcess.java
tiagodolphine: personally, I'm not a fan of synchronized main in the method, as I said the AtomicReference should be an option, like checking if `.get()` is null before the `.compareAndSet`, or even using the `ReentrantLock`, but anyway, this is the impl detail.

tiagodolphine (approved):
LGTM, and thanks for applying the usage of `Processes` to avoid adding a new map.

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcess.java
fjtirado: yes, Im not fan of synchronized either, changed to specific lock

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcess.java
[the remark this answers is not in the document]
danielezonca: Let's review process() in another ticket, can you create it?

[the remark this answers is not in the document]
fjtirado: The whole process code generator is based on the assumption that process() return kie process definitions. 
It is fine for process to return a new definition each time (think of process as a factory method),  I think the problem is that process() should not have been public, but protected and that we should have had a public method like the one I added with this JIRA (that ensures process() protected method is just called once) 
If I open a JIRA, it will be to convert process() into protected and replace it by get()
PS: process() and get() names are changeables, but the fact that we need two methods (one protected and generated, another public and hardcoded) is pretty clear to me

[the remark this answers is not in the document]
danielezonca: Fine for me to make process() protected, my main point is to avoid different and equivalent available/public codepaths that can produce different results :)

cristianonicolai:
@fjtirado is there docs and release notes for this work?

PR #1603, opened 2021-09-14
KOGITO-5861: fix filter in EventDrivenDecisionController

kostola:
* [Jira task](https://issues.redhat.com/browse/KOGITO-5861)

## Description of the problem

A refactoring in `EventDrivenDecisionController` slightly changed the internal behavior removing the filter on CloudEvent type. This causes unwanted CloudEvents to be processed by the controller.

The pre-existing `EventDrivenDecisionControllerTest` can't identify the error, even if there is a test for that case, because it uses an internal private method that wasn't changed instead of the public API.

## Changes in this PR
* Restored the filter in `EventDrivenDecisionController`.
* Refactored `EventDrivenDecisionControllerTest` to test the whole loop: EventReceiver -> internal event processing -> EventEmitter.
* In order to refactor the test I also slightly refactored the event internals, unifying the convert + dispatch logic into `AbstractEventReceiver` that I could later use in the test.


fjtirado (changes requested):
Hi, 
As dicussed with @kostola, I feel the refactor of completablefuture handling both for quarkus and sprinboot might affect the peformance testing that we manually performed several weeks ago. So I will prefer to split this PR in two: one that just fix the type filter issue in EventDrivenDecisionControler and another one with the refactor of the completable future for the event receiver implementations. 
Let me elaborate a bit more. 
The reason the code is different from Quarkus and Sprinboot is because for Quarkus we commit messages indiviaully when the completable future is completed (thats what the callback is doing, commiting the message)   and in Sprinboot we let the framework commit the whole batch (and that why we need to wait for completable future to finish, hence the sequential treatment). 

The refactor wontwork for Quarkus because the returned completablefuture is not completed yet, so it will block delivery of next event till all async completable are done and that will kill performance of start process. 

The refactor wont work for SpringBoot because it will commit the records before the event has been dispatched by Kogito

quarkus/addons/messaging/common/src/main/java/org/kie/kogito/addon/quarkus/messaging/common/AbstractQuarkusCloudEventReceiver.java
```diff
@@ -52,27 +43,7 @@
...
-                future = future.thenCompose(f -> subscription.getConsumer().apply(object));
-            } catch (IOException e) {
-                LOGGER.info("Cannot convert to {} from {}, ignoring type {}, exception message is {}", subscription.getInfo().getOutputClass(), message,
-                        subscription.getInfo().getType(), e.getMessage());
-            }
-        }
-        if (callback != null) {
-            future.whenComplete(callback);
-        }
-        return result;
```
fjtirado, line 70: This is key for performance througput to return a completed future to smallrye so it allow next event in the poll to be processed. with this change we might potentially return a not completed future and that will block event processing, killint througput

springboot/addons/messaging/implementation/src/main/java/org/kie/kogito/addon/cloudevents/spring/SpringKafkaCloudEventReceiver.java
```diff
@@ -16,66 +16,42 @@
...
-                try {
-                    futures.add(consumer.getConsumer().apply(consumer.getInfo().getConverter().apply(message, consumer.getInfo().getOutputClass())));
-                } catch (IOException e) {
-                    log.info("Cannot convert to {} from {}, ignoring type {}, exception message is {}", consumer.getInfo().getOutputClass(), message, consumer.getInfo().getType(),
-                            e.getMessage());
-                }
-            }
-        }
-        // wait for this batch to complete
-        log.debug("Waiting for all operations in batch to complete");
```
fjtirado, line 71: we need to synchronoulsy wait here for two reasons:
1) let sprinboot automatically commit the batch
2) ensure order of polls is kept

kostola:
@fjtirado I reverted all the extra changes keeping only the minimum set required to fix the filter in `EventDrivenDecisionController` and update the corresponding test. Please take a look.

quarkus/addons/messaging/common/src/main/java/org/kie/kogito/addon/quarkus/messaging/common/AbstractQuarkusCloudEventReceiver.java
kostola: Thanks for the detailed explanation :pray:

springboot/addons/messaging/implementation/src/main/java/org/kie/kogito/addon/cloudevents/spring/SpringKafkaCloudEventReceiver.java
kostola: Thanks for the detailed explanation :pray:

fjtirado (approved):
LGTM!

danielezonca (approved):
LGTM, just a minor comment

addons/common/events/decisions/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
```diff
@@ -84,27 +84,21 @@ protected void setup() {
                 CloudEvent.class));
     }
 
-    void handleEvent(String event) {
-        CloudEventUtils.decode(event)
```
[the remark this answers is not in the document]
kostola: It's used quite a lot around..expecially in tests

api/kogito-events-api/src/main/java/org/kie/kogito/event/AbstractEventReceiver.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.function.Function;
+
+public abstract class AbstractEventReceiver implements EventReceiver {
+
+    @Override
+    public <T> void subscribe(Function<T, CompletionStage<Void>> consumer, SubscriptionInfo<String, T> info) {
+        subscriptions.add(new Subscription<>(consumer, info));
+    }
+
+    private final Collection<Subscription<?>> subscriptions = new CopyOnWriteArrayList<>();
```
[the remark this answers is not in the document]
kostola: It's gone :)

addons/common/events/decisions/src/main/java/org/kie/kogito/eventdriven/decision/EventDrivenDecisionController.java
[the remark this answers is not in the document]
danielezonca: Ok let's do it in another PR if necessary 👍

PR #1608, opened 2021-09-16
[KOGITO-5368] Refactor duplicated code for REST WIH and OpenAPI

fjtirado:
This PI will be a multi commit one;

1. WorkItemHandlerParamResolver has been promoted to first class citizen, so when the WorkItem includes a parameter which value is a WorkItemHandlerParamREsolver it automatically resolve its value
2. Working on merging  JSonNodePArameterResolver and JsonNodeResultHander logic with JsonPathResolver and DefaultRestWorkItemHandlerResult  and removing two of the four.


fjtirado:
jenkins run examples tests

fjtirado:
jenkins run runtimes tests

jbpm/process-workitems/src/main/java/org/kie/kogito/process/workitems/impl/KogitoWorkItemImpl.java
```diff
@@ -270,4 +269,262 @@ public void setNodeInstance(KogitoNodeInstance nodeInstance) {
...
+        }
+
+        @Override
+        public void remove() {
+            iter.remove();
+        }
+    }
+
+    private <T> Object processValue(T obj) {
+        return obj instanceof WorkItemParamResolver ? ((WorkItemParamResolver) obj).apply(this) : obj;
```
ricardozanini, line 527: Do we need to implement the entire `Map`  interface just because of this method?

kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/handlers/CompositeContextNodeHandler.java
```diff
@@ -192,51 +193,44 @@ protected CompositeContextNodeHandler(S state, Workflow workflow, RuleFlowNodeCo
...
         }
+        return map;
     }
 
-    private Object processWorkItemValue(JsonNode jsonNode, String paramName) {
+    private static Object processValue(JsonNode jsonNode) {
         if (jsonNode.isTextual()) {
-            String str = jsonNode.asText();
-            return isJsonPathExpr(str) ? new JsonPathExprSupplier(str, paramName) : str;
+            return jsonNode.asText();
```
ricardozanini, line 232: Are you evaluating the expression in another path? I remember that this check was here because we need to verify whether it will be an expression evaluation. Nevertheless, I liked this approach better if the evaluation of the value is in another path.

kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/handlers/CompositeContextNodeHandler.java
```diff
@@ -250,8 +244,15 @@ private Object processWorkItemValue(JsonNode jsonNode, String paramName) {
...
 
-    private boolean isJsonPathExpr(String str) {
-        return str.trim().startsWith("$");
+    private void processArgs(WorkItemNodeFactory<CompositeContextNodeFactory<P>> workItemFactory,
+            JsonNode functionArgs, String paramName, Class<? extends AbstractJsonPathResolver> clazz) {
+
+        Map<String, Object> map = functionsToMap(functionArgs);
+        map.entrySet().forEach(
+                entry -> workItemFactory
+                        .workParameter(entry.getKey(), AbstractServiceTaskDescriptor.processWorkItemValue(entry.getValue(), paramName, clazz, CompositeContextNodeHandler::isJsonPathExpr))
```
ricardozanini, line 253: Ok, that answers my question. :)

ricardozanini (approved):
Looks good! My only concern is about this map adapter that we created. Is it really necessary?

jbpm/process-workitems/src/main/java/org/kie/kogito/process/workitems/impl/KogitoWorkItemImpl.java
fjtirado: There are two alternative to all the boilerplate code which I'm also not fan of, but they look worse to me:

1. Extend HashMap (or AbstractHashMap), but the issue is setParameters(Map map) method, where user can pass any map, so I have to proxy it. Hence this is not really an alternative (unless we do put all in setParameters, but this means cheating :), and can have side effects since right now the semantics assume that changes in the Map passed to setParameters are automatically reflected in the workItem, for exampe in Protobuf code). So this is not really an alternative
2. Do the same I did using Java Proxy, but Im against that by principle, because, although a lot of code is saved, proxy is intendd to extends an interface you have not access in your code, which is not the case. 

Thats why, despite me being lazy ;), I wrote all the code needed for delegation (you know write once and forget principle;))

tiagodolphine (commented):
LGTM, just left a comment about the place for `kogito-jsonpath-util`

kogito-workitems/kogito-openapi-workitem/src/main/java/org/kogito/workitem/openapi/JsonNodeResultHandler.java
```diff
@@ -25,17 +28,31 @@
...
+    private JsonNode fromModel(Object inputModel) {
+        if (inputModel instanceof JsonNode) {
+            return (JsonNode) inputModel;
+        }
+        if (inputModel instanceof String) {
+
+            try {
+                return mapper.readTree((String) inputModel);
+            } catch (JsonProcessingException e) {
+
```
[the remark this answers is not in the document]
fjtirado: Not really, if the parsing of the string fails, we fallback to valueToTree that will return something

kogito-workitems/kogito-jsonpath-utils/pom.xml
```diff
@@ -0,0 +1,27 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.kie.kogito</groupId>
+    <artifactId>kogito-workitems</artifactId>
+    <version>2.0.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>kogito-jsonpath-utils</artifactId>
```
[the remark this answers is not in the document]
fjtirado: I asked myself the same question, right now jsonpath is only used by serverless workflow, so maybe is better to put it there

fjtirado:
jenkins run examples tests

fjtirado:
jenkins run runtimes tests

fjtirado:
jenkins retest this

ricardozanini (approved):

fjtirado:
jenkins run apps tests

tiagodolphine (approved):

PR #1554, opened 2021-08-24
[KOGITO-5719] Transform Kogito Add-Ons to Quarkus Extensions

ricardozanini:
Signed-off-by: Ricardo Zanini <zanini@redhat.com>

See https://issues.redhat.com/browse/KOGITO-5719

Linked PRs:

- https://github.com/kiegroup/kogito-examples/pull/867
- https://github.com/kiegroup/kie-docs/pull/3852

In this PR, we introduce a "small" change to our Quarkus based add-ons. Now, everyone is also a Quarkus Extension. This is important to **unlock** the new Add-On Codegen API since we will leverage the ability of Quarkus to inject the *-deployment modules on-demand on end-users projects during code generation. The idea is to add all the codegen related code for add-ons on the *-deployment modules backed by this new shared API. This way, SB can leverage this mechanism as well.

As a plus, we now can appear at code.quarkus.io as an option in the Quarkus project creation. Thus, users will be able to select "kogito-rules" and "kogito-monitoring-prometheus" from a list of available extensions, for example. 

All our internal builder scripts on the cloud can also leverage this change to add add-ons on-demand simply using the Quarkus maven plugin.

This change is a step forward in leveraging the Quarkus platform.

I've also deprecated the "generic" add-ons. I kept only the "human-task-prediction" since we need to know what to do with this add-on. Seems unused, maybe a good candidate for deprecation.

I haven't created/renamed the generic persistence add-ons for SB in this PR because we have a JIRA to deal with it in the short term. For example [KOGITO-5031](https://issues.redhat.com/browse/KOGITO-5031). @cristianonicolai, do you mind linking the others in this JIRA?

The Quarkus persistence add-ons are a simple "redirect" to the actual common ones. It's not in the scope of this PR to refactor them in the "Quarkus way". @cristianonicolai, if you don't mind we can sync up these next changes in the Persistence sync.

That's cool:

```
2021-08-27 13:23:21,022 INFO  [io.quarkus] (main) Installed features: [agroal, cdi, jdbc-postgresql, kogito-addon-persistence-jdbc-extension, kogito-decisions, kogito-predictions, kogito-processes, kogito-rules, narayana-jta, resteasy, resteasy-jackson, servlet, smallrye-context-propagation, smallrye-health, smallrye-openapi, swagger-ui]
```
:sunglasses: 

Many thanks for submitting your Pull Request :heart:! 

Please make sure that your PR meets the following requirements:

- [x] You have read the [contributors guide](CONTRIBUTING.md)
[...]

PR #1612, opened 2021-09-16
Kogito 5921 Support For OracleDB Peristence

Scuilion:
Add support for persisting to OracleDB
Just an initial stab to make sure that I'm on the right track. Will add test on follow up.
https://issues.redhat.com/browse/KOGITO-5921

- [x] Your code is properly formatted according to [this configuration](https://github.com/kiegroup/kogito-runtimes/tree/main/kogito-build/kogito-ide-config)
- [x] Pull Request title contains the target branch if not targeting main: `[0.9.x] KOGITO-XYZ Subject`

<details>
<summary>
</summary>

* <b>Run all builds</b>  

* <b>Quarkus LTS checks</b>  

* <b>Native checks</b>  

</details>


addons/common/persistence/jdbc/pom.xml
```diff
@@ -7,6 +7,7 @@
   </parent>
   <artifactId>kogito-addons-persistence-jdbc</artifactId>
   <name>Kogito :: Add-Ons :: Persistence :: JDBC</name>
+  <!-- <version>2.0.1-SNAPSHOT</version> -->
```
danielezonca, line 10: To be removed

addons/common/persistence/jdbc/src/main/java/org/kie/kogito/persistence/jdbc/JDBCProcessInstances.java
```diff
@@ -45,9 +44,30 @@
...
+        POSTGRES("PostgreSQL", "process_instances"),
+        ORACLE("Oracle", "PROCESS_INSTANCES");
+
+        private final String dbIdentifier;
+        private final String tableNamePattern;
+
+        DatabaseType(final String dbIdentifier, final String tableNamePattern) {
+            this.dbIdentifier = dbIdentifier;
+            this.tableNamePattern = tableNamePattern;
+        }
```
danielezonca, line 60: What about making `Repository` impl selection part of this enum?
```suggestion
        POSTGRES("PostgreSQL", "process_instances", new PostgresRepository()),
        ORACLE("Oracle", "PROCESS_INSTANCES", new OracleRepository());

        private final String dbIdentifier;
        private final String tableNamePattern;
        private final Repository repository;
        
        DatabaseType(final String dbIdentifier, final String tableNamePattern, Repository repository) {
            this.dbIdentifier = dbIdentifier;
            this.tableNamePattern = tableNamePattern;
            this.repository = repository;
        }
        
        public Repository getRepository() {
            return this.repository;
        }
```

addons/common/persistence/jdbc/src/main/java/org/kie/kogito/persistence/jdbc/JDBCProcessInstances.java
```diff
@@ -45,9 +44,30 @@
...
+
+        public static DatabaseType create(final String dbIdentifier) {
+            if ("Oracle".equals(dbIdentifier)) {
+                return ORACLE;
+            } else if ("PostgreSQL".equals(dbIdentifier)) {
+                return POSTGRES;
+            }
+            return null;
```
danielezonca, line 68: I think it is better to have a default return value in any case, we can decide to have Postgres as default or any other but I would avoid to return null because it makes impossible to even try a different DB and see if it works (with proper logging message)
```suggestion
            log.info("No specific match found for DB identifier '" + dbIdentifier + "', assuming PostgreSQL as default"); 
            return POSTGRES;
```

addons/common/persistence/jdbc/src/main/java/org/kie/kogito/persistence/jdbc/JDBCProcessInstances.java
```diff
@@ -80,10 +94,34 @@ private void init() {
...
+                throw new Exception("Database (" + dbProductName + ") not suported");
+            }
+            switch (databaseType) {
+                case ORACLE:
+                    repository = new OracleRepository();
+                    break;
+                case POSTGRES:
+                    repository = new PostgresRepository();
+                    break;
+            }
```
danielezonca, line 111: Applying the changes I suggested this whole code becomes way simpler
```suggestion
            DatabaseType databaseType = DatabaseType.create(dbProductName);
            repository = databaseType.getRepository();
```

addons/common/persistence/jdbc/src/main/java/org/kie/kogito/persistence/jdbc/OracleRepository.java
```diff
@@ -0,0 +1,147 @@
...
+import static org.kie.kogito.persistence.jdbc.JDBCProcessInstances.VERSION;
+
+public class OracleRepository extends Repository {
+
+    //changed
+    @Override
+    public void insertInternal(DataSource dataSource, String processId, UUID id, byte[] payload) {
+        try (Connection connection = dataSource.getConnection();
+                PreparedStatement statement = connection.prepareStatement(INSERT)) {
+            statement.setString(1, id.toString());
```
danielezonca, line 40: As far as I see the only difference in Oracle vs Postgres impl for this Repository class is related to `uuid` data type: in Oracle this type doesn't exist so a [`char(36)`](https://github.com/kiegroup/kogito-runtimes/pull/1612/files#diff-cfbaacac33b9fe064663318ed97f5e5d6c64a46ddf6afad10895ebb396897ac9R2) is used (and here `setString`) while in Postgres the Java [UUID](https://github.com/kiegroup/kogito-runtimes/pull/1612/files#diff-dd238f0f762e06b93cff5721cfb367160a7ef2fec440506aaab0eb761dcb94a4R2) is propagated directly to the DB with a `setObject`.

If this is the case what about using `String`/`char` also in Postgres? In this case we can have a single `GenericRepository` impl for now that is the same for both DB. If we do this we might need to provide a migration script for existing user but I don't think the impact is big: we have a dedicated Postgres addon so I expect ~all postgres user are already using it and this change will have no impact for them.
Wdyt?

@Scuilion 
If @cristianonicolai and @tiagodolphine agree with the proposed change can you please refactor the code to have a single `GenericRepository` that extends `Repository` that works for both and remove Oracle/Postgres impl?

addons/common/persistence/jdbc/src/main/resources/sql/create_tables.sql
```diff
@@ -1,10 +0,0 @@
-CREATE TABLE process_instances(id uuid NOT NULL,
```
danielezonca, line 1: @ricardozanini @cristianonicolai @tiagodolphine 
Do you know if we mention this script in the doc or we use it in some other scripts to setup the system?

addons/common/persistence/jdbc/src/main/resources/sql/exists_tables.sql
```diff
@@ -1,3 +0,0 @@
-SELECT EXISTS(
```
danielezonca, line 1: @ricardozanini @cristianonicolai @tiagodolphine 
Same comment here. I'm quite sure this is not an issue and that this was only used in `JDBCProcessInstances` to check table existence and this PR implements a more generic way to perform the check

kogito-test-utils/pom.xml
```diff
@@ -74,6 +74,12 @@
       <version>${version.org.testcontainers}</version>
     </dependency>
 
+    <dependency>
+      <groupId>org.testcontainers</groupId>
+      <artifactId>oracle-xe</artifactId>
+      <version>${version.org.testcontainers}</version>
+    </dependency>
```
danielezonca, line 81: @Scuilion 
Do you plan to add an integration test so this is the reason of this dependency?
If so fine for me, otherwise please remove it :)

danielezonca (commented):
Thanks for the PR, I think overall looks good to me 👍 
I suggested some changes to simplify and reduce the amount of duplicated code.

Please add a [Springboot integration test](https://github.com/kiegroup/kogito-runtimes/tree/main/integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-persistence-it) and a [Quarkus one](https://github.com/kiegroup/kogito-runtimes/tree/main/integration-tests/integration-tests-quarkus-processes-persistence) following the same structure

addons/common/persistence/jdbc/src/main/java/org/kie/kogito/persistence/jdbc/OracleRepository.java
cristianonicolai: @Scuilion many thanks for this work, I agree with @danielezonca, I would prefer to keep this addon as a generic jdbc implementation, so adjusting the script to make use String would I guess avoid most duplication/specific implementation. Ideally this can support not only postgresql, oracle but also any jdbc implementation provided by Quarkus https://quarkus.io/guides/datasource#default-datasource

tiagodolphine: IMHO or we keep this addon generic by not adding any specific detail of the implementations (postgresql, oracle, etc) even creating the PostgresRepository/OracleRepository or we go the approach to create an addon per database like jdbc-postgresql, jdbc-oracle,... and keep the generic parts in a common module to be shared between the different jdbc based impls, in this way makes sense to have the Repository/PostgresRepository/OracleRepository, etc...

kogito-test-utils/pom.xml
Scuilion: Most definitely!

addons/common/persistence/jdbc/src/main/java/org/kie/kogito/persistence/jdbc/OracleRepository.java
Scuilion: I like the idea of swapping the UUID column to eliminate the duplicate repository code.

Just wanted to clarify that the `create_tables_Oracle.sql`, `create_tables_postgreSQL.sql`, along with the `enum` will still be required.

danielezonca: > IMHO or we keep this addon generic by not adding any specific detail of the implementations (postgresql, oracle, etc) even creating the PostgresRepository/OracleRepository or we go the approach to create an addon per database like jdbc-postgresql, jdbc-oracle,... and keep the generic parts in a common module to be shared between the different jdbc based impls, in this way makes sense to have the Repository/PostgresRepository/OracleRepository, etc...

I think we can preserve a single jdbc addon but we need to consider that it is sort of impossible to have a very portable creation SQL script: for example each DB has usually custom/adhoc data types, see for example `blob` vs `bytea`.
The Java code will have very few/none duplication because JDBC Driver already partially solve this abstraction issue: for example the `setBytes` method is then interpreted/converted by PostreSQL driver as `bytea` while Oracle make it a `blob` but the creation script still has the problem.
If we avoid to use the generic `setObject` (like we did for `UUID`) this should ~always work for each DB (my assumption is that all JDBC drivers implement basic Java types convertions)

The final PR (after suggested changes) should have very few "DB specific code" + it solves some existing portability problem (current `main` version is performing a check to verify if the table already exists using `pg_tables` that is _very_ PostgreSQL specific) and provides the "infrastructure" to have DB specific creation scripts when necessary.

Btw I'm also fine with a `jdbc-oracle`, `jdbc-mariadb`, etc approach but at the moment the only real difference seems to be the creation script (and of course the code to select which creation script use) so we risk to create many modules with just common code + a custom SQL script :) 

@tiagodolphine 
Wdyt?

ricardozanini:
Please wait for #1554 to be merged first.

addons/common/persistence/jdbc/src/main/resources/sql/create_tables.sql
Scuilion: I haven't noticed a reference to this script anywhere.

Scuilion:
I think I've made all the changes required for this PR. Let me know otherwise.

addons/common/persistence/jdbc/src/main/java/org/kie/kogito/persistence/jdbc/GenericRepository.java
```diff
@@ -0,0 +1,220 @@
...
+import static org.kie.kogito.persistence.jdbc.JDBCProcessInstances.PAYLOAD;
+import static org.kie.kogito.persistence.jdbc.JDBCProcessInstances.VERSION;
+
+public class GenericRepository extends Repository {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(GenericRepository.class);
+
+    private enum DatabaseType {
+        POSTGRES("PostgreSQL", "process_instances"),
+        ORACLE("Oracle", "PROCESS_INSTANCES");
```
ricardozanini, line 44: I'm just curious about this. Please don't take my comment in the wrong way.

Since this is the generic JDBC, why are we restricting the DB type? Won't it make sense to have a `kogito-addons-quarkus-persistence-oracle` instead?

Otherwise, how do we plan to maintain this in the future? Others will add `MYSQL` support, for instance?

ricardozanini (commented):
Hi! It looks fine, but I'm curious why we specify the database implementation in this PR instead of having a particular set of add-ons for each one instead. Sorry if this has already been discussed.

kogito-test-utils/pom.xml
```diff
@@ -74,6 +74,18 @@
...
+    <dependency>
+      <groupId>org.testcontainers</groupId>
+      <artifactId>oracle-xe</artifactId>
+      <version>${version.org.testcontainers}</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.oracle.database.jdbc</groupId>
+      <artifactId>ojdbc11</artifactId>
+      <version>21.3.0.0</version>
```
danielezonca, line 86: Please move this value to a `version.com.oracle.database.jdbc` property in [this file](https://github.com/kiegroup/kogito-runtimes/blob/main/kogito-build/kogito-dependencies-bom/pom.xml#L71) (NOTE: properties are alphabetically ordered)
```suggestion
      <version>${version.com.oracle.database.jdbc}</version>
```

addons/common/persistence/jdbc/src/test/java/org/kie/persistence/jdbc/OracleProcessInstancesIT.java
```diff
@@ -0,0 +1,106 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+import static org.kie.kogito.internal.process.runtime.KogitoProcessInstance.STATE_ACTIVE;
+import static org.kie.kogito.internal.process.runtime.KogitoProcessInstance.STATE_COMPLETED;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+class OracleProcessInstancesIT extends TestHelper {
```
danielezonca, line 37: Please remove this class and rename back `PostgresProcessInstancesIT` as `JdbcProcessInstancesIT`.
Personally I don't like very much the current approach we have in Kogito for this test because it is not very flexible and it will be probably reviewed but I would like to avoid to clone the class.
The main limitation is that it will start both containers (Postgres/Oracle) for each test class.

My suggestion for now is to start leverage JUnit5 `@ParameterizedTest` tests:
```java
    public static Stream<DataSource> datasources() {
        return Stream.of(ORACLE_DATA_SOURCE, ORACLE_DATA_SOURCE);
    }

    @ParameterizedTest
    @MethodSource("datasources")
    public void testBasicTaskFlow(DataSource datasource) {
        var factory = new TestProcessInstancesFactory(datasource, false);
        ....
    }

    @ParameterizedTest
    @MethodSource("datasources")
    void testBasicFlow(DataSource datasource) {
        var factory = new TestProcessInstancesFactory(datasource, false);
        ....
    }
```

You probably need to add this dependency
```xml
        <dependency>
            <groupId>org.junit.jupiter</groupId>
            <artifactId>junit-jupiter-params</artifactId>
            <scope>test</scope>
        </dependency>
```

addons/common/persistence/jdbc/src/main/java/org/kie/kogito/persistence/jdbc/GenericRepository.java
Scuilion: Comments are welcomed!

The original implementation relied on postgres specific table creation. So the only real way to make this generic would be to remove that logic. 

If you specify a jdbc url of a database type that is not postgres or oracle, then we default to postgres. And if, for instance `MYSQL`, supports the same datatypes and ddl, then you're all good. When that is not the case, then you need to add a `/create_tables_MYSQL.sql` and update the Enum with the jdbc `dbIdentifier`.

None of this matters if you have `auto.ddl=false`.

The other option, of course, is to strip all that creation logic out and start talking about database migration strategy.

danielezonca: I think the real issue is the current impl of `-jdbc` addon is not really generic: it uses Postgres data types and also a Postgres specific table (`pg_tables`) to check if the table exists.
This PR it is true that introduce some DB specific code (this enum) that (we can probably also revisit in the future) but it is way more flexible and if we add a fallback with [ANSI SQL](https://github.com/kiegroup/kogito-runtimes/pull/1612#discussion_r712341606) it should be reasonable flexible for now

addons/common/persistence/jdbc/src/main/java/org/kie/kogito/persistence/jdbc/GenericRepository.java
```diff
@@ -0,0 +1,220 @@
...
+
+        public static DatabaseType create(final String dbIdentifier) {
+            if (ORACLE.getDbIdentifier().equals(dbIdentifier)) {
+                return ORACLE;
+            } else if (POSTGRES.getDbIdentifier().equals(dbIdentifier)) {
+                return POSTGRES;
+            } else {
+                var msg = String.format("Unrecognized DB (%s), defaulting to postgres", dbIdentifier);
+                LOGGER.warn(msg);
+                return POSTGRES;
```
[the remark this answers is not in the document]
Scuilion: Comments are welcomed!

The original implementation relied on postgres specific table creation. So the only real way to make this generic would be to remove that logic. If you specify a jdbc url of a database type that is not postgres or oracle, then we default to postgres. And if, for instance `MYSQL`, supports the same datatypes and ddl, then you're all good. 

None of this matters if you have `auto.ddl=false`.

Which brings up another topic. How do you handle database migration/upgrade?

addons/common/persistence/jdbc/src/main/java/org/kie/kogito/persistence/jdbc/GenericRepository.java
danielezonca: Btw as I mentioned [here](https://github.com/kiegroup/kogito-runtimes/pull/1612#discussion_r711246599) it is fine for me to have also a `oracle-jdbc`/`mariadb`but a part for trying to improve table existence check (+ the ANSI SQL fallback that I suggested) and get rid of the enum the rest of the PR is very generic

kogito-build/kogito-dependencies-bom/pom.xml
```diff
@@ -29,6 +29,7 @@
     <version.com.github.victools>4.18.0</version.com.github.victools>
     <version.com.github.tomakehurst.wiremock>2.27.2</version.com.github.tomakehurst.wiremock>
     <version.com.google.protobuf>3.15.8</version.com.google.protobuf>
+    <version.com.oracle.database.jdbc>21.3.0.0</version.com.oracle.database.jdbc>
```
ricardozanini, line 32: We need to understand the implications of doing that in a productized supported versions matrix. Summon @MarianMacik

addons/common/persistence/jdbc/src/main/java/org/kie/kogito/persistence/jdbc/OracleRepository.java
ricardozanini: We can handle the extensibility of this add-on in a new JIRA/PR. One that users can easily extend by adding their script/any additional steps to make this add-on extensible without having to modify its code.

@Scuilion can I ask you to add a README in the root of this module giving a brief explanation of how to use it? Don't need to extend much, we can do the details in the docs: https://docs.jboss.org/kogito/release/latest/html_single/#_enabling_postgresql_persistence_using_jdbc_driver_for_kogito_services

@tiagodolphine @cristianonicolai who can work with the documentation once this PR is merged?

kogito-build/kogito-dependencies-bom/pom.xml
danielezonca: Well that property is just for the integration test, we can declare it locally to the module or here but it is just a property (i.e. not a depMan declaration)

ricardozanini: Yes, I know. But in the future when someone spots this, questions will be asked :)
And everybody will forget that this is just used by ITs. @Scuillion can you add a comment above this property mentioning this?

addons/common/persistence/jdbc/src/test/java/org/kie/persistence/jdbc/JdbcProcessInstancesIT.java
```diff
@@ -96,7 +103,7 @@ void testBasicFlow() {
     }
 
     @Test
-    void testException() {
+    void testException(DataSource dataSource) {
```
danielezonca, line 106: This is not a `ParameterizedTest` so no parameter necessary
```suggestion
    void testException() {
```

kogito-test-utils/src/main/java/org/kie/kogito/testcontainers/KogitoPostgreSqlContainer.java
```diff
@@ -28,12 +28,13 @@
...
  * PostgreSQL Container for Kogito examples.
  */
-public class KogitoPostgreSqlContainer extends PostgreSQLContainer implements TestResource {
+public class KogitoPostgreSqlContainer extends PostgreSQLContainer<KogitoPostgreSqlContainer> implements TestResource {
 
     public static final String POSTGRESQL_CONNECTION_URI = "kogito.persistence.postgresql.connection.uri";
     private static final Logger LOGGER = LoggerFactory.getLogger(KogitoPostgreSqlContainer.class);
 
     public KogitoPostgreSqlContainer() {
+        super("postgres:9.6.12");
```
danielezonca, line 37: Why this change? Is it related to the PR?

danielezonca (approved):
Only few additional comments to address but the PR is good for me 👍 
Thanks again for the contribution :)

Personally I think we can revisit in the future that specific `enum` (probably modifying check table method) and probably verifying if `ANSI` codepath works for Oracle/Postgres too so that we can remove all remaining DB specific code but it is way more flexible than current `main` code.
My proposal is to merge and postpone other refactoring to future tickets.

@ricardozanini @tiagodolphine @cristianonicolai 
Please have a look and feel free to comment

kogito-test-utils/src/main/java/org/kie/kogito/testcontainers/KogitoPostgreSqlContainer.java
Scuilion: This is adding "resiliency" to the my test case. More than happy to make a separate PR for it.

danielezonca:
@Scuilion 
Can you please verified failed build?
https://github.com/kiegroup/kogito-runtimes/pull/1612/checks?check_run_id=3674578414

```
kiegroup_kogito_runtimes/addons/common/persistence/jdbc/src/test/java/org/kie/persistence/jdbc/JdbcProcessInstancesIT.java:83 | java.lang.RuntimeException: Error finding all process instances, for processId UserTask 	at org.kie.persistence.jdbc.JdbcProcessInstancesIT.testBasicFlow(JdbcProcessInstancesIT.java:83)
kiegroup_kogito_runtimes/addons/common/persistence/jdbc/src/test/java/org/kie/persistence/jdbc/JdbcProcessInstancesIT.java:47 | java.lang.RuntimeException: Error finding all process instances, for processId UserTask 	at org.kie.persistence.jdbc.JdbcProcessInstancesIT.testBasicTaskFlow(JdbcProcessInstancesIT.java:47)
kiegroup_kogito_runtimes/addons/common/persistence/jdbc/src/test/java/org/kie/persistence/jdbc/JdbcProcessInstancesWithLockIT.java:72 | java.lang.RuntimeException: Error finding all process instances, for processId UserTask 	at org.kie.persistence.jdbc.JdbcProcessInstancesWithLockIT.testRemove(JdbcProcessInstancesWithLockIT.java:72)
kiegroup_kogito_runtimes/addons/common/persistence/jdbc/src/test/java/org/kie/persistence/jdbc/JdbcProcessInstancesWithLockIT.java:40 | java.lang.RuntimeException: Error finding all process instances, for processId UserTask 	at org.kie.persistence.jdbc.JdbcProcessInstancesWithLockIT.testUpdate(JdbcProcessInstancesWithLockIT.java:40)
```

Scuilion:
@danielezonca Might be oracle not having a timezone setting. Made a small change. Don't know how to trigger a rebuild.

ricardozanini (approved):

addons/common/persistence/jdbc/src/main/java/org/kie/kogito/persistence/jdbc/Repository.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+package org.kie.kogito.persistence.jdbc;
+
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+import javax.sql.DataSource;
+
+abstract class Repository {
```
cristianonicolai, line 24: @Scuilion maybe this would best fit as an Interface instead?

addons/common/persistence/jdbc/src/main/resources/sql/create_tables_Oracle.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE process_instances(
```
cristianonicolai, line 1: @Scuilion could you review the file structure, take a look in the following module: https://github.com/kiegroup/kogito-runtimes/tree/main/addons/common/persistence/ddl
In there, we're organizing ddl scripts by database, we should probably keep similar thing here, so it is the same file name but different folder per db.
The ultimate artefact that combines all scripts is in this module https://github.com/kiegroup/kogito-apps/tree/main/persistence-commons/kogito-ddl so we also need to verify it will remain consistent for end users.

addons/common/persistence/jdbc/src/test/java/org/kie/persistence/jdbc/TestHelper.java
```diff
@@ -16,105 +16,98 @@
...
 
     public static final String TEST_ID = "02ac3854-46ee-42b7-8b63-5186c9889d96";
 
-    private boolean enableLock;
-
     @BeforeAll
     public static void startContainerAndPublicPortIsAvailable() {
-        container.start();
-        ds = getDataSource(container);
+        PG_CONTAINER.start();
```
cristianonicolai, line 64: @Scuilion we should probably run the test for each container instead of starting both together, that would limit resources and give more clarity when relating to a potential issue.

kogito-test-utils/src/main/java/org/kie/kogito/testcontainers/KogitoOracleSqlContainer.java
```diff
@@ -0,0 +1,80 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(KogitoOracleSqlContainer.class);
+
+    public static final String ORACLE_CONNECTION_URI = "kogito.persistence.oracle.connection.uri";
+
+    public KogitoOracleSqlContainer() {
+        super("oracleinanutshell/oracle-xe-11g");
+        withLogConsumer(getLogger());
+        withLogConsumer(new Slf4jLogConsumer(LOGGER));
+        withStartupTimeout(Constants.CONTAINER_START_TIMEOUT);
+        System.setProperty("oracle.jdbc.timezoneAsRegion", "false");
```
cristianonicolai, line 42: maybe set this at start method and clear the property once stop?

cristianonicolai:
ok to test

addons/common/persistence/jdbc/src/test/java/org/kie/persistence/jdbc/TestHelper.java
danielezonca: Yes current impl is not ideal starting both DB at the same time. I asked him to use a parameterized test so that at least it executes both tests but I think we should review the `TestHelper` class in general. Probably it is better to do it with a different PR 🤔

Scuilion: Yes, please, another PR. This one seems to be going off the rails!

Scuilion:
Same test  is failing in PR-1624, PR-1623, PR-1622, and probably others.

addons/common/persistence/jdbc/src/main/java/org/kie/kogito/persistence/jdbc/Repository.java
Scuilion: I wanted package private.

ricardozanini:
> Same test is failing in PR-1624, PR-1623, PR-1622, and probably others.

Fixed here https://github.com/kiegroup/kogito-examples/pull/898

ricardozanini:
Jenkins rerun examples tests

Scuilion:
Seems as though the code coverage for integration test is not registering.

addons/common/persistence/jdbc/src/main/resources/sql/create_tables_Oracle.sql
danielezonca: @cristianonicolai 
I think it is better to move to a different ticket this aspect because it is more related to a refactor/improvement to better package DDL scripts.
As far as I see it requires to review folder structure and the creation of a different module to package scripts. Finally I expect also the documentation needs to be updated to explain this 🤔 
Currently the PR is mainly changing some internals of the -jdbc addon to make the addon more generic and enable not-Postgres execution so we can probably split in different following ticket a "packaging" refactor.
Wdyt?

danielezonca:
> Seems as though the code coverage for integration test is not registering.

Yes it is possible that some integration tests are not included in sonarcloud code coverage analysis so don't worry about that value :)

As far as I see all comments has been resolved expect this one
https://github.com/kiegroup/kogito-runtimes/pull/1612#discussion_r714405464
but I think it is better to move to a different ticket the resolution of this.
Wdyt?

@MarianMacik 
Do you want review/approve it too?

Scuilion:
> but I think it is better to move to a different ticket the resolution of this.

I'm very willing to pick up that ticket after this one.

addons/common/persistence/jdbc/src/main/resources/sql/create_tables_Oracle.sql
cristianonicolai: @danielezonca my point is that this change will likely break the structure we already have in place, those ddl modules are currently copying the SQL scripts from here. With the current status of the pr, the ddl files are not getting included in the final zip file, https://github.com/kiegroup/kogito-runtimes/blob/main/addons/common/persistence/ddl/src/assembly/db-scripts.xml#L18 would need tweaking as I mentioned, so it creates continues to have a `jdbc` folder, and a new `oracle` one.

danielezonca: Ok thanks for the clarification. 
How are these files loaded?
This PR introduces a [`FileLoader`](https://github.com/kiegroup/kogito-runtimes/pull/1612/files#diff-211ccca180df02b13db540bbae3be6ad239857b63d838f170008c495ee1c2306R24) class that has assumption about the location/mechanism to load the script while the target path seems to be `jdbc` folder. Do we have any integration tests that uses this assembly output?
Is it only for documentation and for letting user manually run the scripts or they are actually executed?
I expect it is only for documentation so it should be enough to update that `db-scripts.xml` file as following
```xml
    <fileSets>
        <fileSet>
            <directory>${path.to.persistence.modules}/postgresql/src/main/resources/sql</directory>
            <outputDirectory>postgresql</outputDirectory>
            <includes>
                <include>*_create.sql</include>
            </includes>
        </fileSet>
    </fileSets>
    <files>
        <file>
            <source>${path.to.persistence.modules}/jdbc/src/main/resources/sql/create_tables_Oracle.sql</source>
            <outputDirectory>oracle</outputDirectory>
            <destName>runtime_create.sql</destName>
        </file>
        <file>
            <source>${path.to.persistence.modules}/jdbc/src/main/resources/sql/create_tables_ansi.sql</source>
            <outputDirectory>jdbc</outputDirectory>
            <destName>runtime_create.sql</destName>
        </file>
    </files>
```
I didn't tested locally but it should work and considering that now we have a single file for each DB wildcards should not be necessary

@cristianonicolai 
Wdyt?

Scuilion:
Anything left for me to do on this PR?

addons/common/persistence/jdbc/src/main/resources/sql/create_tables_Oracle.sql
cristianonicolai: yep, that looks good, and correct there are no tests for the final zip file.

cristianonicolai:
> Anything left for me to do on this PR?

I guess only the ddl scripts think mentioned here https://github.com/kiegroup/kogito-runtimes/pull/1612#discussion_r716594439

addons/common/persistence/jdbc/src/main/resources/sql/create_tables_Oracle.sql
Scuilion: Just to clarify, the `FileLoader` class is the same code originally, just refactored our for clarity. So if we look at the current zip we can confirm that it is in fact, not there.

danielezonca:
@Scuilion 
Can you please apply the changes included in https://github.com/kiegroup/kogito-runtimes/pull/1612#discussion_r716594439 ?
On Monday we have sprint cutoff and I would like to include this PR 👍

Scuilion:
> @Scuilion Can you please apply the changes included in [#1612 (comment)](https://github.com/kiegroup/kogito-runtimes/pull/1612#discussion_r716594439) ? On Monday we have sprint cutoff and I would like to include this PR +1

commit 903d634 creates the following ddl zip

```
/home/kevino/projects/kogito-runtimes/addons/common/persistence ±[●][KOGITO-5921-support-for-oracle] k8sstaging/staging
> find . -name "*.zip"                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
./ddl/target/kogito-ddl-runtimes-2.0.0-SNAPSHOT-db-scripts.zip                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                
/home/kevino/projects/kogito-runtimes/addons/common/persistence ±[●][KOGITO-5921-support-for-oracle] k8sstaging/staging 
zipinfo ./ddl/target/kogito-ddl-runtimes-2.0.0-SNAPSHOT-db-scripts.zip                                                                                                                                                                                                       
Archive:  ./ddl/target/kogito-ddl-runtimes-2.0.0-SNAPSHOT-db-scripts.zip                                                                                                                                                                                                        
Zip file size: 1493 bytes, number of entries: 6                                                                                                                                                                                                                                 
drwxr-xr-x  2.0 unx        0 b- stor 21-Sep-30 13:09 postgresql/                                                                                                                                                                                                                
drwxr-xr-x  2.0 unx        0 b- stor 21-Sep-30 13:09 jdbc/                                                                                                                                                                                                                      
-rw-rw-r--  2.0 unx      467 bl defN 21-Sep-23 10:27 postgresql/runtime_create.sql                                                                                                                                                                                              
-rw-rw-r--  2.0 unx      285 bl defN 21-Sep-30 12:11 jdbc/create_tables_ansi.sql                                                                                                                                                                                                
-rw-rw-r--  2.0 unx      291 bl defN 21-Sep-30 12:11 jdbc/create_tables_PostgreSQL.sql                                                                                                                                                                                          
-rw-rw-r--  2.0 unx      286 bl defN 21-Sep-30 12:11 jdbc/create_tables_Oracle.sql                                                                                                                                                                                              
6 files, 1329 bytes uncompressed, 683 bytes compressed:  48.6%              
```

Scuilion:
Can't seem to see the details of the failed Examples build.

cristianonicolai:
Jenkins rerun examples tests

Scuilion:
woot woot!

cristianonicolai:
> > @Scuilion Can you please apply the changes included in [#1612 (comment)](https://github.com/kiegroup/kogito-runtimes/pull/1612#discussion_r716594439) ? On Monday we have sprint cutoff and I would like to include this PR +1
> 
> commit [903d634](https://github.com/kiegroup/kogito-runtimes/commit/903d6341159be4314bbc026adad9d054fe8aac05) creates the following ddl zip
> 
> ```
> /home/kevino/projects/kogito-runtimes/addons/common/persistence ±[●][KOGITO-5921-support-for-oracle] k8sstaging/staging
> > find . -name "*.zip"                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
> ./ddl/target/kogito-ddl-runtimes-2.0.0-SNAPSHOT-db-scripts.zip                                                                                                                                                                                                                  
>                                                                                                                                                                                                                                                                                 
> /home/kevino/projects/kogito-runtimes/addons/common/persistence ±[●][KOGITO-5921-support-for-oracle] k8sstaging/staging 
> zipinfo ./ddl/target/kogito-ddl-runtimes-2.0.0-SNAPSHOT-db-scripts.zip                                                                                                                                                                                                       
> Archive:  ./ddl/target/kogito-ddl-runtimes-2.0.0-SNAPSHOT-db-scripts.zip                                                                                                                                                                                                        
> Zip file size: 1493 bytes, number of entries: 6                                                                                                                                                                                                                                 
> drwxr-xr-x  2.0 unx        0 b- stor 21-Sep-30 13:09 postgresql/                                                                                                                                                                                                                
> drwxr-xr-x  2.0 unx        0 b- stor 21-Sep-30 13:09 jdbc/                                                                                                                                                                                                                      
> -rw-rw-r--  2.0 unx      467 bl defN 21-Sep-23 10:27 postgresql/runtime_create.sql                                                                                                                                                                                              
> -rw-rw-r--  2.0 unx      285 bl defN 21-Sep-30 12:11 jdbc/create_tables_ansi.sql                                                                                                                                                                                                
> -rw-rw-r--  2.0 unx      291 bl defN 21-Sep-30 12:11 jdbc/create_tables_PostgreSQL.sql                                                                                                                                                                                          
> -rw-rw-r--  2.0 unx      286 bl defN 21-Sep-30 12:11 jdbc/create_tables_Oracle.sql                                                                                                                                                                                              
> 6 files, 1329 bytes uncompressed, 683 bytes compressed:  48.6%              
> ```

@Scuilion thanks, its almost there, the idea is to keep a folder per database and jdbc would be the generic/ANSI version, so in that sense we need to endup with:

- postgresql/runtime_create.sql             
- jdbc/runtime_create.sql             
- oracle/runtime_create.sql 

Be aware that the Postgresql one is already there, so you just need to make sure that content matches.

danielezonca:
@Scuilion 
I have applied my proposed changes directly to your branch, see https://github.com/kiegroup/kogito-runtimes/pull/1612/commits/3981a6643b1f42c0e8e15fb2e4bec2579246f33f

@cristianonicolai 
This the output of the script now. Is it fine for you?
The PostgreSQL script is taken by postgres addon and not from the generic one. We might revisit this if necessary
image

cristianonicolai (approved):

cristianonicolai:
@danielezonca thanks!

kogito-build/kogito-dependencies-bom/pom.xml
MarianMacik: @ricardozanini I think once it is going to be supported, we can revise the version used.

kogito-test-utils/src/main/java/org/kie/kogito/testcontainers/KogitoOracleSqlContainer.java
MarianMacik: +1 to that. For safety we should remove this flag after the tests are executed/container is stopped.

Scuilion:
Build failure details not available.

danielezonca:
> Build failure details not available.

Jenkins build are less reliable compared to GHA so I have approved GHA to see if this is green. I expect it is a random (or unrelated) failure

danielezonca:
I think main branch of examples repo is broken, I got similar failures in this other PR https://github.com/kiegroup/kogito-runtimes/pull/1631

@evacchi 
Can you please have a look? It seems related to new programmatic API examples even if it is not clear to me the relation 🤔

cristianonicolai:
Jenkins rerun examples tests

addons/common/persistence/jdbc/src/test/java/org/kie/persistence/jdbc/TestHelper.java
```diff
@@ -66,12 +67,14 @@ public static void startContainerAndPublicPortIsAvailable() {
 
         ORACLE_CONTAINER.start();
         ORACLE_DATA_SOURCE = getOracleDataSource(ORACLE_CONTAINER);
+        System.setProperty(ORACLE_TIMEZONE_PROPERTY, "false");
```
ricardozanini, line 70: I think it's fine to have this for now, but there's another solution for such use cases: https://www.baeldung.com/java-system-rules-junit#working-with-system-properties

MarianMacik (approved):

debu999:
No documentation updated for oracle. Can that be done?? Only we have PostGres in documentation. Is it worth a look.

PR #1613, opened 2021-09-17
[KOGITO-5394] Workitem error Handling strategy

fjtirado:
Further changes needed for error handling strategy to work

- Change the approach to start subprocess, using Kogito API to do that
- If a workItem is not a UserTask, REST GET tasks API was failing (not all work items are user tasks and tasks api is intended to return just user tasks, so a new method was added to ProcessInstance interface, allowing filtering of work items) 
- Adding test to integration

### DIFF
--- a/jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/DummyKnowledgeRuntime.java
+++ b/jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/DummyKnowledgeRuntime.java
@@ -48,6 +48,7 @@ import org.kie.api.runtime.rule.LiveQuery;
 import org.kie.api.runtime.rule.QueryResults;
 import org.kie.api.runtime.rule.ViewChangedEventListener;
 import org.kie.api.time.SessionClock;
+import org.kie.kogito.Application;
 import org.kie.kogito.internal.process.event.KogitoProcessEventSupport;
 import org.kie.kogito.internal.process.runtime.KogitoProcessInstance;
 import org.kie.kogito.internal.process.runtime.KogitoProcessRuntime;
@@ -247,27 +248,27 @@ class DummyKnowledgeRuntime implements InternalKnowledgeRuntime, KogitoProcessRu
 
     @Override
     public KogitoProcessInstance startProcess(String processId) {
-        return null;
+        return (KogitoProcessInstance) processRuntime.startProcess(processId);
     }
 
     @Override
     public KogitoProcessInstance startProcess(String processId, Map<String, Object> parameters) {
-        return null;
+        return (KogitoProcessInstance) processRuntime.startProcess(processId, parameters);
     }
 
     @Override
     public KogitoProcessInstance startProcess(String processId, AgendaFilter agendaFilter) {
-        return null;
+        return (KogitoProcessInstance) processRuntime.startProcess(processId, agendaFilter);
     }
 
     @Override
     public KogitoProcessInstance startProcess(String processId, Map<String, Object> parameters, AgendaFilter agendaFilter) {
-        return null;
+        return (KogitoProcessInstance) processRuntime.startProcess(processId, parameters, agendaFilter);
     }
 
     @Override
     public ProcessInstance startProcessFromNodeIds(String s, Map<String, Object> map, String... strings) {
-        throw new UnsupportedOperationException("org.jbpm.process.instance.DummyKnowledgeRuntime.startProcessFromNodeIds -> TODO");
+        return processRuntime.startProcessFromNodeIds(s, map, strings);
 
     }
 
@@ -284,17 +285,17 @@ class DummyKnowledgeRuntime implements InternalKnowledgeRuntime, KogitoProcessRu
 
     @Override
     public KogitoProcessInstance startProcessInstance(String processInstanceId) {
-        return null;
+        return processRuntime.getKogitoProcessRuntime().startProcessInstance(processInstanceId);
     }
 
     @Override
     public KogitoProcessInstance startProcessInstance(String processInstanceId, String trigger) {
-        return null;
+        return processRuntime.getKogitoProcessRuntime().startProcessInstance(processInstanceId, trigger);
     }
 
     @Override
     public void signalEvent(String type, Object event) {
-
+        processRuntime.getKogitoProcessRuntime().signalEvent(type, event);
     }
 
     @Override
@@ -305,7 +306,7 @@ class DummyKnowledgeRuntime implements InternalKnowledgeRuntime, KogitoProcessRu
 
     @Override
     public void signalEvent(String type, Object event, String processInstanceId) {
-
+        processRuntime.getKogitoProcessRuntime().signalEvent(type, event, processInstanceId);
     }
 
     @Override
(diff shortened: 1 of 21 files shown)


cristianonicolai:
@fjtirado can you add some test?

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/WorkItemNodeInstance.java
```diff
@@ -593,30 +595,32 @@ public void validateWorkItemResultVariable(String processName, List<DataAssociat
...
+        VariableScopeInstance variableScopeInstance = (VariableScopeInstance) processInstance.getContextInstance(VariableScope.VARIABLE_SCOPE);
+        variableScopeInstance.setVariable("DeploymentId", workItem.getDeploymentId());
+        variableScopeInstance.setVariable("ProcessInstanceId", workItem.getProcessInstanceStringId());
+        variableScopeInstance.setVariable("WorkItemId", workItem.getStringId());
+        variableScopeInstance.setVariable("NodeInstanceId", this.getStringId());
+        variableScopeInstance.setVariable("ErrorMessage", handlerException.getMessage());
+        variableScopeInstance.setVariable("Error", handlerException);
+        for (Map.Entry<String, Object> entry : workItem.getParameters().entrySet()) {
+            variableScopeInstance.setVariable(entry.getKey(), entry.getValue());
+        }
```
tiagodolphine, line 619: minor, but can't you create a Map with all the variables and do something like `process.createInstance(process.createModel().fromMap(variables))`, anyway since here is the jbpm-flow probably it is not an issue to not use the kogito-api... it is just strange that there is a mix of 2 APIs (jbpm-flow + kogito-api) in the same method.

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/ProcessServiceImpl.java
```diff
@@ -132,7 +134,7 @@ public ProcessServiceImpl(Application application) {
     public <T extends Model> Optional<List<WorkItem>> getTasks(Process<T> process, String id, String user, List<String> groups) {
         return process.instances()
                 .findById(id, ProcessInstanceReadMode.READ_ONLY)
-                .map(pi -> pi.workItems(Policies.of(user, groups)));
+                .map(pi -> pi.workItems(ni -> ni instanceof HumanTaskNodeInstance && ((WorkItemNodeInstance) ni).getWorkItem().enforce(Policies.of(user, groups))));
```
tiagodolphine, line 137: just a question, because it seems the policies enforce is done now in here, instead of the ProcessInstance, but in the other method it is still done there...isn't this an issue in the ProcessInstance API itself, because accepting a generic predicate, could allow not enforcing the security policies. One possibility would be the have both parameters like `List<WorkItem> workItems(Predicate<KogitoNodeInstance> p, Policy<?>... policies))`, wdyt?

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/DummyKnowledgeRuntime.java
```diff
@@ -247,27 +248,27 @@ public void removeEventListener(AgendaEventListener listener) {
 
     @Override
     public KogitoProcessInstance startProcess(String processId) {
-        return null;
+        return (KogitoProcessInstance) processRuntime.startProcess(processId);
```
tiagodolphine, line 251: why now is it necessary the processRuntime? I don't have the whole context but the name Dummy seems to me that this should not really execute any method.

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/LightProcessRuntimeContext.java
```diff
@@ -48,11 +43,7 @@ public LightProcessRuntimeContext(Application app, Collection<Process> processes
 
     @Override
     public Optional<Process> findProcess(String id) {
-        Optional<Process> result = processes.stream().filter(p -> p.getId().equals(id)).findAny();
-        if (result.isEmpty() && allProcesses != null) {
-            result = Optional.ofNullable(((ProcessSupplier) allProcesses.processById(id)).get());
-        }
-        return result;
+        return processes.stream().filter(p -> p.getId().equals(id)).findAny();
```
tiagodolphine, line 46: why the usage of `Processes` was removed? just wondering because it was added before for some reason...

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessRuntimeImpl.java
```diff
@@ -200,7 +198,7 @@ public KogitoProcessInstance createProcessInstance(String processId,
 
     @Override
     public ProcessInstance startProcessInstance(long l) {
-        throw new UnsupportedOperationException();
+        return kogitoProcessRuntime.startProcessInstance(Long.toString(l));
```
tiagodolphine, line 201: the same, I'm not sure why do we need to change this now, because the idea to start a processInstance would be through the kogito-api like `kogitoProcessInstance.start()`, right?

tiagodolphine (commented):
LGTM in general but I left some comments.

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/WorkItemNodeInstance.java
fjtirado: I tried process.createModel().fromMap(variables) in earlier version, but it is not backward compatible with existing BPMN subprocess handling errors. Reason is that these process does not define "Error" key as propert, so fromMap(Variables) do not add it to the variables Map.

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/LightProcessRuntimeContext.java
fjtirado: I restored previous version since now process is created in different way, using Kogito API

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/DummyKnowledgeRuntime.java
fjtirado: this is uneeded  changesbut also harmless (I added it in previous version of this PR and do not think needs to be removed, since it actually improved the previous implementation)

jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessRuntimeImpl.java
fjtirado: in that case, I think you are right and is better to throw unsupported

jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/ProcessServiceImpl.java
fjtirado: good point. I was assuming we might want to retrieve work items that does not require policy enforcing (any custom work item handler actually)

tiagodolphine (approved):
Thanks for applying the comments, LGTM.

fjtirado:
@cristianonicolai Examples failure are unrelated, this is ready to go, can you take a look please?

fjtirado:
jenkins rerun examples tests
jenkins rerun apps tests

fjtirado:
jenkins rerun apps tests

PR #1621, opened 2021-09-22
[FAI-591] Reorganize modules

gitgabrio:
@danielezonca @evacchi @jiripetrlik

See https://issues.redhat.com/browse/FAI-591

It is almost exclusively moving/renaming task.
I "abstracted" only two classes:
`org.kie.kogito.tracing.event.model.ModelEvent`
`org.kie.kogito.tracing.event.message.Message`

I did not refactor the tracing-decision module (kogito-addons-springboot-tracing-decision)
I also found out that there are two slightly different `EventEmitter` interfaces:

`org.kie.kogito.tracing.EventEmitter` (tracing-api)
`org.kie.kogito.event.EventEmitter` (kogito-events-api)

Depends on

https://github.com/kiegroup/kogito-apps/pull/1030
https://github.com/kiegroup/kogito-examples/pull/893

As a side note, I found some serialized classes (`TraceEvent`, `TraceHeader`, ecc.) not finals but annotated with `@JsonIgnoreProperties(ignoreUnknown = true)`: this would mean that "sublcasses" of them won't be correctly deserialized - i.e. it won't be possible to deserialize an instance of a class extending one of the above (it will be returned only the parent one)


gitgabrio:
Jenkins retest this

manstis:
@gitgabrio Please excuse my nosiness. I saw you wrote:

> As a side note, I found some serialized classes ... annotated with `@JsonIgnoreProperties(ignoreUnknown = true)` : this would mean that "sublcasses" of them won't be correctly deserialized - i.e. it won't be possible to deserialize an instance of a class extending one of the above (it will be returned only the parent one)

I always believed, since the annotation is not inherited, that sub-classes would need an equivalent annotation to exclude unknown properties. I have no example of this; it is just my understanding. For my education have you experienced the issue you report. I'd like to know what is best for the future.

gitgabrio:
> @gitgabrio Please excuse my nosiness. I saw you wrote:
> 
> > As a side note, I found some serialized classes ... annotated with `@JsonIgnoreProperties(ignoreUnknown = true)` : this would mean that "sublcasses" of them won't be correctly deserialized - i.e. it won't be possible to deserialize an instance of a class extending one of the above (it will be returned only the parent one)
> 
> I always believed, since the annotation is not inherited, that sub-classes would need an equivalent annotation to exclude unknown properties. I have no example of this; it is just my understanding. For my education have you experienced the issue you report. I'd like to know what is best for the future.

@manstis 
Good point. I have to make an actual test for this. I'll let you know....

gitgabrio:
@manstis 
I made a very quick test on a simple project, and TBH what I was referring to is possible with the following declarations:

`@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "className")
@JsonIgnoreProperties(ignoreUnknown = true)
public class TraceEvent {`


`public class SubTraceEvent extends TraceEvent {`


That allow 
(1)

` String toRead = "{\"className\":\"net.test.json.beans.TraceEvent\"....";
  TraceEvent retrieved = new ObjectMapper().readValue(toRead, TraceEvent.class);
`
retrieved is actually a `TraceEvent`

(2)

`String toRead = "{\"className\":\"net.test.json.beans.SubTraceEvent\"....";
TraceEvent retrieved = new ObjectMapper().readValue(toRead, TraceEvent.class);`

retrieved is actually a `SubTraceEvent` and its specific properties are correctly populated.

That way it is not needed to know, at compile time, the actuall class that would be instantiated by the ObjectMapper.
On the other side, this is not possible without the `@JsonTypeInfo` annotation on parent class.

addons/common/tracing/tracing-api/src/main/java/org/kie/kogito/tracing/event/message/Message.java
```diff
@@ -13,33 +13,41 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
 
-@JsonInclude(NON_NULL)
-public class Message {
+/**
+ * <code>Message</code> to be eventually extended by model-specific implementations
+ */
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.EXISTING_PROPERTY, property = "@type", visible = true)
```
manstis, line 27: Did you want `@type` here?

addons/common/tracing/tracing-api/src/main/java/org/kie/kogito/tracing/event/message/models/DecisionMessage.java
```diff
@@ -0,0 +1,46 @@
...
+@JsonInclude(NON_NULL)
+public final class DecisionMessage extends Message {
+
+    private MessageFEELEvent feelEvent;
+
+    private DecisionMessage() {
+        // needed for serialization
+    }
+
+    public DecisionMessage(MessageLevel level, MessageCategory category, String type, String sourceId, String text, MessageFEELEvent feelEvent, MessageExceptionField exception) {
```
manstis, line 37: Shouldn't `type` be hard-coded as a literal as it's used to determine the sub-class when deserialising?

For example `decision`?

addons/common/tracing/tracing-api/src/main/java/org/kie/kogito/tracing/event/model/ModelEvent.java
```diff
@@ -0,0 +1,64 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+/**
+ * Abstract <code>ModelEvent</code> to be extended by actual model-specific implementations
+ */
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.EXISTING_PROPERTY, property = "@type", visible = true)
```
manstis, line 31: Did you want `@type` here too?

manstis: Shouldn't this class also have a `type` field, as you're telling Jackson to use an existing field?

gitgabrio: @manstis 
You're right - I'm fixing that right now. Thanks for spotting that!

addons/common/tracing/tracing-api/src/main/java/org/kie/kogito/tracing/event/message/Message.java
gitgabrio: @manstis 
Yes, I would like to use `@type` to differentiate "service" information (not used by final user) from "business" ones

addons/common/tracing/tracing-api/src/main/java/org/kie/kogito/tracing/event/message/models/DecisionMessage.java
gitgabrio: @manstis 
Here, that String "type" has a different meaning, e.g. it could be one possible value from `InternalMessageType` (see 
[here](https://github.com/kiegroup/kogito-runtimes/blob/dcfccd7593c4f3242360c3025c06c33e4aa1ff7a/addons/common/tracing/decision-common/src/main/java/org/kie/kogito/tracing/decision/event/EventUtils.java#L63).

That's another reason to prefix "service-only" properties with `@`

gitgabrio:
@manstis @danielezonca 
I'm trying to simplify a little bit. For ex, there are different enums that, at the end of the day, express always the same concept (i.e. the operation being executed: decision vs process vs... )
I created a specific enum and I've used it overall, from kogito-runtimes to kogito-apps.

addons/common/tracing/tracing-api/src/main/java/org/kie/kogito/tracing/event/model/ModelEvent.java
```diff
@@ -0,0 +1,75 @@
...
+@JsonSubTypes({
+        @JsonSubTypes.Type(value = DecisionModelEvent.class, name = "DECISION"),
+})
+@JsonIgnoreProperties(ignoreUnknown = true)
```
manstis, line 35: This annotation isn't inherited so you should really put it on concrete sub-classes too.

addons/common/tracing/tracing-api/src/test/java/org/kie/kogito/tracing/event/TracingTestUtils.java
```diff
@@ -0,0 +1,58 @@
...
+        try (BufferedReader br = new BufferedReader(new InputStreamReader(inputStream))) {
+            String line;
+            while ((line = br.readLine()) != null) {
+                resultStringBuilder.append(line).append("\n");
+            }
+        }
+        return resultStringBuilder.toString();
+    }
+
+    public TracingTestUtils() {
```
manstis, line 56: I thought the recommendation was for a `private` constructor on utility classes with `static` methods?

manstis (approved):

addons/common/tracing/tracing-api/src/main/java/org/kie/kogito/tracing/event/model/ModelEvent.java
gitgabrio: @manstis 
This is wanted. I mean: abstract parent may not know of new fields declared from children classes, but on the other side I think it is better that concrete implementations does not allow "unknown" fields: does this make sense ?

addons/common/tracing/tracing-api/src/test/java/org/kie/kogito/tracing/event/TracingTestUtils.java
gitgabrio: :+1: @manstis

addons/common/tracing/tracing-api/src/main/java/org/kie/kogito/tracing/event/model/ModelEvent.java
manstis: OK, understood. I thought you was trying to ignore unknown fields on all sub-classes too :+1:

gitgabrio:
Jenkins retest this

jiripetrlik (approved):

danielezonca (approved):

PR #1645, opened 2021-10-05
KOGITO-5807 Public API: Share ObjectMapper instances with Quarkus

evacchi:
Re-use the Quarkus ObjectMapper instances in the Quarkus API modules.

This makes the ObjectMapper implementation dependent on the run-time environment.
As a result, the behavior may be inconsistent **across** runtimes (i.e. Kogito at some point may behave differently depending if it's running stand-alone, on Quarkus, or Spring Boot), however it will be consistent **with** the runtime.

Contra: this leaks even more that the conversion between types is using Jackson under the hood. However, this is already showing when `DataContext` is used as a param or a return type in JAX-RS endpoints.


api/kogito-api-incubation-common/pom.xml
```diff
@@ -37,14 +37,18 @@
...
         </dependency>
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-engine</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
```
tarilabs, line 51: not entirely sure the intent... is this for `test` scope or is a dependency?

quarkus/extensions/kogito-quarkus-rules-extension/kogito-quarkus-rules/src/main/java/org/kie/kogito/core/rules/incubation/quarkus/support/RuleUnitDataObjectCustomizer.java
```diff
@@ -0,0 +1,35 @@
...
+ * Internal use only. Wraps and configures a custom ObjectMapper for DataSources
+ */
+@ApplicationScoped
+class RuleUnitDataObjectCustomizer implements ObjectMapperCustomizer {
+
+    @Override
+    public void customize(ObjectMapper objectMapper) {
+        objectMapper.registerModule(new RuleUnitDataJacksonModule());
+    }
+}
```
tarilabs, line 35: so does it means I should make an analogous for DMN where I register there the needed serializer/deserializer for Quarkus platform only?
And in the REST codegen for Quarkus use then the Quarkus OM, while leave as-is for Spring Boot ?

tarilabs (commented):
couple of comments, thanks!

/cc @danielezonca

quarkus/extensions/kogito-quarkus-rules-extension/kogito-quarkus-rules/src/main/java/org/kie/kogito/core/rules/incubation/quarkus/support/RuleUnitDataObjectCustomizer.java
evacchi: correct. And when we refactor to share+port APIs to it, we'll do similarly for Spring Boot

api/kogito-api-incubation-common/pom.xml
evacchi: it was already like this, it was just moved (as you can see); however you are right that we should not depend on it, instead I just added a dependency to jackson-annotations (it is used for MapDataContext serialization)

alternatively, we may configure the object mapper programmatically (?) I think depending on the annotations should be fine for now though

quarkus/extensions/kogito-quarkus-extension-common/kogito-quarkus-common/src/main/java/org/kie/kogito/incubation/common/objectmapper/quarkus/QuarkusInternalObjectMapper.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+@Singleton
+public class QuarkusInternalObjectMapper implements InternalObjectMapper {
+
+    @Override
+    public <T> T convertValue(Object self, Class<T> type) {
+        if (type.isInstance(self)) {
+            return type.cast(self);
+        }
+        QuarkusObjectMapperCustomizer customizer =
```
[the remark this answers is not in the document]
evacchi: I reworked this after some advice from @danielezonca

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/MapperLoader.java
```diff
@@ -0,0 +1,28 @@
...
+package org.kie.kogito.incubation.common;
+
+import java.util.ServiceLoader;
+
+import org.kie.kogito.incubation.common.objectmapper.InternalObjectMapper;
+
+class MapperLoader {
+    public static InternalObjectMapper objectMapper() {
+        return ServiceLoader.load(InternalObjectMapper.class).findFirst()
+                .orElseThrow();
```
danielezonca, line 26: I think this should not happen but can you add an error message?

quarkus/extensions/kogito-quarkus-rules-extension/kogito-quarkus-rules/src/main/java/org/kie/kogito/core/rules/incubation/quarkus/support/RuleUnitDataObjectCustomizer.java
```diff
@@ -0,0 +1,35 @@
...
+package org.kie.kogito.core.rules.incubation.quarkus.support;
+
+import javax.enterprise.context.ApplicationScoped;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import io.quarkus.jackson.ObjectMapperCustomizer;
+
+/**
+ * Internal use only. Wraps and configures a custom ObjectMapper for DataSources
```
danielezonca, line 26: What does this comment mean? In reality this customizer is applied to the "shared" ObjectMapper so the additional behaviour is not "internal"

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/MapperLoader.java
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+
+package org.kie.kogito.incubation.common;
+
+import java.util.ServiceLoader;
+
+import org.kie.kogito.incubation.common.objectmapper.InternalObjectMapper;
+
+class MapperLoader {
```
danielezonca, line 23: I think it should be better to make this class `public` so that we can avoid to add the same code 
```java
ServiceLoader.load(InternalObjectMapper.class).findFirst().orElseThrow();
```
Every time we need to load the ObjectMapper (`QuarkusStraightThroughProcessService`, `QuarkusRuleUnitService`)
Wdyt?

api/kogito-api-incubation-common/src/test/java/org/kie/kogito/incubation/common/DataContextTest.java
```diff
@@ -96,7 +96,7 @@ public void nestedValue() {
 
         MapDataContext ctx = u.as(MapDataContext.class);
         assertNotEquals(Address.class, ctx.get("addr").getClass());
-        Address addr = InternalObjectMapper.convertValue(ctx.get("addr"), Address.class);
+        Address addr = new MyObjectMapper().convertValue(ctx.get("addr"), Address.class);
```
danielezonca, line 99:
```suggestion
        Address addr = MapperLoader.objectMapper().convertValue(ctx.get("addr"), Address.class);
```

danielezonca (approved):
Few minor comments but overall looks good to me

quarkus/extensions/kogito-quarkus-rules-extension/kogito-quarkus-rules/src/main/java/org/kie/kogito/core/rules/incubation/quarkus/support/RuleUnitDataObjectCustomizer.java
evacchi: outdated comment. will update

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/MapperLoader.java
evacchi: Since it is a one-liner I would rather not make it public

evacchi: I decided I can move it directly to `InternalObjectMapper` itself, since it must be public anyway

tarilabs (approved):
okay with me thank you

tiagodolphine (approved):

PR #1649, opened 2021-10-06
[KOGITO-6052] Missing Process instance

fjtirado:
Process instance should be added to the list of instances before process
execution start.


jbpm/jbpm-flow/src/main/java/org/kie/kogito/process/impl/AbstractProcessInstance.java
```diff
@@ -222,9 +222,9 @@ public void start(String trigger, String referenceId) {
         getProcessRuntime().getProcessInstanceManager().addProcessInstance(this.processInstance);
         this.id = processInstance.getStringId();
         addCompletionEventListener();
-        KogitoProcessInstance processInstance = getProcessRuntime().getKogitoProcessRuntime().startProcessInstance(this.id, trigger);
-        addToUnitOfWork(pi -> ((MutableProcessInstances<T>) process.instances()).create(pi.id(), pi));
```
fjtirado, line 226: why were doing that? there should be a good readon, but I cannot find any. Lets recap:
1) id process changing after process execution? that should not happen
2) persistence? I hardly see how it should affect this
My point is that the list needs to be filled before we start real execution or things like the one reported in the jira scenario will occur

PR #1650, opened 2021-10-07
Increase GHA timeout for kogito-examples

evacchi:
PRs are timing out for kogito-examples because they now exceed 120 minutes.


radtriste:
I mean here: https://github.com/kiegroup/kogito-examples/blob/stable/.github/workflows/ci-pr.yml#L10

PR #1663, opened 2021-10-13
KOGITO-5841  Public API: Allow extra-metadata in DataContext

evacchi:
**This is intended also as a RFC**.

related: https://github.com/kiegroup/kogito-examples/pull/941

Introduce the `ExtendedDataContext` that services may opt-in to return. 

An `ExtendedDataContext` contains a `meta()` and a `data()` section. 

```java
public final class ExtendedDataContext implements DataContext {
      public DataContext meta();
      public DataContext data();
} 
```

`ExtendedDataContext.as(T)` is equivalent to `ExtendedDataContext.data().as(T)` unless `T=ExtendedDataContext`

Thus, to access MetaData users may write:

```
DataContext ctx = svc.evaluate(id, input);
ExtendedDataContext ectx = ctx.as(ExtendedDataContext.class);
DataContext meta = ectx.meta() 
```

`meta` and `data` are both `DataContexts` so they may be converted to another type as any other `DataContext`.

For instance, engines may provide their own typed context; e.g:

```java
DMNMetaDataContext meta = ectx.meta().as(DMNMetaDataContext.class); 
```

If a service does not return an `ExtendedDataContext`, then conversion to `ExtendedDataContext` is equivalent to wrapping `ExtendedDataContext` around the original context; as such:

```java
MapDataContext mdc = MapDataContext.create();
mdc.set(......);
ExtendedDataContext edc = mdc.as(ExtendedDataContext.class);
// edc.data().equals(mdc)
```

### DIFF
--- a/quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions-integration-test/src/test/java/org/kie/kogito/quarkus/dmn/CustomEndpointIT.java
+++ b/quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions-integration-test/src/test/java/org/kie/kogito/quarkus/dmn/CustomEndpointIT.java
@@ -33,10 +33,10 @@ public class CustomEndpointIT {
                 .contentType(ContentType.JSON)
                 .accept(ContentType.JSON)
                 .body(DMNIT.ADULT_PAYLOAD)
-                .post("/dmnModel")
+                .post("/custom")
                 .then()
                 .statusCode(200)
-                .body("d.Hello", Matchers.is("Hello Luca"));
+                .body("data.d.Hello", Matchers.is("Hello Luca"));
     }
 
 }
--- a/quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions/src/main/java/org/kie/kogito/core/decision/incubation/quarkus/support/QuarkusDecisionService.java
+++ b/quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions/src/main/java/org/kie/kogito/core/decision/incubation/quarkus/support/QuarkusDecisionService.java
@@ -39,7 +39,7 @@ public class QuarkusDecisionService implements DecisionService {
     Instance<DecisionModels> decisionModelsInstance;
 
     @Override
-    public DataContext evaluate(LocalId decisionId, DataContext inputContext) {
+    public ExtendedDataContext evaluate(LocalId decisionId, DataContext inputContext) {
         LocalDecisionId localDecisionId;
         LocalDecisionServiceId decisionServiceId = null;
         if (decisionId instanceof LocalDecisionId) {
--- a/quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions/src/main/java/org/kie/kogito/core/decision/incubation/quarkus/support/QuarkusDecisionService.java
+++ b/quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions/src/main/java/org/kie/kogito/core/decision/incubation/quarkus/support/QuarkusDecisionService.java
@@ -26,6 +26,7 @@ import org.kie.kogito.decision.DecisionModel;
 import org.kie.kogito.decision.DecisionModels;
 import org.kie.kogito.dmn.rest.DMNJSONUtils;
 import org.kie.kogito.incubation.common.DataContext;
+import org.kie.kogito.incubation.common.ExtendedDataContext;
 import org.kie.kogito.incubation.common.LocalId;
 import org.kie.kogito.incubation.common.MapDataContext;
 import org.kie.kogito.incubation.decisions.LocalDecisionId;
@@ -69,6 +70,8 @@ public class QuarkusDecisionService implements DecisionService {
                     ctx, decisionServiceId.serviceId());
         }
 
-        return MapDataContext.of(dmnResult.getContext().getAll());
+        MapDataContext meta = MapDataContext.of(dmnResult.getContext().getMetadata().asMap());
+        MapDataContext data = MapDataContext.of(dmnResult.getContext().getAll());
+        return ExtendedDataContext.of(meta, data);
     }
 }
--- a/quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions/src/main/java/org/kie/kogito/core/decision/incubation/quarkus/support/QuarkusDecisionService.java
+++ b/quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions/src/main/java/org/kie/kogito/core/decision/incubation/quarkus/support/QuarkusDecisionService.java
@@ -16,19 +16,19 @@
 
 package org.kie.kogito.core.decision.incubation.quarkus.support;
 
+import java.util.Map;
+
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.inject.Instance;
 import javax.inject.Inject;
 
 import org.kie.dmn.api.core.DMNContext;
+import org.kie.dmn.api.core.DMNMetadata;
 import org.kie.dmn.api.core.DMNResult;
 import org.kie.kogito.decision.DecisionModel;
 import org.kie.kogito.decision.DecisionModels;
 import org.kie.kogito.dmn.rest.DMNJSONUtils;
-import org.kie.kogito.incubation.common.DataContext;
-import org.kie.kogito.incubation.common.ExtendedDataContext;
-import org.kie.kogito.incubation.common.LocalId;
-import org.kie.kogito.incubation.common.MapDataContext;
+import org.kie.kogito.incubation.common.*;
 import org.kie.kogito.incubation.decisions.LocalDecisionId;
 import org.kie.kogito.incubation.decisions.LocalDecisionServiceId;
 import org.kie.kogito.incubation.decisions.services.DecisionService;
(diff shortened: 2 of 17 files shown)


gitgabrio:
Hi @evacchi Thanks for the PR :+1: 
I am doubtful about the usage of `DataContext`. I have the impression (I may be wrong) that almost every kind of "DTO" is going to extend it. At the same time, I think that "MetaData" are semantically pretty different then "calculation" data. I'm pretty sure that MetaData will be used for completely different stuff then "calculation" Data, so they have to express different behavior to have some usefulness. To dig further, I can easily imagine "MetaData" used for "meta-analysis", and mostly managed as string (e.g. inside an ELK setup) while "Data" are used/meant for actual evaluation of results; and possible query/aggregation operations on them would be different. At the same time, the code should "forbid" usage of one instead of the other.
The issues I see are:
1) inside `ExtendedDataContext` it is only the order of arguments to define if the argument is a "MetaData" or a "Data"
2) nothing prevent to (wrongly) use a "MetaData" as "Data", or the other way around
3) for my understanding of what a "MetaData" may be and may be used for, I'm not even sure it make sense for a MetaData to be "castable" (that is the only actual "behavior" of a DataContext).

As far as I may understand, my first suggestion would be to completely separate the hierarchy of "Data" and "MetaData". If, ever, the twos should express some common behavior, I think it would be better to use a composition pattern - i.e. an utility class that may be used by both, keeping them anyway separated. Does this make sense ?

quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions/src/main/java/org/kie/kogito/core/decision/incubation/quarkus/support/QuarkusDecisionService.java
```diff
@@ -69,6 +70,8 @@ public DataContext evaluate(LocalId decisionId, DataContext inputContext) {
                     ctx, decisionServiceId.serviceId());
         }
 
-        return MapDataContext.of(dmnResult.getContext().getAll());
+        MapDataContext meta = MapDataContext.of(dmnResult.getContext().getMetadata().asMap());
```
tarilabs, line 73: I am afraid this will always be empty, since a few lines above you are passing only `inputContext.as(MapDataContext.class).toMap()` to the DMNContext builder, so the incoming metadata are lost at that stage?

If it's true, then maybe easier to build `MapDataContext meta` based on the input `DataContext inputContext` is it's containing metadata?

Wdyt?

quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions/src/main/java/org/kie/kogito/core/decision/incubation/quarkus/support/QuarkusDecisionService.java
```diff
@@ -69,6 +70,8 @@ public DataContext evaluate(LocalId decisionId, DataContext inputContext) {
...
+        MapDataContext data = MapDataContext.of(dmnResult.getContext().getAll());
+        return ExtendedDataContext.of(meta, data);
```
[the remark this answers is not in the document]
tarilabs: imho, we should show `.data()` in the body of the custom examples, so not to loose the "kogito ergo domain" which was always put forward as an imperative by all the Kogito Leads.

wdyt?

quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions-integration-test/src/test/java/org/kie/kogito/quarkus/dmn/CustomEndpointIT.java
```diff
@@ -33,10 +33,10 @@ public void testHelloEndpoint() {
                 .contentType(ContentType.JSON)
                 .accept(ContentType.JSON)
                 .body(DMNIT.ADULT_PAYLOAD)
-                .post("/dmnModel")
+                .post("/custom")
                 .then()
                 .statusCode(200)
-                .body("d.Hello", Matchers.is("Hello Luca"));
+                .body("data.d.Hello", Matchers.is("Hello Luca"));
```
tarilabs, line 39: ref comment

tarilabs (commented):
I think this is going in a good direction; thanks!
Few comments as requested

quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions/src/main/java/org/kie/kogito/core/decision/incubation/quarkus/support/QuarkusDecisionService.java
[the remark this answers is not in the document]
evacchi: if I have understood your question, it is: https://github.com/kiegroup/kogito-examples/pull/941/files

quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions/src/main/java/org/kie/kogito/core/decision/incubation/quarkus/support/QuarkusDecisionService.java
evacchi: I think you are right. Let me look into it!

quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions/src/main/java/org/kie/kogito/core/decision/incubation/quarkus/support/QuarkusDecisionService.java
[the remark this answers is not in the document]
tarilabs: about https://github.com/kiegroup/kogito-runtimes/pull/1663#discussion_r740970901 that is what I would expect, but then I don't understand what about this: https://github.com/kiegroup/kogito-runtimes/pull/1663#discussion_r740957580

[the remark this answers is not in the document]
evacchi: gotcha. As far as "hand-written" end point are concerned, I'd say the "domain-specific" part is all in the writer's hands. 

The integration test is testing that returning the entire payload (including metadata) behaves as it should; users will then choose whether they want to return the entire data+meta section or just data.

I can also make it return `data`. I honestly don't have a specific preference

quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions/src/main/java/org/kie/kogito/core/decision/incubation/quarkus/support/QuarkusDecisionService.java
evacchi: as I was doing it, I realized that I wasn't even accounting for meta-data in the _input_. I suppose either I forgot or I was planning to revisit it later. Either way, with the last commit it should work for the DMN. The only detail that I have left out, there is no `MetaDataContext#as(Class<T>)`, but for now it is possible to use `MapDataContext.from(mdc)`. 

For this last point because I am still unconvinced whether we want to have the `.as()` method on all such classes, although it looks nice and it is intuitive to use; I should probably investigate this as part of https://github.com/kiegroup/kogito-runtimes/pull/1655

quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions/src/main/java/org/kie/kogito/core/decision/incubation/quarkus/support/QuarkusDecisionService.java
```diff
@@ -60,7 +60,17 @@ public ExtendedDataContext evaluate(LocalId decisionId, DataContext inputContext
...
+        ExtendedDataContext extendedDataContext = inputContext.as(ExtendedDataContext.class);
+
+        Map<String, Object> map = extendedDataContext.data().as(MapDataContext.class).toMap();
+        DMNContext ctx = DMNJSONUtils.ctx(decisionModel, map);
+        MetaDataContext inputMeta = extendedDataContext.meta();
+        MapDataContext mapInputMeta = MapDataContext.from(inputMeta);
+        DMNMetadata metadata = ctx.getMetadata();
+        for (Map.Entry<String, Object> kv : mapInputMeta.toMap().entrySet()) {
+            metadata.set(kv.getKey(), kv.getValue());
+        }
```
tarilabs, line 72: discussed with Edo and this could be reviewed later for ad-hoc implementation 👍

tarilabs (approved):

quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions-integration-test/src/test/java/org/kie/kogito/quarkus/dmn/CustomEndpointIT.java
tarilabs: this is a change to Quarkus IT inside of -runtimes, 👍

quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions/src/main/java/org/kie/kogito/core/decision/incubation/quarkus/support/QuarkusDecisionService.java
tarilabs: agreed, https://github.com/kiegroup/kogito-runtimes/pull/1663#discussion_r741765116

danielezonca (approved):

PR #1696, opened 2021-11-05
[KOGITO-6195] Adding JQ Expression native support

fjtirado:
Example in https://github.com/kiegroup/kogito-examples/pull/969


ricardozanini (approved):
Looks neat! My only concern is about the new `kogito-jackson-utils` module. I don't have a better idea, thou. I can see a `utils` parent module in the future.

fjtirado:
jenkins rerun kogito-apps tests

PR #1710, opened 2021-11-11
[DROOLS-6645] remove xml dependencies from drools core and compiler

mariofusco:
JIRA: https://issues.redhat.com/browse/DROOLS-6645

Relate pull requests:
https://github.com/kiegroup/drools/pull/4006
https://github.com/kiegroup/kogito-runtimes/pull/1710
https://github.com/kiegroup/optaplanner/pull/1666


api/kogito-legacy-api/pom.xml
```diff
@@ -52,5 +52,9 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.drools</groupId>
+      <artifactId>drools-xml-support</artifactId>
+    </dependency>
```
danielezonca, line 58: Why? I don't expect this module to have XML-related behaviour 🤔 
Is it because `kogito-legacy-api` requires user to create `kmodule.xml` and this enables the parsing of it?
If so please add a comment

danielezonca: Btw `kmodule` parsing is done at compile time as far as I remember so it should be possible to have this dependency only at compile time (i.e. [`kogito-codegen-rules`](https://github.com/kiegroup/kogito-runtimes/tree/main/kogito-codegen-modules/kogito-codegen-rules))

PR #1692, opened 2021-11-04
BXMSPROD-1540 composite actions centralized github actions

Ginxo:

PR #1729, opened 2021-11-17
BXMSPROD-1540 composite actions centralized github actions

radtriste:
let's try again https://github.com/kiegroup/kogito-runtimes/pull/1692 and find what is the problem with the runtimes build.

We need to know that this problem does not happen on other repositories but only on runtimes.
Noticeable change: Use https://github.com/s4u/setup-maven-action but this is for all.

Checkout hierarchy:
- https://github.com/s4u/setup-maven-action checkout kogito-runtimes in `/home/runner/work/kogito-runtimes/kogito-runtimes`
- build-chain checkout drools in `/home/runner/work/kogito-runtimes/kogito-runtimes/kiegroup_drools`
- build-chain checkout kogito-runtimes in `/home/runner/work/kogito-runtimes/kogito-runtimes/kiegroup_kogito_runtimes`

### DIFF
--- a/.github/workflows/kogito-apps-pr.yml
+++ b/.github/workflows/kogito-apps-pr.yml
@@ -22,52 +22,33 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         java-version: [11]
+        maven-version: ['3.8.1']
       fail-fast: false
     runs-on: ${{ matrix.os }}
     name: Maven Build
     steps:
-      - name: Disk space report before modification
-        shell: bash
-        run: |
-          echo "Available storage:"
-          df -h
-      # Inspired to maximize-build-space action https://github.com/easimon/maximize-build-space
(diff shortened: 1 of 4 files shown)


radtriste:
@Ginxo so it seems indeed the bug was introduced with the use of https://github.com/s4u/setup-maven-action
(see kogito-runtimes GHA build)

radtriste:
@Ginxo should we fallback the composite action and use the "old style" maven/java installation ?

Ginxo:
@radtriste I agree (just for the java/maven step) adding a comment there explaining the reason to keep it. ok?

radtriste:
updated https://github.com/radtriste/kogito-pipelines/blob/test_composite/.ci/actions/maven/action.yml, let's what the kogito-runtimes check is doing :)

radtriste:
Follow-up on https://github.com/kiegroup/kogito-runtimes/pull/1834

PR #1735, opened 2021-11-22
[KOGITO-6282] Deleting "/metrics" classes. Refactored MetricsResources

gitgabrio:
@danielezonca @evacchi  @jiripetrlik

See https://issues.redhat.com/browse/KOGITO-6282

Depends on 

https://github.com/kiegroup/kogito-examples/pull/990


jiripetrlik (approved):

addons/common/monitoring/prometheus/src/main/java/org/kie/kogito/monitoring/prometheus/common/rest/MetricsResource.java
```diff
@@ -15,11 +15,9 @@
  */
 package org.kie.kogito.monitoring.prometheus.common.rest;
 
-import org.kie.kogito.monitoring.prometheus.common.PrometheusRegistryProvider;
-
-public abstract class MetricsResource {
+/**
+ * Needed only to notify AddonsConfigDiscovery about prometheus availability
```
danielezonca, line 19: We used [a similar comment](https://github.com/kiegroup/kogito-runtimes/blob/main/addons/common/persistence/postgresql/src/main/java/org/kie/kogito/persistence/KogitoProcessInstancesFactory.java#L25) in other addons 
```suggestion
 * This class must always have exact FQCN as <code>org.kie.kogito.monitoring.prometheus.common.rest. MetricsResource </code>
```

danielezonca (approved):
Minor comment but LGTM 👍

gitgabrio:
@danielezonca 
On second thought, I think it would be better to keep a "metrics" endpoint, doing redirection to actual ones (/q/metrics vs /actuator/prometheus)
That way
1) the environment looks more consistent
2) the "implementation detail" is hidden

WDYT ?

danielezonca:
> @danielezonca On second thought, I think it would be better to keep a "metrics" endpoint, doing redirection to actual ones (/q/metrics vs /actuator/prometheus) That way
> 
> 1. the environment (kogito) looks more consistent
> 2. the "implementation detail" (metrics provider) is hidden
> 
> See last commit [f96b16e](https://github.com/kiegroup/kogito-runtimes/commit/f96b16e82e73c36bc562f74053d0df5df0a9f8ec)

The idea for monitoring is to leverage the platform (Quarkus/SpringBoot) for configuration and metrics exposition so I don't like to have an additional endpoint just for Kogito: what if other projects follow the same approach? I.e. Camel produce a different metrics endpoint etc.
About point 2 I'm not sure to understand, which implementation detail are we exposing?

gitgabrio:
> > @danielezonca On second thought, I think it would be better to keep a "metrics" endpoint, doing redirection to actual ones (/q/metrics vs /actuator/prometheus) That way
> > 
> > 1. the environment (kogito) looks more consistent
> > 2. the "implementation detail" (metrics provider) is hidden
> > 
> > See last commit [f96b16e](https://github.com/kiegroup/kogito-runtimes/commit/f96b16e82e73c36bc562f74053d0df5df0a9f8ec)
> 
> The idea for monitoring is to leverage the platform (Quarkus/SpringBoot) for configuration and metrics exposition so I don't like to have an additional endpoint just for Kogito: what if other projects follow the same approach? I.e. Camel produce a different metrics endpoint etc. About point 2 I'm not sure to understand, which implementation detail are we exposing?

Quarkus/Springboot are orthogonal respect to Camel (and other frameworks) AFAIK - so other frameworks (camel, kafka, whatever) are out of this discussion.
We are offering "kogito" as developing platform, with two possible context/environments; Quarkus/Springboot.
Final user should be able to switch easily between them (as most as possible).
Now, we are also exposing "metrics" in our environment: regardless of actual implementation (quarkus vs srpingboot) user should not be forced to 
1) going through documentation to find out the correct endpoint for the given environment
2) changing them all over his/her code whenever he/she try to switch to a different environment

The common "metrics" endpoint  was provided before KOGITO-5931, and by itself (i.e. regardless of the implementation) I think it was a good thing.

Keeping in mind all the modifications required on our side, and maybe also on customer side, for the "metrics" removal should give you an idea of the impact it has.

Beside that, keeping the "/metrics" endpoint (behind redirect) makes this PR backward compatible, which is not a bad thing IMO.....

gitgabrio:
@danielezonca @vaibhavjainwiz: The dependent jira (https://issues.redhat.com/browse/KOGITO-6293) does not need this one. The new metrics endpoints have been created in a previous Ticket (https://issues.redhat.com/browse/KOGITO-5931), so switch from "/metrics" to "/q/metrics" or "/actuator/prometheus" should already work in master. This PR is only about removing the "/metrics" endpoint, it does not add anything new

vaibhavjainwiz:
> @danielezonca @vaibhavjainwiz: The dependent jira (https://issues.redhat.com/browse/KOGITO-6293) does not need this one. The new metrics endpoints have been created in a previous Ticket (https://issues.redhat.com/browse/KOGITO-5931), so switch from "/metrics" to "/q/metrics" or "/actuator/prometheus" should already work in master. This PR is only about removing the "/metrics" endpoint, it does not add anything new

ok, thanks for the update. We will merge operator PR

PR #1737, opened 2021-11-22
[DROOLS-6707] move sessions to the new drools-kiesession module

mariofusco:
**JIRA**: https://issues.redhat.com/browse/DROOLS-6707

Related PRs
https://github.com/kiegroup/drools/pull/4035
https://github.com/kiegroup/kogito-runtimes/pull/1737


drools/drools-core/pom.xml
```diff
@@ -42,6 +42,10 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.drools</groupId>
+      <artifactId>drools-kiesession</artifactId>
+    </dependency>
```
danielezonca, line 48: With this dependency we should revisit at least the name of this module (if we cannot remove it) because it is no more a `drools-core` equivalent.
We might use `drools-core-engine` or `drools-core-integration` or `drools-wrapper` or similar 🤔 
Btw we can do this review in another ticket, I personally think we should review (and remove) all/most of [these modules](https://github.com/kiegroup/kogito-runtimes/tree/main/drools) now

drools/drools-core/src/main/resources/META-INF/kie.conf
```diff
@@ -1,2 +1,2 @@
-org.drools.core.reteoo.KieComponentFactoryFactory=org.drools.core.kogito.factory.KogitoComponentFactoryFactory
+org.drools.core.reteoo.RuntimeComponentFactory=org.drools.core.kogito.factory.KogitoKieComponentFactory;2
```
danielezonca, line 1: Please rename `KogitoKieComponentFactory` as `KogitoRuntimeComponentFactory`

mariofusco: Done.

PR #1741, opened 2021-11-23
KOGITO-5962: Job Service Messaging API for Jobs Scheduling

wmedvede:

- [x] Your code is properly formatted according to [this configuration](https://github.com/kiegroup/kogito-runtimes/tree/main/kogito-build/kogito-ide-config)
- [x] Pull Request title contains the target branch if not targeting main: `[0.9.x] KOGITO-XYZ Subject`

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

* <b>Run all builds</b>  

  Please add comment: <b>Jenkins (re)run [kogito-runtimes|optaplanner|kogito-apps|kogito-examples|optaplanner-quickstarts|optaweb-employee-rostering|optaweb-vehicle-routing] tests</b>
 
* <b>Quarkus LTS checks</b>  

  Please add comment: <b>Jenkins (re)run [kogito-runtimes|optaplanner|kogito-apps|kogito-examples|optaplanner-quickstarts|optaweb-employee-rostering|optaweb-vehicle-routing] LTS</b>

* <b>Native checks</b>  

  Please add comment: <b>Jenkins (re)run [kogito-runtimes|optaplanner|kogito-apps|kogito-examples|optaplanner-quickstarts|optaweb-employee-rostering|optaweb-vehicle-routing] native</b>

</details>

### DIFF
--- a/addons/common/jobs/api/pom.xml
+++ b/addons/common/jobs/api/pom.xml
@@ -13,6 +13,22 @@
 
   <dependencies>
     <dependency>
+      <groupId>io.cloudevents</groupId>
+      <artifactId>cloudevents-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.cloudevents</groupId>
+      <artifactId>cloudevents-json-jackson</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.datatype</groupId>
+      <artifactId>jackson-datatype-jsr310</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.kie.kogito</groupId>
+      <artifactId>kogito-events-api</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.junit.jupiter</groupId>
       <artifactId>junit-jupiter</artifactId>
       <scope>test</scope>
--- a/addons/common/jobs/api/src/main/java/org/kie/kogito/jobs/api/event/ProcessInstanceContextJobCloudEvent.java
+++ b/addons/common/jobs/api/src/main/java/org/kie/kogito/jobs/api/event/ProcessInstanceContextJobCloudEvent.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.kogito.jobs.api.event;
+
+import org.kie.kogito.event.CloudEventExtensionConstants;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public abstract class ProcessInstanceContextJobCloudEvent<T> extends JobCloudEvent<T> {
+
+    //Process context extensions
+    @JsonProperty(CloudEventExtensionConstants.PROCESS_INSTANCE_ID)
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    private String processInstanceId;
+
+    @JsonProperty(CloudEventExtensionConstants.PROCESS_ID)
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    private String processId;
+
+    @JsonProperty(CloudEventExtensionConstants.PROCESS_ROOT_PROCESS_INSTANCE_ID)
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    private String rootProcessInstanceId;
+
+    @JsonProperty(CloudEventExtensionConstants.PROCESS_ROOT_PROCESS_ID)
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    private String rootProcessId;
+
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    @JsonProperty(CloudEventExtensionConstants.ADDONS)
+    private String kogitoAddons;
+
+    protected ProcessInstanceContextJobCloudEvent() {
+        // marshalling constructor.
+    }
+
+    public String getProcessInstanceId() {
+        return processInstanceId;
+    }
+
+    public String getProcessId() {
+        return processId;
+    }
+
+    public String getRootProcessInstanceId() {
+        return rootProcessInstanceId;
+    }
+
+    public String getRootProcessId() {
+        return rootProcessId;
+    }
+
+    public String getKogitoAddons() {
+        return kogitoAddons;
+    }
+
+    public void setProcessInstanceId(String processInstanceId) {
+        this.processInstanceId = processInstanceId;
+    }
+
+    public void setProcessId(String processId) {
+        this.processId = processId;
+    }
+
+    public void setRootProcessInstanceId(String rootProcessInstanceId) {
+        this.rootProcessInstanceId = rootProcessInstanceId;
+    }
+
+    public void setRootProcessId(String rootProcessId) {
+        this.rootProcessId = rootProcessId;
+    }
+
+    public void setKogitoAddons(String kogitoAddons) {
+        this.kogitoAddons = kogitoAddons;
+    }
+
+    @Override
+    public String toString() {
+        return "ProcessInstanceContextJobCloudEvent{" +
+                "processInstanceId='" + processInstanceId + '\'' +
+                ", processId='" + processId + '\'' +
+                ", rootProcessInstanceId='" + rootProcessInstanceId + '\'' +
+                ", rootProcessId='" + rootProcessId + '\'' +
+                ", kogitoAddons='" + kogitoAddons + '\'' +
+                "} " + super.toString();
+    }
+}
--- a/addons/common/jobs/api/src/main/java/org/kie/kogito/jobs/api/event/serialization/JobCloudEventDeserializer.java
+++ b/addons/common/jobs/api/src/main/java/org/kie/kogito/jobs/api/event/serialization/JobCloudEventDeserializer.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.kogito.jobs.api.event.serialization;
+
+import java.io.IOException;
+import java.util.Objects;
+
+import org.kie.kogito.jobs.api.Job;
+import org.kie.kogito.jobs.api.event.CancelJobRequestEvent;
+import org.kie.kogito.jobs.api.event.CreateProcessInstanceJobRequestEvent;
+import org.kie.kogito.jobs.api.event.JobCloudEvent;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import io.cloudevents.CloudEvent;
+import io.cloudevents.CloudEventData;
+
+import static org.kie.kogito.jobs.api.event.serialization.JobCloudEventSerializer.DEFAULT_OBJECT_MAPPER;
+
+public class JobCloudEventDeserializer {
+
+    private final ObjectMapper objectMapper;
+
+    public JobCloudEventDeserializer() {
+        this.objectMapper = DEFAULT_OBJECT_MAPPER;
+    }
+
+    public JobCloudEventDeserializer(ObjectMapper objectMapper) {
+        this.objectMapper = objectMapper;
+    }
+
+    public JobCloudEvent<?> deserialize(byte[] data) {
+        try {
+            CloudEvent cloudEvent = objectMapper.readValue(data, CloudEvent.class);
+            CloudEventData cloudEventData = Objects.requireNonNull(cloudEvent.getData(), "JobCloudEvent data field must not be null");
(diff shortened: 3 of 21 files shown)


wmedvede:
Hi @cristianonicolai @tiagodolphine would you mind review this PR?

Tests comes in upcoming additional commit.

Thanks!

addons/common/jobs/api/pom.xml
```diff
@@ -12,6 +12,22 @@
...
+      <artifactId>cloudevents-json-jackson</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.datatype</groupId>
+      <artifactId>jackson-datatype-jsr310</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.kie.kogito</groupId>
+      <artifactId>kogito-events-api</artifactId>
+    </dependency>
```
tiagodolphine, line 30: So basically we are adding cloud event dependencies under the API, in this way even if the user does not want to use events like using the "REST Job Addon" it will have unnecessary dependencies, so I'm wondering if it would make sense to create something like `job-events-api` that would have the `job-api` as a dependency or something like that... just brainstorming, wdyt?

tiagodolphine (commented):
LGTM in general, I just added com comments.

addons/common/jobs/api/src/main/java/org/kie/kogito/jobs/api/event/ProcessInstanceContextJobCloudEvent.java
```diff
@@ -0,0 +1,100 @@
...
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    private String rootProcessInstanceId;
+
+    @JsonProperty(CloudEventExtensionConstants.PROCESS_ROOT_PROCESS_ID)
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    private String rootProcessId;
+
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    @JsonProperty(CloudEventExtensionConstants.ADDONS)
+    private String kogitoAddons;
```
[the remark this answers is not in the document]
wmedvede: This comes from the cloud events spec. As part of the cloud event extensions, information about the "context" where the event was produced can be sent. (this is something that is also being done/used in the case of the data-index related events)
The spec admits this kind of "duplication". The idea is that the piece of information about the "context" (passed as part of extensions and not in the payload), can be used for example for doing "routing/pre-processing" or stuff at the receiver side, even before having to parse the information sent in the "data/payload" of the event itself. (the Job).

So let's say that I'm following this two things, "context" information not in the payload and somehow also following the pattern we already have with the process node and tasks events.

Opening a small thread @tiagodolphine @cristianonicolai, but considering that we aren't sending too much information I believe this is something we can implement. But I'm open so suggestions.

addons/common/jobs/api/pom.xml
wmedvede: yes, we'll probably have refactorings in next stages. But I'd prefer to no add new modules right now, until I'm in the new addon implementation. Mostly because something that I might believe is good today, maybe when I start the addon implementation I reach to a better name or structuring. So this is why I prefer to keep this in that module. But yeah, "job-events-api" can be a good place, I haven't thought a new name yet to be honest. As commented the idea was to not introduce new modules until the addon implementation iteration.

addons/common/jobs/api/src/main/java/org/kie/kogito/jobs/api/event/serialization/JobCloudEventDeserializer.java
```diff
@@ -0,0 +1,66 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+
+import io.cloudevents.CloudEvent;
+import io.cloudevents.CloudEventData;
+import io.cloudevents.jackson.JsonFormat;
+
+public class JobCloudEventDeserializer {
+
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper()
```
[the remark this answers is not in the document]
wmedvede: Not sure if we need a potential differentiation of this core serializing API depending if we are are in SprinBoot/Quarkus.
I mean, this module, basically relies on:

      io.cloudevents:cloudevents-api
      io.cloudevents:cloudevents-json-jackson
      com.fasterxml.jackson.datatype:jackson-datatype-jsr310

     and com.fasterxml.jackson in general.

When we go to SpringBoot the providers e.g. for jackson should be also be available there.

See this example: 

https://github.com/kiegroup/kogito-runtimes/blob/2df50ff62f7d0e9301f59662a6f6b11b6e7cc7b8/springboot/addons/events/kafka/pom.xml

So I believe that, unless we have a specific issue/requiriment se should use the same strategy.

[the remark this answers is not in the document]
tiagodolphine: what I mean, is to reause the already ObjectMapper instance under quarkus CDI context, the same used by the endpoints and whatever... instead of creating a new instance here, it would be injected.

addons/common/jobs/api/src/main/java/org/kie/kogito/jobs/api/event/ProcessInstanceContextJobCloudEvent.java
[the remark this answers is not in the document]
tiagodolphine: well but this is not a context, it is part of a payload, similar to the payload in the REST APIs when we do the POST... I just trying to make the APIs less coupled with processes, to be more generic because at some point we could even allow job scheduling for different resources other than processes. So let's say we could send an event to schedule a job, for a `rule` or any other kind of resource, other than process, but still use the same topic. Anyway, just a comment, I know, that job-service has processes information in many places...

addons/common/jobs/api/pom.xml
cristianonicolai: yeah I think is a good concern, but it is fine to iterate and evolve this to a separated module later on.

addons/common/jobs/api/src/main/java/org/kie/kogito/jobs/api/event/ProcessInstanceContextJobCloudEvent.java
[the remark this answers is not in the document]
cristianonicolai: @tiagodolphine I think this is fine is just metadata that will be sent out and ignored by the Jobs service. In case we do a decision event, we can take a similar approach, with attributes that are specific to a decision as metadata but the payload would be the same for Jobs service. @wmedvede sounds right?

addons/common/jobs/api/src/main/java/org/kie/kogito/jobs/api/event/serialization/JobCloudEventDeserializer.java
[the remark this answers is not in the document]
cristianonicolai: @wmedvede we should try to reuse the mapper available vi DI, which allows extra flexibility for end-users.

addons/common/jobs/api/src/main/java/org/kie/kogito/jobs/api/event/ProcessInstanceContextJobCloudEvent.java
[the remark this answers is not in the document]
wmedvede: @cristianonicolai yes, this is my understanding about sending metadata regarding the context where the event was produced. But the information about what to do is always in the payload. So I think that yes, maybe if in the future, if this event is produced by other things different than a process, then yes, metadata could include other attributes.

@tiagodolphine regarding to have something less coupled with the processes, I agree. But note that even the Job that is passed thorough the REST API is already coupled to processes. https://github.com/kiegroup/kogito-runtimes/blob/07bd4c9f75c5f3ef32e8e5772737204caaf90999/addons/common/jobs/api/src/main/java/org/kie/kogito/jobs/api/Job.java .

At some point I believe it could be nice to do a refactoring and enabling the jobs service, to manage jobs from different sources and completely decoupled from the source itself.

Some ideas:

    1) a REST job type that invokes whatever REST point that was passed as part of the create job request information, and passes the received payload back as the POST information when the endpoint is invoked.
        (looks similar to what we have now, but it's different, because, with this new approach we are kind of "returning" the payload to the interested party)
      
    2) a job that publishes the payload passed as part of the create job request information to an arbitrary kafka topic that was passed as parameter. (enables a kind of time programmed publishing kafka messages)

    3) etc.

But, going back to what we have now, I believe that it's fine to go with this proposal. And, probably even before 1) and 2) we should try to make what we have not to work with N instances of the job service running. Since our first goal is to support the Timer based events and also make all this more scalable, etc.

addons/common/jobs/api/src/main/java/org/kie/kogito/jobs/api/event/serialization/JobCloudEventDeserializer.java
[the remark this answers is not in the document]
wmedvede: The serializer/deserializer were refactored to have the chance to pass an ObjectMapper on instantiation.
Let's now go with this, and at the time of working in the add-ons side, these serializers can be created by passing the ObjectMapper provided by Quarkus/Spring as suggested. Maybe by providing an factory or whatever, but yeah, they can now be be instantiated by passing an ObjectMapper. (there's a default shared mapper in cases where we want to use them directly)

wmedvede:
Jenkins please retest this

wmedvede:
Jenkins please retest this

addons/common/jobs/api/src/main/java/org/kie/kogito/jobs/api/event/serialization/JobCloudEventSerializer.java
```diff
@@ -0,0 +1,55 @@
...
+import io.cloudevents.jackson.JsonFormat;
+
+public class JobCloudEventSerializer {
+
+    /**
+     * Shared default object mapper with the minimal required setups that can be used by the JobCloudEventDeserializer
+     * and JobCloudEventSerializer classes. Interested parties that doesn't want to use this object mapper can use
+     * their own provided mappers by using the proper constructors.
+     */
+    public static final ObjectMapper DEFAULT_OBJECT_MAPPER = new ObjectMapper()
```
cristianonicolai, line 33: @wmedvede could we reuse the objectmapper thats avaiable via CDI instead?

wmedvede: Hi @cristianonicolai , yes, this is what I've tried to explain. 
Right now, we support to create the Serializer by passing the ObjectMapper or not. By I don't have any injectable ObjectMapper at this point, since it'll be available when I start coding the event based plugin.
So, as we commented before, when I start the event based plugin, two things will happen:
1) this classes can be moved another package, probably job-events-api as suggested.
2) As part of the  event based plugin we'll be able to inject the ObjectMapper from:

   a) Quarkus -> @Inject ObjectMapper objectMapper;
   b) SpringBoot ->     @Autowired ObjectMapper mapper; 

But both injections are related with the event based plugin implementation so I prefer to tackle this at this point.

Do you agree?

cristianonicolai: @wmedvede yep, sounds good.

cristianonicolai (approved):

addons/common/jobs/api/src/test/java/org/kie/kogito/jobs/api/event/AbstractJobCloudEventTest.java
```diff
@@ -0,0 +1,56 @@
...
+package org.kie.kogito.jobs.api.event;
+
+import java.net.URI;
+import java.time.ZonedDateTime;
+
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+abstract class AbstractJobCloudEventTest<E extends JobCloudEvent<T>, T> {
```
MarianMacik, line 26: I don't see the second type parameter used anywhere, do we need it? Wouldn't something like:
```suggestion
abstract class AbstractJobCloudEventTest<E extends JobCloudEvent<?>> {
```
also work?

addons/common/jobs/api/src/test/java/org/kie/kogito/jobs/api/event/AbstractProcessInstanceContextJobCloudEventTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+abstract class AbstractProcessInstanceContextJobCloudEventTest<E extends ProcessInstanceContextJobCloudEvent<T>, T> extends AbstractJobCloudEventTest<E, T> {
+
+    static final String PROCESS_ID = "PROCESS_ID";
+    static final String PROCESS_INSTANCE_ID = "PROCESS_INSTANCE_ID";
+    static final String ROOT_PROCESS_INSTANCE_ID = "ROOT_PROCESS_INSTANCE_ID";
+    static final String ROOT_PROCESS_ID = "ROOT_PROCESS_ID";
+    static final String ADDONS = "ADDONS";
+
+    void assertFields(E event) {
```
MarianMacik, line 29: We could add `@Override` here.

MarianMacik (approved):
Looks good, just 2 minor improvements. Thanks!

addons/common/jobs/api/src/test/java/org/kie/kogito/jobs/api/event/AbstractJobCloudEventTest.java
wmedvede: added, thanks.

addons/common/jobs/api/src/test/java/org/kie/kogito/jobs/api/event/AbstractProcessInstanceContextJobCloudEventTest.java
wmedvede: added, thanks.

mbiarnes:
Jenkins run kogito-apps test

mbiarnes:
Jenkins retest this

wmedvede:
weird, it complains about  
Could not find artifact org.kie.kogito:kogito-quarkus-extension-spi:jar:2.0.0-SNAPSHOT -> [Help 2]
however the PR was re-re-re-re-rebased to incorporate this changes. Maybe this artifact wasn't updated into maven repo properly, however this shouldn't be a problem since the artifact belongs to this project

wmedvede:
Jenkins please retest this

cristianonicolai:
Jenkins rerun kogito-apps tests

cristianonicolai:
Jenkins rerun kogito-runtimes tests

wmedvede:
Jenkins please retest this

wmedvede:
Jenkins rerun kogito-runtimes tests

wmedvede:
Jenkins rerun kogito-apps tests

wmedvede:

cristianonicolai:
Jenkins rerun kogito-apps tests

cristianonicolai:

PR #1689, opened 2021-11-02
KOGITO-5508 Data Index Dev Service

cristianonicolai:
To test this, start a Kogito project that contains the `kogito-quarkus-processes` extension.
That should start Data Index in the background. To access the data index UI, jump to Quarkus Dev mode UI:
Selection_101

Selection_100

Clicking on Data Index GraphQL UI bring the UI allowing to query for process instances and tasks.
Start a new process instance and check if results show up in the UI once querying.

PR #1752, opened 2021-11-26
KOGITO-6322 Restore jandex on processes quarkus extension + lazily initialize DataIndexEventPublisher

evacchi:
Jandex was not configured on the quarkus extension (which causes weird startup issues with the new API I am working on, but only in some cases; go figure). I am setting here jandex again, but that causes the dev services to be instantiated. If the config property is not set, because it is not optional, then it throws an error. 

~~I am refactoring the event emitter so that it takes a delegate.~~

EDIT: no need! by annotating with `@UnlessBuildProperty(name = DataIndexEventPublisher.KOGITO_DATA_INDEX, stringValue = "")` the bean won't be created unless the property is set


cristianonicolai:
@evacchi I intentionally disabled that, see https://github.com/kiegroup/kogito-runtimes/pull/1689#discussion_r740782459, at the time, it seems to me that most Quarkus extensions create beans manually/selectively in the codegen. Would that approach work in your use case?

evacchi:
ok I see, I will look into it

evacchi:
so that's
     
```java
@BuildStep
    public AdditionalBeanBuildItem additionalBeans() {
        return AdditionalBeanBuildItem.builder().addBeanClasses(QuarkusStraightThroughProcessService.class, QuarkusProcessIdFactory.class).build(); 
```        

why `QuarkusStraightThroughProcessService` kept working. That was driving me totally crazy. Sure that'll work. I must have missed that comment there

evacchi:
well that makes this patch irrelevant then!

PR #1754, opened 2021-11-26
[KOGITO-5723] Implementing ForEach state on SWF

fjtirado:
Besides implementing ForEach functionality for SWF, this PR also fixes existing ForEach issues for BPMN and potential issues related with wrong container factory that causes errors like the one reported in https://issues.redhat.com/browse/KOGITO-6345

See example at https://github.com/kiegroup/kogito-examples/pull/1001


jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ForEachNodeVisitor.java
```diff
@@ -65,16 +71,41 @@ public void visitNode(String factoryField, ForEachNode node, BlockStmt body, Var
...
+                            new StringLiteralExpr(node.getOutputVariableType().getStringType())))));
+        }
+
+        if (node.getExpressionLanguage() != null) {
+            body.addStatement(getFactoryMethod(getNodeId(node), "expressionLanguage", new StringLiteralExpr(node.getExpressionLanguage())));
+        }
+
+        if (node.getCompletionAction() instanceof Supplier) {
+            body.addStatement(getFactoryMethod(getNodeId(node), "completionAction", ((Supplier<Expression>) node.getCompletionAction()).get()));
+        }
```
ricardozanini, line 89: Isn't too specific to Serverless Workflow?

jbpm/process-workitems/src/main/java/org/kie/kogito/process/workitems/impl/expr/ExpressionHandlerUtils.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.Optional;
+
+public class ExpressionHandlerUtils {
+
+    private ExpressionHandlerUtils() {
+    }
+
+    public static Optional<String> fallbackVarToName(String expr) {
+        int indexOf = expr.lastIndexOf('.');
+        return indexOf >= 0 ? Optional.of(expr.substring(indexOf + 1)) : Optional.empty();
```
ricardozanini, line 27: Isn't too specific to jsonpath/jq? Just playing the devil's advocate here. If we are only using it on SW, maybe we should start layering the engine to SW.

kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/handlers/ForEachStateHandler.java
```diff
@@ -0,0 +1,63 @@
...
+import io.serverlessworkflow.api.states.ForEachState;
+
+public class ForEachStateHandler extends CompositeContextNodeHandler<ForEachState> {
+
+    private final String outputVarName;
+
+    protected ForEachStateHandler(ForEachState state, Workflow workflow, ParserContext parserContext) {
+        super(state, workflow, parserContext);
+        outputVarName =
+                workflowAppContext.getApplicationProperty(ServerlessWorkflowUtils.APP_PROPERTIES_BASE + ServerlessWorkflowUtils.APP_PROPERTIES_STATES_BASE + "foreach.outputVarName", "_swf_eval_temp");
```
ricardozanini, line 37: Just wondering if we are documenting these properties. :)

ricardozanini (commented):
Well done, man! Many thanks! Can you please also update the README under the `kogito-serverless-workflow` module? Now we do support `ForEach`. :)

jbpm/jbpm-flow-builder/src/main/java/org/jbpm/compiler/canonical/ForEachNodeVisitor.java
fjtirado: I do not think so, we are adding a completion (or collection) action to ForEachNode, that extends default functionality (assigning the collection a variable in the model). 
The way to customize the behaviour is though an Action (already existing object in the engine) that implements a Supplier (provide the means to generate the code to be present in the process definition), so it is generic.

jbpm/process-workitems/src/main/java/org/kie/kogito/process/workitems/impl/expr/ExpressionHandlerUtils.java
fjtirado: In that case this is specific to any expression language which uses dot to start the expression. So, yes, strictly speaking this should be located in a project which is common to both jq/jsonpath. On the other hand, this does not require any specific 3pt, it is a fallback (does not prevent other expression to provide their implementation as part of the expression impementaion) and just one shared method (probably the only one they will ever  had) does not justify the addition of a shared library between jsonpath/jq, in my opinion

kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/handlers/ForEachStateHandler.java
fjtirado: The name of the output variable in SWF is a "maybe modified in some really strange case" constant. 
For this type of "constant", I prefer to not document it in public documents (we do not want anyone to use it, except if they have a variable called "_swf_eval_temp" in their model, which is almost impossible), but in internal consumption ones, so maybe one day we need to use that (and do not have to modify the code base)

jbpm/process-workitems/src/main/java/org/kie/kogito/process/workitems/impl/expr/ExpressionHandlerUtils.java
fjtirado: Since I agree with that having that code there is not a good idea, I have moved SFW utils stuff that was in the parser project to a different project (so there is some substance is this new project) and put the logic there

ricardozanini (approved):

kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/handlers/ForEachStateHandler.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    protected ForEachNodeFactory<?> makeNode(RuleFlowNodeContainerFactory<?, ?> factory) {
+        ForEachNodeFactory<?> result =
+                factory.forEachNode(parserContext.newId()).sequential(false).waitForCompletion(true).expressionLanguage(workflow.getExpressionLang()).collectionExpression(state.getInputCollection())
```
tiagodolphine, line 43: it is fine to be parallel, but in case of any async node in the foreach, this won't work, but for "sync" nodes this is fine... just to remember the restriction.

tiagodolphine (approved):

kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/handlers/ForEachStateHandler.java
fjtirado: yes, hopefully in SWF we will only have sync states, lets see!! ;)

fjtirado:
jenking rerun kogito-examples tests

fjtirado:
jenkins rerun kogito-examples tests

PR #1771, opened 2021-12-07
[KOGITO-3917] Adding state filtering

fjtirado:
Adding state filtering support
See modified example here https://github.com/kiegroup/kogito-examples/pull/1017

### DIFF
--- a/kogito-serverless-workflow/kogito-serverless-workflow-builder/src/test/resources/exec/multiple-operations.sw.json
+++ b/kogito-serverless-workflow/kogito-serverless-workflow-builder/src/test/resources/exec/multiple-operations.sw.json
@@ -44,10 +44,6 @@
       "name": "UpdateArg",
       "type": "operation",
       "actionMode": "sequential",
-      "stateDataFilter": {
-        "input": "{{ $.trace }}",
-        "output": "{{ $.trace }}"
-      },
       "actions": [
         {
           "functionRef": {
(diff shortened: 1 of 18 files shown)


kogito-workitems/kogito-jackson-utils/src/main/java/org/kie/kogito/jackson/utils/MergeUtils.java
```diff
@@ -15,29 +15,128 @@
...
+        } else {
+            targetArrayNode = JsonNodeFactory.instance.arrayNode();
+            if (targetNode != null) {
+                existingNodes = skipDuplicates ? Collections.singleton(targetNode) : Collections.emptySet();
+                targetArrayNode.add(targetNode);
+            } else {
+                existingNodes = Collections.emptySet();
+            }
+        }
+        if (srcEntry.getValue() != targetArrayNode) {
```
fjtirado, line 114: We need a custom implementation in merge to be able to control assignment of an array to self, in which case there is not need to do anythign

kogito-serverless-workflow/kogito-serverless-workflow-builder/src/test/resources/exec/multiple-operations.sw.json
```diff
@@ -44,10 +44,6 @@
       "name": "UpdateArg",
       "type": "operation",
       "actionMode": "sequential",
-      "stateDataFilter": {
```
fjtirado, line 47: This is removed because it makes a legacy test to fail. Since the test was testing another thing, it is better to just remove this, which affects the test expectations in terms of numbers of generated nodes

ricardozanini (approved):

PR #1804, opened 2021-12-17
[KOGITO-6411] Using serviceloader to load ExpressionHandler

fjtirado:
I also changed ExpressionHandler interface to avoid double compilation when checking if expression is valid

### DIFF
--- a/jbpm/process-workitems/src/main/java/org/kie/kogito/process/workitems/impl/expr/ExpressionHandlerFactory.java
+++ b/jbpm/process-workitems/src/main/java/org/kie/kogito/process/workitems/impl/expr/ExpressionHandlerFactory.java
@@ -15,40 +15,21 @@
  */
 package org.kie.kogito.process.workitems.impl.expr;
 
-import java.util.HashMap;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.ServiceLoader;
 
 public class ExpressionHandlerFactory {
 
     private ExpressionHandlerFactory() {
     }
 
-    private static final Map<String, String> expHandlerClasses = new HashMap<>();
-    private static final Map<String, ExpressionHandler> expHandlerInstances = new ConcurrentHashMap<>();
+    private static final ServiceLoader<ExpressionHandler> serviceLoader = ServiceLoader.load(ExpressionHandler.class);
 
-    private static final String JSONPATH_CLASSNAME = "org.kie.kogito.expr.jsonpath.JsonPathExpressionHandler";
-    private static final String JSONPATH_LANG = "jsonpath";
-
-    private static final String JQ_CLASSNAME = "org.kie.kogito.expr.jq.JqExpressionHandler";
-    private static final String JQ_LANG = "jq";
-
-    static {
-        expHandlerClasses.put(JSONPATH_LANG, JSONPATH_CLASSNAME);
-        expHandlerClasses.put(JQ_LANG, JQ_CLASSNAME);
-    }
-
-    public static ExpressionHandler get(String lang) {
-        return expHandlerInstances.computeIfAbsent(lang, c -> {
-            try {
-                return Class.forName(expHandlerClasses.getOrDefault(c, JSONPATH_CLASSNAME)).asSubclass(ExpressionHandler.class).getConstructor().newInstance();
-            } catch (ReflectiveOperationException e) {
-                throw new IllegalStateException(e);
-            }
-        });
+    public static Expression get(String lang, String expr) {
+        return serviceLoader.stream().filter(p -> p.get().lang().equals(lang)).findFirst().orElseThrow(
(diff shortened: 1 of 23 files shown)


kogito-serverless-workflow/kogito-jq-expression/src/main/java/org/kie/kogito/expr/jq/JqExpression.java
```diff
@@ -161,4 +159,23 @@ public JsonNode getResult() {
...
+    }
+
+    @Override
+    public boolean isValid() {
+        if (isValid == null) {
+            try {
+                compile();
+                isValid = true;
+            } catch (JsonQueryException e) {
+                isValid = false;
```
cristianonicolai, line 176: maybe add a warn log error, could be useful to understand any error in the expression.

fjtirado: hmmm, not really, remember here we are just trying to find out if what the user put in the file is an expression or not, so a warning will confuse the user that is not writing an expression there.

jbpm/process-workitems/src/main/java/org/kie/kogito/process/workitems/impl/expr/ExpressionHandlerFactory.java
```diff
@@ -15,40 +15,21 @@
...
+import java.util.ServiceLoader;
 
 public class ExpressionHandlerFactory {
 
     private ExpressionHandlerFactory() {
     }
 
-    private static final Map<String, String> expHandlerClasses = new HashMap<>();
-    private static final Map<String, ExpressionHandler> expHandlerInstances = new ConcurrentHashMap<>();
+    private static final ServiceLoader<ExpressionHandler> serviceLoader = ServiceLoader.load(ExpressionHandler.class);
```
ricardozanini, line 25: We just have to make sure to include it in the Native compilation mode: https://github.com/quarkusio/quarkus/blob/main/core/deployment/src/main/java/io/quarkus/deployment/builditem/nativeimage/ServiceProviderBuildItem.java

kogito-serverless-workflow/kogito-jq-expression/src/main/java/org/kie/kogito/expr/jq/JqExpressionHandler.java
```diff
@@ -17,47 +17,43 @@
...
     @Override
-    public boolean isExpr(String expr) {
-        try {
-            JsonQuery.compile(expr, Versions.JQ_1_6);
-            return true;
-        } catch (JsonQueryException e) {
-            return false;
-        }
+    public Expression get(String expr) {
+        return new JqExpression(scopeSupplier, expr);
```
ricardozanini, line 52: This will only work with the correct supplier if called `setScopeSupplier` first. I assume on quarkus we are doing this way?

ricardozanini: I think we can write an integration test to verify this behavior.

ricardozanini (commented):
I liked it! Seems good, my comments are just observations. But the service loader one should be considered seems it will fail with native.

jbpm/process-workitems/src/main/java/org/kie/kogito/process/workitems/impl/expr/ExpressionHandlerFactory.java
fjtirado: Hi Ricardo, 
I tested with native mode (besided the automatic test, I do that manually)
It is working because of this https://github.com/kiegroup/kogito-runtimes/pull/1804/files#diff-1e0a6dfe8aab829b606823d0a0546f92374bc8a2df5f3c02c4b31ee070c783e9R43

kogito-serverless-workflow/kogito-jq-expression/src/main/java/org/kie/kogito/expr/jq/JqExpressionHandler.java
fjtirado: We have a startup bean here https://github.com/kiegroup/kogito-runtimes/blob/main/quarkus/extensions/kogito-quarkus-serverless-workflow-extension/kogito-quarkus-serverless-workflow/src/main/java/org/kie/kogito/expr/jq/JqQuarkusExtension.java#L25

jbpm/process-workitems/src/main/java/org/kie/kogito/process/workitems/impl/expr/ExpressionHandlerFactory.java
ricardozanini: Oh you already add it. That's fine, I missed.

ricardozanini (approved):

PR #1814, opened 2021-12-21
[KOGITO-6418] Event states filter and merging

fjtirado:
This add filter and merge support for SWF event states. 
In order to achieve that, code generator was slightly changed, so `eventToModel` method is only generated is the target node is a start one. This step is required to not force events to carry the whole model, as was kind of assumed before this change. 
The presence of that optional method has an additional benefit.  It allow us to identify at runtime if the target of the event is start node, so an attempt to start a process should be done or the targer is an intermediate event and a signal method should be called. This allow us to prevent what was happning with https://issues.redhat.com/browse/KOGITO-6356 
To round up this set of changes, name of classes generated for events in a subprocess are now guaranteed to be unique (which was not the case before)  
Finally, I feel we are lacking a way to signal a bunch of active process, we probable need another method to` ProcesService` interface, since right now we only have the one that accepts a process intance Id. I have opened  https://issues.redhat.com/browse/KOGITO-6468 and assigned to @evacchi to evaluate that need, since it affects the process API. 
Callback modified example at https://github.com/kiegroup/kogito-examples/pull/1055
As part of the testing of the examples, some improvements in JsonPathExpression were added.

### DIFF
--- a/jbpm/jbpm-flow/src/main/java/org/kie/kogito/event/impl/DataEventConsumer.java
+++ b/jbpm/jbpm-flow/src/main/java/org/kie/kogito/event/impl/DataEventConsumer.java
@@ -15,6 +15,7 @@
  */
 package org.kie.kogito.event.impl;
 
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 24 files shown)


jbpm/jbpm-flow/src/main/java/org/kie/kogito/event/impl/DataEventConsumer.java
```diff
@@ -25,28 +26,25 @@
...
+    public DataEventConsumer(ProcessService processService, ExecutorService executorService, Optional<Function<D, M>> function) {
         this.processService = processService;
         this.executorService = executorService;
         this.function = function;
     }
 
     @Override
     public CompletionStage<Void> consume(Application application, Process<M> process, Object eventData, String trigger) {
-        return CompletableFuture.runAsync(() -> processService.createProcessInstance(process, function.apply((D) eventData), null, trigger, null), executorService);
+        //TODO right now it is only possible to start a new process instance when not using cloudevent
```
ricardozanini, line 46: Do we have a JIRA for this task?

ricardozanini (approved):
We are making good progress here!

jbpm/jbpm-flow/src/main/java/org/kie/kogito/event/impl/DataEventConsumer.java
fjtirado: yes, I opened https://issues.redhat.com/browse/KOGITO-6468

fjtirado:
jenkins retest this

fjtirado:
jenkins rerun kogito-runtimes tests

PR #1815, opened 2021-12-22
[DROOLS-6721] fix missing LogManager in springboot examples

mariofusco:
JIRA: https://issues.redhat.com/browse/DROOLS-6721


mariofusco:
Spring examples shouldn't need the jboss logmanager. There is something wrong in our dependency tree that could be possibly fixed by this PR https://github.com/kiegroup/kogito-examples/pull/1052

PR #1816, opened 2021-12-22
KOGITO-6448 async findById processInstance used by CloudEventConsumer

tiagodolphine:
Async findById in ProcessInstances used by CloudEventConsumero avoid Thread Blocked Exception.
This issue was noticed using the latest PostgreSQL 14+ with the reactive client.
https://issues.redhat.com/browse/KOGITO-6448

- [ ] Your code is properly formatted according to [this configuration](https://github.com/kiegroup/kogito-runtimes/tree/main/kogito-build/kogito-ide-config)
- [ ] Pull Request title contains the target branch if not targeting main: `[0.9.x] KOGITO-XYZ Subject`

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

* <b>Run all builds</b>  

  Please add comment: <b>Jenkins (re)run [kogito-runtimes|optaplanner|kogito-apps|kogito-examples|optaplanner-quickstarts|optaweb-employee-rostering|optaweb-vehicle-routing] tests</b>
 
* <b>Quarkus LTS checks</b>  

  Please add comment: <b>Jenkins (re)run [kogito-runtimes|optaplanner|kogito-apps|kogito-examples|optaplanner-quickstarts|optaweb-employee-rostering|optaweb-vehicle-routing] LTS</b>

* <b>Native checks</b>  

  Please add comment: <b>Jenkins (re)run [kogito-runtimes|optaplanner|kogito-apps|kogito-examples|optaplanner-quickstarts|optaweb-employee-rostering|optaweb-vehicle-routing] native</b>

</details>

### DIFF
--- a/jbpm/jbpm-flow/src/main/java/org/kie/kogito/event/impl/CloudEventConsumer.java
+++ b/jbpm/jbpm-flow/src/main/java/org/kie/kogito/event/impl/CloudEventConsumer.java
@@ -62,30 +62,39 @@ public class CloudEventConsumer<D, M extends Model, T extends AbstractProcessDat
             logger.debug("Received message with reference id '{}' going to use it to send signal '{}'",
                     cloudEvent.getKogitoReferenceId(),
                     trigger);
-            Optional<ProcessInstance<M>> instance = process.instances().findById(cloudEvent.getKogitoReferenceId());
-            if (instance.isPresent()) {
-                return CompletableFuture.completedFuture(processService.signalProcessInstance((Process) process, cloudEvent.getKogitoReferenceId(), cloudEvent.getData(), "Message-" + trigger));
-            } else if (function.isPresent()) {
-                logger.info("Process instance with id '{}' not found for triggering signal '{}', starting a new one",
-                        cloudEvent.getKogitoReferenceId(),
-                        trigger);
-                return startNewInstance(process, function.get().apply(cloudEvent.getData()), cloudEvent, trigger);
-            } else {
-                logger.info("Process instance with id {} not found for triggering signal {}", cloudEvent.getKogitoReferenceId(), trigger);
-                return CompletableFuture.completedFuture(null);
-            }
-
+            return CompletableFuture.supplyAsync(() -> {
+                Optional<ProcessInstance<M>> instance = findProcessInstance(process, cloudEvent);
+                if (instance.isPresent()) {
+                    return signalProcessInstance(process, trigger, cloudEvent);
+                } else if (function.isPresent()) {
+                    logger.info("Process instance with id '{}' not found for triggering signal '{}', starting a new one",
+                            cloudEvent.getKogitoReferenceId(),
+                            trigger);
+                    return startNewInstance(process, function.get().apply(cloudEvent.getData()), cloudEvent, trigger);
+                } else {
+                    logger.info("Process instance with id {} not found for triggering signal {}", cloudEvent.getKogitoReferenceId(), trigger);
+                    return null;
+                }
+            }, executor);
         } else if (function.isPresent()) {
             logger.debug("Received message without reference id, starting new process instance with trigger '{}'", trigger);
-            return startNewInstance(process, function.get().apply(cloudEvent.getData()), cloudEvent, trigger);
+            return CompletableFuture.supplyAsync(() -> startNewInstance(process, function.get().apply(cloudEvent.getData()), cloudEvent, trigger), executor);
         } else {
             logger.warn("Received not start event without kogito referecence id for trigger {}", trigger);
             return CompletableFuture.completedFuture(null);
         }
     }
 
-    private CompletionStage<Void> startNewInstance(Process<M> process, M model, T cloudEvent, String trigger) {
-        return CompletableFuture.runAsync(() -> processService.createProcessInstance(process, model, cloudEvent.getKogitoStartFromNode(), trigger, cloudEvent.getKogitoProcessinstanceId()), executor);
+    private Optional<M> signalProcessInstance(Process process, String trigger, T cloudEvent) {
+        return processService.signalProcessInstance(process, cloudEvent.getKogitoReferenceId(), cloudEvent.getData(), "Message-" + trigger);
+    }
+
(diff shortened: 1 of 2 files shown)


cristianonicolai (approved):

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstances.java
```diff
@@ -17,13 +17,23 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
 
 public interface ProcessInstances<T> {
 
     default Optional<ProcessInstance<T>> findById(String id) {
         return findById(id, ProcessInstanceReadMode.MUTABLE);
     }
 
+    default CompletionStage<Optional<ProcessInstance<T>>> findByIdAsync(String id) {
```
fjtirado, line 29: Do we really need this new method? 
I mean, I think it will be enough to change CloudEventConsumer to invoke existing sync findById inside the existing async block. 
So in order to fix this we just need to move findById to the async executor

jbpm/jbpm-flow/src/main/java/org/kie/kogito/event/impl/CloudEventConsumer.java
```diff
@@ -63,24 +62,27 @@ public CloudEventConsumer(ProcessService processService, ExecutorService executo
...
-            Optional<ProcessInstance<M>> instance = process.instances().findById(cloudEvent.getKogitoReferenceId());
-            if (instance.isPresent()) {
-                return CompletableFuture.completedFuture(processService.signalProcessInstance((Process) process, cloudEvent.getKogitoReferenceId(), cloudEvent.getData(), "Message-" + trigger));
-            } else {
-                logger.warn("Process instance with id '{}' not found for triggering signal '{}', starting a new one",
-                        cloudEvent.getKogitoReferenceId(),
-                        trigger);
-                return startNewInstance(process, model, cloudEvent, trigger);
-            }
+            return process.instances()
```
fjtirado, line 65: As stated in previous comment, rather than adding a new findByIdAyns I will  move findById call to the callback and create a chaing of completablestate here.

fjtirado (commented):
Nice job, but I think we do not need to change ProcessService interface and just rearrange call to findById in CloudEventConsumer. 
Maybe is better to wait to perform this rearrange till this is merged
https://github.com/kiegroup/kogito-runtimes/pull/1814/files#diff-c36c00514b9a60b48080d4000e5f426d8de6e8e768fb036930d936868e769fe3R68

jbpm/jbpm-flow/src/main/java/org/kie/kogito/event/impl/CloudEventConsumer.java
tiagodolphine: this new method is the whole point of this PR, an exception of `Blocked Thread` is being thrown, because the findById is blocking, and since the PostgreSQL add-on is using reactive, we can leverage async APIs that is the proper behavior. Anyway, if we don't go using async API, we **must** change the `QuarkusCloudEventReceiver` to say that it is blocking, in this way Quarkus change the execution to the worker thread pool instead of the IO threads.

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstances.java
tiagodolphine: yes this is another option, to wrap the blocking calls to be executed like CompletableFuture.supplyAsync... or similar... I just tried to leverage the reactive impl we have for PostgreSQL addon... anyway, we need to decide, there is no perfect solution I'd say...

tiagodolphine:
> I've tested it locally and works fine. Regarding the changes LGTM, however I'm not 100% familiar the persistence mechanism and all the potential implicances of this change. So I prefer @cristianonicolai to take a look too.
> 
> Be sure please that this change is covered by the tests please.

Yeah, I can add some tests for the newly added methods, I just want to confirm that we go with this approach and it works. The impact should be minimal since the new methods are just wrapping the normal blocking call with CompletableFuture as the default impl.
see https://github.com/kiegroup/kogito-runtimes/pull/1816#discussion_r774491655
basically, we need to take a decision :)

api/kogito-api/src/main/java/org/kie/kogito/process/ProcessInstances.java
fjtirado: I prefer to keep ProcessService as it is and do the wrapping in the caller (addon or consume event). 
Reason is that the wrapping is just one line (we do not even need an utility method to avoid code duplication because of that)
Another one is that if you check the usage, they are different in every caller, and thats why you are returning a CompletionStage rather than providing a callback argument (a Consumer object) to the existing findById api (which is what Kafka API is doing in a similar scenario)
Third and final reason is that findById implementation is going to be hardly different than just a wrapper.

jbpm/jbpm-flow/src/main/java/org/kie/kogito/event/impl/CloudEventConsumer.java
fjtirado: No, I did not mean that, what Im proposing is to invoke existing findById in the async executor thread, as we are doing with start process instance, thats all

tiagodolphine:
@fjtirado to make the PR simpler and avoid modification in the ProcessInstances, I just changed the CloudEventConsumer to wrap the execution to run async in the executed already configured in this class.

But for the future to work more integrated with reactive libraries like reactive pg client and smallrye reactive messaging, we need to come with reactive APIs under ProcessInstances at least... and at some point the integration with the engine itself, but this is a more complex story.

tiagodolphine:
jenkins retest this

tiagodolphine:
Jenkins run kogito-apps tests

fjtirado (approved):

tiagodolphine:
@cristianonicolai missing your review this one with the last changes.

tiagodolphine:
all green @cristianonicolai, ok to merge?

cristianonicolai (approved):

PR #1757, opened 2021-11-29
KOGITO-6131 Enable codegen when RESTEasy reactive is used

tkobayas:
https://issues.redhat.com/browse/KOGITO-6131

This is a WIP branch to continue https://github.com/kiegroup/kogito-runtimes/pull/1669

Applied AdditionalStaticResourceBuildItem following https://github.com/quarkusio/quarkus/issues/20846#issuecomment-962932557

Use @QuarkusIntegrationTest instead of @QuarkusTest to handle static resources generated by codegen
https://github.com/kiegroup/kogito-examples/pull/1004

- [x] You have read the [contributors guide](CONTRIBUTING.md)
- [x] Your code is properly formatted according to [this configuration](https://github.com/kiegroup/kogito-runtimes/tree/main/kogito-build/kogito-ide-config)
- [x] Pull Request title is properly formatted: `KOGITO-XYZ Subject`
- [x] Pull Request title contains the target branch if not targeting main: `[0.9.x] KOGITO-XYZ Subject`
- [x] Pull Request contains link to the JIRA issue
- [x] Pull Request contains link to any dependent or related Pull Request
- [x] Pull Request contains description of the issue
- [x] Pull Request does not include fixes for issues other than the main ticket

<details>
<summary>
How to replicate CI configuration locally?
</summary>

Build Chain tool does "simple" maven build(s), the builds are just Maven commands, but because the repositories relates and depends on each other and any change in API or class method could affect several of those repositories there is a need to use [build-chain tool](https://github.com/kiegroup/github-action-build-chain) to handle cross repository builds and be sure that we always use latest version of the code for each repository.
 
[build-chain tool](https://github.com/kiegroup/github-action-build-chain) is a build tool which can be used on command line locally or in Github Actions workflow(s), in case you need to change multiple repositories and send multiple dependent pull requests related with a change you can easily reproduce the same build by executing it on Github hosted environment or locally in your development environment. See [local execution](https://github.com/kiegroup/github-action-build-chain#local-execution) details to get more information about it.
</details>

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* <b>Run all builds</b>  
  Please add comment: <b>Jenkins retest this</b>

* <b>Run (or rerun) specific test(s)</b>  
  Please add comment: <b>Jenkins (re)run [kogito-runtimes|optaplanner|kogito-apps|kogito-examples|optaplanner-quickstarts|optaweb-employee-rostering|optaweb-vehicle-routing] tests</b>
 
[...]

PR #1823, opened 2021-12-27
[KOGITO-5094] Allowing byte[] as serializer/deserializer for Quarkus

fjtirado:
This JIRA is a first step to achieve integration with Apache Avro. 
I have created a follow up JIRA to replace DataEvent by CloudEvent https://issues.redhat.com/browse/KOGITO-6476
The overall idea in this one is to infer the right marshaller from the channel property. 
If StringSerializer or ByteSeralizer we use a default marshaller that converts the Json to byte[]/string (this keeps backward compatibility) 
If different serializer we assume the serializer deals properly with the event object and we use a marshaller that keep the object intact
See existing modified example at https://github.com/kiegroup/kogito-examples/pull/1057
In addition to that, this PR fixes following issues: 
- A missing managed dependency for testing
- EventConverterProducer and EventMarshallerProducer are not longer generated (and are now declared as DefaultBeans). This will allow users to extend them and also to "specialize" them, providing their own marshallers/unmarshallers for custom data types
- A problem with future version of Quarkus. We are not longing mengling with Smallrye internals, but treating default channels as regular channels:
  -  MediatorConfig and ChannelRegistar are not longer used to provide a default emitter. Now all meaningful emitters are generated from configuration.  For covering such cases where an Emitter and Receiver is being inject into a class but no configuration is provided to create the proper channel (for example in QuarkusEventDrivenDecission and knative example), a default dummy bean is produced. 
  - Before these changes, emitter/receiver for quarkus were generated as part of processes codegen, now they are generated by the addon processor. This requires changes in the existing main buildsteps in order to allow an addon to modify a generated class. The main buildstep in KogitoAssetProcessor was split in two, one that generates the sources and another that generate the classes


kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/producer/EventMarshallerProducerSpringTemplate.java
```diff
@@ -31,7 +35,7 @@
     ObjectMapper mapper;
 
     @Bean
-    public EventMarshaller eventMarshaller(){
-        return new DefaultEventMarshaller(mapper);
+    public EventMarshaller<String> stringEventMarshaller(){
```
fjtirado, line 38: For Spring, we keep just string event marshaller at the moment

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/producer/EventMarshallerProducerQuarkusTemplate.java
```diff
@@ -32,7 +34,17 @@
     ObjectMapper mapper;
     
     @Produces
-    public EventMarshaller eventMarshaller(){
-        return new DefaultEventMarshaller(mapper);
+    public EventMarshaller<String> stringEventMarshaller(){
```
fjtirado, line 37: Quarkus will inject the proper marshaller depending on the generated emitter type

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/DefaultChannelMappingStrategy.java
```diff
@@ -57,27 +58,41 @@
...
+    }
+
+    private static String getClassName(Optional<String> serializerClassName) {
+        if (serializerClassName.isPresent()) {
+            String value = serializerClassName.get();
+            int indexOf = value.lastIndexOf(".");
+            if (indexOf >= 0) {
+                value = value.substring(indexOf + 1);
+            }
+            if (value.startsWith("String")) {
```
fjtirado, line 86: If the serializer (which is a required property in smallrye) contains String or ByteArray, weconvert the Object to a json String or byte using Jackson 
If a different serialized we skip any marshalling and the serializer will be responsible to convert the object into the byte[] that is published by Kafka

api/kogito-services/src/main/java/org/kie/kogito/services/event/impl/ByteArrayEventMarshaller.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
```
ricardozanini, line 2:
```suggestion
 * Copyright 2021 Red Hat, Inc. and/or its affiliates.
```

api/kogito-services/src/main/java/org/kie/kogito/services/event/impl/NoOpEventMarshaller.java
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
```
ricardozanini, line 2:
```suggestion
 * Copyright 2021 Red Hat, Inc. and/or its affiliates.
```

api/kogito-services/src/main/java/org/kie/kogito/services/event/impl/DefaultJsonConverter.java
```diff
@@ -15,25 +15,32 @@
...
+    public <T> T apply(Object value, Class<T> clazz) throws IOException {
         logger.debug("Converting event with payload {} to class {} ", value, clazz);
-        return objectMapper.readValue(value, clazz);
+        if (clazz.isInstance(value)) {
+            return clazz.cast(value);
+        } else if (value instanceof byte[]) {
+            return objectMapper.readValue((byte[]) value, clazz);
+        } else {
+            return objectMapper.readValue(value.toString(), clazz);
+        }
```
ricardozanini, line 44: We just need to make sure to have test cases to cover all these branches and corner cases

ricardozanini (commented):
Good job! It looks like a simple change to keep things working and add the byte serializer. Do you mind adding some test cases?

api/kogito-services/src/main/java/org/kie/kogito/services/event/impl/DefaultJsonConverter.java
fjtirado: yes, the String combination is covered by existing examples. I modified one of them for covering the byte[] one. 
As specified in the description, this JIRA scope is just Quarkus.

kogito-build/kogito-dependencies-bom/pom.xml
```diff
@@ -327,6 +327,13 @@
         <version>${version.org.assertj}</version>
         <scope>test</scope>
       </dependency>
+      
+      <dependency>
+        <groupId>org.junit-pioneer</groupId>
+        <artifactId>junit-pioneer</artifactId>
```
evacchi, line 333: I think this may be a leftover?

evacchi (commented):
as much as I like that we are going towards this direction, I am a bit worried that we may turn this into a collection of special cases, so as a general note, I hope that, as we go forward, we will delegate more and more of the serialization aspect to the underlying implementation, and avoid as much as possible special-casing.

kogito-build/kogito-dependencies-bom/pom.xml
fjtirado: This is just moving to the right place a non managed dependency we had from the past.

fjtirado:
> as much as I like that we are going towards this direction, I am a bit worried that we may turn this into a collection of special cases, so as a general note, I hope that, as we go forward, we will delegate more and more of the serialization aspect to the underlying implementation, and avoid as much as possible special-casing.

Yes, I had the same concern. 
If this JIRA is merging
We are doing the assumption that the object we will receive from smallrye (or sprigboot kafka) will either be a json  string, a json byte[] or the the right cloud event instance (this is better that current situation in the sense that we are not just assuming we receive a json string)
We currently have a hack for emitting, we are somehow inferring from the smallrye configuration which conversion we have to perform, if a stringserializer generate a json string, if byte serializer a byte[] string, if any other serializer (or no serialized, mean no kafka) we assume the broker knows how to deal with the POJO. (this is also better than current situation, where we assume json string. 
But we are still lacking support for dealing with string or byte[] that are not json or generate string/byte[] that are not json. So we still need a mechanism to override the default marshaller in a generic and easily configurable way. 
This is why this JIRA is just a first step and before doing that I will prefer to refactor the whole thing to use the CloudEvent SDK first. Lets discuss in a meetting the way to go further

ricardozanini:
As a side note, in this PR I push the CE SDK to the `kogito-events-api`: https://github.com/kiegroup/kogito-runtimes/pull/1828

Which will help us refactor the `DataEvent` types later.

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
```diff
@@ -281,7 +281,8 @@ public ProcessCodegen(KogitoBuildContext context, Collection<? extends Process>
             }
             this.processes.put(process.getId(), (KogitoWorkflowProcess) process);
         }
-        mappingStrategy = context().getAddonsConfig().useCloudEvents() ? Optional.of(getChannelMappingStrategy()) : Optional.empty();
+        context.hasDI();
```
danielezonca, line 284: Leftover? This value is ignored

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/MessageProducerJavaTemplate.java
```diff
@@ -30,7 +30,7 @@
 
     Optional<Boolean> useCloudEvents = Optional.of(true);
 
-    EventMarshaller marshaller = new DefaultEventMarshaller(new ObjectMapper());
+    EventMarshaller<String> marshaller = new StringEventMarshaller(new ObjectMapper());
```
danielezonca, line 33: Is this code used or it is just to have a Java version?

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
```diff
@@ -281,7 +281,8 @@ public ProcessCodegen(KogitoBuildContext context, Collection<? extends Process>
...
+        mappingStrategy = context().getAddonsConfig().useQuarkusCloudEvents() ? Optional.of(getChannelMappingStrategy()) : Optional.empty();
```
danielezonca, line 285: It seems that this change is breaking spring support: now it is using `useQuarkusCloudEvents`

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/DefaultChannelMappingStrategy.java
```diff
@@ -35,49 +35,54 @@
...
```
danielezonca, line 76: What if `java.lang.String`? Isn't it possible?

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/DefaultChannelMappingStrategy.java
```diff
@@ -35,49 +35,54 @@
...
+                return "String";
+            } else if (value.startsWith("ByteArray")) {
```
danielezonca, line 78: Same here

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/DefaultChannelMappingStrategy.java
danielezonca: Can you add this clarification to a comment?

fjtirado: @danielezonca Im confused about the policy comment to be followed. When I entered Redhat, I was told that I should put things that I usually write as comments in the PR. I think this is a wise thing, because left the code pretty clean. On the other hand, I agree this (and not only this) qualify to a comment in the code, but, the truth is that we do not really have comments in the code right now. So, what should be the policy?

api/kogito-services/src/main/java/org/kie/kogito/services/event/impl/ByteArrayEventMarshaller.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Override
+    public <T> byte[] marshall(T event) {
+        logger.debug("Marshalling event {}", event);
+        try {
+            return mapper.writeValueAsBytes(event);
+        } catch (JsonProcessingException e) {
+            logger.error("Error marshalling event {}", event);
+            throw new IllegalStateException(e);
```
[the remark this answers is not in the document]
fjtirado: good catch

api/kogito-services/src/main/java/org/kie/kogito/services/event/impl/JsonEventConverter.java
```diff
@@ -15,25 +15,32 @@
...
-    public <T> T apply(String value, Class<T> clazz) throws JsonProcessingException {
+    public <T> T apply(Object value, Class<T> clazz) throws IOException {
         logger.debug("Converting event with payload {} to class {} ", value, clazz);
-        return objectMapper.readValue(value, clazz);
+        if (clazz.isInstance(value)) {
+            return clazz.cast(value);
+        } else if (value instanceof byte[]) {
+            return objectMapper.readValue((byte[]) value, clazz);
+        } else {
+            return objectMapper.readValue(value.toString(), clazz);
```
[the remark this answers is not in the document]
fjtirado: An event should always has a value, isnt it?

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
fjtirado: yes, typo

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/MessageProducerJavaTemplate.java
fjtirado: This is a recurrent question. I do not think is used at all, but I do not really know

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/ProcessCodegen.java
fjtirado: Not really, channel resolution does not apply to Spring. It was working before becuase channel returned was empty, but once we need to generate emitter for default channel for quarkus, this this check needed in order Spring to work

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/DefaultChannelMappingStrategy.java
fjtirado: It is checking for Kafka StringSerializer and ByteArraySerializer, no java.lang.string

kogito-codegen-modules/kogito-codegen-processes/src/main/java/org/kie/kogito/codegen/process/DefaultChannelMappingStrategy.java
danielezonca: I don't think we have a strict policy, I think it is hard to define a similar policy because it is sort of impossible to have a "fixed" definition. We can probably raise a similar point during next sprint retrospective

In this specific context I think it is important to mention that the third branch (the fallback) returns a NoOp serializer that assumes the event emitter to be able to serialize the message

fjtirado: I added a brief comment

fjtirado:
Working on an approach that work for decission. 
Problem of removing the smallrye hack is that we also need generation for rules, and the generation code is on process, I need to move the generation somewhere else

cristianonicolai (approved):

quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions/pom.xml
```diff
@@ -27,6 +27,11 @@
             <groupId>org.kie.kogito</groupId>
             <artifactId>kogito-dmn</artifactId>
         </dependency>
+        
+         <dependency>
+           <groupId>io.quarkus</groupId>
+           <artifactId>quarkus-undertow</artifactId>
```
ricardozanini, line 33: I think with Toshiya's PR (#1757), we can remove this.

quarkus/extensions/kogito-quarkus-extension-common/kogito-quarkus-common-deployment/pom.xml
```diff
@@ -50,31 +49,33 @@
...
             <artifactId>quarkus-resteasy-jackson-deployment</artifactId>
-            <version>${version.io.quarkus}</version>
             <scope>provided</scope>
         </dependency>
 
+        <!-- Undertow is needed so that the static resource serving can correctly locate CP resources from `META-INF/resources` of the application, as it would be normally expected.
+             See https://issues.redhat.com/browse/KOGITO-3477 -->
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-undertow-deployment</artifactId>
```
ricardozanini, line 76: Same thing, can you try after rebasing with Toshiya's PR? I think he was working to remove this dependency, so we won't need to add it here anymore.

quarkus/extensions/kogito-quarkus-predictions-extension/kogito-quarkus-predictions/pom.xml
```diff
@@ -28,6 +28,10 @@
             <groupId>org.kie.kogito</groupId>
             <artifactId>kogito-api-incubation-predictions-services</artifactId>
         </dependency>
+        <dependency>
+          <groupId>io.quarkus</groupId>
+           <artifactId>quarkus-undertow</artifactId>
+         </dependency>
```
ricardozanini, line 34: Same here

ricardozanini (commented):
Pretty nice! The only thing to review is the usage of the undertow dependency. After #1757 I think we can get rid of them. Can you please try? It was merged this morning.

quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions/pom.xml
fjtirado: ok, Ill try again, I have to add this change this morning in order it to work

fjtirado: @ricardozanini @tkobayas  If I remove this undertow dependency, I get this failure
INFO] --- quarkus-bootstrap-maven-plugin:2.6.0.Final:extension-descriptor (default) @ kogito-quarkus-predictions ---
[ERROR] Quarkus Extension Dependency Verification Error
[ERROR] The deployment artifact org.kie.kogito:kogito-quarkus-predictions-deployment::jar depends on the following Quarkus extension deployment artifacts whose corresponding runtime artifacts were not found among the dependencies of org.kie.kogito:kogito-quarkus-predictions:jar:2.0.0-SNAPSHOT:
[ERROR]  org.kie.kogito:kogito-quarkus-predictions-deployment:jar:2.0.0-SNAPSHOT
[ERROR]    org.kie.kogito:kogito-quarkus-common-deployment:jar:2.0.0-SNAPSHOT
[ERROR] *    io.quarkus:quarkus-undertow-deployment:jar:2.6.0.Final

tkobayas: Hi @fjtirado , the error message is caused by an inconsistency between runtime module dependency and deployment module dependency. (e.g. you removed `quarkus-undertow-deployment` from `kogito-quarkus-common-deployment` but `quarkus-undertow` in `kogito-quarkus-predictions` remains)

I see this PR added undertow dependency in these 6 pom.xml files. If you remove these undertow dependencies altogether, the build should work.

quarkus/addons/messaging/runtime/pom.xml
quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions/pom.xml
quarkus/extensions/kogito-quarkus-extension-common/kogito-quarkus-common-deployment/pom.xml
quarkus/extensions/kogito-quarkus-predictions-extension/kogito-quarkus-predictions/pom.xml
quarkus/extensions/kogito-quarkus-processes-extension/kogito-quarkus-processes/pom.xml
quarkus/extensions/kogito-quarkus-rules-extension/kogito-quarkus-rules/pom.xml

Thanks,

tkobayas: Btw, if you find that a test fails to access generated static resources (= `GeneratedFileType.Category.STATIC_HTTP_RESOURCE`), the test needs to be `@QuarkusIntegrationTest` (or `@QuarkusTest` + `quarkus-undertow` dependency (test scope)). Please refer to https://github.com/kiegroup/kogito-runtimes/pull/1757#issuecomment-998560583  I'll file a PR to document it in https://github.com/kiegroup/kogito-runtimes/blob/main/CONTRIBUTING.md

fjtirado: @tkobayas 
But undertow seems to be needed in kogito-quarkus-common-deployment
```
  <!-- Undertow is needed so that the static resource serving can correctly locate CP resources from `META-INF/resources` of the application, as it would be normally expected.
             See https://issues.redhat.com/browse/KOGITO-3477 -->
        <dependency>
            <groupId>io.quarkus</groupId>
            <artifactId>quarkus-undertow-deployment</artifactId>
        </dependency>
```

ricardozanini (approved):

tkobayas (approved):

danielezonca:
@fjtirado 
I see you are still pushing new code chances to the PR, is it ready for review or still WIP?
Build is green so if you are fixing some issue I would suggest to also add some additional test to prevent regression in the future

fjtirado:
> ild is green so if you are fixing some issue I would suggest to also add some additional test to prevent regression in the future

Build is green now, after my lastest change, merging.

PR #1778, opened 2021-12-09
KOGITO-6290 Public API: stateful process api

evacchi:
Introduce the first part of the Process API. 

- processes can be started, updated and aborted
- user tasks are supported 
   - attachments and comments are still not implemented (present only as an Identifier, no service available for them, yet)
  

## Examples

these are also available as test cases (more examples in test cases too)

### Creating a process

```java
        var id = appRoot.get(ProcessIds.class).get("approvals");
        var ctx = mapper.readValue(body, MapDataContext.class);

        var result = processSvc.create(id, ctx);
```

### Aborting a process

```java
        var result = processSvc.abort(pid);
```

### Completing a task

```java
        var taskMeta = TaskMetaDataContext.of(
                Policy.of("admin", List.of("managers")));

        // get active tasks
        var tasks = taskSvc.get(pid.tasks(), taskMeta);

        // get first task id
        var taskId = tasks.meta().as(TaskWorkItemDataContext.class).tasks().get(0);
        
        var result = taskSvc.complete(taskId,
               // pass extra data to complete the task
                ExtendedDataContext.of(
                        taskMeta,
                        MapDataContext.of(Map.of("some", "value"))));

```

PR #1826, opened 2021-12-29
KOGITO-5850 Public API: Stateful Rule Unit API

evacchi:
This is a first working round of the "new" rule unit API. 
**All the following choices are, as usual, still open for debate**, but not a blocker for the merging of this PR.
I am marking this as *ready for review* but I don't want to rush it, so please share your feedback.

In this initial version I had to mix the `org.kie.kogito.rules.*` API with the incubation API, in particular the `RuleUnitData` definition and the `DataSource` declarations need to be mixed. More PRs will make the 2 worlds more consistent to each other (basically we will remove `org.kie.kogito.rules.*`)

## DataSourceService

- a `DataSourceId` is an identifier to a data source
- a `DataId` is an identifier that represents an object added to a data source (similarly to the data handle)
- a `DataId` is always nested to a `DataSourceId`

e.g. consider the rule unit:

```java
package com.example;
class MyUnit {
    DataStore<Person>
}
```
then you may access the data source with

```java
// /rule-units/com.example.MyUnit/data-sources/students
DataSourceId dsid = appRoot.get(RuleUnitIds.class).get(MyUnit.class).dataSources().get("students");
```

you may add data to the data source with:

```java
// /rule-units/com.example.MyUnit/data-sources/students/data/12437247823
DataId dataId = dsSvc.add(dsid, new Person("Paul", 79)); 
```

the service may throw if the data source is not capable of the command; e.g. if `dsid` is a stream:

```java
dsSvc.remove(dsid, dataId);
```

### "Floating" data sources (not implemented yet)

**Note:** data sources may live "outside" of a rule unit and be defined *regardless* of them. In that case,
they may also be referred to using a different API 

```java
// /data-sources/students
DataSourceId dsid = appRoot.get(DataSourceIds.class).get("students");
```

In this case, wiring of the data source may occur in the rule unit by ID; e.g.:

```java
package com.example;
class MyUnit {
    @Named("kogito:///data-sources/students") DataStore<Person>
}
```

the "other" ID (`/rule-units/com.example.MyUnit/data-sources/students`) would still be valid but would point to the same DS.

Using "floating" DS allows them to be shared.

**Note:** this makes DSs very similar to Quarkus channels and it is worth investigating how to conflate the two.

## StatefulRuleUnitService

Similarly to the Stateless `RuleUnitService` (we may want to rename these two, making "RuleUnitService" the stateful one), we get the ID to a unit:

```java
var id = appRoot.get(RuleUnitIds.class).get(AnotherService.class);
```

we then create an instance of the ruleUnitData (as we do in codegen)

```java
public class AnotherService implements RuleUnitData, ReferenceContext {
    // StringHolder is just a domain object: 
    // a wrapper for a String for easier pattern matching in rules
    DataStore<StringHolder> strings; 
    DataStore<StringHolder> greetings;
    ...
}

AnotherService ruleUnitData =
    new AnotherService(
            DataSource.createStore(),             
            DataSource.createStore());
```

**Note:** this is `org.kie.kogito.rules.DataSource` not the "new" version in this PR! needs work to collapse the two!

We `create` the instance using the service:

```java
var id = appRoot.get(RuleUnitIds.class).get(AnotherService.class);
var result = ruleUnitService.create(id, ExtendedReferenceContext.ofData(ruleUnitData));
```

the rule unit service gets a *reference* to the computation context (read more below about `DataContext` vs. `ReferenceContext`) and holds onto it; i.e. any change to `ruleUnitData` will be reflected immediately on the state of the `ruleUnitInstance` that gets created.

the rule unit instance ID is returned in the result:

```java
var meta = result.as(RuleUnitMetaDataContext.class);
// /rule-units/org.kie.kogito.quarkus.drools.AnotherService/instances/27387289423
var instanceId = meta.id(RuleUnitInstanceId.class);
assertEquals("/rule-units/org.kie.kogito.quarkus.drools.AnotherService",
        instanceId.ruleUnitId().asLocalUri().path());
```

we then add data to a datasource *directly* (we may also use the data source service)

```java
    ruleUnitData.getStrings().add(new StringHolder("hello folks"));
    ruleUnitData.getStrings().add(new StringHolder("hello people"));
    ruleUnitData.getStrings().add(new StringHolder("hello Mario"));
    ruleUnitData.getStrings().add(new StringHolder("helicopter"));
```

we construct a queryId using the returned `instanceId`; let's assume `Strings` is the name of a query for unit `org.kie.kogito.quarkus.drools.AnotherService`:

```java
// /rule-units/org.kie.kogito.quarkus.drools.AnotherService/instances/27387289423/queries/Strings
    InstanceQueryId queryId = instanceId.queries().get("Strings");
    Stream<ExtendedDataContext> result = ruleUnitService.query(queryId, ExtendedReferenceContext.ofData(EmptyDataContext.Instance));
```

### Adapters

The "most natural" way for users to interact with rule units is to get a reference to a `RuleUnit`/`RuleUnitInstance`s and `DataSource`s without having to fiddle with IDs.

I have introduced here "adapters", which were conceived from the beginning as an optional part of the new API design.

An adapter turns:

```java
svc.action(id, data);
```

into:

```java
obj.action()
```

for instance:

```java
var dID1 = dataSourceSvc.add(dsId, data);
var res1 = ruleSvc.fire(ruleUnitId);
data.setSomething("...");
var dID2 = dataSourceSvc.update(dsId, dataId);
var res2 = ruleSvc.fire(ruleUnitId);
```

may become:

```java
var dataHandle = ruleUnitInstance.context().myDataSource().add(data);
var res1 = ruleUnitInstance.fire();
data.setSomething("...");
dataHandle.update();
var res2 = ruleUnitInstance.fire();
```

An experimental version of this is already provided; we use CDI to create 

1) the rule unit computational context + data sources:

```java
@ApplicationScoped
public class AnotherService implements RuleUnitData, ReferenceContext {
    @Inject
    DataStore<StringHolder> strings;
    @Inject
    DataStore<StringHolder> greetings;
```

then inject an instance

```java
@Inject
RuleUnitInstance<AnotherService> instance;
```

**Note:** this creates a singleton for `AnotherService` and a singleton for its `RuleUnitInstance`. This may be too restrictive for your use cases, and the restriction may be liftable with a notion of scope, and/or by allowing naming rule unit instances explicitly (which is expected in the future)

```java
AnotherService ctx = instance.context();

ctx.getStrings().add(new StringHolder("hello folks"));
ctx.getStrings().add(new StringHolder("hello people"));
ctx.getStrings().add(new StringHolder("hello Mario"));
ctx.getStrings().add(new StringHolder("helicopter"));
```

and query:

```java
Stream<ExtendedDataContext> result = instance.query("Strings", ExtendedReferenceContext.ofData(EmptyDataContext.Instance));
```

**Note:** DataSource types are still `org.kie.kogito.*` types for the reasons already described above.

## DataContext

`DataContext` was previously defined as a computational context that can be converted _into_ another `DataContext` and that is supposedly serializable into JSON. The conversion does not preserve data and may be lossy. For instance:

```
Student(String name, int age, String school)
```

may be *converted* into

```
Person(String name, int age)
```

but if `Person` gets converted back into a `Student`, then `school` may be `null` (it not just may in this implementation: it always will, or it may even throw).

The `DataContext` concepts serves us well in all those situations were we are not really *binding* data to a computation, but really transferring data back and forth from an engine. 

This is always true for "stateless" engines such as PMML and DMN, and for the "stateless" versions of Rules and Processes ("Straight-through processes"). 

It is also basically true for **stateful processes**: in the API that we are exposing #1778 we _always_ pass data in. This is reasonable because of the way processes are exposed in Kogito via REST endpoints, where **data is always received and returned in its entirety** from/to an external serializable, payload.

However, this model is limiting for rules.

## ReferenceContext

I am introducing a new concept that I have called `ReferenceContext` which represents a more general abstraction; i.e. it represents **a computational context** but without the "serializability" requirement. The term **Reference** is to indicate "Java references" i.e. pointers to data.

Because it is more general, `DataContext` now extends `ReferenceContext`.

(Note: I think we will have to rename DataContext into something like "SerializableContext" or "ConvertibleContext" and use DataContext to indicate ReferenceContext as this is the more general, "root" concept)

The reason is that, in stateful rules, the computational context is "bound" to the rule unit instance (or "session"), and that binding is really a binding on a reference. Users are supposed to interact with that reference. 

For instance:

```
School ( DataStore<Student> students
...
school.students.add( new Student( ... )
```

Now, it is true that `School` may be somewhat serializable, but serializing a `ReferenceContext` won't really produce a meaningful result: the expected result may be "the contents of the `students` data store", while it would be more serialized into a place-holder value or an identifier/address of that data source.


api/kogito-api-incubation-rules-services/src/main/java/org/kie/kogito/incubation/rules/services/StatefulRuleUnitService.java
```diff
@@ -0,0 +1,37 @@
...
+    MetaDataContext dispose(LocalId ruleUnitInstanceId);
+
+    MetaDataContext fire(LocalId ruleUnitInstanceId);
+
+    /**
+     * @param queryId identifier of the query
+     * @param params params + meta-data for the query
+     * @return a stream of results (data + meta)
+     */
+    Stream<ExtendedDataContext> query(LocalId queryId, ExtendedReferenceContext params);
```
mariofusco, line 35: In general I don't like methods (especially in public API) that return `java.util.stream.Stream`. Doing so invites users to reference that Stream with a variable and then potentially call more than one terminal operation on it, thus leading to a runtime error.

api/kogito-api-incubation-rules-services/src/main/java/org/kie/kogito/incubation/rules/services/adapters/DataStore.java
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.kogito.incubation.rules.services.adapters;
+
+public interface DataStore<T> {
+    DataHandle add(T value);
```
mariofusco, line 19: Should this also have `update` and `remove`, or did we already decide to keep these operations only on the `DataHandle`? Also should there be a method to retrieve the `DataHandle` of an already added object?

api/kogito-api-incubation-rules-services/src/main/java/org/kie/kogito/incubation/rules/services/adapters/DataStream.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.kogito.incubation.rules.services.adapters;
+
+import org.kie.kogito.incubation.rules.data.DataId;
+
+public interface DataStream<T> {
+    DataId append(T value); // void?
```
mariofusco, line 21: Yes, I agree this should be void.

api/kogito-api-incubation-common/src/main/java/org/kie/kogito/incubation/common/ExtendedReferenceContext.java
```diff
@@ -0,0 +1,63 @@
...
+
+    void setMeta(MetaDataContext meta) {
+        this.meta = meta;
+    }
+
+    void setData(DataContext data) {
+        this.data = data;
+    }
+
+    public MetaDataContext meta() {
```
mariofusco, line 55: I'm not understanding what a `MetaDataContext` is and the difference with a `DataContext`.

evacchi: - MetaDataContext should contain meta-data e.g. ID of the unit, extra config options, diagnostic info, etc.
- DataContext contains the actual data :-)

api/kogito-api-incubation-rules-services/src/main/java/org/kie/kogito/incubation/rules/services/StatefulRuleUnitService.java
evacchi: uhm, I see. A similar API is already present on the stateless "`RuleUnitService`". I didn't want to return a `Collection` because it may give a wrong expectation (e.g. it is all available). 

- Probably the right abstraction would be a Reactive Stream (Provider?). But that would basically make Reactive Streams a hard-dependency (?) 🤔 
- Maybe we should return a DataStream (?) 🤔 

needs some thought.

evacchi: (maybe `Iterable`?)

api/kogito-api-incubation-rules-services/src/main/java/org/kie/kogito/incubation/rules/services/DataSourceService.java
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * @param id identifier of the data source
+     * @param ctx data that should be inserted into the data source
+     * @return id of the inserted fact
+     */
+    // "/data-sources/my-data-source", val
+    DataId add(LocalId id, DataContext ctx);
+
+    // "/data-sources/7574598375943/data/7525792847584395"
+    DataContext get(DataId id);
```
evacchi, line 35: most likely need inverse operation:

```
DataId fromContext(DataContext ctx)
```

notice that in a distributed context, this implies EQUALS semantics (because of serialization over the wire, to query the service); in a local context, this may be ok for IDENTITY semantics as well

notice2: is IDENTITY/EQUALS now a property of the DataSource?

evacchi:
jenkins retest this

evacchi:
jenkins run examples test

evacchi:
jenkins run kogito-examples tests

evacchi:
jenkins run kogito-examples tests

evacchi:
@mariofusco @danielezonca this is green

quarkus/extensions/kogito-quarkus-rules-extension/kogito-quarkus-rules/src/main/java/org/kie/kogito/core/rules/incubation/quarkus/support/StatefulRuleUnitServiceImpl.java
```diff
@@ -0,0 +1,118 @@
...
+
+    @Override
+    public MetaDataContext create(LocalId localId, ExtendedReferenceContext extendedDataContext) {
+        RuleUnitId ruleUnitId;
+        if (localId instanceof RuleUnitId) {
+            ruleUnitId = (RuleUnitId) localId;
+        } else
+            throw new IllegalArgumentException("cannot parse rule unit id");
+
+        RuleUnitData ruleUnitData = (RuleUnitData) extendedDataContext.data();
```
danielezonca, line 50: Also this cast can fail

quarkus/extensions/kogito-quarkus-rules-extension/kogito-quarkus-rules-integration-test/src/main/java/org/kie/kogito/quarkus/drools/AnotherService.java
```diff
@@ -0,0 +1,52 @@
...
+import javax.inject.Inject;
+
+import org.kie.kogito.incubation.common.ReferenceContext;
+import org.kie.kogito.rules.DataStore;
+import org.kie.kogito.rules.RuleUnitData;
+
+import io.quarkus.arc.Unremovable;
+
+@ApplicationScoped
+@Unremovable // temporary workaround to allow injecting RuleUnitInstance<AnotherService>
```
danielezonca, line 28: Do we have a ticket to solve/revisit it?

drools/kogito-ruleunits/src/main/java/org/kie/kogito/rules/units/ListDataStore.java
```diff
@@ -46,6 +46,16 @@ public DataHandle add(T t) {
         return dh;
     }
 
+    public DataHandle findHandle(long id) {
```
danielezonca, line 49: Why is this necessary?

quarkus/extensions/kogito-quarkus-rules-extension/kogito-quarkus-rules/src/main/java/org/kie/kogito/core/rules/incubation/quarkus/support/DataSourceServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    private DataSource<DataContext> getDataSource(DataSourceId dataSourceId) {
+        try {
+            RuleUnitInstanceId instanceId = dataSourceId.ruleUnitInstanceId();
+            RuleUnitId ruleUnitId = instanceId.ruleUnitId();
+            Class<RuleUnitData> ruleUnitDataClass = toClass(ruleUnitId);
+            RuleUnitInstance<?> registeredInstance = ruleUnits.getRegisteredInstance(instanceId.ruleUnitInstanceId());
+            RuleUnitData data = registeredInstance.ruleUnitData();
+            // this may not be necessary at all if we define an actual registry of data sources
```
danielezonca, line 104: Yes please create a follow-up ticket because we should avoid reflection and this naming convention to retrieve the data source

danielezonca (commented):
Few minor comments, mainly to create follow-up tickets

quarkus/extensions/kogito-quarkus-rules-extension/kogito-quarkus-rules/src/main/java/org/kie/kogito/core/rules/incubation/quarkus/support/adapters/RuleUnitInstanceProvider.java
```diff
@@ -0,0 +1,76 @@
...
+            InjectionPoint ip) {
+
+        Type t = ip.getType();
+        if (!(t instanceof ParameterizedType))
+            throw new IllegalArgumentException(t + " must be a ParameterizedType");
+        ParameterizedType pt = (ParameterizedType) t;
+
+        Type[] actualTypeArguments = pt.getActualTypeArguments();
+        if (actualTypeArguments.length < 1)
+            throw new IllegalArgumentException(t + " must have at least a type parameter");
```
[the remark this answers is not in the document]
evacchi: good point

drools/kogito-ruleunits/src/main/java/org/kie/kogito/rules/units/ListDataStore.java
evacchi: to implement [`DataSourceService#get(DataId id)`](https://github.com/kiegroup/kogito-runtimes/pull/1826/files/a5be129cadf2d3c683f7aefbbd122f6a1cb6390b#diff-79487cf1c96c509ed1af0fc0526e243ab0fbea345b9954ef69128d184af30223R45) which returns the DataHandle for a given DataId. This is required by the current implementation of the DataHandle which embeds the ID as well. When we will separate the concepts in the impl this won't be necessary anymore.

quarkus/extensions/kogito-quarkus-rules-extension/kogito-quarkus-rules-integration-test/src/main/java/org/kie/kogito/quarkus/drools/AnotherService.java
evacchi: https://issues.redhat.com/browse/KOGITO-6529

quarkus/extensions/kogito-quarkus-rules-extension/kogito-quarkus-rules/src/main/java/org/kie/kogito/core/rules/incubation/quarkus/support/DataSourceServiceImpl.java
evacchi: https://issues.redhat.com/browse/KOGITO-6530

quarkus/extensions/kogito-quarkus-rules-extension/kogito-quarkus-rules/src/main/java/org/kie/kogito/core/rules/incubation/quarkus/support/StatefulRuleUnitServiceImpl.java
evacchi: I am adding a more explicit type check + error, but this will go away when we actually collapse RuleUnitData over DataContext (it _will_ be a DataContext).

(we may actually make `RuleUnitData` extend `DataContext` but I don't want to twist too much the chain of dependencies for now)

evacchi:
jenkins run kogito-examples tests

evacchi:
jenkins run kogito-examples tests

danielezonca (approved):
LGTM 👍

mariofusco (approved):

evacchi:
jenkins run kogito-examples tests

evacchi:
jenkins run kogito-examples tests

PR #1856, opened 2022-01-13
KOGITO-6534 Always pull latest image version of Data Index

cristianonicolai:


manstis:
Well.. this minimises the versioning problem :smiley: 

I'm doing the same change to trusty.

radtriste (approved):

manstis:
@cristianonicolai With this change, I found I can no longer use local images. Can you verify?

For example when the image name is `org.kie.kogito/trusty-service-infinispan:2.0.0-SNAPSHOT` and I confirm it's in my local cache, Docker still attempts to pull from `org.kie.kogito`. See https://stackoverflow.com/questions/20481225/how-can-i-use-a-local-image-as-the-base-image-with-a-dockerfile

cristianonicolai:
@manstis updated so it only pulls if image tag is `latest`.

ricardozanini (approved):

cristianonicolai:

PR #1864, opened 2022-01-13
[KOGITO-6456] Implementing recognition of children exceptions

fjtirado:
This will allow to use more generic exception in this example 
https://github.com/kiegroup/kogito-examples/blob/stable/serverless-workflow-error-quarkus/src/main/resources/error.sw.json#L43
After this change, it will also work with java.lang.RuntimeException

### DIFF
--- a/jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/context/exception/ExceptionScope.java
+++ b/jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/context/exception/ExceptionScope.java
@@ -15,8 +15,11 @@
  */
 package org.jbpm.process.core.context.exception;
 
+import java.util.Collection;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
+import java.util.Map.Entry;
 
 import org.jbpm.process.core.Context;
 import org.jbpm.process.core.context.AbstractContext;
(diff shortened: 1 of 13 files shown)


cristianonicolai (approved):

jbpm/jbpm-flow/src/main/java/org/jbpm/process/core/context/exception/ExceptionScope.java
```diff
@@ -45,6 +46,22 @@ public ExceptionHandler getExceptionHandler(String exception) {
         return result;
     }
 
+    public ExceptionHandler getExceptionHandler(Throwable exception) {
+        Class<?> exceptionClass = exception.getClass();
+        ExceptionHandler handler = exceptionHandlers.get(exceptionClass.getName());
+        while (handler == null && (exception = exception.getCause()) != null) {
+            exceptionClass = exception.getClass();
+            handler = exceptionHandlers.get(exceptionClass.getName());
+        }
```
elguardian, line 55: as discussed offline I think this is not really common behavior. maybe set this as configurable as it could lead to a very complex behavior or even set the behavior in one way or another.

fjtirado: In order to be able to configure this behaviour in future (throug a serviceclassloader or an alternative mechanism), ExceptionHandlerPolicy interface has been added

ricardozanini (approved):

fjtirado:
jenkins rerun kogito-runtimes tests

PR #1868, opened 2022-01-14
[KOGITO-6548] Support business key when starting process through event

fjtirado:


api/kogito-events-api/src/main/java/org/kie/kogito/event/cloudevents/CloudEventExtensionConstants.java
```diff
@@ -39,6 +39,7 @@
     public static final String PROCESS_USER_TASK_INSTANCE_STATE = "kogitousertaskist";
     public static final String RULE_UNIT_ID = "kogitoruleunitid";
     public static final String RULE_UNIT_QUERY = "kogitoruleunitquery";
+    public static final String BUSINESS_KEY = "kogitobusinesskey";
```
ricardozanini, line 42: Don't we want this key to be included in the `KogitoProcessExtension`? So that will be propagated to the OpenAPI calls as well. https://github.com/kiegroup/kogito-runtimes/blob/main/api/kogito-events-api/src/main/java/org/kie/kogito/event/cloudevents/extension/KogitoProcessExtension.java#L37

Hopefully, with your refactor to CE we could review all these duplications.

fjtirado: This will also imply adding a getBusinessKey method to KogitoProcessInstance. @ricardozanini @cristianonicolai are we fine with that?

cristianonicolai: I think that's ok.

fjtirado: Ok, done

ricardozanini (approved):
Thanks for considering my comments! :)

danielezonca (approved):

cristianonicolai (approved):

OC87484:
Hi there, can this work for an intermediate message as well? For example, if we receive a CloudEvent via Kafka with a businessKey on a topic that the intermediate message is listening to, is there a way for this to trigger Kogito to continue?

PR #1928, opened 2022-01-24
[1.11.x] Review prod modules

cristianonicolai:


jstastny-cz:
I suspect kogito-maven-plugin (which is needed for use with Spring Boot AFAIK) might be problematic, cause it depends on:
* kogito-codegen-processes
* jbpm-flow-builder
* jbpm-bpmn2
So it might not be possible to exclude these modules from productized profile.

cristianonicolai:
@jstastny-cz yeah there is still a lot of cross dependency there, rollback as it basically needs to include most modules

cristianonicolai:
Jenkins rerun kogito-runtimes tests

cristianonicolai:
Jenkins rerun kogito-runtimes tests

cristianonicolai:
jenkins please retest this

cristianonicolai:

cristianonicolai:
Jenkins run examples tests

cristianonicolai:
Jenkins run optaplanner tests

cristianonicolai:
Jenkins run runtimes tests

cristianonicolai:
Jenkins run apps tests

integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-decisions-it/pom.xml
```diff
@@ -0,0 +1,135 @@
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>org.kie.kogito</groupId>
+  <artifactId>integration-tests-springboot-decisions-it</artifactId>
```
jstastny-cz, line 6: I assume this module is a copy of an existing one. To be 100% on the safe side regarding test results reporting - we should adjust java package of the Test clasess - so that we don't overwrite results while running both projects. I understand it's not being run at the moment, so this is just precaution regarding possible inclusion in future.

integration-tests/integration-tests-kogito-plugin/src/it/integration-tests-kogito-plugin-it/src/test/java/org/acme/travels/ManagementAddOnTest.java
```diff
@@ -1,115 +0,0 @@
...
-import org.junit.jupiter.api.Test;
-
-import static io.restassured.RestAssured.given;
-import static org.hamcrest.CoreMatchers.equalTo;
-import static org.hamcrest.CoreMatchers.is;
-import static org.hamcrest.CoreMatchers.not;
-import static org.hamcrest.Matchers.emptyOrNullString;
-
-@QuarkusTest
-class ManagementAddOnTest {
```
MarianMacik, line 31: Is there a way to put this test back? Maybe using an invoker property to include this test only if `-Dproductized` is not specified?

integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-decisions-it/pom.xml
MarianMacik: I think it was a split rather than copying. The only class I noticed that is duplicated is the `BaseRestTest` abstract class.

cristianonicolai: correct, I simply split process from decisions tests and only one class was copied over.

integration-tests/integration-tests-kogito-plugin/src/it/integration-tests-kogito-plugin-it/src/test/java/org/acme/travels/ManagementAddOnTest.java
cristianonicolai: @MarianMacik this is actually tested on process related modules (integration-tests-springboot-processes-it) so this seems not necessary here, especially as it brings this extra deps to an addon.

integration-tests/integration-tests-springboot/src/it/integration-tests-springboot-decisions-it/pom.xml
jstastny-cz: Alright, thank you for explanation.

integration-tests/integration-tests-kogito-plugin/src/it/integration-tests-kogito-plugin-it/src/test/java/org/acme/travels/ManagementAddOnTest.java
MarianMacik: @cristianonicolai But there it is tested on SB, here on Quarkus. Isn't that a different integration or it doesn't matter?

cristianonicolai: @MarianMacik sorry I meant to say it is on both SB ( integration-tests-springboot-processes-it: ManagementAddOnTest ) and Quarkus ( integration-tests-quarkus-processes: ManagementAddOnIT )

MarianMacik: That's fine then.

PR #1929, opened 2022-01-24
PLANNER-2641 OptaPlanner depends on Drools, not Kogito

triceo:
Upstream PR:
https://github.com/kiegroup/kogito-pipelines/pull/351

Downstream PRs:
https://github.com/kiegroup/optaplanner/pull/1800

### DIFF
--- a/.ci/jenkins/dsl/jobs.groovy
+++ b/.ci/jenkins/dsl/jobs.groovy
@@ -37,29 +37,13 @@ Map getMultijobPRConfig() {
                     DISABLE_SONARCLOUD: !Utils.isMainBranch(this),
                 ]
             ], [
-                id: 'optaplanner',
-                dependsOn: 'kogito-runtimes',
-                repository: 'optaplanner',
-            ], [
                 id: 'kogito-apps',
-                dependsOn: 'optaplanner',
+                dependsOn: 'kogito-runtimes',
                 repository: 'kogito-apps'
             ], [
                 id: 'kogito-examples',
-                dependsOn: 'Optaplanner',
+                dependsOn: 'kogito-runtimes',
                 repository: 'kogito-examples'
-            ], [
-                id: 'optaweb-employee-rostering',
-                repository: 'optaweb-employee-rostering'
-            ], [
-                id: 'optaweb-vehicle-routing',
-                repository: 'optaweb-vehicle-routing'
-            ], [
-                id: 'optaplanner-quickstarts',
-                repository: 'optaplanner-quickstarts',
-                env : [
-                    OPTAPLANNER_BUILD_MVN_OPTS_UPSTREAM: '-Dfull'
-                ]
             ]
         ],
     ]


Ginxo (approved):
LGTM, but @radtriste is the most appropriated person to review it.

.ci/jenkins/dsl/jobs.groovy
```diff
@@ -38,28 +38,16 @@ Map getMultijobPRConfig() {
                 ]
             ], [
                 id: 'optaplanner',
-                dependsOn: 'kogito-runtimes',
+                dependsOn: 'drools',
                 repository: 'optaplanner',
```
radtriste, line 42: if `optaplanner` does not depend on `kogito-runtimes` anymore, then the optaplanner check can also be removed] I would say.
And `kogito-apps` and `kogito-examples`, would depend on `kogito-runtimes`

.ci/jenkins/dsl/jobs.groovy
```diff
@@ -38,28 +38,16 @@ Map getMultijobPRConfig() {
...
+                dependsOn: 'optaplanner',
                 repository: 'kogito-examples'
-            ], [
-                id: 'optaweb-employee-rostering',
-                repository: 'optaweb-employee-rostering'
-            ], [
-                id: 'optaweb-vehicle-routing',
-                repository: 'optaweb-vehicle-routing'
-            ], [
-                id: 'optaplanner-quickstarts',
```
radtriste, line 58: no quickstart using kogito ? I guess they are in kogito-examples ?

.ci/jenkins/dsl/jobs.groovy
triceo: But `kogito-examples` still has OptaPlanner code inside.

.ci/jenkins/dsl/jobs.groovy
triceo: Exactly.

.ci/jenkins/dsl/jobs.groovy
radtriste: here we are building PR jobs which should be executed in case of a change in `kogito-runtimes`, so `optaplanner` does not need to be built as it does not depend on `kogito-runtimes`.
(we are only talking about job dependencies here, and not project dependencies, which is in fact not so relevant right now as the jobs are all executed in parallel... but we need to keep it up-to-date)
So from a `kogito-runtimes` change, we only need to check `kogito-apps` and `kogito-examples`, that is why those 2 would depend on `kogito-runtimes`.

BUT in an optaplanner PR, it would need to check both kgoito-apps and kgoito-examples, as already defined in https://github.com/kiegroup/optaplanner/blob/main/.ci/jenkins/dsl/jobs.groovy#L30. So if you do a change in optaplanner, it still checks kgoito-apps/examples.

triceo:
@radtriste Done.

radtriste (approved):

triceo:
Jenkins retest this

PR #2126, opened 2022-04-04
KOGITO-6936 - Drop Xstream

cristianonicolai:
Replacement for https://github.com/kiegroup/kogito-runtimes/pull/1959

PR #2142, opened 2022-04-08
[1.13.x] KOGITO-6936 - Drop Xstream from process modules

cristianonicolai:
Cherry-pick https://github.com/kiegroup/kogito-runtimes/pull/2126
Removing only the process use of Xstream and upgrading to latest available as drools-core make use of it.


cristianonicolai:
@evacchi not sure if the drools-core can be reviewed to drop xstream or we simply leave it as is.

mareknovotny (approved):
lgtm

PR #2164, opened 2022-04-22
Revert "[KOGITO-6989] Error dumping files to disk should not break build"

danielezonca:
Reverts kiegroup/kogito-runtimes#2157


radtriste (approved):

fjtirado (changes requested):
Better use https://github.com/kiegroup/kogito-runtimes/pull/2165

PR #2205, opened 2022-05-12
KOGITO-7152 Add predefined logger for processes when dev mode is enabled

cristianonicolai:


ricardozanini (approved):

krisv:
PR looks good!  Wondering if we should also add a way to override this when not running in dev mode?  Thinking about running this on an actual kube but still for testing, could we maybe use a property to override the behavior that this would only log in dev mode, to still enable the logger?  I believe that would be similar to how swagger-ui typically only shows up in dev mode but can be overridden?

cristianonicolai:
@krisv make sense to me. I'll merge this as is and work on a property to allow including it beyond dev mode.

PR #2208, opened 2022-05-18
Rename SW test not getting triggered and resources copied causing it …

cristianonicolai:
…to show in git list


fjtirado (approved):
Thanks Cristiano
@davidesalerno FYI

PR #2214, opened 2022-05-23
[1.13.x] KOGITO-7201 - Filter starters and addons based on profile

cristianonicolai:
Creating a project without including a starter will generate the following error:

`[ERROR] Failed to execute goal org.apache.maven.plugins:maven-archetype-plugin:3.2.1:generate (default-cli) on project standalone-pom: Invalid Kogito starter, please use one of the following: rules, decisions, predictions -> [Help 1]`

### DIFF
--- a/springboot/archetype/src/main/resources/META-INF/archetype-post-generate.groovy
+++ b/springboot/archetype/src/main/resources/META-INF/archetype-post-generate.groovy
@@ -22,6 +22,52 @@ import java.nio.file.Paths
 
 log = LoggerFactory.getLogger("org.apache.maven")
 
+includeAllArtefacts = ${artefacts.all}
+
+validStarters = [
+        [id: "rules", starter: "kogito-rules-spring-boot-starter"],
+        [id: "decisions", starter: "kogito-decisions-spring-boot-starter"],
+        [id: "predictions", starter: "kogito-predictions-spring-boot-starter"]
+]
+
+if (includeAllArtefacts) {
+    validStarters.addAll([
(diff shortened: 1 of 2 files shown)


springboot/archetype/src/main/resources/META-INF/archetype-post-generate.groovy
```diff
@@ -22,20 +22,58 @@ import java.nio.file.Paths
...
+            [id: "processes", starter: "kogito-processes-spring-boot-starter"],
+            [id: "serverless-workflows", starter: "kogito-serverless-workflows-spring-boot-starter"]
+    ])
+}
+
+// this list should be maintained manually for now for each generic/spring boot add-on we create
+// see: https://issues.redhat.com/browse/KOGITO-5619
+validAddons = [
+        [id: "messaging", addon: "kogito-addons-springboot-messaging"],
+        [id: "events-decisions", addon: "kogito-addons-springboot-events-decisions"],
```
radtriste, line 44: what about rules and predictions defined in https://github.com/kiegroup/kogito-runtimes/blob/1.13.x/springboot/addons/events/pom.xml#L18-L20 ?

radtriste (commented):
see my comment. else tested ok

springboot/archetype/src/main/resources/META-INF/archetype-post-generate.groovy
cristianonicolai: @radtriste added

ricardozanini (approved):

PR #2263, opened 2022-06-14
KOGITO-7382 - Allow to override Data Index Dev Service image

cristianonicolai:


quarkus/extensions/kogito-quarkus-processes-extension/kogito-quarkus-processes-deployment/pom.xml
```diff
@@ -12,6 +12,10 @@
   <artifactId>kogito-quarkus-processes-deployment</artifactId>
   <name>Kogito :: Quarkus Processes Extension :: Deployment</name>
 
+  <properties>
+    <data-index-ephemeral.image>quay.io/kiegroup/kogito-data-index-ephemeral</data-index-ephemeral.image>
```
mareknovotny, line 16: i would not set the full path of the image in the property. 
First it is hard to use on command line and second we could invite users/devs to completely overwrite it with no check? Just checking
```suggestion
    <data-index.registry>quay.io</data-index.registry>
    <data-index-ephemeral.image>{data-index.registry}/kiegroup/kogito-data-index-ephemeral</data-index-ephemeral.image>
```

nmirasch (approved):

krisv (approved):
Probably still want to look at if we could override this in productized build somehow so users of productized jars don't have to always override to get the productized image.

quarkus/extensions/kogito-quarkus-processes-extension/kogito-quarkus-processes-deployment/pom.xml
cristianonicolai: @mareknovotny this is consumed during build time for Maven filtering so users can't really change that when using the extension. As for splitting image and registry, I think that's probably unnecessary, as it adds complexity to deal with for instance not providing a repository. It just relies on Docker image naming convention.

cristianonicolai:
> 

@krisv this is exactly the idea here. This property should be provided during the productized build and will be used by the generated artefacts, end-users won't need to override this on their end.

ricardozanini (approved):

krisv:
If productization could override this during prod build without further "magic" on our side, than I'm all good ;)

quarkus/extensions/kogito-quarkus-processes-extension/kogito-quarkus-processes-deployment/pom.xml
cristianonicolai: @spolti that's why I'm keeping a single String parameter there, so during productization build, it only needs to set _-Ddata-index-ephemeral.image=new image name_, and that image will be used by default in the generated artefacts.

spolti (approved):

PR #2264, opened 2022-06-14
KOGITO-4015 - Add integration tests for custom async WIH with persist…

cristianonicolai:
…ence

### DIFF
--- a/integration-tests/integration-tests-quarkus-processes-persistence/integration-tests-processes-persistence-common/src/test/java/org/kie/kogito/it/PersistenceTest.java
+++ b/integration-tests/integration-tests-quarkus-processes-persistence/integration-tests-processes-persistence-common/src/test/java/org/kie/kogito/it/PersistenceTest.java
@@ -16,6 +16,7 @@
 package org.kie.kogito.it;
 
 import java.math.BigDecimal;
+import java.time.Duration;
 import java.time.Instant;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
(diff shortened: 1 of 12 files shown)


cristianonicolai:
@ricardozanini this IT should replicate the scenario you described in the JIRA. This issue has been fixed by previous changes so just adding the test case.

integration-tests/integration-tests-quarkus-processes-persistence/integration-tests-processes-persistence-common/src/test/java/org/kie/kogito/it/PersistenceTest.java
```diff
@@ -225,4 +229,26 @@ void testMultipleInstance() {
...
+                .pathParam("processId", PROCESS_ASYNC_WIH)
+                .when()
+                .post("/{processId}")
+                .then()
+                .statusCode(201)
+                .body("id", not(emptyOrNullString()))
+                .extract()
+                .path("id");
+
+        await().atMost(TIMEOUT)
```
jstastny-cz, line 245: The course of test is here following?
1. Start process with custom workitem, get id.
2. Check that process with that id has finished

?

From the Jira I got the impression we would like to allow for a more complex behavior:
```
Marshallers are calling disconnect every time a process is persisted. Once we call this method, the current ProcessManager removes the persisted instance from memory.

If a Async work item handler is trying to use that manager to complete a work item when the future task is completed, it will fail because it want be able to find the process instance, something that did not happen when persistence is not active
```

So the original reported issue would most probably fail silently in this test wouldn't it? Cause it's the async wih that fails - will that result in any relevant outcome in the course of this test case? I don't see it.

cristianonicolai: @jstastny-cz the process would simply not finish if there is any issue while executing the WIH. As the WIH execution is async, the process state is persisted on the node where the WIH will be later executed. If it fails to execute, the process will simply get stuck there and not complete.

jstastny-cz: Thank you for explaining.

jstastny-cz (approved):
Thanks.

PR #2269, opened 2022-06-16
[KOGITO-7397] Fixing bug in order of parameters

fjtirado:
The HashMap should be a LinkedHashMap or the order of the parameters will not necessarily be respected. 
This is difficult to test, since it will fail or not depending on the name of the parameters (the order in the hash in the map), so I added and extra parameter to the multi param scenario

### DIFF
--- a/quarkus/extensions/kogito-quarkus-serverless-workflow-extension/kogito-quarkus-serverless-workflow-integration-test/src/main/resources/pojoServiceFilter.sw.json
+++ b/quarkus/extensions/kogito-quarkus-serverless-workflow-extension/kogito-quarkus-serverless-workflow-integration-test/src/main/resources/pojoServiceFilter.sw.json
@@ -27,7 +27,8 @@
             "refName": "getPerson",
             "arguments": {
               "name": ".name",
-              "age": 27
+              "age": ".age",
+              "income": 20000.50
             }
           },
           "actionDataFilter": {
--- a/kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/handlers/CompositeContextNodeHandler.java
+++ b/kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/handlers/CompositeContextNodeHandler.java
@@ -15,8 +15,8 @@
  */
 package org.kie.kogito.serverless.workflow.parser.handlers;
 
-import java.util.HashMap;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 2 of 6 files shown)


quarkus/extensions/kogito-quarkus-serverless-workflow-extension/kogito-quarkus-serverless-workflow-integration-test/src/main/resources/pojoServiceFilter.sw.json
```diff
@@ -27,7 +27,8 @@
             "refName": "getPerson",
             "arguments": {
               "name": ".name",
-              "age": 27
+              "age": 27,
+              "income": 20000.50
```
wmedvede, line 31: Hi Francisco,
in general looks good to me, but see my note here:

Around the filters you are repeating "**name**", right?

          "actionDataFilter": {
             "results" : "{name: .name}"
          } 
        }],
      "stateDataFilter" : {
        "output" : "{name: .name}"
      },

However this is not a good strategy for this type of test.
Why?
Because the pojo that is returned from the service already has a filed "name", and thus, there might be situations of a bug where the filtering is now working well but the final result sill gets a field "name" because this was the name in the pojo.

Please, use something like this instead: (will take 5 min)

          "actionDataFilter": {
             "results" : "{name_to_the_result: .name}"
          } 
        }],
      "stateDataFilter" : {
        "output" : "{name_to_the_output: .name_to_the_result}"
      },

And please, in the test, check that final workflow data contains:

1) a field "name_to_the_output" with the "expected_name"
2) but also that other fields that were passed originally in the workflow data are not being overwrited by the 
stateDataFilter.

for example

if you started the workflow with 

workflowdata: {

"param1" : "the_param1"

}

the final result must be

workflowdata: {

"param1" : "the_param1" ,
"name_to_the_output": "expected_name"

}


Pleas add this considerations, I takes 5 minutes and makes the test much stronger.

wmedvede (commented):
Looks good, just a small thing to improve.
Thanks.

quarkus/extensions/kogito-quarkus-serverless-workflow-extension/kogito-quarkus-serverless-workflow-integration-test/src/main/resources/pojoServiceFilter.sw.json
wmedvede: true thing, the param won't survive in this execution. (my mistake)
But to repeat "name" around the filter, independent that you are now passing "input_name",  I think is not a good idea.
That's all I can say.

fjtirado: I agree the test was incomplete (it was testing that the stateDataFilter worked but it might still work if the actionDataFilter fails). In order to avoid this false green, it is enough to append a value to the original name in the service and check that the target name value has indeed changes in the test (and this way we still test that it is possible to modify values with the same name)

wmedvede (approved):

krisv (approved):

PR #2296, opened 2022-06-27
KOGITO-7468 Incubation API: Move ID API into `application` module

evacchi:
In the incubation API there are two logical parts:

- an *addressing* scheme to indicate a resource
- a *data* API to manipulate and transfer data payloads across services

These two components are currently mixed in a weird way inside the `common` module; a part of the addressing API lives alone in the `application` module.

I am moving the addressing part to the `application` module and package; the leftovers (data-related) will stay in common. If it makes sense, then we will also rename those modules in the future.

Note: this is also in preparation to further moving/renaming the addressing API for shared usage in Drools.


evacchi:
close as stale / outdated (will leave the branch for reference)

PR #2306, opened 2022-07-01
KOGITO-7487 - Skip internal variables for SW and for variable change …

cristianonicolai:
…on MergeAction

### DIFF
--- a/kogito-serverless-workflow/kogito-serverless-workflow-runtime/src/main/java/org/kie/kogito/serverless/workflow/actions/MergeAction.java
+++ b/kogito-serverless-workflow/kogito-serverless-workflow-runtime/src/main/java/org/kie/kogito/serverless/workflow/actions/MergeAction.java
@@ -19,6 +19,8 @@ import org.jbpm.process.instance.impl.Action;
 import org.kie.kogito.internal.process.runtime.KogitoProcessContext;
 import org.kie.kogito.jackson.utils.MergeUtils;
 
+import com.fasterxml.jackson.databind.JsonNode;
+
 import static org.kie.kogito.serverless.workflow.actions.ActionUtils.getJsonNode;
 
 public class MergeAction implements Action {
@@ -33,6 +35,15 @@ public class MergeAction implements Action {
 
     @Override
     public void execute(KogitoProcessContext context) throws Exception {
-        MergeUtils.merge(getJsonNode(context, inputName), getJsonNode(context, outputName));
+        JsonNode inputNode = getJsonNode(context, inputName);
+        if (inputNode.isEmpty()) {
+            return;
+        }
+        JsonNode targetNode = getJsonNode(context, outputName);
+        if (inputNode.equals(targetNode)) {
+            return;
+        }
+        JsonNode merge = MergeUtils.merge(inputNode, targetNode);
+        context.setVariable(outputName, merge);
     }
 }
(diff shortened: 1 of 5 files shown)


ricardozanini (approved):

kogito-serverless-workflow/kogito-serverless-workflow-runtime/src/main/java/org/kie/kogito/serverless/workflow/actions/MergeAction.java
```diff
@@ -33,6 +35,7 @@ public MergeAction(String inputName, String outputName) {
 
     @Override
     public void execute(KogitoProcessContext context) throws Exception {
-        MergeUtils.merge(getJsonNode(context, inputName), getJsonNode(context, outputName));
+        JsonNode merge = MergeUtils.merge(getJsonNode(context, inputName), getJsonNode(context, outputName));
+        context.setVariable(outputName, merge);
```
fjtirado, line 39: why this is needed? To trigger the variable change event?

fjtirado (commented):
Im not sure forcing a variable event change is the way to go here. There should be better alternatives to create a log (This is the first new I have we have a logger)
One that come to my mind is to record action execution and create a log from that (rather than from internal events)

kogito-serverless-workflow/kogito-serverless-workflow-runtime/src/main/java/org/kie/kogito/serverless/workflow/actions/MergeAction.java
fjtirado: Note that merge action might have not done anything because the inputvarialbe is empty (it happend when the workflow function returns not data) so we might be reporting variable changes that are not really variable changes

cristianonicolai: @fjtirado thanks, It makes sense to check if input has some value and if it is equals to the target json node, in these two cases we can skip the update.
Regarding setting the variable, this is so we can feed back into the engine about the state change and any hook can be triggered, this is fundamental in my opinion to keep the engine working as expected. We should retrieve and update the variables state within the knowledge of the engine every time. Without it, I believe it is hard to predict the side effects it could cause to the state, by only keeping this in memory. This could, for instance, impact persistence. So listeners, or log, is just one piece of why this change is needed, but overall a way to keep the engine aware of any changes in the state so it can properly react to those changes.

fjtirado: @cristianonicolai I agree is important to keep the engine aware... till a point (Engine does not really do a great job in handling pojo, or any complex object, changes by scripting, which is the scenario Im concerned about) . I would like to discuss another possibilities, since this only partially covered by this change.

fjtirado: One possibility will be to proxy the json node, so this proxy record changes and internal trigger of node handles them
Another possibility is to add listeners to ExpressionHandler (which will do essentially the same than the previous step, record changes and intenal trigger of node handles them)
The problem of this approaches is that you know a JsonNode has changed, but you do not know the name of the variable that changed.

fjtirado (approved):

cristianonicolai:
@krisv @ricardozanini @fjtirado refactored to split into two loggers, using process type. That should make the loggers a bit cleaner and easier to customize later on.

ricardozanini (approved):

PR #2318, opened 2022-07-07
KOGITO-7461 - Add event dispatcher listener to make for the possibility of throttling the message consumer

iurybakov:
This functionality allowed us to register listeners of the event handler execution, which allows us to have metrics of worker thread loading during execution and thereby be able to control the intensity of the event flow. In simple words, this functionality allows us to implement throttling of the event flow. As a result, we were able to solve the 'thread blocked' problem.

### DIFF
--- a/api/kogito-services/src/main/java/org/kie/kogito/services/event/EventDispatcherListener.java
+++ b/api/kogito-services/src/main/java/org/kie/kogito/services/event/EventDispatcherListener.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2022 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


iurybakov:
This functionality allowed us to register listeners of the event handler execution, which allows us to have metrics of worker thread loading during execution and thereby be able to control the intensity of the event flow. In simple words, this functionality allows you to implement throttling of the event flow. As a result, we were able to solve the 'thread blocked' problem.

fjtirado:
> Therefore, our idea is to be able to control the intensity of the flow of events from outside. For example suspending the kafka consumer. And for control, an indication of the exhaustion of the pool of worker threads on the event dispatcher is needed.

Ok, but I think there are better alternatives to achieve that
For example, to give users the possibility of changing the executor, which will be as easy as adding a defaultbean annoation to  @Inject
    @javax.inject.Named(KogitoEventExecutor.BEAN_NAME)
    ExecutorService executorService;
in the generated TravelersMessageConsumer, so you can provide a different executor with a differen rejectionhandler. 

Alternatively, we can change scope for consume method from private to protected, so generated consumer  can be inherited, and consume method override this way

protected CompletionStage<?> consume(Object payload)  {
   super.consume(payload).whenComplete (dosomething);
}

iurybakov:
@fjtirado great, I have understood you. I will try all your options and prepare a PR(s). I'll keep you posted on the results. Thanks!

fjtirado (changes requested):
@iurybakov is working on alternative approach. Thanks!

iurybakov:
In the process of researching the proposed ideas, another alternative solution was found https://github.com/kiegroup/kogito-runtimes/pull/2361

PR #2330, opened 2022-07-14
Align version with Quarkus

cristianonicolai:
- Add script for testing with Quarkus snapshot

### DIFF
--- a/scripts/quarkusLatest.sh
+++ b/scripts/quarkusLatest.sh
@@ -1,0 +1,5 @@
+# This script allows to replicate the changes necessary to run Kogito using Quarkus latest snapshot
+# Ensure to build Quarkus locally first
+mvn -f ../pom.xml versions:set-property -pl :kogito-dependencies-bom -Dproperty=version.io.quarkus -DnewVersion=999-SNAPSHOT -DgenerateBackupPoms=false
+mvn -f ../pom.xml versions:set-property -pl :kogito-dependencies-bom -Dproperty=version.io.quarkus.quarkus-test-maven -DnewVersion=999-SNAPSHOT -DgenerateBackupPoms=false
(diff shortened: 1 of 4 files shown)


scripts/quarkusLatest.sh
```diff
@@ -0,0 +1,5 @@
+# This script allows to replicate the changes necessary to run Kogito using Quarkus latest snapshot
```
radtriste, line 1: We should document the build-chain command instead IMO because it will do it automatically with the pr config

cristianonicolai: could we run that locally then? I guess that is my point here to be able to have something to quickly reproduce locally.

radtriste: You could yes. You can run build-chain for a specific branch. It will take all upstream repo, checkout and install artifacts. That way you don't need to take care of anything.
See https://github.com/kiegroup/github-action-build-chain#execution-build-action---branch-flow-arguments

radtriste: If you need to update this PR, I would suggest to remove the shell script

radtriste: we can update the doc later in another PR

radtriste:
I see none of those updates in https://github.com/kiegroup/kogito-runtimes/pull/2337.
Could it be that quarkus main is updated after ?

radtriste:
well the quarkus-branch check will update automatically those for the check.
But we would need to update to a Final version to have those updated permanently. The compare version is only done and committed when we upgrade to a new fix release ...
Could it be that the update of deps was done after a fix release ?

cristianonicolai:
@radtriste not sure but I think something like a major bump in a version like the Keycloack one wouldn't be something to be changed in a fix/minor release.

radtriste:
so you mean when we updating to Quarkus 2.10, those version were not updated with it ?
That means that the Maven `compare-dependencies` is not enough ?

cristianonicolai:
yeah, it might have some sort of edge cases that we miss some updates.

radtriste:
closing as changes are already on main

PR #2348, opened 2022-07-22
KOGITO-7625 - Add ExceptionScope to native resource

cristianonicolai:


cristianonicolai:
Jenkins run native

fjtirado (approved):
This is a weird, since excepcion scope is not new and the change does not increase the number of instances, so it should have failed before the change if ExcepcionScope was not there in the native set. 
But anyway the change is good, ExceptionScope needs to be there.

PR #2349, opened 2022-07-22
KOGITO-7626 Setup Maven wrapper and build-chain docs

radtriste:
- https://github.com/kiegroup/kogito-pipelines/pull/565
- https://github.com/kiegroup/kogito-runtimes/pull/2349


.github/ci-prerequisites.sh
```diff
@@ -12,7 +12,8 @@
 current_path=$(pwd)
 
 cd ../drools
-mvn clean install -Dquickly
+# TODO setup mvn wrapper on Drools
```
radtriste, line 15: TODO

.github/workflows/kogito-runtimes-pr.yml
```diff
@@ -44,6 +44,7 @@ jobs:
         with:
           annotations-prefix: ${{ runner.os }}-${{ matrix.java-version }}/${{ matrix.maven-version }}
           github-token: "${{ secrets.GITHUB_TOKEN }}"
+          definition-file: 'https://raw.githubusercontent.com/radtriste/kogito-pipelines/kogito-7626-maven-wrapper/.ci/pull-request-config.yaml'
```
radtriste, line 47: To revert before merging. This is test only

gitgabrio:
HI @radtriste - thanks for that PR.
I just have two question:
1) I was expecting somewhere the removal of "3.8.6" hardcoded check inside `kogito-dependencies-bom`: am I missing it ? Is it still needed ?
2) if it is needed to keep that enforce rule, would this mean that also on lacla machine we will have to use the mvnw (just asking) ?
Thanks!

PR #2361, opened 2022-07-26
KOGITO-7461: add throttling publisher decorator

iurybakov:
Hello!

This addon allow to us fix "The current thread cannot be blocked".

This PR is a continuation of the research of the problem solution, previously created PR and discussion of the problem https://github.com/kiegroup/kogito-runtimes/pull/2318.

### DIFF
--- a/quarkus/addons/messaging/throttling/src/main/java/org/kie/kogito/addon/quarkus/messaging/throttling/ThrottledMultiOperator.java
+++ b/quarkus/addons/messaging/throttling/src/main/java/org/kie/kogito/addon/quarkus/messaging/throttling/ThrottledMultiOperator.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2022 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.kogito.addon.quarkus.messaging.throttling;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.ConfigProvider;
+import org.eclipse.microprofile.reactive.messaging.Message;
+import org.kie.kogito.event.KogitoEventExecutor;
+
+import io.smallrye.mutiny.Multi;
+import io.smallrye.mutiny.operators.multi.AbstractMultiOperator;
+import io.smallrye.mutiny.operators.multi.MultiOperatorProcessor;
+import io.smallrye.mutiny.subscription.MultiSubscriber;
+
+public class ThrottledMultiOperator extends AbstractMultiOperator<Message<?>, Message<?>> {
+
+    private final int limit;
+
+    private final AtomicInteger processedMessageCount = new AtomicInteger(0);
+
+    public ThrottledMultiOperator(Multi<? extends Message<?>> upstream) {
+        super(upstream);
+        final Config config = ConfigProvider.getConfig();
+        final int factor = config
+                .getOptionalValue("kogito.quarkus.events.throttling.threadQueueSizeFactor", Integer.class)
+                .orElse(2);
+        this.limit =
+                config.getOptionalValue(KogitoEventExecutor.MAX_THREADS_PROPERTY, Integer.class)
+                        .orElse(KogitoEventExecutor.DEFAULT_MAX_THREADS_INT)
+                        +
+                        (config.getOptionalValue(KogitoEventExecutor.QUEUE_SIZE_PROPERTY, Integer.class)
+                                .orElse(KogitoEventExecutor.DEFAULT_QUEUE_SIZE_INT) / factor);
+    }
+
+    @Override
+    public void subscribe(MultiSubscriber<? super Message<?>> downstream) {
+        upstream.subscribe().withSubscriber(new CustomProcessor(downstream));
+    }
+
+    private class CustomProcessor extends MultiOperatorProcessor<Message<?>, Message<?>> {
+
+        CustomProcessor(MultiSubscriber<? super Message<?>> downstream) {
+            super(downstream);
+        }
+
+        @Override
+        public void onItem(Message<?> item) {
+            processedMessageCount.incrementAndGet();
+            super.onItem(new MessageProxy<>(item, (r, ex) -> {
+                if (processedMessageCount.decrementAndGet() == limit) {
+                    super.request(1);
+                }
+            }));
+        }
+
+        @Override
+        public void request(final long n) {
+            if (processedMessageCount.get() <= limit) {
+                super.request(1);
+            }
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


iurybakov:
This is simple solution and a working solution that demonstrates the idea well. I am grateful in advance for the comments and remarks.

quarkus/addons/messaging/throttling/src/main/java/org/kie/kogito/addon/quarkus/messaging/throttling/ThrottlingPublisherDecorator.java
```diff
@@ -0,0 +1,39 @@
...
+
+import io.smallrye.mutiny.Multi;
+import io.smallrye.reactive.messaging.providers.PublisherDecorator;
+
+@ApplicationScoped
+public class ThrottlingPublisherDecorator implements PublisherDecorator {
+
+    @Override
+    public Multi<? extends Message<?>> decorate(Multi<? extends Message<?>> multi, String s) {
+        final boolean isEventsThrottlingEnabled = ConfigProvider.getConfig()
```
fjtirado, line 31: I do not think this property is needed. 
If the addon is enable and this property is disable, the addon does nothing, so there is not point on having that, adding the addon should be enough. If the addon is not needed, then just not add it to the pom of the project

fjtirado (commented):
I think we need a usage example. 
I still this should be better handled by providing a different rejectionhandled at executor level.

quarkus/addons/messaging/throttling/src/main/java/org/kie/kogito/addon/quarkus/messaging/throttling/ThrottledMultiOperator.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    private class CustomProcessor extends MultiOperatorProcessor<Message<?>, Message<?>> {
+
+        CustomProcessor(MultiSubscriber<? super Message<?>> downstream) {
+            super(downstream);
+        }
+
+        @Override
+        public void onItem(Message<?> item) {
```
fjtirado, line 62: Let me try to see if I understood the approach, In this method we track the number of message being currently processed. 
In the callback that is called when the process has been process, if the count already went beyond the limit, the message that decrease the count notify the consumer to request one more item.

tiagodolphine (commented):
@iurybakov thanks for bringing this issue, I'd say for now the most important thing would be to create an integration test to reproduce the issue, in this PR or in a different one. Once we have it we can open a Jira to work on the solution, probably adding a new addon is not necessary it could be done in the core... we need to enable the backpressure in some way to stop consuming events once the executor is full, with no available threads.
@fjtirado fyi

fjtirado:
@tiagodolphine @iurybakov The JIRA is https://issues.redhat.com/browse/KOGITO-7259
I can work in a solution that does not require an addon, but it would be nice, as Tiago is saying, to have an integration test to verify.

quarkus/addons/messaging/throttling/src/main/java/org/kie/kogito/addon/quarkus/messaging/throttling/ThrottledMultiOperator.java
```diff
@@ -0,0 +1,78 @@
...
+            processedMessageCount.incrementAndGet();
+            super.onItem(new MessageProxy<>(item, (r, ex) -> {
+                if (processedMessageCount.decrementAndGet() == limit) {
+                    super.request(1);
+                }
+            }));
+        }
+
+        @Override
```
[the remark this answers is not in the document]
iurybakov: Method request called from reactive stream logic, in method "drainLoop". It quite complicated logic.

Ignoring n - yes, great remark. I fix it, and test again.

quarkus/addons/messaging/throttling/src/main/java/org/kie/kogito/addon/quarkus/messaging/throttling/ThrottledMultiOperator.java
iurybakov: Yes, exactly.

quarkus/addons/messaging/throttling/src/main/java/org/kie/kogito/addon/quarkus/messaging/throttling/ThrottlingPublisherDecorator.java
iurybakov: Ok, good remark. Thanks!

And next:

" in terms of memory consumption is better that using an executor with an unbouded queue, because Im not sure how the kafka emitter really works underneath (maybe it has an internal queue that will fill up if events are not emitted after a while) If it is not better in terms of memory comsumption, then using an unbounded queue for the executor will be basically equivalent in terms of performance than this implementation"

In my opinion, not really. For Kafka client there is some subscription logic exists [KafkaRecordStreamSubscription](https://github.com/smallrye/smallrye-reactive-messaging/blob/8d1c0515f0d1e52cc0c446987a670981d53c72ef/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/impl/KafkaRecordStreamSubscription.java#L132).
Where there is a buffering message queue. And there is even a logic for suspending topic reading if the queue is full. But there is no connection with the downstream into which messages from the queue are thrown. In other words, the kafka client does not care about the capabilities and state of its downstream. Based on this implementation, an infinite queue of tasks for threads will create a guaranteed OOM under heavy load.

iurybakov:
Yeah, I prepare the project, where you can reproduce problem.

fjtirado:
@iurybakov  Im working on PR without addon. Ill try to change default rejection policy so rather than blocking the thread when the executor is exhausted, we stop emitter before that.

iurybakov:
@fjtirado, it great idea. I wanted to do the same, but I couldn't do this properly. I will be glad if you have a good solution.

quarkus/addons/messaging/throttling/src/main/java/org/kie/kogito/addon/quarkus/messaging/throttling/ThrottlingPublisherDecorator.java
fjtirado: ok fine. Im using your idea to implement similar logic in KogitoEventExecutor, so there is not need to new addon.
the idea is rather than executing in the same thread (which lead to the blocking thread exception) tell the emitter to stop emitting as you are doing in this addon.

iurybakov:
@tiagodolphine @fjtirado An [example](https://github.com/iurybakov/workflow-engines-load-test/tree/kogito-bpmn-sync-kafka-throttling/kogito-bpmn-sync-kafka-throttling), where you can reproduce problem.

quarkus/addons/messaging/throttling/src/main/java/org/kie/kogito/addon/quarkus/messaging/throttling/ThrottlingPublisherDecorator.java
iurybakov: Looks great!:) One thing, the many BackpressureProcessor and many KogitoThreadPoolExecutor use one static status KogitoEmitterStatus. So, I think one BackpressureProcessor must relate to one KogitoThreadPoolExecutor by their own KogitoEmitterStatus, otherwise free channel will be suspended by another one loaded channel.

iurybakov: And maybe instead of synchronized the better choice will be [concurrent deque](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentLinkedDeque.html), how do you think?

fjtirado:
@iurybakov PR was merged, please try and let us know

iurybakov:
Yeah, thanks! Good job! We will do it:)

PR #2425, opened 2022-08-31
KOGITO-7878 - Override workitemhandler.getName method with work item …

cristianonicolai:
…handler name, using fixed name instead of class name.


fjtirado (commented):
There is something I do not get, the default implementation for getName does this
```
 default String getName() {
        return getClass().getSimpleName();
    }
```
which should be equivalent to classname

Why this is fixing the issue? I guess is related with quarkus proxies (we put a fixed name there and that will be it), but this should basically not happen.

ricardozanini (approved):

cristianonicolai:
@fjtirado correct, the Quarkus proxy ends up changing the class name.

PR #2443, opened 2022-09-06
KOGITO-7890 - Use operation id as work item handler name

cristianonicolai:
Related to https://github.com/quarkiverse/quarkus-openapi-generator/pull/137

### DIFF
--- a/kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/rest/GeneratedRestOperationHandler.java
+++ b/kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/rest/GeneratedRestOperationHandler.java
@@ -23,17 +23,17 @@ import io.serverlessworkflow.api.functions.FunctionDefinition;
 
 public class GeneratedRestOperationHandler implements RestOperationHandler {
 
-    private final String className;
+    private final String workItemHandler;
 
-    public GeneratedRestOperationHandler(String className) {
-        this.className = className;
+    public GeneratedRestOperationHandler(String workItemHandler) {
+        this.workItemHandler = workItemHandler;
     }
 
     @Override
     public <T extends RuleFlowNodeContainerFactory<T, ?>> WorkItemNodeFactory<T> fillWorkItemHandler(WorkItemNodeFactory<T> node,
             Workflow workflow,
             FunctionDefinition actionFunction) {
-        return node.workName(className);
+        return node.workName(workItemHandler);
     }
 
 }
--- a/kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/rest/RestOperationHandlerFactory.java
+++ b/kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/rest/RestOperationHandlerFactory.java
@@ -27,8 +27,8 @@ public class RestOperationHandlerFactory {
 
     public static RestOperationHandler get(ParserContext parserContext, WorkflowOperationId id) {
         KogitoBuildContext context = parserContext.getContext();
-        String className = ServerlessWorkflowUtils.getOpenApiClassName(id.getFileName(), id.getOperation());
-        return context.getGeneratedHandlers().contains(className) ? new GeneratedRestOperationHandler(className) : new DescriptorRestOperationHandler(parserContext, id);
+        String workItemHandlerName = ServerlessWorkflowUtils.getOpenApiWorkItemName(id.getFileName(), id.getOperation());
+        return context.getGeneratedHandlers().contains(workItemHandlerName) ? new GeneratedRestOperationHandler(workItemHandlerName) : new DescriptorRestOperationHandler(parserContext, id);
     }
 
 }
--- a/kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/utils/ServerlessWorkflowUtils.java
+++ b/kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/utils/ServerlessWorkflowUtils.java
@@ -191,6 +191,10 @@ public class ServerlessWorkflowUtils {
         return StringUtils.ucFirst(getValidIdentifier(removeExt(fileName.toLowerCase())) + '_' + methodName);
     }
 
+    public static String getOpenApiWorkItemName(String fileName, String methodName) {
+        return removeExt(fileName) + '_' + methodName;
+    }
+
     public static String removeExt(String fileName) {
         return fileName.replaceFirst(REGEX_NO_EXT, "");
     }
--- a/quarkus/extensions/kogito-quarkus-serverless-workflow-extension/kogito-quarkus-serverless-workflow-deployment/src/main/java/org/kie/kogito/quarkus/serverless/workflow/WorkflowCodeGenUtils.java
+++ b/quarkus/extensions/kogito-quarkus-serverless-workflow-extension/kogito-quarkus-serverless-workflow-deployment/src/main/java/org/kie/kogito/quarkus/serverless/workflow/WorkflowCodeGenUtils.java
@@ -60,13 +60,13 @@ public class WorkflowCodeGenUtils {
                 .map(Optional::get);
     }
 
-    public static String getRefHandler(GeneratedFile generatedFile) {
-        String fileName = generatedFile.path().getFileName().toString();
-        return fileName.substring(0, fileName.lastIndexOf('.'));
-    }
-
     public static GeneratedFile fromCompilationUnit(KogitoBuildContext context, CompilationUnit unit, String className) {
         return new GeneratedFile(GeneratedFileType.SOURCE, Path.of("", context.getPackageName().split("\\.")).resolve(className + ".java"),
+                unit.toString());
+    }
+
+    public static WorkflowHandlerGeneratedFile fromCompilationUnit(String workItemHandlerName, KogitoBuildContext context, CompilationUnit unit, String className) {
+        return new WorkflowHandlerGeneratedFile(workItemHandlerName, GeneratedFileType.SOURCE, Path.of("", context.getPackageName().split("\\.")).resolve(className + ".java"),
                 unit.toString());
     }
 
--- a/quarkus/extensions/kogito-quarkus-serverless-workflow-extension/kogito-quarkus-serverless-workflow-deployment/src/main/java/org/kie/kogito/quarkus/serverless/workflow/deployment/ServerlessWorkflowAssetsProcessor.java
+++ b/quarkus/extensions/kogito-quarkus-serverless-workflow-extension/kogito-quarkus-serverless-workflow-deployment/src/main/java/org/kie/kogito/quarkus/serverless/workflow/deployment/ServerlessWorkflowAssetsProcessor.java
@@ -31,6 +31,7 @@ import org.kie.kogito.process.expr.ExpressionHandler;
 import org.kie.kogito.quarkus.common.deployment.KogitoAddonsPreGeneratedSourcesBuildItem;
 import org.kie.kogito.quarkus.common.deployment.KogitoBuildContextBuildItem;
 import org.kie.kogito.quarkus.serverless.workflow.WorkflowCodeGenUtils;
+import org.kie.kogito.quarkus.serverless.workflow.WorkflowHandlerGeneratedFile;
 import org.kie.kogito.quarkus.serverless.workflow.WorkflowHandlerGenerator;
 import org.kie.kogito.quarkus.serverless.workflow.openapi.WorkflowOpenApiHandlerGenerator;
 import org.kie.kogito.quarkus.serverless.workflow.rpc.WorkflowRPCHandlerGenerator;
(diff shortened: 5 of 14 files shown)


ricardozanini (approved):

kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/rest/GeneratedRestOperationHandler.java
```diff
@@ -23,17 +23,17 @@
 
 public class GeneratedRestOperationHandler implements RestOperationHandler {
 
-    private final String className;
+    private final String workItemHandler;
```
fjtirado, line 26: I do not think this class needs to be changed.

kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/rest/RestOperationHandlerFactory.java
```diff
@@ -27,8 +27,8 @@ private RestOperationHandlerFactory() {
 
     public static RestOperationHandler get(ParserContext parserContext, WorkflowOperationId id) {
         KogitoBuildContext context = parserContext.getContext();
-        String className = ServerlessWorkflowUtils.getOpenApiClassName(id.getFileName(), id.getOperation());
-        return context.getGeneratedHandlers().contains(className) ? new GeneratedRestOperationHandler(className) : new DescriptorRestOperationHandler(parserContext, id);
+        String workItemHandlerName = ServerlessWorkflowUtils.getOpenApiWorkItemName(id.getFileName(), id.getOperation());
```
fjtirado, line 30: same here, this change is not need, I guess the intention is increase readiblity, but I do not really think it increases it

quarkus/extensions/kogito-quarkus-serverless-workflow-extension/kogito-quarkus-serverless-workflow-deployment/src/main/java/org/kie/kogito/quarkus/serverless/workflow/WorkflowCodeGenUtils.java
```diff
@@ -60,11 +60,6 @@ public static Stream<Workflow> getWorkflows(Stream<Path> files) {
                 .map(Optional::get);
     }
 
-    public static String getRefHandler(GeneratedFile generatedFile) {
```
fjtirado, line 63: This is unrelated isnt it?

quarkus/extensions/kogito-quarkus-serverless-workflow-extension/kogito-quarkus-serverless-workflow-deployment/src/main/java/org/kie/kogito/quarkus/serverless/workflow/deployment/ServerlessWorkflowAssetsProcessor.java
```diff
@@ -69,7 +68,6 @@ void addWorkItemHandlers(KogitoBuildContextBuildItem contextBI, CombinedIndexBui
         for (WorkflowHandlerGenerator generator : generators) {
             for (GeneratedFile generated : generator.generateHandlerClasses(context, index)) {
                 generatedFiles.add(generated);
-                context.addGeneratedHandler(WorkflowCodeGenUtils.getRefHandler(generated));
```
fjtirado, line 72: This forces for every future workfllowhandler generatior to invoke context.addGeneratedHandler in the generated code. 
This is not good.

fjtirado (commented):
We need to discuss the implementation. 
I was not aware user has to tweak the operationid name in the flow file, but I think we should fix in a different way.

quarkus/extensions/kogito-quarkus-serverless-workflow-extension/kogito-quarkus-serverless-workflow-deployment/src/main/java/org/kie/kogito/quarkus/serverless/workflow/openapi/WorkflowOpenApiHandlerGenerator.java
```diff
@@ -77,8 +77,7 @@ protected Stream<GeneratedFile> generateHandler(KogitoBuildContext context, Anno
 
     private GeneratedFile generateHandler(KogitoBuildContext context, ClassInfo classInfo, String fileName, MethodInfo m) {
         final String packageName = context.getPackageName();
-        final String methodName = m.annotation(generatedMethod).value().asString();
```
fjtirado, line 80: Not needed, isnt it?
You can still use methodName both for getOpenClassName and getOpenApiWorkItemNane
This also helps to review history, I got really confused because of this change

quarkus/extensions/kogito-quarkus-serverless-workflow-extension/kogito-quarkus-serverless-workflow-deployment/src/main/java/org/kie/kogito/quarkus/serverless/workflow/deployment/ServerlessWorkflowAssetsProcessor.java
fjtirado: I think we need a WorkflowGeneratedFile specialization of GenerateFile to include the handler name (which is not necessaruly the handler class)

quarkus/extensions/kogito-quarkus-serverless-workflow-extension/kogito-quarkus-serverless-workflow-deployment/src/main/java/org/kie/kogito/quarkus/serverless/workflow/rpc/WorkflowRPCHandlerGenerator.java
```diff
@@ -73,6 +73,7 @@ private GeneratedFile generateHandler(KogitoBuildContext context, String service
         constructor.setBody(new BlockStmt().addStatement(new MethodCallExpr(null, "super").addArgument("enumDefault").addArgument("streamTimeout")));
         clazz.addMethod("getName", Keyword.PUBLIC).setType(parseClassOrInterfaceType(String.class.getCanonicalName()))
                 .setBody(new BlockStmt().addStatement(new ReturnStmt(new StringLiteralExpr(className))));
+        context.addGeneratedHandler(className);
```
fjtirado, line 76: This duplication between WorkflowRPCHandlerGenearteor and WorkfflowOpenHandlerGenerator I would like to avoid by adding WorkflowGeneratedFile and forcing the generator to provide a handler name (different or equal to classname)

kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/utils/ServerlessWorkflowUtils.java
```diff
@@ -191,6 +191,10 @@ public static String getOpenApiClassName(String fileName, String methodName) {
         return StringUtils.ucFirst(getValidIdentifier(removeExt(fileName.toLowerCase())) + '_' + methodName);
```
fjtirado, line 191: As you can verify, methodName is not really sanitized here, so once the annotation value is not sanitized, this PR is not really needed, but see my general comment

kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/utils/ServerlessWorkflowUtils.java
```diff
@@ -191,6 +191,10 @@ public static String getOpenApiClassName(String fileName, String methodName) {
...
     }
 
+    public static String getOpenApiWorkItemName(String fileName, String methodName) {
+        return removeExt(fileName) + '_' + methodName;
```
fjtirado, line 195: The only difference between this method and the original one is that file name portion (which is not relevant for operationid resolution, the problem reported in the JIRA) will not be sanitized, so I do not really think we need this new method

quarkus/extensions/kogito-quarkus-serverless-workflow-extension/kogito-quarkus-serverless-workflow-deployment/src/main/java/org/kie/kogito/quarkus/serverless/workflow/openapi/WorkflowOpenApiHandlerGenerator.java
fjtirado: ok, there was  indeed a potential problem, if the opeationid includes a not valid java idenfitier (and since it is not sanitized, the class name will be invalid). 
According to spec, operationid is  Unique string used to identify the operation. The id MUST be unique  among all operations described in the API. The operationId value is case-sensitive.  Tools and libraries MAY use the operationId to uniquely identify an  operation, therefore, it is RECOMMENDED to follow common programming  naming conventions.

So, although not strictly prohibited by the spec, this is very unlikely to happen

kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/rest/GeneratedRestOperationHandler.java
cristianonicolai: @fjtirado if the goal for this variable is to define the work item handler name, I think it is a valid change for make it more readable.

quarkus/extensions/kogito-quarkus-serverless-workflow-extension/kogito-quarkus-serverless-workflow-deployment/src/main/java/org/kie/kogito/quarkus/serverless/workflow/deployment/ServerlessWorkflowAssetsProcessor.java
cristianonicolai: that could be an improvement for sure, but I dont think it is a must have for the PR. I can review that in a follow up.

cristianonicolai:
> ink that with q

@fjtirado Im aware we need to update our tests, but that will only work once we upgrade to a new openapi codegen version with the related change. I'll take care of that once it is released. I can keep this JIRA open to track that.

Regarding the impl, we do need to change things in our end. The operationId name is not a sanitized string by default, we shouldnt expect that, and there is no reason for that. We can decloupe the class name from the WIH name, which can then match the operationId.

Regarding uniqueness, operationId is true that openapi codegen enforces uniquiness in the same OpenAPI spec file, but not across multiple files, hence we need to append the file name in the WIH name to avoid conflict.

kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/utils/ServerlessWorkflowUtils.java
cristianonicolai: Non sanitized string was used as the class name, and so it would generate invalid classes. We need to make a distinction between WIH class name, which,  to be honest, is mostly irrelevant to the WIH name, which should match the operationId, as a plain string, without any manipulation from our side. So I don't see how this would work without these changes.

quarkus/extensions/kogito-quarkus-serverless-workflow-extension/kogito-quarkus-serverless-workflow-deployment/src/main/java/org/kie/kogito/quarkus/serverless/workflow/openapi/WorkflowOpenApiHandlerGenerator.java
cristianonicolai: @fjtirado we dont need to rely on function name as the class name anymore. The link is based on the WIH name now, so using the method name, which is sanitized ( by openapi codegen ) version of operationId, seems fine to me.

kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/rest/GeneratedRestOperationHandler.java
fjtirado: ok, lets use workName as variable name.

quarkus/extensions/kogito-quarkus-serverless-workflow-extension/kogito-quarkus-serverless-workflow-deployment/src/main/java/org/kie/kogito/quarkus/serverless/workflow/deployment/ServerlessWorkflowAssetsProcessor.java
fjtirado: I disagree, the changes in the PR (which are just needed to avoid a corner case, when the unsanitized name is not a valid) introduce an issue in our design that was not there

fjtirado:
@cristianonicolai 
I agree this change is neededd to avoid the problem with unsanitized method name. 
What I propose is to keep this JIRA open till the quarkiverse release is updated, then add the integration test and in the meantime do the change to add the new class with the workitem name.

kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/utils/ServerlessWorkflowUtils.java
fjtirado: what I was trying to point out is that there are two issues: the mapping and the unsanitize, the mapping being the important one. And the mapping will work witouth this PR (when the quarkiverse one is channve), but if the unsanitize operationId is invald string (which will be a rare ocurrence), then this PR is needed. 
I also agree we need the distinciton between workItemName and className and I vote for making it explicit by the introduction of a new class that needs to be generated by every WorkflowWorItemGenerator

cristianonicolai:
jenkins run quarkus-lts

fjtirado (approved):

fjtirado:
@cristianonicolai Thanks for adding the new class, I approved, but we still need the change in the pom and the integration test (not a new test but change the opeationid in one of the existing ones)

PR #2458, opened 2022-09-12
KOGITO-7911 - Allow to override Data Index image version

cristianonicolai:
Unless `placeholder` string is used, DevServices will now use the full string that is defined in the pom.xml, including the version part of it.

### DIFF
--- a/quarkus/extensions/kogito-quarkus-processes-extension/kogito-quarkus-processes-deployment/pom.xml
+++ b/quarkus/extensions/kogito-quarkus-processes-extension/kogito-quarkus-processes-deployment/pom.xml
@@ -13,7 +13,7 @@
   <name>Kogito :: Quarkus Processes Extension :: Deployment</name>
 
   <properties>
-    <data-index-ephemeral.image>quay.io/kiegroup/kogito-data-index-ephemeral</data-index-ephemeral.image>
+    <data-index-ephemeral.image>quay.io/kiegroup/kogito-data-index-ephemeral:{placeholder}</data-index-ephemeral.image>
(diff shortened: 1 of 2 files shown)


ricardozanini (approved):

ricardozanini:
Maybe `latest` instead of `placeholder`? The idea is to have a fixed version?

quarkus/extensions/kogito-quarkus-processes-extension/kogito-quarkus-processes-deployment/pom.xml
```diff
@@ -13,7 +13,7 @@
   <name>Kogito :: Quarkus Processes Extension :: Deployment</name>
 
   <properties>
-    <data-index-ephemeral.image>quay.io/kiegroup/kogito-data-index-ephemeral</data-index-ephemeral.image>
+    <data-index-ephemeral.image>quay.io/kiegroup/kogito-data-index-ephemeral:placeholder</data-index-ephemeral.image>
```
radtriste, line 16: nitpik but could we use something like `{placeholder}` ?

cristianonicolai: @radtriste done
@ricardozanini it has to be an invalid string for the image name, so we can override it. `latest` is a valid one that we might want to use.

cristianonicolai:
jenkins run quarkus-lts

radtriste (approved):

mareknovotny (approved):

PR #2497, opened 2022-09-23
KOGITO-7941 Sonarcloud - (Java) Methods should not be empty (97)

Kbowers-redhat:
https://issues.redhat.com/browse/KOGITO-7941

Goal here is to reduce the number of code smells in Kogito-runtimes.

Addressed in this PR is the code smell, "(Java) Methods should not be empty (97)"
https://sonarcloud.io/project/issues?resolved=false&rules=java%3AS1186&id=org.kie.kogito%3Akogito-runtimes

I have remedied this by adding nested comments as per sonarclouds recommendation, which explain why the method is empty.

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

<b>Jenkins retest this</b>

</details>

### DIFF
--- a/api/kogito-api/src/main/java/org/kie/kogito/process/flexible/Milestone.java
+++ b/api/kogito-api/src/main/java/org/kie/kogito/process/flexible/Milestone.java
@@ -37,6 +37,7 @@ public class Milestone extends ItemDescription {
         private Status status;
 
         public Builder() {
+            //intentionally blank constructor
         }
 
         public Builder withId(String id) {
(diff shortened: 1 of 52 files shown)


mariofusco (commented):
I'm sorry but I absolutely see no value, in adding an explicit comment to an empty method just to make it not empty.

mariofusco:
Just to be clear: doing something without any real value and with the only goal of making a tool happy is not the right way of using that tool.

PR #2539, opened 2022-10-05
KOGITO-8067 - Add common NodeInstance data to dynamic node

cristianonicolai:


tiagodolphine (approved):
Nice, thanks for the fix, so the persistence marshaling was not working for dynamic nodes

ricardozanini (approved):

PR #2624, opened 2022-11-23
[BXMSPROD-1881] dependabot for release and blue branches

Ginxo:


radtriste:
Please include `8.30.x` also.
And the next productized branch should be included in there. Gonna think about something :)

mareknovotny:
we talked about it on our mtg that the new dependabot scanning should be done only on long lived branches, 8.30.x will not be used anymore and we move to new one for future releases so everything what is scanned on main is the vehicle there

mareknovotny (approved):

PR #3112, opened 2023-07-06
[KOGITO-9500] - Remove pmml dependencies from SWF

VaniHaripriya:
Jira : https://issues.redhat.com/browse/KOGITO-9500


fjtirado (commented):
@VaniHaripriya Unfortunaltely I do not think this is the rigth approach. 
why not trying to declare pmml-model as optional in jbpm-flow pom first, so we do not have to exclude it in several places?

VaniHaripriya:
> @VaniHaripriya Unfortunaltely I do not think this is the right approach to achieve the purpose of removing pmml from workflow. why not trying to declare pmml-model as optional in jbpm-flow pom first, so we do not have to exclude it in several places?
@fjtirado  Thank you Francisco, that approach looks more cleaner. Will create a new PR with the change.

PR #3263, opened 2023-10-25
[KOGITO-9914] Setting LD_LIBRARY_PATH to include Jep

fjtirado:
Fixing native image generation issue when using python. The Jep library was not included if LD_LIBRARY_PATH is not set.


jstastny-cz:
I am trying a native build in CI first, will approve based on the outcome.

jstastny-cz (approved):
The testing native build passed, LGTM. Thanks @fjtirado

PR #3266, opened 2023-10-26
Try to reduce the log file size of SWF IT

fjtirado:
This sets the level for container and kafka to warn (this generates a log of trace that does not really tell us anything unless debugging kafka or container issue, which is not the point of the test runt)

There is still an annoying system.out that is printing all kafka events, but I do not think that can be fixed in this repo


ricardozanini (approved):

jstastny-cz (approved):
LGTM.

PR #3291, opened 2023-11-23
Quarkus 3 migration

mariofusco:


ricardozanini:
@mariofusco just adding me here to follow up the work and help closing the gaps.

addons/common/monitoring/core/pom.xml
```diff
@@ -25,7 +25,7 @@
   <parent>
     <artifactId>kogito-addons-monitoring-parent</artifactId>
     <groupId>org.kie.kogito</groupId>
-    <version>2.0.0-SNAPSHOT</version>
+    <version>3.0.0-SNAPSHOT</version>
```
fjtirado, line 28: I know is too late, but I think it would have been easier to review the whole thing if we have leave the kogito version untouched

quarkus/extensions/kogito-quarkus-serverless-workflow-extension/kogito-quarkus-serverless-workflow-deployment/src/main/java/org/kie/kogito/quarkus/serverless/workflow/asyncapi/AsyncAPIInfoConverter.java
```diff
@@ -27,9 +27,9 @@
 import org.kie.kogito.serverless.workflow.asyncapi.AsyncInfo;
 import org.kie.kogito.serverless.workflow.asyncapi.AsyncInfoConverter;
 
-import com.asyncapi.v2.model.AsyncAPI;
-import com.asyncapi.v2.model.channel.ChannelItem;
-import com.asyncapi.v2.model.channel.operation.Operation;
+import com.asyncapi.v2._6_0.model.AsyncAPI;
```
fjtirado, line 30: Good catch!

addons/common/monitoring/core/pom.xml
tkobayas: I'm not sure if we will make it finally `3.0.0-SNAPSHOT` or `2.0.0-SNAPSHOT`. I think we can set it back to `2.0.0-SNAPSHOT` for the ease of review. I have filed https://github.com/mariofusco/kogito-runtimes/pull/9 @mariofusco if you think it's reasonable, please apply.

PR #3323, opened 2023-12-11
[kie-issues#754] Use UTF-8 on important places during Quarkus build

baldimir:
Fixes https://github.com/apache/incubator-kie-issues/issues/754

[Dependant Drools PR](https://github.com/apache/incubator-kie-drools/pull/5620)

Uses UTF-8 on important places during the Quarkus build. I suggest we should do a broader encoding review during next year. There are more places, like e.g. calling String.getBytes() without encoding and similar on many places. Unfortunately this is very hard to test, as the behaviour I was fixing only happens on an environment that doesn't use UTF-8, so there are no new test cases added. If you know about some option how to add such tests, please let me know and I will add them.


gitgabrio (approved):
LGTM!

mariofusco (approved):

PR #3349, opened 2024-01-12
[Fix_@818] Fixing native python test on Java 17

fjtirado:
Fixes https://github.com/apache/incubator-kie-issues/issues/818
LD_LIBRARY_PATH should point either to graalvm_home|java_home/server/lib and some extra classes (mainly jdk ones) are needed in the jni set.


pefernan (approved):
@fjtirado great! This fixes the it test. 

However it's important to know that the produced binary uses the python installed in the system and `LD_LIBRARY_PATH` env should be properly set.

PR #3358, opened 2024-01-17
[Fix_#3353] Adding source to process definition data event

fjtirado:
Fixes


quarkus/extensions/kogito-quarkus-extension-common/kogito-quarkus-common/src/main/java/org/kie/kogito/quarkus/registry/ProcessDefinitionRegistration.java
```diff
@@ -37,9 +40,10 @@ public class ProcessDefinitionRegistration {
     ProcessDefinitionEventRegistry processDefinitionRegistry;
 
     @Inject
-    public ProcessDefinitionRegistration(Application application, KogitoRuntimeConfig runtimeConfig, Instance<Processes> processes) {
+    public ProcessDefinitionRegistration(Application application, KogitoRuntimeConfig runtimeConfig, Instance<Processes> processes, Instance<SourceFilesProvider> sourceFilesProvider) {
```
fjtirado, line 43: We keep this optional, so source addon should still be present for source to be added to the event. 
The alternative is to make this mandatory and move SourceFilesProviderFactory to quarkus common extension. 
I wanted to be conservative since we are close to the release deadline

api/kogito-services/src/main/java/org/kie/kogito/services/registry/ProcessDefinitionEventRegistry.java
```diff
@@ -88,9 +98,16 @@ private Function<Process<?>, ProcessDefinitionDataEvent> mapProcessDefinition(Se
...
-            return definitionDataEvent;
+                    .setMetadata(metadata);
+            sourceFilesProvider.flatMap(provider -> provider.getProcessSourceFile(p.id())).map(s -> {
+                try {
+                    return new String(s.readContents());
+                } catch (IOException e) {
+                    LOGGER.warn("Error reading source for process {}", p.id(), e);
+                    return null;
+                }
+            }).ifPresentOrElse(s -> builder.setSource(s), () -> LOGGER.warn("Not source found for process id {}", p.id()));
```
tiagodolphine, line 109: You can do like other fields using the builder, like `.setMetadata(metadata).source(getSource(...))...` just to extract this into a method, similar `.setNodes(getNodesDefinitions(p))` just to keep it cleaner.

tiagodolphine (commented):
LGTM, a good improvement would be to add an IT for it, maybe on https://github.com/apache/incubator-kie-kogito-apps/blob/main/apps-integration-tests/integration-tests-data-index-service/integration-tests-data-index-service-common/src/test/java/org/kie/kogito/index/AbstractProcessDataIndexIT.java#L173

api/kogito-services/src/main/java/org/kie/kogito/services/registry/ProcessDefinitionEventRegistry.java
fjtirado: if source (in case the addon is not in the set of dependencies) is not there, getSource() will return null, and we will be setting source to null in the builder when it is not really needed. 
However, I agree that this can be done more readable, please check the new version

fjtirado:
> LGTM, a good improvement would be to add an IT for it, maybe on https://github.com/apache/incubator-kie-kogito-apps/blob/main/apps-integration-tests/integration-tests-data-index-service/integration-tests-data-index-service-common/src/test/java/org/kie/kogito/index/AbstractProcessDataIndexIT.java#L173

@tiagodolphine Thanks for the hint. Added in the linked PR

ricardozanini (approved):

tiagodolphine (approved):

PR #817, opened 2020-10-09
KOGITO-3601 DMN integrate programmatically generated OAS definitions

tarilabs:
This is a DRAFT PR for now, to better clarify one of the many usages of https://issues.redhat.com/browse/DROOLS-5714

**referenced Pull Requests**: part of an ensemble of PRs:

* https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1493
* https://github.com/kiegroup/drools/pull/3155
* (DRAFT) this PR

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* <b>a pull request</b> please add comment: <b>Jenkins retest this</b>
 
* <b>a full downstream build</b> please add comment: <b>Jenkins run fdb</b>
  
* <b>a compile downstream build</b> please  add comment: <b>Jenkins run cdb</b>

* <b>a full production downstream build</b> please add comment: <b>Jenkins execute product fdb</b>

* <b>an upstream build</b> please add comment: <b>Jenkins run upstream</b>
</details>

PR #3369, opened 2024-01-23
#817 - removing oracle database dependency

porcelli:
This PR is related to https://github.com/apache/incubator-kie-issues/issues/817

Related PRs:
  - https://github.com/apache/incubator-kie-kogito-apps/pull/1968


pefernan (approved):

mariofusco (approved):

mariofusco:
I approved this PR not because I'm a fan of removing potentially useful features in this way, but because I didn't see any alternative if we want to move forward with the first Apache release.

Thinking twice, if we understand correctly, what it is not compliant with Apache policies is the use of the Oracle driver, that however we only use for tests (and in all honesty is highly questionable if a dependency with test scope should be considered among our actual dependencies and block the release). If that's the case the other alternative would be to keep the addon as it is and just remove the test. It's true that in this way we will end up with an untested addon, but I would argue that keeping this feature untested would be better than totally removing it. And on the longer term we could move this test, and maybe other with similar problems, into another project (maybe back in `kiegroup`?) so it won't be under the Apache control, but still we will be notified if a future change will break the addon.

@porcelli please let me know if the alternative that I'm suggesting makes sense.

fjtirado:
Folllowing Marios comment, which I fully concur with, iif we interpret the legal restriction to also bind test dependecies, which is highly debatable, this PR (and the related one in Apps)  should just have disable the oracle test (which I think does not depend on the driver for compilation) and removal of the test dependency from the pom. Since we are already disabling flaky test, that should not cause any issue.

fjtirado:
Already implemented

PR #3381, opened 2024-01-29
Removing oracle driver and associated tests

fjtirado:
Removing oracle jdbc driver from kogito_test_utils because of legal issues
Merge with https://github.com/apache/incubator-kie-kogito-apps/pull/1971


porcelli (changes requested):
I think there are still some left over oracle references that I don't think it makes sense to carry over like DB-scripts and the Enum with explict reference to it.

ricardozanini:
@porcelli but these references are not linked to the actual driver jar. We can't even mention _oracle_?

nmirasch (approved):

fjtirado:
@porcelli Which else do you propose to remove?. Can you please point the place for the scripts? I was not aware of its existance, I just change the minimal to get rid of the driver (there are not specific test for Oracle in runtimes)

fjtirado:
@porcelli About the enum, it might be useful to keep it for people that actually use Oracle with out stuff. Same rationale probably apply to the scripts, but I want to take a look to them first.

mariofusco:
@porcelli I agree that we shouldn't left anything behind, but I would approve and merge this PR regardless, at least to resolve this _emergency_. We could address any other leftover with another follow up PR, but at least this wouldn't be blocker for the release anymore.

porcelli:
@mariofusco if this was the only remaining blocker for release; I'd agree with you. But since we still have time due other issues being worked.. I think we could have an adicional cycle to get this properly cleaned up.

fjtirado:
@porcelli I do not understand why we need to remove a script that works, but since I do not want to discuss any longer, I will remove it

addons/common/persistence/jdbc/src/main/java/org/kie/kogito/persistence/jdbc/DatabaseType.java
```diff
@@ -1,67 +0,0 @@
...
-import java.sql.DatabaseMetaData;
-import java.sql.SQLException;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-public enum DatabaseType {
-    ANSI("ansi", "process_instances"),
-    ORACLE("Oracle", "PROCESS_INSTANCES"),
-    POSTGRES("PostgreSQL", "process_instances");
```
porcelli, line 31: why instead of completely remove the enum, use something more generic... better aligned with the intention. Something like:

- JSONB_DATABASE
- BLOB_DATABASE
- ANSI_DATABASE

porcelli (approved):

porcelli:
thank you @fjtirado for the changes!

mariofusco (approved):

PR #3393, opened 2024-02-06
[kie-issues#839] Reproducible build

baldimir:
Fixes https://github.com/apache/incubator-kie-issues/issues/839

Changes to make kogito-runtimes build reproducible. 
- Upgrades jandex-maven-plugin, so it generates reproducible index files. 
- Applies similar fix as in Drools and OptaPlanner to remove timestamp comments from generated text files. 
    - drools-util is only added as the exec-maven-plugin dependency, not as a normal dependency. 
    - First argument of the RemoveCommentsMain class is to ignore non-existing files. This enables to have one common configuration and not individual configurations in modules. 

If you want to have a bit of a laugh, check the comment names :) (please squash before merging, if this gets approved)


mariofusco (approved):

yesamer (approved):
Thank you @baldimir, well done!

PR #3400, opened 2024-02-12
[KIE-922] remove kogito-rules quarkus extension

mariofusco:
Fixes https://github.com/apache/incubator-kie-issues/issues/922


lucamolteni (approved):

mariofusco:
I was wrong and there are still features in the kogito-rules extension that are not covered by the drools one, in particular Grafana Dashboard integration, see https://github.com/apache/incubator-kie-kogito-runtimes/blob/main/kogito-codegen-modules/kogito-codegen-rules/src/main/java/org/kie/kogito/codegen/rules/RuleUnitDashboardCodegen.java

I believe that stuff actually belongs to kogito and I don't want to get rid of it also because we have a few module in kogito-examples and apps relying on it, and then causing these failures https://github.com/apache/incubator-kie-kogito-examples/pull/1871#issuecomment-1939449702

For this reason I'm closing this pull request and all related works.

PR #3407, opened 2024-02-16
[Fix #3406] Avoid publishing internal variables

### LINKED ISSUE #3406: Internal variables changes should not be published externally
### Description

Variable marked as internal by metadata should not be published to external systems 

### Implementation ideas

_No response_

fjtirado:
Fixes https://github.com/apache/incubator-kie-kogito-runtimes/issues/3406

### DIFF
--- a/api/kogito-events-core/src/main/java/org/kie/kogito/event/impl/ProcessInstanceEventBatch.java
+++ b/api/kogito-events-core/src/main/java/org/kie/kogito/event/impl/ProcessInstanceEventBatch.java
@@ -120,7 +120,9 @@ public class ProcessInstanceEventBatch implements EventBatch {
     }
 
     private void handleProcessVariableEvent(ProcessVariableChangedEvent event) {
-
+        if (event.getTags().contains("internal")) {
+            return;
+        }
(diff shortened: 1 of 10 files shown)


api/kogito-events-core/src/main/java/org/kie/kogito/event/impl/ProcessInstanceEventBatch.java
```diff
@@ -120,7 +120,9 @@ private void addDataEvent(ProcessEvent event) {
     }
 
     private void handleProcessVariableEvent(ProcessVariableChangedEvent event) {
-
+        if (event.getTags().contains("internal")) {
```
nmirasch, line 123: @fjtirado it wont be better to use the existing constant in bpmn variables?

wmedvede (approved):
agree with @nmirasch

api/kogito-events-core/src/main/java/org/kie/kogito/event/impl/ProcessInstanceEventBatch.java
fjtirado: good catch! done

nmirasch (approved):
thanks!

PR #3428, opened 2024-03-05
[Fix apache/incubator-kie-issues#987] New onError event

fjtirado:
Fixes https://github.com/apache/incubator-kie-issues/issues/987
Ensure that starting event is the first one and complete process events is the last one
Unlink error event from state event so, if there is an error is published always and not only if it happens in the sync part of the process. 
Merge with https://github.com/apache/incubator-kie-drools/pull/5766 and https://github.com/apache/incubator-kie-kogito-apps/pull/2011


nmirasch (approved):

wmedvede (approved):

elguardian:
@fjtirado @wmedvede @nmirasch  try to follow naming convetions when create new events please.

fjtirado:
@elguardian Good to know for the future. Which should be the naming convention in this case?

elguardian:
@fjtirado ProcessErrorEvent as you are in Kie-api and belongs to the process error.

PR #3455, opened 2024-03-22
[Fix #3453] Setting overflow strategy to unbounded for process events

### LINKED ISSUE #3453: Allow adjusting backpressure for Reactive Event Publisher
### Description

In a large workflow consisting on many nodes, when the UnitOfWork call EventPublisher, the Event batch might be large enough to cause exception` SRMSG00034: Insufficient downstream requests to emit item` and events being discarded. 
To fix this we should put a larger buffer by default and make this policy configurable. 

### Implementation ideas

OnOverflow annotation only admits constant values, so probably the best approach is to change strategy to unbounded for process instances events

fjtirado:
Fixes https://github.com/apache/incubator-kie-kogito-runtimes/issues/3453
Unfortunately OnOverflow annotation requires contant value, so this cannot be make configurable without code generation. Therefore I just changed overflow strategy to unbounded.


ricardozanini (approved):

gmunozfe (approved):
Looks good to me, nicely done @fjtirado !

PR #3468, opened 2024-04-09
[Fix_#3465] Adds DMN support to SonataFlow

fjtirado:
Fixes https://github.com/apache/incubator-kie-kogito-runtimes/issues/3465
A new DMN custom type (called "dmn") has been added to swf parser. 
The mapping is either automatically performed from the workflow model (if the name matches), so there is not need to provide args. If the workflow model names does not match, the workflow writer should provide the mapping as argument of the function call. 

This is the workflow definition used in the[ unit test](https://github.com/apache/incubator-kie-kogito-runtimes/pull/3468/files#diff-bd9f4f159022434c89f13107535be3ba2f2c3245380656262a11b07bb7503535R31-R36) (which just calls the traffiction violation decision from the workflow)
```
{
  "id" : "traffic-violation",
  "name" : "traffic-violation",
  "version" : "1_0",
  "start" : "traffic-violation",
  "functions" : [ {
    "name" : "DMNTest",
    "operation" : "dmn",
    "type" : "custom",
    "metadata" : {
      "file" : "Traffic Violation.dmn",
      "namespace" : "https://github.com/kiegroup/drools/kie-dmn/_A4BCA8B8-CF08-433F-93B2-A2598F19ECFF",
      "model" : "Traffic Violation"
    }
  } ],
  "states" : [ {
    "actions" : [ {
      "functionRef" : "DMNTest"
    } ],
    "name" : "traffic-violation",
    "type" : "operation",
    "end" : true
  } ]
}
```
See example https://github.com/apache/incubator-kie-kogito-examples/pull/1906


baldimir:
Hi, I opened a thread on the dev mailing list about this, because I didn't saw one there. Let's see if we have an agreement about this there.

kogito-serverless-workflow/kogito-serverless-workflow-dmn-parser/src/test/resources/Traffic Violation.dmn
```diff
@@ -0,0 +1,230 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<dmn:definitions xmlns:dmn="http://www.omg.org/spec/DMN/20180521/MODEL/" xmlns="https://github.com/kiegroup/drools/kie-dmn/_A4BCA8B8-CF08-433F-93B2-A2598F19ECFF" xmlns:di="http://www.omg.org/spec/DMN/20180521/DI/" xmlns:kie="http://www.drools.org/kie/dmn/1.2" xmlns:feel="http://www.omg.org/spec/DMN/20180521/FEEL/" xmlns:dmndi="http://www.omg.org/spec/DMN/20180521/DMNDI/" xmlns:dc="http://www.omg.org/spec/DMN/20180521/DC/" id="_1C792953-80DB-4B32-99EB-25FBE32BAF9E" name="Traffic Violation" expressionLanguage="http://www.omg.org/spec/DMN/20180521/FEEL/" typeLanguage="http://www.omg.org/spec/DMN/20180521/FEEL/" namespace="https://github.com/kiegroup/drools/kie-dmn/_A4BCA8B8-CF08-433F-93B2-A2598F19ECFF">
```
gitgabrio, line 2: HI @fjtirado 
We created a"testing" module containing all the dmn files used for testing purposes [here](https://github.com/apache/incubator-kie-drools/tree/main/kie-dmn/kie-dmn-test-resources)
We did that to be sure that the files used around are perfectly compatible (and validated) by our engine. And also to stop a bit this uncontrolled proliferation of dmns 😄 
(Side note: Traffic Violation.dmn is probably the most used models all over)

kogito-bom/pom.xml
```diff
@@ -1941,6 +1941,28 @@
...
+        <artifactId>kogito-serverless-workflow-dmn-parser</artifactId>
+        <version>${project.version}</version>
+        <classifier>sources</classifier>
+      </dependency>
+      <dependency>
+        <groupId>org.kie.kogito</groupId>
+        <artifactId>kogito-serverless-workflow-dmn</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
```
gitgabrio, line 1960: @fjtirado 
OUt of curiosity: why the dependency on `sources`

gitgabrio (changes requested):
Hi @fjtirado 
There is only one bit I would prefer to be changed (the usage of dmn from shared model)
Thanks!

kogito-bom/pom.xml
fjtirado: This is done for all dependencies on the Kogito bom (which includes all modules defined in kogito)
I guess it make easier to download sources on IDES automatically.

gitgabrio (approved):

nmirasch (approved):

PR #3477, opened 2024-04-19
Quarkus LTS Upgrade to 3.8.4

porcelli:
Related https://github.com/apache/incubator-kie-issues/issues/958

Ensemble:
https://github.com/apache/incubator-kie-drools/pull/5853
https://github.com/apache/incubator-kie-optaplanner/pull/3076/files
https://github.com/apache/incubator-kie-kogito-runtimes/pull/3477
https://github.com/apache/incubator-kie-kogito-apps/pull/2039
https://github.com/apache/incubator-kie-kogito-examples/pull/1908
https://github.com/apache/incubator-kie-kogito-images/pull/1760


mariofusco (approved):

quarkus/addons/messaging/common/src/main/java/org/kie/kogito/addon/quarkus/messaging/common/QuarkusTopicDiscovery.java
```diff
@@ -68,7 +68,15 @@ private String extractChannelName(String property, String prefix) {
         if (channelName.contains(".")) {
             channelName = channelName.substring(0, channelName.indexOf("."));
         }
-        final Optional<String> topicName = ConfigProvider.getConfig().getOptionalValue(prefix + channelName + TOPIC_SUFFIX, String.class);
+        final Optional<String> topicName = getOptionalValue(prefix + channelName + TOPIC_SUFFIX);
         return topicName.orElse(channelName);
     }
+
+    Iterable<String> getPropertyNames() {
```
fjtirado, line 75: I think this method and getOptionalValue should be protected

quarkus/addons/messaging/common/src/test/java/org/kie/kogito/addon/quarkus/messaging/common/QuarkusTopicDiscoveryTest.java
```diff
@@ -37,44 +33,66 @@
...
+            final Map<String, String> properties = new HashMap<>() {
+                {
+                    put("mp.messaging.outgoing.processedtravellers.connector", "quarkus-http");
+                    put("mp.messaging.outgoing.processedtravellers.url", "http://localhost:8080/");
+                    put("mp.messaging.incoming.kogito_incoming_stream.connector", "smallrye-kafka");
+                    put("mp.messaging.incoming.kogito_incoming_stream.topic", "mycooltopic");
+                }
+            };
+
+            Iterable<String> getPropertyNames() {
```
fjtirado, line 51: Missing override?

quarkus/addons/messaging/common/src/test/java/org/kie/kogito/addon/quarkus/messaging/common/QuarkusTopicDiscoveryTest.java
```diff
@@ -37,44 +33,66 @@
...
+                    put("mp.messaging.outgoing.processedtravellers.connector", "quarkus-http");
+                    put("mp.messaging.outgoing.processedtravellers.url", "http://localhost:8080/");
+                    put("mp.messaging.outgoing.processedtravellers.topic", "mycooltopic");
+                    put("mp.messaging.incoming.kogito_incoming_stream.connector", "smallrye-kafka");
+                    put("mp.messaging.incoming.kogito_incoming_stream.topic", "mycooltopic");
+                }
+            };
+
+            Iterable<String> getPropertyNames() {
+                return properties.keySet();
```
fjtirado, line 82: same here, actually, to avoid code duplication (which I agree in test is mostly irrelevant) this probably deserve an inner class

fjtirado (approved):
Just really minor commens related with QuarkusTopicsDiscovered
Great job!

kogito-test-utils/pom.xml
```diff
@@ -54,8 +54,13 @@
       <version>${version.org.testcontainers}</version>
       <scope>compile</scope>
     </dependency>
-    
-    
+
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter</artifactId>
+      <scope>compile</scope>
```
fjtirado, line 61: hmmm, shoul not be test scope?

fjtirado: Actually, why JUnit has to be added to the dependency set? Or the other way around, how this was working without this dependency?

quarkus/addons/messaging/common/src/main/java/org/kie/kogito/addon/quarkus/messaging/common/QuarkusTopicDiscovery.java
porcelli: why protected and not default?

fjtirado: Sorry, I forgot to mention why ;). 
It is just styling. I prefer to use package for methods that might be called only within the package and protected for methods that might be inherited. 
It is true that package is more restrictive (a protected can be inherited outside the original package) and you are only overriding from the unit test within the same package, theferore both are fine.

porcelli: as this is in the personal preference realm, I'll leave as is as my personal preference is default scope as it's more restrictive and supposed to be used only by unit test to avoid using environment config

quarkus/addons/messaging/common/src/test/java/org/kie/kogito/addon/quarkus/messaging/common/QuarkusTopicDiscoveryTest.java
```diff
@@ -83,21 +75,49 @@ void verifyTopicsWithNoPropertiesSet() {
...
+            }
+
+            @Override
+            Optional<String> getOptionalValue(String key) {
+                return Optional.ofNullable(properties.get(key));
+            }
+        };
+    }
+
```
porcelli, line 122: @fjtirado improved test code

kogito-test-utils/pom.xml
porcelli: I can't remember what I had to add this, I'll try remove and see how it goes.

yesamer:
I guess we could manage the maven update to 3.9.6 in this PR as well, @porcelli ?
https://github.com/apache/incubator-kie-kogito-runtimes/pull/3479
WDYT? 
(kie-apps and kogito-examples should be updated to maven 3.9.6 too)

baldimir (approved):

PR #3482, opened 2024-04-23
[incubator-kie-issues-1108] Add capability to create node instances (node states) from the nodes

elguardian:
Closes: https://github.com/apache/incubator-kie-issues/issues/1108

### DIFF
--- a/jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/factory/DefaultNodeInstanceFactoryProvider.java
+++ b/jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/factory/DefaultNodeInstanceFactoryProvider.java
@@ -1,0 +1,104 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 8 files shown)


martinweiler (approved):

jbpm/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/factory/DefaultNodeInstanceFactoryProvider.java
```diff
@@ -0,0 +1,104 @@
...
+import org.jbpm.workflow.instance.node.ThrowLinkNodeInstance;
+import org.jbpm.workflow.instance.node.TimerNodeInstance;
+import org.jbpm.workflow.instance.node.WorkItemNodeInstance;
+
+public class DefaultNodeInstanceFactoryProvider implements NodeInstanceFactoryProvider {
+
+    @Override
+    public List<NodeInstanceFactory> provide() {
+        List<NodeInstanceFactory> nodeInstanceFactoryList = new ArrayList<>();
+        nodeInstanceFactoryList.add(new DefaultNodeInstanceFactory(RuleSetNode.class, RuleSetNodeInstance::new));
```
fjtirado, line 78: @elguardian As we discussed as part of the proposal, I do not think RuleSetNodeInstance should be part of the default factory provider. 
That way DRL/DMN wont be a dependency of jbpm-flow
Are we planning to do this split with a follow up Issue/PR?

elguardian: the scope of the PR is just adding the mechanism. @fjtirado

fjtirado (approved):

PR #3485, opened 2024-04-24
[Fix #3483] Adding state duration summary

### LINKED ISSUE #3483: Add node duration metric to monitoring addon
### Description

It would be interesting to track the duration of certain nodes, as we are tracking process instances duration and all work items duration

### Implementation ideas

In BPMN and SWF parsers, add a metadata key called "Metric" to those nodes that it will be interesting to track duration for.
In monitoring addon, add a new Distributionsummary object called ""kogito_node_instance_duration_seconds" which will record duration for those nodes flagged with the metadata key.

fjtirado:
Fixes https://github.com/apache/incubator-kie-kogito-runtimes/issues/3483
Add "kogito_node_instance_duration_milliseconds" distribution summary. 
Different to the other two summaries, times are collected in milliseconds rather than seconds


elguardian:
@fjtirado IMO we should record also the number this node has been executed. The reason is because you can calculate the mean (duration) / hits = average so we have an idea of the duration of the node....
min max would be interesting as well.

fjtirado:
> @fjtirado IMO we should record also the number this node has been executed. The reason is because you can calculate the mean (duration) / hits = average so we have an idea of the duration of the node.... min max would be interesting as well.

@elguardian I think this is automatically done by the Distribution summary micrometer.

elguardian (approved):

PR #1131, opened 2021-03-11
Remove duplicated version.archetype.plugin property

jomarko:
The same file *kogito-build-parent/pom.xml* contained the same property twice.

PR #3505, opened 2024-05-10
Process test migration from v7 legacy runtime to code generation (Issue #1131)

Abhitocode:
Migrated some BPMN processes which are being compiled using legacy v7 to code generation approach.

The tests can be identified by referring to Activitytest.java:
<https://github.com/apache/incubator-kie-kogito-runtimes/blob/main/jbpm/jbpm-tests/src/test/java/org/jbpm/bpmn2/ActivityTest.java>

**Issue**: <https://github.com/apache/incubator-kie-issues/issues/1131>

### DIFF
--- a/jbpm/jbpm-tests/src/test/java/org/jbpm/bpmn2/ActivityTest.java
+++ b/jbpm/jbpm-tests/src/test/java/org/jbpm/bpmn2/ActivityTest.java
@@ -26,6 +26,7 @@ import java.util.List;
 import java.util.Map;
 
 import org.drools.compiler.rule.builder.PackageBuildContext;
+import org.jbpm.bpmn2.flow.*;
 import org.jbpm.bpmn2.handler.ReceiveTaskHandler;
 import org.jbpm.bpmn2.handler.SendTaskHandler;
 import org.jbpm.bpmn2.handler.ServiceTaskHandler;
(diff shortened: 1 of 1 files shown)


elguardian:
hi @Abhitocode few notes in here: your branch should be called as the issue in this case 

"incubator-kie-issues-1131"

the subjects needs to change to 

[incubator-kie-issues-1131] Process test migration from v7 legacy runtime to code generation

the rest of the comments inline

jbpm/jbpm-tests/src/test/java/org/jbpm/bpmn2/ActivityTest.java
```diff
@@ -936,6 +950,7 @@ public void testBusinessRuleTaskWithContionalEvent() throws Exception {
         assertThat(list).hasSize(1);
     }
 
+    // The model does not have set myVar
```
elguardian, line 953: avoid to add noise in here. If you want to annotate something like this won't be migrated but was want to focus attention to it. just add 
@Category with something like V7_ENGINE

Abhitocode: Okay, will make the changes.

PR #3548, opened 2024-06-04
[Fix_#3546] NodeInstanceFactoryProvided not included in native build

fjtirado:
Fix https://github.com/apache/incubator-kie-kogito-runtimes/issues/3546. This is an issue in native build consequence of the refactor done by https://github.com/apache/incubator-kie-kogito-runtimes/pull/3482
I think including the service loader classes should be enough for native compiler to realize it should include all node classes, but @domhanak please let us know if the issue is fixed by this change

### DIFF
--- a/quarkus/extensions/kogito-quarkus-extension-common/kogito-quarkus-common-deployment/src/main/java/org/kie/kogito/quarkus/common/deployment/KogitoAssetsProcessor.java
+++ b/quarkus/extensions/kogito-quarkus-extension-common/kogito-quarkus-common-deployment/src/main/java/org/kie/kogito/quarkus/common/deployment/KogitoAssetsProcessor.java
@@ -42,6 +42,8 @@ import org.jboss.jandex.DotName;
 import org.jboss.jandex.IndexView;
 import org.jboss.jandex.Indexer;
 import org.jboss.logging.Logger;
+import org.jbpm.ruleflow.core.factory.provider.NodeFactoryProvider;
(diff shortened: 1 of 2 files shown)


fjtirado:
@domhanak Stil not working. It seems more stuff needs to be added. Working on that

fjtirado:
Working now

ricardozanini (approved):

elguardian:
@fjtirado this was fixed in documentation as for service loader it is required 

[uarkus.native.auto-service-loader-registration](https://es.quarkus.io/guides/building-native-image#quarkus-native-pkg-native-config_quarkus-native-auto-service-loader-registration)

This fixes this classloaders and futures ones. propably it is better solution to add this into an application.properties

fjtirado:
> @fjtirado this was fixed in documentation as for service loader it is required
> 
> [uarkus.native.auto-service-loader-registration](https://es.quarkus.io/guides/building-native-image#quarkus-native-pkg-native-config_quarkus-native-auto-service-loader-registration)
> 
> This fixes this classloaders and futures ones. propably it is better solution to add this into an application.properties

Ill try and let you know (Im not sure if setting that property in the quarkus extension automatically scan the META-INF of all dependencies, in which case is indeed a better solution or it just scans the extension one, in which case it wont work)

fjtirado:
Updated to use quarkus property rather than explicitly naming the service loader involved, thanks @elguardian

elguardian (approved):

PR #3577, opened 2024-07-10
[incubator-kie-issues-1259] RestWorkItemHandler logger set to debug

Abhitocode:
The logging output from RestWorkItemHandler generates too many logs which is not helpful. Changed  the log level from info to debug.

Closes: [https://github.com/apache/incubator-kie-kogito-runtimes/pull/3559](url)


kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
```diff
@@ -152,27 +152,27 @@ public void executeWorkItem(KogitoWorkItem workItem, KogitoWorkItemManager manag
                 path += "?" + query;
             }
         } catch (MalformedURLException ex) {
-            logger.info("Parameter endpoint {} is not valid uri {}", endPoint, ex.getMessage());
+            logger.debug("Parameter endpoint {} is not valid uri {}", endPoint, ex.getMessage());
```
fjtirado, line 155: This one should not be a frequent situation, so having it as info seems proper

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
```diff
@@ -229,7 +229,7 @@ private Class<?> getType(KogitoWorkItem workItem, String varName) {
                 return variable.getType().getObjectClass();
             }
         }
-        logger.info("Cannot find definition for variable {}", varName);
+        logger.debug("Cannot find definition for variable {}", varName);
```
fjtirado, line 232: This one also seems a valid info situation.

fjtirado:
Link [https://github.com/apache/incubator-kie-kogito-runtimes/pull/3559](https://github.com/apache/incubator-kie-kogito-runtimes/pull/url) does not exist

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
Abhitocode: Okay will compare and update.

Abhitocode:
Issue link updated.

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
fjtirado: After checking the issue description, this one is indeed debug (I think is highly debatable if all logs should be info or debug for RestWorkItemHandler, but accepting they should be debug, this particular one lies in the same category that the host, port and path)

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
fjtirado: After checking the issue, this particular one I think should remain info, because thats a non frequent situation the user should be aware of, event when not debugging

kogito-workitems/kogito-rest-workitem/src/main/java/org/kogito/workitem/rest/RestWorkItemHandler.java
Abhitocode: I understand, will update based on further inputs.

martinweiler (approved):
LGTM @Abhitocode. 

I think it is justified to change the log level to DEBUG in the `executeWorkItem` method, as the messages are logged for the fallback scenario. So if you define a partial URL and the handler works as expected, you don't want these messages to be logged at INFO on each invocation.

@fjtirado thanks for your feedback on this PR!

pefernan (approved):
+1, going to try to have a green before merging :D

PR #3588, opened 2024-07-23
[Fix #3573] configKey function metadata

### LINKED ISSUE #3573: Dynamic OpenAPI configKey
### Description

OpenAPI properties depends on `configKey`
For example, the property for the URL is of the form 
`quarkus.rest-client.<configKey>.url`
`configKey` is currently calculated at build time (there are different alternatives based on the name and content of the openapi spec file being parsed, by default `configKey` is the sanitized file name) 
The proposal is to allow dynamic calculation of `configKey` at runtime. This will allow appending extra information to the configKey calculated at build time. 
For example, you can have different url properties depending on a cluster name
```
quarkus.rest-client.<buildTimeConfigKey>.cluster1.url=<cluster1Address>
quarkus.rest-client.<buildTimeConfigKey>.cluster2.url=<cluster2Address>
```
In order SonataFlow to know which information should be used to calculate configKey, users should add a new property, called `sonataflow.rest-client.dynamic-config-key`, which value must be a jq expression which evaluation will return the value of the dynamic `configKey`

For the previous example, assuming the workflow input has a `clusterName ` which value is `cluster1` or `cluster2`, users must add this line to application.properties.

`sonataflow.rest-client.dynamic-config-key=.clusterName`

### Implementation ideas

_No response_

fjtirado:
Fix https://github.com/apache/incubator-kie-kogito-runtimes/issues/3573

This add configKey property to function definition metadata. 
configKey value is an expression that is evaluated againts the workflow model when the rest call is performed for the function that defines it. 
If expression returns null, an IllegalArgumentException is launched, failing the workflow execution. 
Othewise, the string result of the expression is added to the configKey calculated at build time, allowing usage of different endpoints for the same workflow definition


wmedvede (approved):

domhanak (commented):
:+1:

PR #3594, opened 2024-07-30
Removing unneeded dependency

fjtirado:
The dependency added as workaround is not longer needed

### DIFF
--- a/springboot/integration-tests/pom.xml
+++ b/springboot/integration-tests/pom.xml
@@ -76,10 +76,6 @@
     </dependency>
     <dependency>
       <groupId>org.kie.kogito</groupId>
-      <artifactId>kogito-serverless-workflow-openapi-common</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.kie.kogito</groupId>
       <artifactId>kogito-maven-plugin</artifactId>
     </dependency>
     <dependency>


baldimir:
Hi @fjtirado, will this break the reproducibility again please?

fjtirado:
@baldimir Im checking. Let you know when I have news

fjtirado:
@baldimir Once the new version of openapi-common has been finally uploaded to maven, it seems to work fine, so I think we can go ahead and remove the dependency

fjtirado:
@gitgabrio 
Yes, Im referring to kogito-serverless-workflow-openapi-common. Once this file has been uploaded to the maven repo, the build works, so it can be removed from the springboot it. 
As I tried to explain in the previous PR, reproducible build compare goal is trying to compare all artifacts, not just sprinboot it dependencies. Springboot IT dependnecies (build in that load) are available for the comparison, so compare does no try to download them. But not springboot IT dependencies (for example, kogito-serverless-workflow-openapi-common) will have to be retrieved from maven central (because they are not installed in the local repo used by the sprinboot it, once you make it a dependency, it is installe and pr passed)
I agree we have a huge problem with this and thats why I proposed to disable compare goal for springboot it, since any change in a non sprinboot it dependency jar will provoke a red flag in the PR. 
Now that the jar is now in central, we can removed the dependency, which makes no sense for SpringBoot IT

fjtirado:
Not longer needed, see https://github.com/apache/incubator-kie-kogito-runtimes/pull/3604

PR #3635, opened 2024-08-23
[incubator-kie-issues-1131] test migration from V7 to code generation-23

Abhitocode:
Issue's with **testEventBasedSplit3** and **testEventBaedSplit5** tests within IntermediateEventTest.java

The test can be identified by referring to IntermediateEventTest.java:
https://github.com/apache/incubator-kie-kogito-runtimes/blob/main/jbpm/jbpm-tests/src/test/java/org/jbpm/bpmn2/IntermediateEventTest.java

Closes https://github.com/apache/incubator-kie-issues/issues/1131.

**Issue 1:** 

`testEventBasedSplit3`

Error: java.lang.NullPointerException: Cannot invoke "org.drools.core.common.InternalAgenda.isRuleActiveInRuleFlowGroup(String, String, String)" because the return value of "org.drools.core.common.InternalKnowledgeRuntime.getAgenda()" is null
at org.jbpm.bpmn2.IntermediateEventTest.testEventBasedSplit3(IntermediateEventTest.java:479)

org.opentest4j.AssertionFailedError: 
expected: 1
 but was: 5
 
In the v7, we use:

`kruntime.getKieSession().insert(jack);`	

But there appears to be an issue because the jack variable  doesn't exist in our data model. Could you please guide on how we should handle this.

**Issue 2:**

`testEventBasedSplit5`

 The receive task handler needs a kruntime object. Need guidance on how to create this handler in v9 as we don't have the kruntime object.
 
 ` ReceiveTaskHandler receiveTaskHandler = new ReceiveTaskHandler(null);
        ProcessTestHelper.registerHandler(app, "Receive Task", receiveTaskHandler);`

### DIFF
--- a/jbpm/jbpm-tests/src/test/java/org/jbpm/bpmn2/IntermediateEventTest.java
+++ b/jbpm/jbpm-tests/src/test/java/org/jbpm/bpmn2/IntermediateEventTest.java
@@ -41,6 +41,10 @@ import org.jbpm.bpmn2.event.BoundaryTimerCycleISOVariableModel;
 import org.jbpm.bpmn2.event.BoundaryTimerCycleISOVariableProcess;
 import org.jbpm.bpmn2.handler.ReceiveTaskHandler;
 import org.jbpm.bpmn2.handler.SendTaskHandler;
+import org.jbpm.bpmn2.intermediate.EventBasedSplit3Model;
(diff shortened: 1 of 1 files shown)


elguardian:
we are not supporting this sort of scenario with drools. set this as standby. I need to fix this in the engine

Abhitocode:
Okay

PR #3683, opened 2024-09-30
 [Fix #3674] Signalling compensation on abort

### LINKED ISSUE #3674: Workflow finalizer / abort hook
### Description

The way we currently to it is by updating the state via a graphQL mutation request sent to the DI so there is no known way to me to execute a state that is executed before aborting the workflow.

This would be nice to be able to set a finalizer / abort hook that is always executed when canceling the workflow: if the workflow already started some external execution or made some changes, it would be nice to be able to revert those.

### Implementation ideas

_No response_

fjtirado:
Fix https://github.com/apache/incubator-kie-kogito-runtimes/issues/3674


ricardozanini (approved):

gmunozfe (approved):
Looks good to me, nice work @fjtirado !

PR #3710, opened 2024-10-08
[Fix #3709] Inject action node should be cloned before merge

### LINKED ISSUE #3709: Combination of InjectAction and empty Input parameters generate a memory leak
### Description

when invoking a workflow with no input parameter which first action is an Inject state (for example, our beloved Hello World workflow) the node object associate to the action is reused in all workflows, causing a huge memory leak. 

### Implementation ideas

This is happening because I do not obey a sacred rule: Do not keep **mutable** state in the action object.

fjtirado:
Fix https://github.com/apache/incubator-kie-kogito-runtimes/issues/3709


gmunozfe (approved):
nice work @fjtirado !
A small change in code that leads a great improvement in performance for this kind of workflows like hello-world example

domhanak (commented):
:+1:

PR #3756, opened 2024-10-29
[Fix #3733] Collecting more than one error message

### LINKED ISSUE #3733: ServerlessWorkflow Parser to return a comprenhensive list of errors
### Description

The existing serverless workflow parser is designed to fail as soon as possible to not waste resources. 
So, when an error is found the parser is aborted. 
That means that if a workflow contain several errors, only the first one is reported. 
Thats not good in some situations where iteration of changes is costly, like IA generators. 
Therefore, we should modify the parser to collect as much errors as possible without changing the overall design. 

### Implementation ideas

_No response_

fjtirado:
Fix https://github.com/apache/incubator-kie-kogito-runtimes/issues/3733

### DIFF
--- a/kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/handlers/ActionNodeUtils.java
+++ b/kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/handlers/ActionNodeUtils.java
@@ -34,16 +34,21 @@ public class ActionNodeUtils {
         return embeddedSubProcess.actionNode(context.newId()).name(functionDef.getName());
     }
 
-    public static void checkArgs(FunctionRef functionRef, String... requiredArgs) {
+    public static boolean checkArgs(ParserContext context, FunctionRef functionRef, String... requiredArgs) {
         JsonNode args = functionRef.getArguments();
+        boolean isOk = true;
         if (args == null) {
-            throw new IllegalArgumentException("Arguments cannot be null for function " + functionRef.getRefName());
-        }
-        for (String arg : requiredArgs) {
-            if (!args.has(arg)) {
-                throw new IllegalArgumentException("Missing mandatory " + arg + " argument for function " + functionRef.getRefName());
+            context.addValidationError("Arguments cannot be null for function " + functionRef.getRefName());
+            isOk = false;
+        } else {
+            for (String arg : requiredArgs) {
+                if (!args.has(arg)) {
+                    context.addValidationError("Missing mandatory " + arg + " argument for function " + functionRef.getRefName());
+                    isOk = false;
+                }
             }
         }
+        return isOk;
     }
 
     private ActionNodeUtils() {
--- a/kogito-serverless-workflow/kogito-serverless-workflow-executor-core/src/test/java/org/kie/kogito/serverless/workflow/executor/StaticWorkflowApplicationTest.java
+++ b/kogito-serverless-workflow/kogito-serverless-workflow-executor-core/src/test/java/org/kie/kogito/serverless/workflow/executor/StaticWorkflowApplicationTest.java
@@ -51,7 +51,7 @@ import ch.qos.logback.classic.spi.ILoggingEvent;
 import ch.qos.logback.core.read.ListAppender;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.assertj.core.api.Assertions.catchThrowableOfType;
 import static org.kie.kogito.serverless.workflow.utils.ServerlessWorkflowUtils.getWorkflow;
 import static org.kie.kogito.serverless.workflow.utils.ServerlessWorkflowUtils.writeWorkflow;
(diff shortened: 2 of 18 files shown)


kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/handlers/ActionNodeUtils.java
```diff
@@ -34,16 +34,21 @@ public class ActionNodeUtils {
         return embeddedSubProcess.actionNode(context.newId()).name(functionDef.getName());
     }
 
-    public static void checkArgs(FunctionRef functionRef, String... requiredArgs) {
+    public static boolean checkArgs(ParserContext context, FunctionRef functionRef, String... requiredArgs) {
         JsonNode args = functionRef.getArguments();
+        boolean isOk = true;
```
gmunozfe, line 39: Instead of setting `isOk` to false and continuing the checks, you could return `false` immediately after adding a validation error. This will remove the need for the `isOk` flag entirely, making the code more readable. Wdyt?

```suggestion
```

kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/handlers/ActionNodeUtils.java
```diff
@@ -34,16 +34,21 @@ public class ActionNodeUtils {
...
+            isOk = false;
+        } else {
+            for (String arg : requiredArgs) {
+                if (!args.has(arg)) {
+                    context.addValidationError("Missing mandatory " + arg + " argument for function " + functionRef.getRefName());
+                    isOk = false;
+                }
             }
         }
+        return isOk;
```
gmunozfe, line 51: This part without the `isOk` flag would be:

```suggestion
            context.addValidationError("Arguments cannot be null for function " + functionRef.getRefName());
            return false;
        } 
        for (String arg : requiredArgs) {
            if (!args.has(arg)) {
                context.addValidationError("Missing mandatory " + arg + " argument for function " + functionRef.getRefName());
                return false;
            }
        }
        return true;
```

kogito-serverless-workflow/kogito-serverless-workflow-executor-core/src/test/java/org/kie/kogito/serverless/workflow/executor/StaticWorkflowApplicationTest.java
```diff
@@ -107,6 +109,17 @@ void interpolationFile(String fileName, WorkflowFormat format) throws IOExceptio
         }
     }
 
+    @Test
+    void testValidationError() throws IOException {
+        try (Reader reader = new InputStreamReader(Thread.currentThread().getContextClassLoader().getResourceAsStream("wrong.sw.json"));
+                StaticWorkflowApplication application = StaticWorkflowApplication.create()) {
+            Workflow workflow = getWorkflow(reader, WorkflowFormat.JSON);
+            ValidationException validationException = catchThrowableOfType(() -> application.process(workflow), ValidationException.class);
+            assertThat(validationException.getErrors()).hasSizeGreaterThanOrEqualTo(4);
```
gmunozfe, line 118: Why also greater? Shouldn't it be equal to 4 always?

gmunozfe (approved):
Looks good to me, wonderful work @fjtirado !
Just some comments if you want to simplify and enhance readability of `checkArgs` method

kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/handlers/ActionNodeUtils.java
fjtirado: But I want to collect all missing mandatory parameters, thats the reson for the isOk

kogito-serverless-workflow/kogito-serverless-workflow-executor-core/src/test/java/org/kie/kogito/serverless/workflow/executor/StaticWorkflowApplicationTest.java
fjtirado: The validation errors are a mix of Sonataflow parser errors (which we control) and engine validation errors (which we do not control), so, in order to not tie the unit test too much with the engine implementation, we consider it is ok is there are at least 4 validation errors (one for transition, one for missing error, one for not existing function and another for not connected state) and also check the final exception message contains the four significant strings

kogito-serverless-workflow/kogito-serverless-workflow-builder/src/main/java/org/kie/kogito/serverless/workflow/parser/handlers/ActionNodeUtils.java
fjtirado: see response below

ricardozanini (approved):

PR #3768, opened 2024-11-12
[Fix #3767] Supporting open api functions with File as parameter

### LINKED ISSUE #3767: Sonataflow OpenAPI UploadFile API
### Description

An [OpenAPI](https://swagger.io/specification/) operation that is defined in the spec file as
```
 operationId: uploadJar
      requestBody:
        content:
          application/x-java-archive:
            schema:
              type: string
              format: binary
```
when processed by openapi [quarkiverse](https://github.com/quarkiverse/quarkus-openapi-generator) extension is transformed to the following Java method 
```

  @POST
    @Path("/jars/upload")
    @Consumes({"application/x-java-archive"})
    @Produces({"application/json"})
    @GeneratedMethod ("uploadJar")
    public JarUploadResponseBody uploadJar(
        File body

    );

```
Currently, SonataFlow implementation is not able to invoke that method, since the generated briged to perform the operation will never work (there is not way to convert a JsonObject to a File withtou aditional logic) 

 ```
protected Object internalExecute(org.kie.kogito.openapi.flinkapi.api.DefaultApi openApiRef, java.util.Map<java.lang.String, java.lang.Object> parameters) {
        return openApiRef.uploadJar(buildBody(parameters, java.io.File.class));
    }
```
This issue has been opened to add the misisng logic to create a [File](https://docs.oracle.com/javase/7/docs/api/java/io/File.html) object from the function call arguments. In particular, for an OpenAPI operation which has a File as argument, the argument of the function call in the workflow definition must be a string with the file location in disk

A YAML example:

```
 functionRef:
          refName: uploadJar
          arguments: /uploading/file/path/file.txt

```

After this change, it will be possible to upload files that are accessible in the local file system.

### Implementation ideas

When the target class of the buildBody call is a File, assume that the paramters map contains one entry of type string and build the File object from that string.

fjtirado:
Fix https://github.com/apache/incubator-kie-kogito-runtimes/issues/3767


ricardozanini (approved):

gmunozfe (approved):
Looks good to me, just a couple of minor comments
Excellent work @fjtirado !

PR #3774, opened 2024-11-13
[Fix #3772] Duplicate function error

### LINKED ISSUE #3772: Duplicated function in workflow definition is ignored without message
### Description

If a workflow definition contains the same function definition duplicated, the first one in the list is used, the second one is ignored.
After some thoughts, @ricardozanini and myself agrees that it will be better to throw an error for users to correct it.

### Implementation ideas

_No response_

fjtirado:
Fix https://github.com/apache/incubator-kie-kogito-runtimes/issues/3772

### DIFF
--- a/kogito-serverless-workflow/kogito-serverless-workflow-executor-core/src/test/resources/wrong.sw.json
+++ b/kogito-serverless-workflow/kogito-serverless-workflow-executor-core/src/test/resources/wrong.sw.json
@@ -19,6 +19,11 @@
       "name": "logInfo",
       "type": "custom",
       "operation": "sysout:INFO"
+    },
+    {
+      "name": "pushData",
+      "type": "custom",
+      "operation": "script:python:print('javierito')"
     }
   ],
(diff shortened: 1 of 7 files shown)


wmedvede (approved):

kogito-serverless-workflow/kogito-serverless-workflow-executor-core/src/test/resources/wrong.sw.json
```diff
@@ -19,6 +19,11 @@
       "name": "logInfo",
       "type": "custom",
       "operation": "sysout:INFO"
+    },
+    {
+      "name": "pushData",
+      "type": "custom",
+      "operation": "script:python:print('javierito')"
```
gmunozfe, line 26: Different operation but same name (which is the one to be checked) makes them duplicated, clear example!

gmunozfe (approved):
Looks good to me, nice implementation, taking care for the best performance!

PR #3784, opened 2024-11-19
Connection leak in persistence jdbc common

elguardian:
Closes: https://github.com/apache/incubator-kie-issues/issues/1625

### DIFF
--- a/addons/common/persistence/jdbc/src/main/java/org/kie/kogito/persistence/jdbc/GenericRepository.java
+++ b/addons/common/persistence/jdbc/src/main/java/org/kie/kogito/persistence/jdbc/GenericRepository.java
@@ -22,8 +22,8 @@ import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
-import java.util.ArrayDeque;
-import java.util.Deque;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Optional;
 import java.util.Spliterator;
 import java.util.Spliterators;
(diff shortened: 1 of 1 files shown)


martinweiler (approved):

pefernan (approved):

addons/common/persistence/jdbc/src/main/java/org/kie/kogito/persistence/jdbc/GenericRepository.java
```diff
@@ -160,74 +160,33 @@ Optional<Record> findByBusinessKey(String processId, String processVersion, Stri
...
+        try (Connection connection = dataSource.getConnection();
+                PreparedStatement statement = connection.prepareStatement(sqlIncludingVersion(FIND_ALL, processVersion))) {
             statement.setString(1, processId);
             if (processVersion != null) {
                 statement.setString(2, processVersion);
             }
-            ResultSet resultSet = close.nest(statement.executeQuery());
-            return StreamSupport.stream(new Spliterators.AbstractSpliterator<Record>(
-                    Long.MAX_VALUE, Spliterator.ORDERED) {
+            List<Record> record = new ArrayList<>();
```
fjtirado, line 171: Hi, the pointof the previous PR was to not go through the whole list, avoiding a perfomrance issue is the list of process instances is huge. 
Cant we fix the leak in closeablewrapper without changing this?

fjtirado (changes requested):

addons/common/persistence/jdbc/src/main/java/org/kie/kogito/persistence/jdbc/GenericRepository.java
elguardian: the performance issue was caused due to the lack of pagination. opening a cursor without close it properly causes connection leaks.

addons/common/persistence/jdbc/src/main/java/org/kie/kogito/persistence/jdbc/GenericRepository.java
```diff
@@ -160,74 +160,33 @@ Optional<Record> findByBusinessKey(String processId, String processVersion, Stri
...
+            try (ResultSet resultSet = statement.executeQuery()) {
+                while (resultSet.next()) {
+                    record.add(from(resultSet));
+                }
+            }
+            return StreamSupport.stream(new Spliterators.AbstractSpliterator<Record>(Long.MAX_VALUE, Spliterator.ORDERED) {
```
gmunozfe, line 177: not sure why is needed to overwrite `tryAdvance` method here, passing a List<Record> that already has ordering and a known size. 

Why don't directly `return record.stream();`

addons/common/persistence/jdbc/src/main/java/org/kie/kogito/persistence/jdbc/GenericRepository.java
fjtirado: No, the performance issue is caused because if the result set is large, you have to upload the whole result set in memory anyway.

PR #3794, opened 2024-11-28
[Fix #3793] Remove business key uniqueness

### LINKED ISSUE #3793: Bussiness_key uniqueness constraint
### Description

https://github.com/apache/incubator-kie-kogito-runtimes/pull/3412 added the possibility of resuming a workflow instance execution by using its bussines key
Since business key was not explicitly in db schema but within payload, this implied a linear search, therefore https://github.com/apache/incubator-kie-kogito-runtimes/issues/3413 was used to modify the schema. As part of this change, since business key is suppose to be unique, business key was done primary key of the mapping table. However, it turns out that business key was also used to identify a group of process instances. Therefor this uniqueness constraint causes a backward compatibility issue (althught the functionality being broken was build on the wrong assumption that business key is not unique)
After some analysis, it seems that the more advisable path is to relax the contraint.

### Implementation ideas

Add a new flyway script that modifty the existing uniqueness constraint

fjtirado:
Fix https://github.com/apache/incubator-kie-kogito-runtimes/issues/3793

### DIFF
--- a/addons/common/persistence/jdbc/src/main/resources/kie-flyway/db/persistence-jdbc/postgresql/V10.0.2__remove_business_key_uniqueness.sql
+++ b/addons/common/persistence/jdbc/src/main/resources/kie-flyway/db/persistence-jdbc/postgresql/V10.0.2__remove_business_key_uniqueness.sql
@@ -1,0 +1,21 @@
+--
+-- Licensed to the Apache Software Foundation (ASF) under one
+-- or more contributor license agreements.  See the NOTICE file
+-- distributed with this work for additional information
+-- regarding copyright ownership.  The ASF licenses this file
+-- to you under the Apache License, Version 2.0 (the
+-- "License"); you may not use this file except in compliance
+-- with the License.  You may obtain a copy of the License at
+--
+--   http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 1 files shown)


addons/common/persistence/jdbc/src/main/resources/kie-flyway/db/persistence-jdbc/postgresql/V10.0.2__remove_business_key_uniqueness.sql
```diff
@@ -0,0 +1,21 @@
...
+--
+-- Unless required by applicable law or agreed to in writing,
+-- software distributed under the License is distributed on an
+-- "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+-- KIND, either express or implied.  See the License for the
+-- specific language governing permissions and limitations
+-- under the License.
+--
+
+DROP CONSTRAINT business_key_mapping_pkey;
```
gmunozfe, line 20: There's a compilation error `ERROR: syntax error at or near "CONSTRAINT"`

```suggestion
ALTER TABLE business_key_mapping DROP CONSTRAINT business_key_mapping_pkey;
```

gmunozfe (commented):
Check the syntax @fjtirado

fjtirado:
Closing since the current script is fine according to business_key purpose. 
Will patch for this particular user on midstream

PR #3832, opened 2025-01-22
[incubator-kie-issues#1775] Springboot apps - Conditionally overwrite code-related context properties only if twin system properties are not null

gitgabrio:
Fixes https://github.com/apache/incubator-kie-issues/issues/1775

This PR:
1. overwrite the code-gen context properties only if related system properties are not null/empty
2. unit test newly added methods/checks

<details>
<summary>
</summary>

</details>

### DIFF
--- a/kogito-maven-plugin/src/main/java/org/kie/kogito/maven/plugin/AbstractKieMojo.java
+++ b/kogito-maven-plugin/src/main/java/org/kie/kogito/maven/plugin/AbstractKieMojo.java
@@ -73,19 +73,19 @@ public abstract class AbstractKieMojo extends AbstractMojo {
     @Parameter(required = true, defaultValue = "${project.basedir}")
     protected File baseDir;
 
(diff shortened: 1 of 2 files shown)


yesamer (commented):
@gitgabrio Thank you for the fix and adding the Unit Tests. So, with this change, we are now saying the code-gen MUST be explicitly enabled, otherwise is disabled by default. Is this change consistent with all the examples/documentation we have?

gitgabrio:
@yesamer 
code-gen is already enabled by default (see [here](https://github.com/apache/incubator-kie-kogito-runtimes/blob/61795df025308bd4bdd9da1a651cbe43bb1721d3/kogito-codegen-modules/kogito-codegen-api/src/main/java/org/kie/kogito/codegen/api/Generator.java#L72)). This fix allows user to disable it on demand.

baldimir (approved):

pefernan (approved):

PR #3836, opened 2025-01-29
[NO-ISSUE][DEP UPDATE] Updating logback to 1.5.16

LightGuard:
Follow-up for https://github.com/apache/incubator-kie-issues/issues/1787. We're updating SpringBoot, which is using LogBack 1.5.16. It makes sense to keep the versions the same everywhere.

See also:
https://github.com/apache/incubator-kie-tools/pull/2884
https://github.com/apache/incubator-kie-drools/pull/6236


gitgabrio:
Hi @LightGuard 
I'm slightly confused about this PR:
1 IINW,, it seems to replicate changes done in the springboot one
2 it is not clear why there is a new dependency declared
3 I can't understand why it fails 😄

PR #3837, opened 2025-01-29
[kie-issues 1788] Updating to Keycloak 26.1.0

LightGuard:
Fixes https://github.com/apache/incubator-kie-issues/issues/1788.

Dashbuilder Appformer also uses keycloak, but I think this takes care of it as well. We'll need to do a full build to make sure everything is good.

### DIFF
--- a/kogito-build/kogito-dependencies-bom/pom.xml
+++ b/kogito-build/kogito-dependencies-bom/pom.xml
@@ -111,7 +111,7 @@
     <version.jakarta.persistence-api>3.1.0</version.jakarta.persistence-api>
 
     <version.org.jboss.resteasy>6.2.7.Final</version.org.jboss.resteasy>
-    <version.org.keycloak>25.0.6</version.org.keycloak>
+    <version.org.keycloak>26.1.0</version.org.keycloak>
     <!-- It seems that the confluent kafka cannot replace wurstmeister/kafka so easily. See FAI-729 -->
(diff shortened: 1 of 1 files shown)


baldimir (approved):

yesamer:
> Dashbuilder Appformer also uses keycloak, but I think this takes care of it as well. We'll need to do a full build to make sure everything is good.

Not sure about that

baldimir:
I see this as green. Are there any other checks to be made, please? Or can we merge?

LightGuard:
I believe we're good to merge.

LightGuard:
I've found some other places where we pull in Keycloak transitively in the data-index-persistence addon. It includes quarkus-oidc artifacts, which in turn pull in Keycloak. The biggest problem is that there is no version using a more recent version of keycloak :( I have my doubts of getting it fixed upstream in the 3.15 branch, I could try though. 

Just looked, the more recent versions have changed the pom, so that probably won't work.

Do we exclude keycloak in that dependency and get our own version?

baldimir:
@LightGuard thanks for checking. I think we could just align with what is in Quarkus now and I guess there will be another Quarkus bump soon as as far as I know there is a very recent new LTS release.

LightGuard:
Okay, then we're good with this change for now.

LightGuard:
I believe we're good to merge

PR #3844, opened 2025-02-11
[incubator-kie-issues#1815] Data Index doesn't index any event form runtime when transitioning UserTasks

josedee:
Closes https://github.com/apache/incubator-kie-issues/issues/1815

When transitioning a usertask to the next phase using the new /usertask/instance/{id}/transition endpoint, the process engine events are not indexed in the Data Index.

This is a regression introduced after adding the support to Deadlines in User Tasks (https://github.com/apache/incubator-kie-kogito-runtimes/pull/3752)


pefernan:
@josedee nice catch! Looks like the issue was that the UserTaskService wasn't wrapping the task transition in a UOW so event's weren't emitted.

Let's reformat this code and see if the CI gets green.

Thank you!

pefernan (approved):

pefernan:
Thank you @josedee works like a charm

fjtirado (approved):

PR #3846, opened 2025-02-17
[incubator-kie-issues#1817] Check that topic is disabled

fjtirado:
Fix https://github.com/apache/incubator-kie-issues/issues/1817

### DIFF
--- a/springboot/addons/events/kafka/src/main/java/org/kie/kogito/events/spring/KafkaEventPublisher.java
+++ b/springboot/addons/events/kafka/src/main/java/org/kie/kogito/events/spring/KafkaEventPublisher.java
@@ -25,7 +25,6 @@ import org.kie.kogito.event.EventPublisher;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.beans.factory.annotation.Value;
 import org.springframework.core.env.Environment;
 import org.springframework.kafka.core.KafkaTemplate;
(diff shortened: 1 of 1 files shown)


gmunozfe (approved):
Looks good to me. Well done @fjtirado !
Only a suggestion if you want to add a debug log in case of the topic is disabled and there's no publish call for any of the 3 cases.

springboot/addons/events/kafka/src/main/java/org/kie/kogito/events/spring/KafkaEventPublisher.java
```diff
@@ -89,15 +81,19 @@ public void publish(Collection<DataEvent<?>> events) {
...
+            try {
+                String eventString = json.writeValueAsString(event);
+                logger.debug("Event payload '{}'", eventString);
+                eventsEmitter.send(env.getProperty(String.format(TOPIC_FORMAT, topic, "topic"), topic), eventString);
+                logger.debug("Successfully published event {} to topic {}", event, topic);
+            } catch (Exception e) {
+                logger.error("Error while publishing event to Kafka topic {} for event {}", topic, event, e);
+            }
+        } else {
+            logger.debug("Event {} won't be published becuase topic {} is disabled", event, topic);
```
gmunozfe, line 96: typo
```suggestion
            logger.debug("Event {} won't be published because topic {} is disabled", event, topic);
```

gmunozfe (approved):

pefernan (approved):

PR #3860, opened 2025-03-07
 [incubator-kie-issues#1752] Lambda function returns null - changes in runtimes

bncriju:
Added the fix for this problem in incubator-kie-drools
Fixes: https://github.com/apache/incubator-kie-issues/issues/1752

Related PR: https://github.com/apache/incubator-kie-drools/pull/6261
This required the changes to be done in runtimes repo also.


yesamer (approved):

pibizza:
Question - why we have these dependencies from the internal parts of DMN/Drools inside kogito runtimes? The utility class involved does not seem part of an interface, more part of internal packages. In general, do we have an API for what we should export from the drools project and what we should nto export?

gitgabrio:
> Question - why we have these dependencies from the internal parts of DMN/Drools inside kogito runtimes? The utility class involved does not seem part of an interface, more part of internal packages. In general, do we have an API for what we should export from the drools project and what we should nto export?

In general terms, kogito-runtimes build/adapt the API from drools to be exposed by the kogito environment.
But since I'm sure you know that, what are the ones you specifically referring to ?

baldimir (approved):

PR #3863, opened 2025-03-18
Including Python modules in regular build (but not IT)

fjtirado:
This includes python serverless workflow addon in regular modules, but keep Python IT module just for full build. 
This way community users will be able to test python examples at their own risk, without compromising release because of the inherent flaky nature of pip/jep relationship

### DIFF
--- a/quarkus/addons/pom.xml
+++ b/quarkus/addons/pom.xml
@@ -61,21 +61,8 @@
     <module>process-definitions</module>
     <module>dynamic</module>
     <module>jbpm-usertask-storage-jpa</module>
+    <module>python</module>
   </modules>
-
-  <profiles>
-    <profile>
-      <id>full</id>
-      <activation>
-        <property>
-          <name>full</name>
-        </property>
-      </activation>
-      <modules>
-        <module>python</module>
-       </modules>
-     </profile>
-  </profiles>
 
   <build>
     <pluginManagement>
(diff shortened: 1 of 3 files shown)


wmedvede (approved):

jstastny-cz:
@fjtirado did you actually try building without the full profile being active? I think you broke the build, because of the reactor missing following artifacts when NOT doing full build:
https://github.com/apache/incubator-kie-kogito-runtimes/blob/main/kogito-serverless-workflow/pom.xml#L72
```
  <profiles>
    <profile>
      <id>full</id>
      <activation>
        <property>
          <name>full</name>
        </property>
      </activation>
      <modules>
        <module>kogito-serverless-workflow-python-runtime</module>
        <module>kogito-serverless-workflow-executor-python</module>
      </modules>
    </profile>
  </profiles>
```
Please fix and thoroughly test the build both with full profile active or not (with clean local .m2 repo ideally).

gitgabrio:
HI @fjtirado 
sorry, but this should not have been merged, IMO.
Beside other considerations, now the kogito-runtimes build fails for everyone who does not have the required python stack installed on its machine.
Python usage has always been a sort of _stretched_ use case, but at least the `full` profile isolated it.
Could you pleas revert this ?
Thanks!

fjtirado:
@gitgabrio well, it did not failed in my new laptop (empty .m2). I would have never merged it if I was aware it was breaking anything.

PR #3872, opened 2025-03-27
Incubator-kie-issues-3871] Repeating timer gets executed immediately

abhijithumbe:
https://github.com/apache/incubator-kie-kogito-runtimes/issues/3871

Using timer node with timer expression R/PT5M, the job gets executed immediately, then in the configured 5min intervals.

### DIFF
--- a/jbpm/jbpm-flow/src/test/java/org/jbpm/process/core/timer/DateTimeUtilsTest.java
+++ b/jbpm/jbpm-flow/src/test/java/org/jbpm/process/core/timer/DateTimeUtilsTest.java
@@ -18,9 +18,11 @@
  */
 package org.jbpm.process.core.timer;
 
+import java.time.Duration;
 import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
+import java.time.temporal.ChronoUnit;
 
 import org.jbpm.test.util.AbstractBaseTest;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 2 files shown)


jbpm/jbpm-flow/src/test/java/org/jbpm/process/core/timer/DateTimeUtilsTest.java
```diff
@@ -135,7 +135,7 @@ public void testParseRepeatablePeriodOnly() {
 
         assertThat(parsedRepeatable[0]).isEqualTo(-1L);
         // Default delay time is 1000ms
-        assertThat(parsedRepeatable[1]).isEqualTo(1000L);
```
abduljaleelak, line 138: If we need to cover this targeted logic, we can add below junit 

```

@Test
    public void testDefaultDelay() {
        String isoString = "R/PT0M";
        long[] parsedRepeatable = DateTimeUtils.parseRepeatableDateTime(isoString);
        // Default delay time is 1000ms
        assertThat(parsedRepeatable[1]).isEqualTo(1000L);
    }

```

martinweiler: I think we should have the same test here as it was used in v7:
https://github.com/kiegroup/jbpm/blob/main/jbpm-flow/src/test/java/org/jbpm/process/core/timer/DateTimeUtilsTest.java#L126-L136

In the new syntax, this would be:
```
        assertThat(parsedRepeatable[0]).isEqualTo(-1L);
        assertThat(parsedRepeatable[1] <= MINUTE_IN_MILLISECONDS).as("Parsed delay is bigger than " + MINUTE_IN_MILLISECONDS).isTrue();
        assertThat(parsedRepeatable[1] > FIFTY_NINE_SECONDS_IN_MILLISECONDS)
                .as("Parsed delay is too low! Expected value is between " + MINUTE_IN_MILLISECONDS + " and " + FIFTY_NINE_SECONDS_IN_MILLISECONDS + " but is " + parsedRepeatable[1]).isTrue();
        assertThat(parsedRepeatable[2]).as("Parsed period should be one minute in milliseconds but is " + parsedRepeatable[2]).isEqualTo(MINUTE_IN_MILLISECONDS);
```

Not directly linked to this change, but as you are on it, I think it might be a good idea to add the additional tests from v7 (starting in L138) to the codebase as well.

abhijithumbe: @martinweiler  Thank you for looking into this. I have included additional tests as you suggested.Please take a look.

martinweiler (approved):
Great job @abhijithumbe

pefernan (approved):

PR #3922, opened 2025-05-15
[kie-issues#1964] Fix locking mechanism to use processinstanceid instead of workitemid

martinweiler:
Closes https://github.com/apache/incubator-kie-issues/issues/1964

Thanks @elguardian for spotting this issue!


elguardian (approved):
LGTM

PR #3967, opened 2025-06-25
[kie-issues#2014] Ensure user tasks gets completed after PIM

martinweiler:
Closes https://github.com/apache/incubator-kie-issues/issues/2014

### DIFF
--- a/addons/common/persistence/jdbc/src/test/java/org/kie/persistence/jdbc/AbstractProcessInstancesIT.java
+++ b/addons/common/persistence/jdbc/src/test/java/org/kie/persistence/jdbc/AbstractProcessInstancesIT.java
@@ -252,6 +252,40 @@ abstract class AbstractProcessInstancesIT {
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


jomarko (commented):
Thank you for this immediate fix @martinweiler . I wonder if we can test this change somehow. I see we have:

https://github.com/apache/incubator-kie-kogito-runtimes/blob/main/addons/common/persistence/jdbc/src/test/java/org/kie/persistence/jdbc/AbstractProcessInstancesIT.java



with `testMigrateAll` and `testMigrateSingle`

martinweiler:
Thanks for your input @jomarko !

I added a test case that covers the scenario of migrating a process instance containing a user task, to verify if it completes correctly on the target process definition after the migration. However, you might notice that this test case doesn't exactly verify the code change of the fix, as it doesn't use the `usertask` module with the `transition` method, but rather the `completeWorkItem` call.

I need to check with @pefernan when he is back on how to improve this test and wire the application with the `usertask` module.

PR #3970, opened 2025-06-26
[Fix #3969] Making sure log method is unique

### LINKED ISSUE #3969: Using log() fluent method in two different states cause error
### Description

A duplicate function validation error is thrown. 
This does not happen if the log is withing the same state, but different actions

### Implementation ideas

_No response_

fjtirado:
Fix https://github.com/apache/incubator-kie-kogito-runtimes/issues/3969


kogito-serverless-workflow/kogito-serverless-workflow-fluent/src/main/java/org/kie/kogito/serverless/workflow/fluent/ActionBuilder.java
```diff
@@ -118,12 +118,13 @@ public static ActionBuilder call(String functionName, JsonNode args) {
         return new ActionBuilder(new Action().withFunctionRef(new FunctionRef().withRefName(functionName).withArguments(args)));
     }
 
+    @Deprecated
```
ricardozanini, line 121: Can you add a note to use the other one?

ricardozanini (commented):
LGTM, just add a comment to the deprecated annotations to alert what to use instead. 🙏

kogito-serverless-workflow/kogito-serverless-workflow-fluent/src/main/java/org/kie/kogito/serverless/workflow/fluent/ActionBuilder.java
fjtirado: sure, done

ricardozanini (approved):

PR #3983, opened 2025-07-15
[incubator-kie-issues-2032] Clear actual owner after reassign

josedee:
This issue has been identified in some tests by Santander devs while evaluating the reassignments feature.

Santander want to have User tasks reassigned to a group at the end of day if the task isn't completed. To do so they are adding reassignments on the Usertasks.

Apparently, after the reassignment is applied, the task is released (not reserved by the user) but the actual owner is still set.


josedee:
Added tests @martinweiler

PR #3989, opened 2025-07-21
[Fix apache/incubator-kie-issues#1986] Forcing runtime read of config

fjtirado:
Fix https://github.com/apache/incubator-kie-issues/issues/1986


gmunozfe (approved):
Looks good to me, astounding fix @fjtirado !

PR #3991, opened 2025-07-22
[incubator-kie-issues-2036] Workflow Engine: Process instance getting stuck in Join node intermittently with subprocess based model

elguardian:
closes: https://github.com/apache/incubator-kie-issues/issues/2036


martinweiler (approved):
Nice solution @elguardian

PR #4008, opened 2025-08-03
[incubator-kie-issues#2043] Fix text annotation

alishaaaqil:
Closes: https://github.com/apache/incubator-kie-issues/issues/2043

### DIFF
--- a/jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/TextAnnotationHandler.java
+++ b/jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/TextAnnotationHandler.java
@@ -1,3 +1,21 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
 package org.jbpm.bpmn2.xml;
 
 import java.util.HashMap;
--- a/jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/core/TextAnnotation.java
+++ b/jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/core/TextAnnotation.java
@@ -1,3 +1,21 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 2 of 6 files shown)


jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/TextAnnotationHandler.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    public Object start(String uri, String localName, Attributes attrs, Parser parser) throws SAXException {
+        parser.startElementBuilder(localName, attrs);
+        String id = attrs.getValue("id");
+        TextAnnotation annotation = new TextAnnotation();
+        annotation.setId(id);
+        // Register the TextAnnotation by ID in ProcessBuildData
+        ((ProcessBuildData) parser.getData()).setMetaData(id, annotation);
```
elguardian, line 48: This is not good. You can collide in ehre with other stuff.
the text of the annotation is not there. You need to add it.

more like

Map<String, TextAnnotation> annotations
((ProcessBuildData) parser.getData()).setMetaData(ANNOTATIONS, annotations);

jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/TextAnnotationHandler.java
```diff
@@ -0,0 +1,63 @@
...
+            this.validPeers.add(Interface.class);
+            this.validPeers.add(Escalation.class);
+            this.validPeers.add(Error.class);
+            this.validPeers.add(Signal.class);
+            this.validPeers.add(DataStore.class);
+            this.validPeers.add(RuleFlowProcess.class);
+            this.validPeers.add(Node.class);
+            this.validPeers.add(SequenceFlow.class);
+            this.validPeers.add(TextAnnotation.class);
+            this.validPeers.add(MetaDataHandler.MetaDataWrapper.class);
```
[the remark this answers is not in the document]
alishaaaqil: i have removed it , but kept the sequenceFlow, as its causing error without it.

jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/TextAnnotationHandler.java
alishaaaqil: Sure wll resolve this, missed it

elguardian (changes requested):
plz address the constants

alishaaaqil:
Added unit test along with few other changes , since some cases where text annotation on different tasks ,timers where failing .

jbpm/jbpm-tests/src/test/java/org/jbpm/bpmn2/test/TextAnnotationTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.jbpm.bpmn2.JbpmBpmn2TestCase;
+import org.jbpm.bpmn2.core.TextAnnotation;
+import org.junit.jupiter.api.Test;
+import org.kie.kogito.internal.process.runtime.KogitoProcessInstance;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.jbpm.ruleflow.core.Metadata.TEXT_ANNOTATIONS;
+
+public class TextAnnotationTest extends JbpmBpmn2TestCase {
```
martinweiler, line 13: Thanks for adding the test class. Please convert the tests to use the codegen models, instead of the legacy runtime. Thanks!

elguardian (approved):
test only check they can process the test annotation. but they might not be there in the codegen. Anyway they are useless from processing point of view

PR #4017, opened 2025-08-12
[incubator-kie-issues#2060] Generate SourceFilesProvider when there are no models in the Business Service project

gitgabrio:
Fixes https://github.com/apache/incubator-kie-issues/issues/2060

This PR:
1. introduce a `DummyProcess` class to instantaite as placeholder when there are no bpmn models in the application but the kie-addons-quarkus-source-files is present
2. in the above case, all check and code-generation are skept
3. the generated `SourceFilesProviderProducer` does not have the static initializer, so that the retrieved `SourceFilesProvider` does not have any content
4. I also tested locally native build§
5. this is meant as temporary solution to unlock other PRs, because IMO the overall approach and design has to be reviewed by SMEs (e.g. - all the add-ons are actually "plugins", but their code is implemented in the core itself)

<details>
<summary>
</summary>

</details>


baldimir (approved):

tiagobento:
Build is green. Two approvals are in. Merging to unblock https://github.com/apache/incubator-kie-tools/pull/3219.

PR #4023, opened 2025-08-13
[incubator-kie-issues#4022] Memory leak with external signal scope

abhijithumbe:
Closes #[4022](https://github.com/apache/incubator-kie-kogito-runtimes/issues/4022)


jbpm/jbpm-flow/src/main/java/org/jbpm/process/instance/LightWorkItemManager.java
```diff
@@ -146,6 +146,7 @@ public void completeWorkItem(String workItemId, Map<String, Object> data, Policy
         KogitoProcessInstance processInstance = processInstanceManager.getProcessInstance(workItem.getProcessInstanceStringId());
         workItem.setState(KogitoWorkItem.COMPLETED);
         processInstance.signalEvent("workItemCompleted", workItem);
+        internalRemoveWorkItem(workItemId);
```
elguardian, line 149: the transition work item will take of this already if we reached the point of completed.

martinweiler (approved):
Nice work, thank you @abhijithumbe

PR #4026, opened 2025-08-14
[Fix #4024] Allowing multiple parameters on python invocation

### LINKED ISSUE #4024: using python functions works only for a single param function
### Describe the bug

i am refering to funcitons with positional params. take ctime and strftime as an example. ctime has one param. strftime has two params. If I write in the workflow `arguments: "$WORKFLOW.python.pyNowSecs"` when calling ctime then it works. if i want to call strftime and i use array arguments: `["%Y-%m-%d %H:%M:%S", "$WORKFLOW.python.pyNowSecs"]` then it fails with `jep.JepException - <class 'TypeError'>: strftime() argument 1 must be str, not ArrayList`

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Output of `uname -a` or `ver`

_No response_

### Output of `java -version`

_No response_

### GraalVM version (if different from Java)

_No response_

### Kogito version or git rev (or at least Quarkus version if you are using Kogito via Quarkus platform BOM)

_No response_

### Build tool (ie. output of `mvnw --version` or `gradlew --version`)

_No response_

### Additional information

_No response_

fjtirado:
Fix https://github.com/apache/incubator-kie-kogito-runtimes/issues/4024

In order to specify multiple positional parameters, user should pass an array.

Note that keyword parameters was already supported

### DIFF
--- a/kogito-serverless-workflow/kogito-serverless-workflow-python-runtime/src/main/java/org/kie/kogito/serverless/workflow/python/PythonServiceWorkItemHandler.java
+++ b/kogito-serverless-workflow/kogito-serverless-workflow-python-runtime/src/main/java/org/kie/kogito/serverless/workflow/python/PythonServiceWorkItemHandler.java
@@ -18,6 +18,7 @@
  */
 package org.kie.kogito.serverless.workflow.python;
 
+import java.util.Collection;
 import java.util.Map;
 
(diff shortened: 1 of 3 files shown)


gmunozfe (approved):
Looks good to me, just a minor comment, awesome work @fjtirado !

PR #4028, opened 2025-08-20
[incubator-kie-issues#2070]Entities not discovered and missing transactional boundaries

alishaaaqil:
Closes: https://github.com/apache/incubator-kie-issues/issues/2070


elguardian:
this should work. Not very sure why it is passing CI

PR #4062, opened 2025-09-12
NO-ISSUE: fix test failures on windows

jstastny-cz:
Fixing PredictionCodegenFactory class which was using File.separator matches against an instance of PortablePath. PortablePath though contains forward slashes `/` even on Windows, this such File.separator match fails there and results in error during code generation. (The only non-test fix in this PR).

Fixing several problems in tests causing failures in Windows:
* resource path has always forward slash `/` no need for `File.separator` in there. Replacing such File.separator references with plain `/` string/char.
* URL#getPath method returns resource path - not a valid path on Windows. Switching to Paths#get method signature that accepts URI parameter instead.
* flaky DeadlineHelperTest has perhaps a too strict check on expiration time, using `isAfter` assertion, which IMHO is susceptible to cases when the test is too fast (assumption). Adding a few millis to stabilize.
* replacing string comparison on paths with Path instances comparisons.

### DIFF
--- a/jbpm/jbpm-tests/src/test/java/org/jbpm/bpmn2/BrokenStructureRefTest.java
+++ b/jbpm/jbpm-tests/src/test/java/org/jbpm/bpmn2/BrokenStructureRefTest.java
@@ -18,6 +18,7 @@
  */
 package org.jbpm.bpmn2;
 
+import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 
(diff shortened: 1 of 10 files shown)


jstastny-cz:
@pibizza would you be willing to review this one? There are some changes to the test assertions you've rewritten recently.

pefernan (approved):

yesamer:
@jstastny-cz @pefernan Is this PR still valid? Should we merge it?

jstastny-cz:
@yesamer yes, all the fixes should be still relevant, I've rebased and restarted the PR checks.

pibizza (approved):

PR #4091, opened 2025-10-13
[kie-issues#2141] Improve OpenAPI v3 spec endpoints

yesamer:
Closes https://github.com/apache/incubator-kie-issues/issues/2141

Relates to: https://github.com/apache/incubator-kie-drools/pull/6484

Scope of this PR is:
Improve the generated OpenAPI v3 spec with enriched data, in details:
- Using the TAG annotation to classify the endpoints by model and type. The same approach should be used for all the model types (Rule, DMN, Process). Used pattern is:
     TYPE - MODEL NAME (Eg. Rule - HelloQuery, Decision - Traffic Violation, Process - Orders)
- Using the Operation tag to identify an endpoint in an unique way

The result of the TAG usage is:

DMN Endpoints BEFORE

dmn-quarkus-before

AFTER
Quarkus
dmn-quarkus-after
Springboot
dmn-spingboot-after

Process

BEFORE
Screenshot 2025-10-14 at 10 03 13
Screenshot 2025-10-14 at 10 07 04
AFTER
Screenshot 2025-10-14 at 11 05 17

### DIFF
--- a/springboot/addons/process-instance-migration/src/main/java/org/kie/kogito/process/management/ProcessInstanceMigrationRestController.java
+++ b/springboot/addons/process-instance-migration/src/main/java/org/kie/kogito/process/management/ProcessInstanceMigrationRestController.java
@@ -63,7 +63,7 @@ public class ProcessInstanceMigrationRestController extends BaseProcessInstanceM
 
     @Override
     @PostMapping(value = "{processId}/migrate", produces = APPLICATION_JSON_VALUE)
-    public ResponseEntity migrateAllInstances(String processId, ProcessMigrationSpec migrationSpec) {
+    public ResponseEntity migrateAllInstances(@PathVariable("processId") String processId, ProcessMigrationSpec migrationSpec) {
         return doMigrateAllInstances(processId, migrationSpec);
     }
 


fjtirado:
@yesamer By disabling this https://github.com/apache/incubator-kie-kogito-runtimes/pull/4091/files#diff-51bf67d0f9bd0e476c3c17809373974eb0824e72f2cd5c2534f8bf48211068c3L224-R227 (more precisely, by adding Process - to the tag), the mapping between input schema and workflow was break here  https://github.com/apache/incubator-kie-kogito-runtimes/blob/main/quarkus/extensions/kogito-quarkus-serverless-workflow-extension/kogito-quarkus-serverless-workflow-deployment/src/main/java/org/kie/kogito/serverless/workflow/parser/schema/OpenApiModelSchemaGenerator.java#L137-L141

Hence my question, is the "Process - " prefix really needed in the tag name, cannot be moved to the description?

fjtirado:
> Hi @fjtirado, apologies for the bad consequence of applying this change in that part of the code. As I wrote in the description, the idea is to unify the Tag management for all the model types we expose. The idea of having the TYPE - MODEL NAME pattern is to clearly state to which model (type and name) that subset of endpoints belongs. In the case of Kogito project with DMN, RULE and BPMN models, the previous approach didn't provide so much clarity on that. With the TYPE - MODEL NAME, the end user can now immediately know that information. I think that moving the TYPE in the description doesn't fully accomplish that goal of proving clarity to the end user: Indeed the information is present, but having it a tag level provides more visibility, in swagger. Moreover, in the process case, SpringBoot-based examples expose the tag in a different way than Quarkus-based example (see images above: person vs person-resources). Said so, my suggestion is to keep the new pattern and apply the change in the code you pointed out. I volunteer to do it, in case. WDYT?

@yesamer I think just keeping the tag name as it was will do the trick. Basically removing the "Process -" prefix from the resource template. I need to do a PR anyway (because this functionality was not only broken by this change, but  also broken by changes to [migrate](https://github.com/apache/incubator-kie-kogito-runtimes/pull/3935/files#diff-dbe9b4e217720b3b0fd2c9cb8bd8a5e83848069f7a91f17402f57c303f7ee4ffR49-R71) to quarkus 3.20 ) so if the pattern is really needed (as it seems), then I will change the code myself. 

Implementation note (for future ones): One problem of moving the code to the template if that you have to copy the "Process -" constant prefix in two different places: the template and the code that needs to check for it, I think it was better to use TagGenerator.

yesamer:
@fjtirado Thank you too for your cooperation :)

PR #4059, opened 2025-09-11
[Incubator-kie-issues-2093] Workflow Runtime: Problems when emitting events inside a @Transactional endpoint

elguardian:
Closes https://github.com/apache/incubator-kie-issues/issues/2093
ensmble:
https://github.com/apache/incubator-kie-kogito-runtimes/pull/4059
https://github.com/apache/incubator-kie-kogito-examples/pull/2134

PR #4112, opened 2025-10-27
[incubator-kie-issues-2093] remove forgotten reference to kie-addons-…

jstastny-cz:
…quarkus-messaging-deployment

Follows-up on #4059 which removed artifact `org.kie:kie-addons-quarkus-messaging-deployment`, though by accident left its usage in one of integration tests that are part of the build reactor.

This problem was not apparent during PR check build as this was hidden by the fact that there was an older artifact downloaded from snapshots repository. The apparent failure would happen if the project version is changed, e.g. during a release build, then it would fail to locate the artifact and reactor build would fail.


baldimir:
Merging as this may eventually fix a potential build problem. The Jenkins CI problem looked not related. All other PR checks pass.

PR #4184, opened 2026-02-05
[incubator-kie-issues#2239] Process compilation fails if process has event subprocess with TimerStartEvent

Abhitocode:
Closes: https://github.com/apache/incubator-kie-issues/issues/2239

Process execution fails if process has event subprocess with TimerStartEvent. 

Error:
```
Caused by: java.lang.IllegalStateException: Process could not be validated ![Process 'scripts' [scripts]: Node 'Start' [_D7F366D7-ABE9-41E8-A3C0-CC1C7A37C1F1] Start in Event SubProcess 'Event Sub-process' [_BC671924-5590-42CC-A8FA-EC1A4FB50869] must contain a trigger (event definition).]
        at org.jbpm.ruleflow.core.RuleFlowProcessFactory.validate(RuleFlowProcessFactory.java:245)
        at org.acme.travels.ScriptsProcess.process(ScriptsProcess.java:171)
```

**Root cause as I understood:**
The code generation path was missing logic to add a trigger to the StartNode and register the timer on the EventSubProcessNode for timer start events. The XML parsing path handles this correctly, but the code generation path did not, causing validation to fail.

**Fix :**
Added trigger creation for timer events in StartNodeVisitor when StartNode is inside an EventSubProcessNode
Added timer registration on EventSubProcessNode in EventSubProcessNodeVisitor
Addedtimer()method in EventSubProcessNodeFactory to support timer registration

**Action is "null" error when the timer expire**s :
for Event Subprocess timers, the Action is never compiled or wired during process compilation, it's always null. The code was calling action.execute() without a null check, causing the error.

Fix: Added null check

### DIFF
--- a/jbpm/jbpm-tests/src/test/java/org/jbpm/bpmn2/IntermediateEventTest.java
+++ b/jbpm/jbpm-tests/src/test/java/org/jbpm/bpmn2/IntermediateEventTest.java
@@ -101,7 +101,6 @@ import org.jbpm.process.workitem.builtin.ReceiveTaskHandler;
 import org.jbpm.process.workitem.builtin.SendTaskHandler;
 import org.jbpm.process.workitem.builtin.SystemOutWorkItemHandler;
 import org.jbpm.test.util.NodeLeftCountDownProcessEventListener;
-import org.jbpm.test.util.NodeTriggeredProcessEventListener;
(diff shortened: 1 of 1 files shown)


pefernan (approved):
great @Abhitocode!

PR #4229, opened 2026-03-26
[incubator-kie-issues#2253] Replace instantiation of streams inside generated DecisionModels with creation of modelPaths.txt file that is parsed at DMNRuntime instantiation

### LINKED ISSUE #2253: DMN: refactor DecisionModels codegeneration
The following snippet is an example of code-generated "DecisionModels":

```java
    static {
        init(/* arguments provided during codegen */
        null, null, java.util.Set.of(), true, readResource(Application.class.getResourceAsStream("/valid_models/DMNv1_5/Importing_EmptyNamed_Model_With_Href_Namespace.dmn"), "UTF-8"), readResource(Application.class.getResourceAsStream("/valid_models/DMNv1_5/Imported_Model_Unamed.dmn"), "UTF-8"), readResource(Application.class.getResourceAsStream("/valid_models/DMNv1_5/TypeConstraintsChecks.dmn"), "UTF-8"), readResource(Application.class.getResourceAsStream("/valid_models/DMNv1_5/AllowedValuesChecksInsideCollection.dmn"), "UTF-8"));
    }
```

The problem with that class are:
1. `readResource(InputStream, String)` returns an `InputStreamReader` that contains an `InputStream`, which could lead to not-efficient resource consuming
2. it has been reported at least once the case where at least 3500 DMN files are automatically generated; in this case, for each DMN, a `readResource` entry is written in the class, leading to “Error during processing model classes: code too large” error during compilation

So, we need to devise a different approach to:
1. limit the concurrently-opened InputStreams
2. limit the size of codegenerated class

gitgabrio:
Fixes https://github.com/apache/incubator-kie-issues/issues/2253

The original issue was mainly that, with lot (~3500) dmn models to manage, the generated DecisionModels class become too big to be read.
That also raised the doubt about having so many streams opened at once.

The proposed solution is:

1. create a "modelPaths.txt" file that contains the path of each model and its charset (e.g. 

```
/valid_models/DMNv1_5/Importing_EmptyNamed_Model_With_Href_Namespace.dmn:UTF-8
/valid_models/DMNv1_5/Imported_Model_Unamed.dmn:UTF-8
/valid_models/DMNv1_5/TypeConstraintsChecks.dmn:UTF-8
/valid_models/DMNv1_5/AllowedValuesChecksInsideCollection.dmn:UTF-8
```
inside DecisionCodeGenUtils.generateModelPathsFile()
2. create new methods to allow this different codepath without breaking backward compatibility
3.  parse that file to a Map only soon after the dmn runtime gets instantiated (inside AbstractDecisionModels.init)
4. instantiate the dmnRuntime with the map provided above

Caveat: while this could resolve the initial issue, it is to be verified how the rest of the execution would behave with so many models to manage.

@AthiraHari77 @ChinchuAjith @mumi

<details>
<summary>
</summary>

</details>


pefernan (approved):
Not expert in the area, but looks like a good improvement and everything seems to work great.

thank you @gitgabrio

drools/kogito-dmn/src/main/java/org/kie/kogito/dmn/DMNKogito.java
```diff
@@ -96,4 +124,57 @@ public static KogitoDMNResult evaluate(DMNRuntime dmnRuntime, String modelNamesp
         return new KogitoDMNResult(modelNamespace, modelName, evaluationResult.result);
     }
 
+    static Optional<InputStream> getDMNModelStream(String path) {
+        logger.info("getDMNModelStream for {}", path);
```
jstastny-cz: A bit concerned about the log level here, @gitgabrio you mentioned high numbers of dmn models causing the problem - the logs will be very long, right?

drools/kogito-dmn/src/main/java/org/kie/kogito/dmn/DMNKogito.java
```diff
@@ -96,4 +124,57 @@ public static KogitoDMNResult evaluate(DMNRuntime dmnRuntime, String modelNamesp
...
+        InputStream toReturn = Application.class.getResourceAsStream(path);
+        if (toReturn != null) {
+            return Optional.of(toReturn);
+        } else {
+            logger.warn("DMN model stream not found in Application.class.getResourceAsStream");
```
jstastny-cz: Also here, from how I read this method, there are various "fallback" ways to locate the resource, if each of the models produces 3 warn log messages, but eventually succeeds, the warn logs are then not relevant really.

quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions-integration-test/src/main/resources/application.properties
```diff
@@ -18,3 +18,4 @@
 #
 
 kogito.decisions.stronglytyped=true
+quarkus.log.category."org.kie.kogito".level=DEBUG
```
jstastny-cz: Is this intended change or was just necessary during development?

jstastny-cz (commented):
@gitgabrio a few things caught my attention, please consider if my comments are relevant.

drools/kogito-dmn/src/main/java/org/kie/kogito/dmn/DMNKogito.java
gitgabrio: yeah, true... thanks for spotting that!

quarkus/extensions/kogito-quarkus-decisions-extension/kogito-quarkus-decisions-integration-test/src/main/resources/application.properties
gitgabrio: sorry, left-over....

kogito-codegen-modules/kogito-codegen-decisions/src/main/java/org/kie/kogito/codegen/decision/DecisionCodegenUtils.java
```diff
@@ -334,6 +360,25 @@ static void generateAndStoreDecisionModelResourcesProvider(Collection<GeneratedF
...
     }
 
+    private static Optional<String> determineEncoding(CollectedResource resource) {
+        try {
+            BufferedReader br = new BufferedReader(resource.resource().getReader());
+            StringBuilder sb = new StringBuilder(br.readLine());
+            sb.append(br.readLine());
+            String head = sb.toString();
+            boolean prologUTF8 = head.startsWith("<?xml version=\"1.0\" encoding=\"UTF-8\"") || head.startsWith("<?xml version=\"1.0\" encoding=\"utf-8\"");
+            boolean kogitoDMNEditor = head.contains("xmlns:kie=\"http://www.drools.org/kie/dmn");
```
yesamer, line 371: @gitgabrio Just out of curiosity, why is this logic required?

yesamer: AH ok, I noticed that code was already there, you just moved it.

yesamer (approved):

yesamer (approved):

yesamer:
@gitgabrio As this PR covers DMN only, should the txt file be named dmnModelPaths.txt? Or are you planning to use that file for any model?

PR #4273, opened 2026-05-04
[incubator-kie-issues#2308] Improve usertask endpoint

### LINKED ISSUE #2308: Improve the usertasks endpoint
The technical API provides the endpoint

http://localhost:8080/usertasks/instance?user=jdoe
returning a list of tasks, including the task details.

This API could be improved:

/instance is indicating the endpoint refers to a single id, which is not the case. http://localhost:8080/usertasks/ should be the list endpoint
It should be possible to narrow the scope of the query by process id, process instance id, etc.
The amount of data returned on the list is quite verbose and could be limited, eg. returning taskid, processid, name, etc is enough for the list, and then use the /usertasks/instance/id to get all detail
Note that any changes need to be done with backwards compatibility in mind.

GopikaReghunath:
Closes https://github.com/apache/incubator-kie-issues/issues/2308

This PR introduces performance optimizations and enhanced filtering capabilities for user task REST endpoints through lightweight DTOs and improved API design.

**1️⃣ Refactored Main Endpoint: `GET /usertasks/instance`**

Purpose: Unified endpoint for listing user tasks with flexible response format

```
UserTaskInfo Fields (Short Format)
{
  "id": "task-123",
  "userTaskId": "approval-task",
  "taskName": "Approve Request",
  "taskDescription": "Review and approve the request",
  "taskPriority": "1",
  "status": { "name": "Ready" },
  "actualOwner": "john",
  "processId": "hiring",
  "processInstanceId": "instance-456",
  "processVersion": "1.0"
}
```

Query Parameters:

- format -
  - long (default): List - Complete task information including inputs, outputs, metadata
  - short: List - Lightweight DTO with essential fields only
- user : Filter by user identity
- group : Filter by group membership
- processId : Filter by process definition
- processInstanceId : Filter by process instance
- status : Filter by task status
- taskName : Filter by task name (case-insensitive partial match)

```
Examples

Basic query with lightweight response:
GET /usertasks/instance?user=john&format=short

Advanced filtering:
GET /usertasks/instance?user=john&group=managers&status=Ready&processId=hiring&format=short

Multiple statuses:
GET /usertasks/instance?user=john&status=Ready&status=Reserved&status=InProgress

```

**2️⃣ Optimized Input/Output Endpoints**

- Input Endpoint

```
PUT /usertasks/instance/{taskId}/inputs
Request: Map<String, Object>
Response: UserTaskInputsView
```

UserTaskInputsView - Contains only id + inputs map

```
{
  "id": "task-123",
  "inputs": {
    "applicant": "John Doe",
    "position": "Senior Developer"
  }
}
```

- Output Endpoint

```
PUT /usertasks/instance/{taskId}/outputs
Request: Map<String, Object>
Response: UserTaskOutputsView
```

UserTaskOutputsView - Contains only id + outputs map

```
{
  "id": "task-123",
  "outputs": {
    "approved": true,
    "comments": "Excellent candidate"
  }
}
```

Benefits: Reduced payload size compared to full UserTaskView response


addons/common/jbpm-usertask-storage-jpa/src/main/java/org/jbpm/usertask/jpa/repository/UserTaskInstanceRepository.java
```diff
@@ -42,6 +48,56 @@ public List<UserTaskInstanceEntity> findByIdentity(IdentityProvider identityProv
         return query.getResultList();
     }
 
+    public List<UserTaskInstanceEntity> findByIdentityAndFilter(IdentityProvider identityProvider, UserTaskFilter filter) {
+        // Start with identity-based query
+        List<UserTaskInstanceEntity> identityFiltered = findByIdentity(identityProvider);
+
+        // If no filter, return identity-filtered results
+        if (filter == null) {
+            return identityFiltered;
```
martinweiler: Isn't this kind of inefficient to create this first query, and in case of further filters, execute another query to apply the additional filters? Could we do something like the following instead?
```
if(filter == null) {
    return findByIdentity(identityProvider);
}

// ... create single query combining identity and filter predicates
```

addons/common/jbpm-usertask-storage-jpa/src/test/java/org/jbpm/usertask/jpa/JPAUserTaskInstancesFilterTest.java
```diff
@@ -0,0 +1,147 @@
...
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/**
+ * Test class for JPAUserTaskInstances.findByIdentityAndFilter() method.
+ * This test verifies that the method correctly delegates to the repository
+ * and maps the results using the entity mapper.
+ */
+@ExtendWith(MockitoExtension.class)
+public class JPAUserTaskInstancesFilterTest {
```
martinweiler, line 51: I only see mocked tests in this PR, but no actual integration-tests that verify the correctness of the filter queries. Can you add those @GopikaReghunath ?

class-templates/usertask/RestResourceUserTaskQuarkusTemplate.java
```diff
@@ -0,0 +1,276 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper.DefaultTyping;
+import com.fasterxml.jackson.databind.cfg.MapperConfig;
+import com.fasterxml.jackson.databind.deser.DeserializationProblemHandler;
+import com.fasterxml.jackson.databind.jsontype.PolymorphicTypeValidator;
+import com.fasterxml.jackson.databind.jsontype.TypeIdResolver;
+import com.fasterxml.jackson.databind.jsontype.PolymorphicTypeValidator.Validity;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+
+@Path("/usertasks/instance")
+public class UserTasksResource {
```
martinweiler: Is this file required here? Aren't the templates already in `kogito-codegen-modules`?

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/usertask/RestResourceUserTaskQuarkusTemplate.java
```diff
@@ -86,20 +102,54 @@ public void init() {
...
     @Produces(MediaType.APPLICATION_JSON)
-    public List<UserTaskView> list(@QueryParam("user") String user, @QueryParam("group") List<String> groups) {
-        return userTaskService.list(identityProviderFactory.getOrImpersonateIdentity(user, groups));
+    public Object list(
+            @QueryParam("user") String user,
+            @QueryParam("group") List<String> groups,
+            @QueryParam("format") String format,
+            @QueryParam("processId") String processId,
+            @QueryParam("processInstanceId") String processInstanceId,
+            @QueryParam("status") String status,
```
martinweiler: Should we support a list of `status` elements?

GopikaReghunath: changes done, supports a list of statuses for filtering

class-templates/usertask/RestResourceUserTaskQuarkusTemplate.java
GopikaReghunath: removed the unwanted file

addons/common/jbpm-usertask-storage-jpa/src/test/java/org/jbpm/usertask/jpa/JPAUserTaskInstancesFilterTest.java
GopikaReghunath: integration tests added

addons/common/jbpm-usertask-storage-jpa/src/main/java/org/jbpm/usertask/jpa/repository/UserTaskInstanceRepository.java
GopikaReghunath: comment addressed

quarkus/integration-tests/integration-tests-quarkus-usertasks/src/test/java/org/jbpm/userTask/jpa/it/UserTaskFilterIT.java
```diff
@@ -0,0 +1,395 @@
...
+        // Given - Start a process instance
+        Traveller traveller = new Traveller("John", "Doe", "john.doe@example.com", "American",
+                new Address("main street", "Boston", "10005", "US"));
+        startProcessInstance(traveller);
+
+        // When - Query with uppercase taskName (should match case-insensitively)
+        given().contentType(ContentType.JSON)
+                .queryParam("user", "manager")
+                .queryParam("group", "department-managers")
+                .queryParam("taskName", "APPROVAL")
```
martinweiler, line 209: So the task name match is a partial, case-insensitive match, correct? How about task names with spaces, are these treated correctly (both exact match as well as partial match)?

GopikaReghunath: Yes, The task name match is partial and case-insensitive.

For task names with spaces, I verified this behavior by temporarily modifying the BPMN file to use task names with spaces ("First Line Approval" instead of "firstLineApproval") and running the integration tests, which confirmed that all matching scenarios work correctly including:

Test Scenarios Verified:

Partial match from start: Search "First" → Matches "First Line Approval" ✅
Partial match from middle: Search "Line" → Matches "First Line Approval" ✅
Partial match from end: Search "Approval" → Matches "First Line Approval" ✅
Exact match: Search "First Line Approval" → Matches "First Line Approval" ✅
Partial match with space: Search "Line App" → Matches "First Line Approval" ✅
Case-insensitive match: Search "first LINE approval" → Matches "First Line Approval" ✅
Empty string: Search "" → Matches all tasks ✅
Non-existent: Search "nonExistentTask" → Returns no results ✅
All tests passed successfully, confirming that task names with spaces are handled correctly.

martinweiler (approved):
Great work @GopikaReghunath

josedee:
@GopikaReghunath Please create the issue in https://github.com/apache/incubator-kie-issues/issues

GopikaReghunath:
> @GopikaReghunath Please create the issue in https://github.com/apache/incubator-kie-issues/issues

Issue created https://github.com/apache/incubator-kie-issues/issues/2308

quarkus/integration-tests/integration-tests-quarkus-usertasks/src/test/java/org/jbpm/userTask/jpa/it/UserTaskFilterIT.java
josedee: What do you think about `status` filter being case insensitive. Would it make more sense. Just a thought

GopikaReghunath: Since status is having predefined values I kept it as case sensitive to match those predefined values. Let me know if this needs to be changed. @martinweiler Let me know your thoughts as well

martinweiler: I don't have a strong opinion on making `status` filter case insensitive, although I can see that this could be a small improvement in the usability of the query.

GopikaReghunath: PR comment addressed

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/usertask/RestResourceUserTaskQuarkusTemplate.java
```diff
@@ -273,4 +333,65 @@ public Attachment getAttachment(
                 .orElseThrow(() -> new UserTaskInstanceNotFoundException("Attachment " + attachmentId + " not found"));
     }
 
+    @GET
+    @Path("/{taskName}/schema")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Map<String, Object> getSchemaByTaskName(
+            @PathParam("taskName") String taskName,
+            @QueryParam("user") String user,
+            @QueryParam("group") List<String> groups) {
```
josedee: @GopikaReghunath I have my doubts over this implementation. What happens if we have
- Two different processes with same task name and different schema
- Same process with two tasks with same task name

Aren't these scenarios possible. Shouldn't we use something unique like task definition id

cc: @martinweiler @pefernan Would like to hear your thoughts

josedee (dismissed):

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/usertask/RestResourceUserTaskQuarkusTemplate.java
GopikaReghunath: We tried the scenarios as mentioned by @josedee  and below observations were made example used is process-user-tasks-subsystem (hiring.bpmn is the process)

Two different processes with same task name and different schema
		created a test process (test.bpmn) with taskName hr_interview input: interview_in output: interview_out, while hitting schema endpoint with taskName as hr_interview only the schema details from hiring.bpmn process is returned and the test process is ignored

Same process with two tasks with same task name
		To test this we renamed it_interview to hr_interview in hiring.bpmn (The input/output of both these tasks are different the original hr_interview has hr_approval while it_interview which is renamed to hr_interview has it_approval) and hit the schema endpoint. The response had both it_approval and hr_approval, Attaching a sample response for second scenario 

`URL : http://localhost:8080/usertasks/hr_interview/schema`
		
`{
  "$schema": "https://json-schema.org/draft/2019-09/schema",
  "$defs": {
    "Candidate": {
      "type": "object",
      "properties": {
        "firstName": {
          "type": "string"
        },
        "lastName": {
          "type": "string"
        },
        "position": {
          "type": "string"
        }
      }
    }
  },
  "type": "object",
  "properties": {
    "candidate": {
      "$ref": "#/$defs/Candidate",
      "input": true
    },
    "it_approval": {
      "type": "boolean",
      "input": true,
      "output": true
    },
    "hr_approval": {
      "type": "boolean",
      "input": true,
      "output": true
    }
  }
}`

cc: @martinweiler @pefernan

GopikaReghunath: The schema endpoints has been removed as discussed

josedee (approved):

addons/common/jbpm-usertask-storage-jpa/src/main/java/org/jbpm/usertask/jpa/repository/UserTaskInstanceRepository.java
```diff
@@ -42,6 +52,73 @@ public List<UserTaskInstanceEntity> findByIdentity(IdentityProvider identityProv
...
 
+    public List<UserTaskInstanceEntity> findByIdentityAndFilter(IdentityProvider identityProvider, UserTaskFilter filter) {
+        // If no filter, use the optimized named query
+        if (filter == null) {
+            return findByIdentity(identityProvider);
+        }
+
+        // Build single query combining identity and filter predicates using Criteria API
+        CriteriaBuilder cb = getEntityManager().getCriteriaBuilder();
+        CriteriaQuery<UserTaskInstanceEntity> cq = cb.createQuery(UserTaskInstanceEntity.class);
```
jstastny-cz: I would advice not mixing approaches JPQL and Criteria query. Enough that we have such differentiation between components, but bringing it into a single class basically means we have to manage duplicate implementations.

addons/common/jbpm-usertask-storage-jpa/src/main/java/org/jbpm/usertask/jpa/repository/UserTaskInstanceRepository.java
```diff
@@ -42,6 +52,73 @@ public List<UserTaskInstanceEntity> findByIdentity(IdentityProvider identityProv
         return query.getResultList();
     }
 
+    public List<UserTaskInstanceEntity> findByIdentityAndFilter(IdentityProvider identityProvider, UserTaskFilter filter) {
+        // If no filter, use the optimized named query
+        if (filter == null) {
+            return findByIdentity(identityProvider);
```
jstastny-cz: This seems as if this method should be overload of findByIdentity in the first place.

api/kogito-api/src/main/java/org/kie/kogito/usertask/UserTaskFilter.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.List;
+
+import org.kie.kogito.usertask.lifecycle.UserTaskState;
+
+/**
+ * Filter criteria for querying user tasks.
+ * All filters are combined using AND logic.
+ * Null filters are ignored (no filtering applied for that criterion).
+ */
+public class UserTaskFilter {
```
jstastny-cz: This can be a record instead of class + builder. It does not have any additional behavior, just getters and instance creation.

api/kogito-api/src/main/java/org/kie/kogito/usertask/UserTaskInfo.java
```diff
@@ -0,0 +1,126 @@
...
+import org.kie.kogito.usertask.lifecycle.UserTaskState;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+/**
+ * Lightweight data transfer object for user task list operations.
+ * Contains only essential task information without inputs, outputs, or metadata.
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class UserTaskInfo {
```
jstastny-cz: Don't we need equals and hashcode for DTOs?

api/kogito-api/src/main/java/org/kie/kogito/usertask/UserTaskInstances.java
```diff
@@ -32,6 +32,8 @@ public interface UserTaskInstances {
 
     List<UserTaskInstance> findByIdentity(IdentityProvider identityProvider);
 
+    List<UserTaskInstance> findByIdentityAndFilter(IdentityProvider identityProvider, UserTaskFilter filter);
```
jstastny-cz: As apparent from implementation, isn't this rather overload of findByIdentity? As it falls back to call that when filter is not provided.

jstastny-cz (changes requested):
Generally looks good, I would just ask to stick with existing query implementation for given module, not to bring in parallel execution paths to maintain - JPQL vs Criteria API.
As soon as schema changes later on, it would need to be adjusted not only in 2 places, but in 2 implementation approaches.

addons/common/jbpm-usertask-storage-jpa/src/main/java/org/jbpm/usertask/jpa/repository/UserTaskInstanceRepository.java
GopikaReghunath: comment addressed

addons/common/jbpm-usertask-storage-jpa/src/main/java/org/jbpm/usertask/jpa/repository/UserTaskInstanceRepository.java
GopikaReghunath: comment addressed

api/kogito-api/src/main/java/org/kie/kogito/usertask/UserTaskFilter.java
GopikaReghunath: comment addressed

api/kogito-api/src/main/java/org/kie/kogito/usertask/UserTaskInstances.java
GopikaReghunath: comment addressed

addons/common/jbpm-usertask-storage-jpa/src/main/java/org/jbpm/usertask/jpa/model/UserTaskInstanceEntity.java
```diff
@@ -29,25 +29,33 @@
...
+                    "or adminGroup in (:roles) " +
+                    "or userTask.actualOwner = :userId " +
+                    "or (userTask.actualOwner is null " + // checking if task is not reserved, we cannot check by status since lifecycle can be customized
+                    "and :userId not member of userTask.excludedUsers " +
+                    "and (:userId member of userTask.potentialUsers or potentialGroup in (:roles)))) " +
+                    "and (cast(:processId as string) is null or userTask.processInfo.processId = :processId) " +
+                    "and (cast(:processInstanceId as string) is null or userTask.processInfo.processInstanceId = :processInstanceId)";
+
+    // TaskName filter clause - case-insensitive (conditional to avoid bytea issues)
+    public static final String TASKNAME_FILTER_CLAUSE = " and lower(userTask.taskName) like lower(concat('%', :taskName, '%'))";
```
jstastny-cz: do these TASKNAME and STATUS filters hit DB index or resorts in skipping them (concat is always a problem) ?

addons/common/jbpm-usertask-storage-jpa/src/main/java/org/jbpm/usertask/jpa/JPAUserTaskInstances.java
```diff
@@ -61,6 +62,15 @@ public List<UserTaskInstance> findByIdentity(IdentityProvider identityProvider)
                 .toList();
     }
 
+    @Override
+    public List<UserTaskInstance> findByIdentity(IdentityProvider identityProvider, UserTaskFilter filter) {
```
jstastny-cz, line 62: shouldn't `findByIdentity(IdentityProvider identityProvider)` call this overload with null filter parameter?

api/kogito-api/src/main/java/org/kie/kogito/usertask/UserTaskInfo.java
```diff
@@ -0,0 +1,154 @@
...
```
jstastny-cz: why isn't existing UserTaskView enough? Seems like we're creating new view/dto for each case, this might come out of hands soon.

api/kogito-api/src/main/java/org/kie/kogito/usertask/view/UserTaskInputsView.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+/**
+ * Lightweight view for user task inputs.
+ * Used by PUT /usertasks/instance/{id}/inputs endpoint to return only input data.
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class UserTaskInputsView {
```
jstastny-cz, line 30: UserTaskView has both inputs and outputs values, why new view/dto? Isn't it possible to used the existing one?

springboot/integration-tests/integration-tests-springboot-usertasks-it/src/test/java/org/jbpm/userTask/jpa/it/UserTaskFilterIT.java
```diff
@@ -0,0 +1,668 @@
...
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.Matchers.hasItem;
+
+/**
+ * Integration test for user task filtering functionality.
+ * Tests the filter query parameters on the /usertasks/instance endpoint.
+ */
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT, classes = KogitoSpringbootApplication.class)
+@ContextConfiguration(initializers = PostgreSqlSpringBootTestResource.class)
+public class UserTaskFilterIT extends BaseUserTaskIT {
```
jstastny-cz, line 45: Please add tests also to quarkus and springboot modules under addons/jbpm-usertask-storage-jpa testing the storage, not only these rest level tests. They are complementary.

jstastny-cz (changes requested):

api/kogito-api/src/main/java/org/kie/kogito/usertask/view/UserTaskInputsView.java
jstastny-cz: This was an explicit ask during reviews, resolving.

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/usertask/RestResourceUserTaskSpringTemplate.java
```diff
@@ -92,15 +99,54 @@ public void init() {
...
-            @RequestParam(value = "group", required = false) List<String> groups) {
-        return userTaskService.list(identityProviderFactory.getOrImpersonateIdentity(user, groups));
+    public Object list(
+            @RequestParam(value = "user", required = false) String user,
+            @RequestParam(value = "group", required = false) List<String> groups,
+            @RequestParam(value = "format", required = false) String format,
+            @RequestParam(value = "processId", required = false) String processId,
+            @RequestParam(value = "processInstanceId", required = false) String processInstanceId,
+            @RequestParam(value = "status", required = false) List<String> status,
+            @RequestParam(value = "taskName", required = false) String taskName) {
```
jstastny-cz, line 92: What is the rationale behind using task name (free form text) as the filtering parameter for a task? Why not use user task id instead?
With this choice and allowing partial match, IMHO you're beating the "selling" point calling this performance optimization.
When checking jbpm_user_tasks table I can see index only on id column and user_task_id:
```
   "jbpm_user_tasks_pkey" PRIMARY KEY, btree (id)
   "idx_usertasks_tid" btree (user_task_id)
```
So those are IMHO the only ones to be queried in performant and optimized way.
Even if the task name column was indexed - the partial match and concat SQL functions might prevent the query from hitting the index at all.

Problem with user_task_id is that the KIE editor might allow only generated values - but that's IMHO a room for improvement on that front, I don't think it needs to be restricted to that.

addons/common/jbpm-usertask-storage-jpa/src/main/java/org/jbpm/usertask/jpa/model/UserTaskInstanceEntity.java
pefernan: I have some feelings about this too, I'm afraid this will perform very bad with a large amount of tasks in DB. I'd prefer use not use partial and lowercase here. I think that it's better to go for equality in a first iteration and we can see if we need to evolve later if we get feedback on this.

martinweiler: You are raising some valid concerns about potential performance issues. We also don't have an actual request to support `like` queries.

@GopikaReghunath please adjust the code accordingly. Thanks!

pefernan: To be clear, I'd go with exact matches and would try to rely in DB indexes as much as possible

addons/common/jbpm-usertask-storage-jpa/src/main/java/org/jbpm/usertask/jpa/JPAUserTaskInstances.java
pefernan: +1

api/kogito-api/src/main/java/org/kie/kogito/usertask/UserTaskInfo.java
pefernan: I think the idea is have a reduced view of the usertask to not return all task information unless it is required (metadata, inputs, outputs...)

addons/common/jbpm-usertask-storage-jpa/src/main/java/org/jbpm/usertask/jpa/model/UserTaskInstanceEntity.java
pefernan: maybe we should add more indexes to our schema with this new feature

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/usertask/RestResourceUserTaskSpringTemplate.java
pefernan: Task Name is a business name that the user gives to a task (or tasks), I think it is good to have it here.

addons/common/jbpm-usertask-storage-jpa/src/main/java/org/jbpm/usertask/jpa/model/UserTaskInstanceEntity.java
GopikaReghunath: Made taskname filter case sensitive and exact match

addons/common/jbpm-usertask-storage-jpa/src/main/java/org/jbpm/usertask/jpa/JPAUserTaskInstances.java
GopikaReghunath: comment addressed

api/kogito-api/src/main/java/org/kie/kogito/usertask/UserTaskInfo.java
GopikaReghunath: I have moved this class to view package

springboot/integration-tests/integration-tests-springboot-usertasks-it/src/test/java/org/jbpm/userTask/jpa/it/UserTaskFilterIT.java
GopikaReghunath: comment addressed

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/usertask/RestResourceUserTaskSpringTemplate.java
GopikaReghunath: Have made taskname exact match and case sensitive

api/kogito-api/src/main/java/org/kie/kogito/usertask/UserTaskService.java
```diff
@@ -34,13 +37,15 @@ public interface UserTaskService {
...
+
     Optional<UserTaskView> transition(String taskId, String transitionId, Map<String, Object> data, IdentityProvider identity);
 
     List<UserTaskTransitionView> allowedTransitions(String taskId, IdentityProvider identity);
 
-    Optional<UserTaskView> setOutputs(String taskId, Map<String, Object> data, IdentityProvider identity);
+    Optional<UserTaskOutputsView> setOutputs(String taskId, Map<String, Object> data, IdentityProvider identity);
 
-    Optional<UserTaskView> setInputs(String taskId, Map<String, Object> data, IdentityProvider identity);
+    Optional<UserTaskInputsView> setInputs(String taskId, Map<String, Object> data, IdentityProvider identity);
```
cimbalek, line 47: Aren't these chages breaking backward compatibility? And do we need all those new views?

cimbalek: Seems that it was part of assignment, but the question about breaking backwards compatibility stays.

martinweiler: @cimbalek thanks for highlighting this! I agree that this is a breaking change, but one that has been done intentionally. There are other endpoints available which return the full UserTaskView that users can call if needed. Returning the new `UserTaskInputsView` / `UserTaskOutputsView` streamlines the response for the input/output related endpoints.

martinweiler (approved):

cimbalek (approved):

addons/common/jbpm-usertask-storage-jpa/src/main/java/org/jbpm/usertask/jpa/repository/UserTaskInstanceRepository.java
```diff
@@ -36,9 +42,61 @@ public UserTaskInstanceRepository(UserTaskJPAContext context) {
...
+
+            // Only set taskName if we added it to the query
+            if (hasTaskNameFilter) {
+                query.setParameter("taskName", filter.taskName());
+            }
+
+            // Only set statusFilter if we added it to the query
+            if (hasStatusFilter) {
+                List<String> statusFilter = filter.statuses().stream()
+                        .map(UserTaskState::getName)
```
jstastny-cz: entity.setStatus(userTaskInstance.getStatus().getName());

Check the mapper - we have full control over the format of the state value - it's being persisted via getName() of the enum. We should drop any lower-case transformation, especially from the SQL query, and just make sure we pass the correctly formatted values.

addons/common/jbpm-usertask-storage-jpa/src/main/java/org/jbpm/usertask/jpa/model/UserTaskInstanceEntity.java
```diff
@@ -29,25 +29,33 @@
...
+            "select userTask from UserTaskInstanceEntity userTask " +
+                    "left join userTask.adminGroups adminGroup " +
+                    "left join userTask.potentialGroups potentialGroup " +
+                    "where (:userId member of userTask.adminUsers " +
+                    "or adminGroup in (:roles) " +
+                    "or userTask.actualOwner = :userId " +
+                    "or (userTask.actualOwner is null " + // checking if task is not reserved, we cannot check by status since lifecycle can be customized
+                    "and :userId not member of userTask.excludedUsers " +
+                    "and (:userId member of userTask.potentialUsers or potentialGroup in (:roles)))) " +
+                    "and (cast(:processId as string) is null or userTask.processInfo.processId = :processId) " +
```
jstastny-cz: Do we need such casts for a NULL check?
Why don't we append the query similarly to how we handle taskName and status fields? That's IMHO cleaner approach, also you have such logic already applied for taskName and status, why not also processId and processInstanceId?

jstastny-cz (commented):
One thing I still don't completely understand is why we opted to filter by task name? I suppose such value is an open-text value, not checked for uniqueness across BPMN model, what is the intended usage when multiple user tasks have the same task name? Not that it would be a good design, but I think can happen. My assumption is that User Task ID is required to be unique within a single BPMN. (Would be good if you could check that.)
I'd also suggest unifying approach how we implement the conditional presence of the filter conditions, we're mixing bind param NULL check in query with conditional query appending (and parameter binding) - I'd prefer the second option for all.

quarkus/integration-tests/integration-tests-quarkus-usertasks/src/test/java/org/jbpm/userTask/jpa/it/UserTaskFilterIT.java
```diff
@@ -466,98 +466,7 @@ public void testFilterByStatusCaseInsensitive() {
...
-                .queryParam("user", "manager")
-                .queryParam("group", "department-managers")
-                .queryParam("status", "ReSeRvEd")
-                .when()
-                .get(USER_TASKS_ENDPOINT)
-                .then()
-                .statusCode(200)
-                .body("$.size()", is(1))
-                .body("[0].status.name", is("Reserved"));
+                .body("$.size()", is(0));
```
jstastny-cz, line 469: I think this test case either needs renaming or remove? `testFilterByStatusCaseInsensitive` - also based on body contents - now it is case sensitive.

GopikaReghunath: comment addressed

addons/common/jbpm-usertask-storage-jpa/src/main/java/org/jbpm/usertask/jpa/model/UserTaskInstanceEntity.java
GopikaReghunath: comment addressed

addons/common/jbpm-usertask-storage-jpa/src/main/java/org/jbpm/usertask/jpa/repository/UserTaskInstanceRepository.java
GopikaReghunath: comment addressed

springboot/addons/jbpm-usertask-storage-jpa/pom.xml
```diff
@@ -53,5 +53,48 @@
...
+            <groupId>com.h2database</groupId>
+            <artifactId>h2</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.postgresql</groupId>
+            <artifactId>postgresql</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
```
jstastny-cz, line 73: We should rather use 
```
        <dependency>
            <artifactId>kogito-spring-boot-test-utils</artifactId>
            <scope>test</scope>
        </dependency>
```
instead of direct testcontainers dependencies, or why do we not use it?

jstastny-cz (commented):
I finally got to review the tests too, looks good except for the maven dependency use.

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/usertask/RestResourceUserTaskQuarkusTemplate.java
```diff
@@ -87,8 +84,49 @@ public void init() {
...
+            @QueryParam("taskName") String taskName) {
+
+        // Build filter from query parameters
+        UserTaskFilter.Builder filterBuilder = UserTaskFilter.builder()
+                .processId(processId)
+                .processInstanceId(processInstanceId)
+                .taskName(taskName);
+        
+        // Handle multiple status values
+        if (status != null && !status.isEmpty()) {
```
pefernan: @GopikaReghunath Considering that you use the UserTaskState name to build the query [here](https://github.com/apache/incubator-kie-kogito-runtimes/pull/4273/changes#diff-f5872faada1b3bc1899550b79afbd939226b3c39e778fa51ccb370e54195251cR107) I think this conversion is not necessary. 

You can directly pass the states collection to the `filterBuilder`, so we simplify this and both REST templates are simpler.

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/usertask/RestResourceUserTaskQuarkusTemplate.java
```diff
@@ -87,8 +84,49 @@ public void init() {
 
     @GET
     @Produces(MediaType.APPLICATION_JSON)
-    public List<UserTaskView> list(@QueryParam("user") String user, @QueryParam("group") List<String> groups) {
-        return userTaskService.list(identityProviderFactory.getOrImpersonateIdentity(user, groups));
+    public Object list(
```
pefernan: I understand the change, but I still think it should return a List<UserTaskView> instead of object, why not using [jackson views](https://www.baeldung.com/jackson-json-view-annotation) on top of `UserTaskView` to somehow switch between both task presentations. 
Maybe with this approach we can also simplify [this fragment](https://github.com/apache/incubator-kie-kogito-runtimes/pull/4273/changes#diff-3efefca73563302797ac35ae81f359f53ab353d6a23b27d90670e23c60bb35b6R113-R129) in both templates.

pefernan (changes requested):
@GopikaReghunath thank you for this work! I think now it's pretty close, could you please take a look at my comments?

springboot/addons/jbpm-usertask-storage-jpa/pom.xml
GopikaReghunath: comment addressed

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/usertask/RestResourceUserTaskQuarkusTemplate.java
GopikaReghunath: comment addressed

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/usertask/RestResourceUserTaskQuarkusTemplate.java
GopikaReghunath: comment addressed

jstastny-cz (approved):
Thanks @GopikaReghunath for addressing all the concerns. Looks good.

pefernan (approved):

PR #4275, opened 2026-05-05
[incubator-kie-issues#2294] Workflow: Use constants instead of String…

### LINKED ISSUE #2294: Workflow: Use constants instead of Strings for the notification handling
The current implementation of event handling in NotificationEventPublisher relies on comparing fully qualified class names using String values .
This approach is fragile and prone to breaking silently during refactoring (e.g., class renaming, package restructuring), as such changes are not detected at compile time.

abhijithumbe:
…s for the notification handling

 Fixes: https://github.com/apache/incubator-kie-issues/issues/2294


martinweiler (approved):
Thank you @abhijithumbe

PR #4283, opened 2026-05-12
[DO-NOT-MERGE] [incubator-kie-drools-6576] Native Java Compiler as default instead o…

tkobayas:
…f ecj

- Issue: https://github.com/apache/incubator-kie-drools/issues/6576

- Cross-repo PR : https://github.com/apache/incubator-kie-drools/pull/6587

---
- This kogito-runtimes PR only fixes test assertion not to depend on ecj error message
- If we set `-Ddrools.dialect.java.compiler=ECLIPSE` for kogito application:
    - Switches BPMN-script and rule compilation back to ECJ.
    - Does not switch the kogito-maven-plugin's generated-sources compilation back to ECJ.
    - So a kogito project cannot uniformly opt back to ECJ. The kogito-maven-plugin path will always use Native compiler. This is pre-existing, not introduced by this PR.
-  In other words, this drools PR makes uniformly Native compiler as default from Native compiler + ECJ mix.


tkobayas:
We will keep ECJ as the default, based on the performance results (See https://github.com/apache/incubator-kie-drools/issues/6576). Closing.

PR #4302, opened 2026-05-22
incubator-kie-issues#2307 data-isolation by process id and version pair

jstastny-cz:
Fixes:
* apache/incubator-kie-issues/issues/2307

Part of ensemble:
* apache/incubator-kie-kogito-runtimes/pull/4302
* apache/incubator-kie-kogito-apps/pull/2337

Broad changes to accommodate for possible filtering of DB entries by process id and version pairs. Important aspect is that the engine was currently not properly populating the version information throughout the internal models, so there are excessive changes related to this, while the actual data-isolation filtering is small footprint.

Moreover, for sake of properly isolate items belonging to reusable subprocesses, id and version information needs to be exposed also for root process, so via root process id and root process version.

Areas affected:
* Code generation - Processes bean now apart from processIds() method implements also processes() exposing both id and version information.
* Marshalling - process_instances payload column holds byte representation of given process instance, adding fields in the class requires also change in marshalling.
* Data Events - data events were refactored to properly set the new fields, instead of adding yet another constructor with way too many parameters, I implemented builder pattern for the data event objects. This should be now the encouraged way, I have marked the existing multi-parameter constructors as deprecated.
* Database queries - actual database queries are being composed dynamically and depending on local processes, appropriate filtering is appended and parameters bound.
* Conditional enabling of this feature by passing or not Processes bean to Repository class implementations.

Components isolated:
* runtime persistence addon
* jbpm usertasks storage addon


pefernan (approved):
great work @jstastny-cz !!

martinweiler (approved):

PR #4330, opened 2026-07-03
[incubator-kie-issues-2360] Migrate SLAComplianceTest methods from v7 legacy runtime to code generation

### LINKED ISSUE #2360: Migrate SLAComplianceTest methods from v7 legacy runtime to code generation (v9)
Summary
Migrate test methods in SLAComplianceTest.java from the v7 runtime-based API (createKogitoProcessRuntime) to the v9 code generation-based API (ProcessTestHelper + generated process classes).

File
Those test can be told by
https://github.com/apache/incubator-kie-kogito-runtimes/blob/main/jbpm/jbpm-tests/src/test/java/org/jbpm/bpmn2/SLAComplianceTest.java

an example of migration is
https://github.com/apache/incubator-kie-kogito-runtimes/blob/main/jbpm/jbpm-tests/src/test/java/org/jbpm/bpmn2/ResourceTest.java#L37

SUNILathulya:
closes https://github.com/apache/incubator-kie-issues/issues/2360

Migrated SLAComplianceTest which are being compiled using legacy v7 to code generation approach.

The tests can be identified by referring to SLAComplianceTest.java:
https://github.com/apache/incubator-kie-kogito-runtimes/blob/main/jbpm/jbpm-tests/src/test/java/org/jbpm/bpmn2/SLAComplianceTest.java

## Tests Migrated

**testSLAonProcessViolated()** - Line 62
- Migrates process-level SLA violation tracking
- Converts from v7 `createKogitoProcessRuntime()` to v9 generated classes
- Uses `UserTaskWithSLAProcess` and `UserTaskWithSLAModel`
- Captures SLA compliance on process completion via `afterProcessCompleted()` event listener
- Validates SLA is violated and remains violated after process completes

**testSLAonUserTaskViolated()** - Line 140
- Migrates node-level SLA violation tracking on a human task
- Uses `UserTaskWithSLAOnTaskProcess` and `UserTaskWithSLAOnTaskModel`
- Captures node SLA compliance at violation time via `afterSLAViolated()` and after completion via `afterNodeLeft()`
- Validates process-level SLA is `SLA_NA` while node-level SLA is `SLA_VIOLATED`

**testSLAonCatchEventViolated()** - Line 444
- Migrates SLA violation tracking on an intermediate catch event node
- Uses `IntermediateCatchEventSignalWithSLAOnEventProcess` and `IntermediateCatchEventSignalWithSLAOnEventModel`
- Replaces `kruntime.signalEvent()` with `processInstance.send(SignalFactory.of(...))`
- Validates event node SLA is `SLA_VIOLATED` and persists after signal received

## Migration Changes
- Replaced `createKogitoProcessRuntime()` with generated process classes
- Used `ProcessTestHelper` for application, handler and event listener setup
- Replaced `getSLAComplianceForProcessInstance()` with `executeInWorkflowState(p -> p.getSlaCompliance())`
- Replaced `getSLAComplianceForNodeInstance()` with `AtomicInteger` captured via event listeners
- Replaced `assertProcessInstanceFinished()` with direct `processInstance.status()` check
- Updated constants from `KogitoProcessInstance.STATE_*` / `SLA_*` to v9 equivalents

## BPMN Files
- `BPMN2-UserTaskWithSLA.bpmn2` (test 1)
- `BPMN2-UserTaskWithSLAOnTask.bpmn2` (test 2)
- `BPMN2-IntermediateCatchEventSignalWithSLAOnEvent.bpmn2` (test 3)

### DIFF
--- a/jbpm/jbpm-tests/src/test/java/org/jbpm/bpmn2/SLAComplianceTest.java
+++ b/jbpm/jbpm-tests/src/test/java/org/jbpm/bpmn2/SLAComplianceTest.java
@@ -61,22 +61,29 @@ public class SLAComplianceTest extends JbpmBpmn2TestCase {
     @Test
     public void testSLAonProcessViolated() throws Exception {
         CountDownLatch latch = new CountDownLatch(1);
-        final KogitoProcessEventListener listener = new DefaultKogitoProcessEventListener() {
+        AtomicInteger processSlaCompliance = new AtomicInteger(-1);
 
+        KogitoProcessEventListener listener = new DefaultKogitoProcessEventListener() {
             @Override
             public void afterSLAViolated(SLAViolatedEvent event) {
                 latch.countDown();
             }
 
+            @Override
+            public void afterProcessCompleted(ProcessCompletedEvent event) {
+                WorkflowProcessInstanceImpl impl = (WorkflowProcessInstanceImpl) event.getProcessInstance();
+                processSlaCompliance.set(impl.getSlaCompliance());
+            }
         };
-        kruntime = createKogitoProcessRuntime("org/jbpm/bpmn2/sla/BPMN2-UserTaskWithSLA.bpmn2");
 
+        Application app = ProcessTestHelper.newApplication();
         TestWorkItemHandler workItemHandler = new TestWorkItemHandler();
-        kruntime.getKogitoWorkItemManager().registerWorkItemHandler("Human Task", workItemHandler);
-        kruntime.getProcessEventManager().addEventListener(listener);
-
-        KogitoProcessInstance processInstance = kruntime.startProcess("UserTaskWithSLA");
-        assertThat(processInstance.getState()).isEqualTo(KogitoProcessInstance.STATE_ACTIVE);
+        ProcessTestHelper.registerHandler(app, "Human Task", workItemHandler);
+        ProcessTestHelper.registerProcessEventListener(app, listener);
+        org.kie.kogito.process.Process<UserTaskWithSLAModel> processDefinition = UserTaskWithSLAProcess.newProcess(app);
+        ProcessInstance<UserTaskWithSLAModel> processInstance = processDefinition.createInstance(processDefinition.createModel());
+        processInstance.start();
+        assertThat(processInstance.status()).isEqualTo(ProcessInstance.STATE_ACTIVE);
 
         KogitoWorkItem workItem = workItemHandler.getWorkItem();
         assertThat(workItem).isNotNull();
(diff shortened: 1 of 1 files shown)


pefernan (approved):

jbpm/jbpm-tests/src/test/java/org/jbpm/bpmn2/SLAComplianceTest.java
```diff
@@ -85,17 +89,17 @@ public void afterSLAViolated(SLAViolatedEvent event) {
...
+        assertThat(slaCompliance).isEqualTo(org.kie.api.runtime.process.ProcessInstance.SLA_VIOLATED);
 
-        kruntime.getKogitoWorkItemManager().completeWorkItem(workItem.getStringId(), null);
-        assertProcessInstanceFinished(processInstance, kruntime);
+        ProcessTestHelper.completeWorkItem(processInstance, Collections.emptyMap(), "john");
+        assertThat(processInstance.status()).isEqualTo(ProcessInstance.STATE_COMPLETED);
 
-        slaCompliance = getSLAComplianceForProcessInstance(processInstance);
-        assertThat(slaCompliance).isEqualTo(KogitoProcessInstance.SLA_VIOLATED);
+        assertThat(processSlaCompliance.get()).isEqualTo(org.kie.api.runtime.process.ProcessInstance.SLA_VIOLATED);
```
martinweiler, line 102: Why do we need the listener approach here instead of querying the `slaCompliance` directly from the processInstance, similar to how it is done in lines 96-97?

It would also be possible to adapt the `getSLAComplianceForProcessInstance` method to the new style and use it throughout the test case. Or is this not possible because we can't access the process instance anymore once it is completed?

SUNILathulya: @martinweiler  Thanks for review — the listener approach is necessary specifically for asserting SLA compliance after the process completes.

In v9 with InMemoryProcessInstances, once a process reaches STATE_COMPLETED, it is immediately removed from the instances store (via MutableProcessInstances.remove()). At that point, processDefinition.instances().findById(id) returns empty — the process instance is no longer accessible.

The existing executeInWorkflowState(processInstance, p -> p.getSlaCompliance()) on line 96 works because it is called while the process is still ACTIVE (after SLA violation but before the work item is completed). That's the same pattern as the old getSLAComplianceForProcessInstance() — querying on a live instance.

However, for the assertion on line 102, we need the SLA compliance value after completeWorkItem has been called (and the process has completed and been removed from the store). The afterProcessCompleted listener fires while the internal WorkflowProcessInstanceImpl still exists in memory — before the store removes it — so it's the only reliable hook to capture the final SLA compliance value post-completion.

Adapting getSLAComplianceForProcessInstance to the new style would face the same limitation: it can only work on a live/active instance. For post-completion assertions, the listener is the correct approach.

martinweiler: Thanks for the explanation - that makes sense! Please raise this PR on the unified drools repo and we can get it approved and merged!

SUNILathulya: @martinweiler  Sure, Thanks!

PR #4335, opened 2026-07-07
[incubator-kie-issues-2367] PATCH API ignores explicit JSON null values for String process variables

### LINKED ISSUE #2367: PATCH request ignores explicit JSON null values for String process variables
When updating process variables using the generated workflow REST API, a PATCH request containing an explicit JSON null for a String process variable does not update the variable.

Instead, the existing value is retained.

PATCH requests using the literal string "null" update the variable successfully, but store the string "null" rather than clearing the variable.

PUT requests do not exhibit this behavior.

alishaaaqil:
Closes: https://github.com/apache/incubator-kie-issues/issues/2367

Fixes PATCH requests ignoring explicit JSON null values during partial process variable updates.

This approach preserves field presence during deserialization so explicit null values can be distinguished from omitted fields, while keeping the existing public APIs and REST endpoints unchanged. Existing PATCH, PUT, and POST behavior has been validated.


martinweiler:
Thanks for this PR @alishaaaqil 

However, I have the impression that this is overly complicated, and adds a lot of code. Wouldn't it be rather a question of removing the null check here, or are we afraid this could cause any regressions or other unwanted side-effects?
https://github.com/alishaaaqil/incubator-kie-kogito-runtimes/blob/main/api/kogito-api/src/main/java/org/kie/kogito/Model.java#L36

alishaaaqil:
@martinweiler 
replying to your  comment

The issue is that by the time updatePartially() is called, the information about whether a field was omitted from the JSON or explicitly provided as null has already been lost. Both cases have become plain Java null values when the model is converted to a map.

If we simply remove the null filter here, the map contains both omitted fields and explicitly null fields, so update() would clear all omitted fields as well. In effect, PATCH would start behaving like a full replace (PUT) for any omitted variable.

martinweiler (approved):
Thank you @alishaaaqil

kogito-codegen-modules/kogito-codegen-processes/src/main/resources/class-templates/RestResourceQuarkusTemplate.java
```diff
@@ -160,7 +160,9 @@ public class $Type$Resource {
     @Produces(MediaType.APPLICATION_JSON)
     @Operation(operationId = "updateProcessInstance_$name$", summary = "$documentation$", description = "$processInstanceDescription$")
     public $Type$Output updateModel_$name$(@PathParam("id") String id, $Type$Input resource) {
-        return processService.update(process, id, resource.toModel()).orElseThrow(NotFoundException::new);
+        $Type$ model = resource.toModel();
+        model.clearModifiedFields();
```
martinweiler, line 164: Two questions:

1. Wouldn't this be required in `RestResourceSpringTemplate` as well?
2. Can we extend the tests to have both `PATCH` and `PUT` calls on a resource, to make sure this invocation works as expected?

martinweiler (changes requested):

martinweiler (approved):

josedee (commented):
@alishaaaqil Thanks for the PR. I have a question over how patch works in terms of a json field. Consider the below example

```
{
  "id": "9ef22827-5dd7-4301-b323-9cc7745d5daf",
  "supportCase": {
    "product": {
      "name": "Kogito",
      "family": "Middleware"
    },
    "description": "test",
    "engineer": "kelly",
    "customer": "Paco",
    "state": "WAITING_FOR_OWNER",
    "comments": null,
    "questionnaire": null
  },
  "supportGroup": "Kogito"
}
```

Is it possible to update a field like `description`. If yes, how? It may not be directly related to the fix you are aiming at. But if this scenario is valid then maybe your fix should aim this also. 

Maybe complex objects like these are only entirely replaceable, in which case you don't need to bother about this comment.

alishaaaqil:
@josedee 
Currently, PATCH works at the process variable level, not inside the fields of a complex object. For a variable such as supportCase, sending only `{"supportCase":{"description":"..."}} `replaces the entire supportCase object rather than updating just the description field, so the complete supportCase object needs to be sent.

This PR keeps that existing behavior and only fixes how explicit null values are handled for top-level process variables.

If updating individual fields within complex objects is something we want to support, I believe that would require a separate enhancement. Would you prefer that to be tracked  seperately

josedee:
Thanks @alishaaaqil for confirming! Not sure if that feature is needed, I was just curious to understand. Maybe a topic for another day.

josedee (approved):

PR #4338, opened 2026-07-09
[incubator-kie-issues#2370] Duplicate Process Variable available in s…

abhijithumbe:
…wagger page if variable starts with Capital Letter
 
Closes/Fixes/Resolves #https://github.com/apache/incubator-kie-issues/issues/2370


martinweiler (changes requested):
Thanks for raising this PR @abhijithumbe 

I'd like to ask for two changes:
1. The comments are quite verbose. If we think it makes sense to add some explanation to the codebase, can we at least try to trim it down to one line? Otherwise, I think users can also use the git history to understand why a certain code change was done, so we could keep the explanation in the PR description.
2. Would it be possible to create a test case for this?

abhijithumbe:
Thank you @martinweiler . I’ve shortened the comment and added a test for the uppercase variable usecase.

martinweiler (approved):
Thank you @abhijithumbe

